PR #8, opened 2017-01-11
Add programs.fish

### LINKED ISSUE #3: Add programs.fish module
I'm making a separate issue for this.

I've done some hacking locally (and trashed my `/nix/store/` several times) and have a broken implementation I can push to my fork tomorrow or Wednesday.

The main issues are with `/`s in the etc config (see `nixpkgs/nixos/modules/programs/fish.nix`) and by extension the `foreign-env/` config. I'm guessing/hoping you can get it sorted once I push. :smile:

yurrriq:
This is working well enough for me to use locally now, so I figured I'd open a PR. If nothing else, we can discuss publicly here.

Closes #3

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -4,12 +4,21 @@ with lib;
 
 let
 
+  cfg = config.programs.fish;
   cfge = config.environment;
 
-  cfg = config.programs.fish;
+  foreignEnv = pkgs.writeText "fish-foreign-env" ''
+    # TODO: environment.shellInit
+    ${cfge.extraInit}
+  '';
 
-  fishVariables =
-    mapAttrsToList (n: v: ''set -x ${n} "${v}"'') cfg.variables;
+  loginForeignEnv = pkgs.writeText "fish-login-foreign-env" ''
+    # TODO: environment.loginShellInit
+  '';
+
+  interactiveForeignEnv = pkgs.writeText "fish-interactive-foreign-env" ''
+    ${cfge.interactiveShellInit}
+  '';
 
   shell = pkgs.runCommand pkgs.fish.name
     { buildInputs = [ pkgs.makeWrapper ]; }
@@ -23,6 +32,9 @@ let
   fishAliases = concatStringsSep "\n" (
     mapAttrsFlatten (k: v: "alias ${k} '${v}'") cfg.shellAliases
   );
+
+  fishVariables =
+    mapAttrsToList (n: v: ''set -x ${n} "${v}"'') cfg.variables;
 
 in
 
@@ -99,10 +111,6 @@ in
 
   config = mkIf cfg.enable {
 
-    environment.etc."fish/foreign-env/shellInit".text = cfge.shellInit;
-    environment.etc."fish/foreign-env/loginShellInit".text = cfge.loginShellInit;
-    environment.etc."fish/foreign-env/interactiveShellInit".text = cfge.interactiveShellInit;
-
     environment.etc."fish/config.fish".text = ''
       # /etc/fish/config.fish: DO NOT EDIT -- this file has been generated automatically.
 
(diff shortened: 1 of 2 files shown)


modules/programs/fish.nix
```diff
@@ -0,0 +1,149 @@
...
+          Shell script code used to initialise fish prompt.
+        '';
+      };
+
+    };
+  };
+
+  config = mkIf cfg.enable {
+
+    environment.etc."fish/foreign-env/shellInit" = mkIf (hasAttr "shellInit" cfge)
```
LnL7: Only [config.environment. interactiveShellInit](https://github.com/LnL7/nix-darwin/blob/master/modules/environment/default.nix#L106) exists at the moment. That's probably why you ran into issues here.

yurrriq: The trouble is [`nix-daemon` `extraInit`](https://github.com/LnL7/nix-darwin/blob/5a5ec9ac42bcc99a6ddc577704824af225b99bd4/modules/services/nix-daemon.nix#L32) is not valid `fish`.

LnL7: Right, for the `foreign-env` thing you need a file?

LnL7:
What about something like this https://github.com/LnL7/nix-darwin/commit/ef80efd731ef2a6c433abc0900fad26b341b9a00?

yurrriq:
That should work. And we can `fenv` it later, right?

LnL7:
I'm not following

yurrriq:
Sorry, I've been under the weather and out of it. I mean [like this](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/fish.nix#L99).

yurrriq:
I've ticked the *Allow edits from maintainers* box so feel free to commit to my branch and I'll pull before doing anything.

LnL7:
Yes, my changes just use the store paths directly instead of linking those to `/etc`. Only `fish/config.fish` has to exist in etc if I'm not mistaken.

LnL7:
Does this look good to you?

yurrriq:
Yeah, I've been using this awhile and it seems to work. Feel free to make any finishing touches and merge at your leisure.

modules/programs/fish.nix
```diff
@@ -0,0 +1,155 @@
...
+        # TODO: environment.loginShellInit
+        ${cfg.loginShellInit}
+      end
+
+      if status --is-interactive
+        ${fishAliases}
+        ${optionalString (cfge.interactiveShellInit != "") ''
+          fenv source ${interactiveForeignEnv}
+        ''}
+        set -x NIX_REMOTE daemon
```
LnL7: This should come from `system.build.setEnvironment` so it's only enabled when the daemon is configured.

modules/programs/fish.nix
```diff
@@ -0,0 +1,155 @@
...
+  config = mkIf cfg.enable {
+
+    environment.etc."fish/config.fish".text = ''
+      # /etc/fish/config.fish: DO NOT EDIT -- this file has been generated automatically.
+
+      set fish_function_path $fish_function_path ${pkgs.fish-foreign-env}/share/fish-foreign-env/functions
+
+      set PATH ${replaceStrings [":"] [" "] config.environment.systemPath} $PATH
+
+      ${config.system.build.setEnvironment}
```
LnL7: Does this work or should it be moved to `foreignEnv`?

yurrriq:
I'll do a once over later today and be sure to check on those two things.

LnL7:
Great! Everything else looks good to merge, also feel free to add an example of your config.

modules/programs/fish.nix
yurrriq: This works because:

```fish
$ functions export
function export --description 'Set global variable. Alias for set -gx, made for bash compatibility'
	if test -z "$argv"
        set
        return 0
    end
    for arg in $argv
        set -l v (string split -m 1 "=" -- $arg)
        switch (count $v)
            case 1
                set -gx $v $$v
            case 2
                set -gx $v[1] $v[2]
        end
    end
end
```

yurrriq:
Not sure where to put this exactly so here's this:

```nix
{
  # ...

  environment.systemPackages = with pkgs; [
    # ...
    highlight
  ];

  programs.fish.enable = true;

  programs.fish.variables.cfg = "$HOME/.nixpkgs/darwin-config.nix";
  programs.fish.variables.darwin = "$HOME/.nix-defexpr/darwin";
  programs.fish.variables.pkgs = "$HOME/.nix-defexpr/nixpkgs";

  programs.fish.interactiveShellInit = ''
    function hicat -d 'Hackish hicat clone via highlight'
      highlight -O xterm256 $argv | less -cR
    end
  '';

  # ...
}
```

yurrriq:
I'm going to rebase and remove the hack commit, but after that I think this is good enough to merge.

yurrriq:
Ok, I'm done rebasing.

modules/programs/fish.nix
LnL7: ah, interesting.

LnL7:
Thanks! 🎉

PR #18, opened 2017-03-05
Add ability to load custom vim plugins

bkase:


modules/programs/vim.nix
```diff
@@ -39,6 +42,27 @@ in {
       '';
     };
 
+    programs.vim.customPlugins = mkOption {
+      # What should the type be here?
+      # type = ?
```
LnL7: `types.attrsOf types.package`

modules/programs/vim.nix
```diff
@@ -39,6 +42,27 @@ in {
...
+      default = ({});
```
LnL7: you can drop the parentheses here, same with the example.

bkase:
Thanks! Fixed

modules/programs/vim.nix
```diff
@@ -39,6 +42,26 @@ in {
       '';
     };
 
+    programs.vim.customPlugins = mkOption {
+      type = types.attrsOf types.package;
+      default = {};
+      example = {
```
LnL7: Examples referring to derivations should use [`literalExample`](https://github.com/LnL7/nix-darwin/blob/master/modules/nix/nixpkgs.nix#L44-L53). This avoids evaluating the derivation when using darwin-option or when building a manual (if I add one).

bkase: interesting, thanks for the tip!

modules/programs/vim.nix
```diff
@@ -39,6 +42,29 @@ in {
       '';
     };
 
+    programs.vim.customPlugins = mkOption {
```
LnL7: I think `extraKnownPlugins` is a slightly better name.

LnL7:
Thanks

PR #21, opened 2017-03-18
Adds mopidy service with media keys support

peel:
Basic support for mopidy service along with media keys support service.
Currently the mopidy installation via nix is broken, therefore set the default packages to `/usr/local`.

Have a look if that's something you're ok with including :)

### DIFF
--- a/modules/services/mopidy.nix
+++ b/modules/services/mopidy.nix
@@ -1,0 +1,61 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.services.mopidy;
+
+in
+
+{
+  options = {
+    services.mopidy = {
+
+      enable = mkOption {
+        type = types.bool;
+        default = false;
+        description = "Whether to enable the Mopidy Daemon.";
+      };
+
+      package = mkOption {
+        type = types.path;
+        default = pkgs.mopidy;
+        description = "This option specifies the mopidy package to use.";
+      };
+
+      mediakeys = {
+        enable = mkOption {
+          type = types.bool;
+          default = false;
+          description = "Whether to enable the Mopidy OSX Media Keys support daemon.";
+        };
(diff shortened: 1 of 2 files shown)


yurrriq:
:+1: to getting this sorted.

modules/services/mopidy.nix
```diff
@@ -0,0 +1,67 @@
...
+
+      };
+
+    };
+  };
+
+  config = mkMerge [
+    (mkIf cfg.enable {
+      launchd.user.agents.mopidy = {
+        serviceConfig.ProgramArguments = [
```
LnL7: You can use [`serviceConfig.Program`](https://github.com/LnL7/nix-darwin/blob/master/modules/launchd/launchd.nix#L103-L111) if you don't need arguments.

modules/services/mopidy.nix
```diff
@@ -0,0 +1,67 @@
...
+          "${cfg.package}/bin/mopidy"
+        ];
+        serviceConfig.RunAtLoad = true;
+        serviceConfig.KeepAlive = true;
+        serviceConfig.ProcessType = "Interactive";
```
LnL7: Not sure if this should be interactive.

modules/services/mopidy.nix
```diff
@@ -0,0 +1,67 @@
...
+
+      enable = mkOption {
+        type = types.bool;
+        default = false;
+        description = "Whether to enable the Mopidy Daemon.";
+      };
+
+      package = mkOption {
+        type = types.path;
+        default = "/usr/local";
```
LnL7: I would prefer the package from nixpkgs here, you can do this in your configuration until it's fixed there.

peel:
Thanks @LnL7. Any more suggestions? :)

LnL7:
Thanks!

PR #22, opened 2017-03-21
Makes mopidy service interactive

peel:
Noticed that adaptive service causes mopidy to stutter under heavy load. 
Mediakeys though works fine as adaptive.


peel:
Yeah, sorry I failed at this. It seems to be working fine under a heavy load w/o a process type setting.

PR #26, opened 2017-05-12
Don't put profiles in reverse-order in PATH

clhodapp:
Your user-specific profile should take precedence over the system one. If I've gone out of my way to install something into my personal profile, my version should be used instead of the system's.


LnL7:
Thanks!

LnL7:
Since profiles is a list it's still possible to get the previous behaviour like this.

```nix
{ config, lib, pkgs, ... }:

{
  environment.profiles = lib.mkBefore [ "/run/current-system" ];
}
```

PR #28, opened 2017-05-12
Default outputs should include "man"

clhodapp:
If you don't include "man", basic nix commands (e.g. nix-env) basically have broken help/usage text, since they delegate to calling man.

### DIFF
--- a/modules/environment/default.nix
+++ b/modules/environment/default.nix
@@ -51,7 +51,7 @@ in {
 
     environment.extraOutputsToInstall = mkOption {
       type = types.listOf types.str;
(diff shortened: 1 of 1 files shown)


LnL7:
Thanks!

I remember wondering why that's not the default on NixOS. Turns out it's added by the [`programs.man` module](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/man.nix#L27), which enabled by default. 😄

LnL7:
Added the module in b6af291c84da5970d9db165c6cb2834455dc04e8.

Issue #9, opened 2017-01-12
disable nix-daemon?


sellout:
I “accidentally” enabled nix-daemon, and now (after disabling it) whenever I try `darwin-rebuild build`, it errors with “the group ‘nixbld’ specified in ‘build-users-group’ does not exist”

How can I get it to not try to start nix-daemon now?

yurrriq:
FWIW I've written and used [this script](https://github.com/yurrriq/dotfiles/blob/nix/nixbld-setup.sh) (modified from outdated NixOS docs) to set up the proper groups and users for `nix-daemon`.

yurrriq:
N.B. The hard-coded `gid=502` might need to be adjusted on your system.

LnL7:
If you remove `/etc/nix/nix.conf` and stop the daemon you can rebuild again.

sellout:
@LnL7 I’ve removed /etc/nix/nix.conf, stopped the daemon, and I still get the same error whenever I try `darwin-rebuild` or `nix-build '<darwin>' -A system --no-out-link` (as part of `$(nix-build '<darwin>' -A system --no-out-link)/sw/bin/darwin-rebuild build`).

I suppose I may as well use @yurrriq’s script and run the daemon anyway – my only reservation is basically having to remember to run that script whenever my nix config moves to a new machine.

LnL7:
I wrote this when I first installed the daemon on my machine https://gist.github.com/LnL7/ed6a23cd803fdb371e853fe970a848a3, I'm planning to include this in the repo or provide a proper installer sometime.

PR #30, opened 2017-05-31
Bootstrap script

cmacrae:
Figured I'd spend a little time this evening writing a bash script to bootstrap `nix-darwin`.  
It follows the steps from the README, except it uses `curl` and `unzip` to fetch the repo.  

Straightforward stuff, see what you think @LnL7 🙂

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -102,7 +102,6 @@ create_daemon_users(){
 	    echo -e "Creating user: "$YELLOW"nixbld${i}"$ESC"..."
 	    sudo /usr/sbin/sysadminctl -fullName "Nix build user $i" \
 		 -home /var/empty \
-		 -shell /sbin/nologin \
 		 -UID $(expr 30000 + $i) \
 		 -addUser nixbld$i >&2 \
 		|| exit_message "Problem creating user: nixbld${i}"
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -55,7 +55,7 @@ install(){
 		break
 		;;
 	    n|N)
-		echo "Proceeding without upgrading..."
+		echo "Proceeding without upgrade..."
 		break
 		;;
 	    *)
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -2,6 +2,38 @@
 set -e
 set -o pipefail
 
+# Argument parsing
+init(){
+    if [ $# -gt 0 ]; then
+	case "$1" in
+	    -h|--help)
+		usage
+		exit 1
+		;;
+	    -u|--create-daemon-users)
+		export CREATE_DAEMON_USERS=true
+		;;
+	    *)
+		echo -e "ERROR: Unrecognized parameter: '${1}'\nSee usage (--help) for options..." 1>&2
+		exit 1
+		;;
+	esac
+    fi
+}
+
+# Usage
+usage(){
+    cat <<EOF
+ Usage: bootstrap.sh [Options]
+
+ Options:
+
+   -h, --help: Show this message
+   -u, --create-daemon-users: Create users & group for use with the Nix daemon
+EOF
+}
+
+
 # Sanity checks
 sanity(){
     # Ensure the script is running on macOS
@@ -27,7 +59,7 @@ sanity(){
     # Check if nix-darwin is already present
     if [[ $(type darwin-rebuild 2>/dev/null) ]]; then
 	echo -e "It looks like "$YELLOW"nix-darwin"$ESC" is already installed..."
-	exit 1
+	[ ! -z $CREATE_DAEMON_USERS ] && create_daemon_users || exit 1
     fi
 }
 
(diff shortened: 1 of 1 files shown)


yurrriq:
See also:
- https://gist.github.com/LnL7/ed6a23cd803fdb371e853fe970a848a3
- https://github.com/yurrriq/dotfiles/blob/nix/nixbld-setup.sh
- Discussion re: setting up `nix-daemon` on #9

LnL7:
This is awesome!

cmacrae:
Cool @yurrriq - you guys reckon this stuff is worth adding to the script, too?  
Perhaps as a prompted question to the user initially, but maybe also as a script option, in case the user initially doesn't set this stuff up, but then later decides they want to?

yurrriq:
:+1: to the prompt and to the script option. I think the default should be not to set up `nix-dameon`, so long as it remains [disabled by default](https://github.com/LnL7/nix-darwin/blob/d684bbf2028f066bbeb471bfe9c26ab182348eb6/modules/services/nix-daemon.nix#L17).

yurrriq:
I would suggest borrowing heavier from [@Lnl7's script](https://gist.github.com/LnL7/ed6a23cd803fdb371e853fe970a848a3) than mine, as they do basically the same thing but it's more robust and generally in line with your script anyway. I think when I wrote mine I hadn't yet seen it.

cmacrae:
@yurrriq Nice one 🙂   
@LnL7 what d'you think? Should I add these functionalities?

LnL7:
Using the daemon currently breaks the nixops bootstrapping if you don't have distributed builds configured. But just creating the users doesn't hurt and makes it easy for people to try the daemon if they want.

bootstrap.sh
```diff
@@ -0,0 +1,133 @@
...
+
+    # Link run directory
+    echo "Setting up /run..."
+    test -L /run || sudo ln -s /private/var/run /run
+
+    # Fetch the nix-darwin repo as a zip (shouldn't assume presence of git)
+    REPO_DOWNLOAD=/tmp/nix-darwin-$(date +%Y%m%d).zip
+    if [ ! -f $REPO_DOWNLOAD ]; then
+	echo -e ""$YELLOW"Fetching nix-darwin repo..."$ESC""
+	curl -Lo $REPO_DOWNLOAD https://github.com/LnL7/nix-darwin/archive/master.zip || \
```
LnL7: the nix package includes `nix-prefetch-url`, with that it's possible to setup something similar to `nix-channel` for a git repo

bootstrap.sh
```diff
@@ -0,0 +1,133 @@
...
+    # Source generated bashrc
+    . /etc/static/bashrc
+
+    # Run first darwin-rebuild switch
+    echo -e "Running first "$YELLOW"darwin-rebuild switch"$ESC"..."
+    darwin-rebuild switch && echo -e ""$GREEN"Success!"$ESC"" || exit_message "Problem running darwin-rebuild switch"
+
+    echo -e ""$BLUE"You're all done!"$ESC""
+    echo -e "Take a look at "$YELLOW"~/.nixpkgs/darwin-configuration.nix"$ESC" to get started."
+    echo -e "See the README for more information: "$BLUE_UL"https://github.com/LnL7/nix-darwin/blob/master/README.md"$ESC""
```
LnL7: this should probably also print a warning and some suggestions on how to make sure the environment will get loaded, since `/etc/bashrc` won't be replaced by default

cmacrae: Sure, I'll add some more information :)

bootstrap.sh
cmacrae: Sorry, I'm not sure I follow - I'm pretty new to Nix, so you'll have to explain this one

LnL7: Yeah, let me make an example.

LnL7:
```
nix-env -p "/nix/var/nix/profiles/per-user/$USER/darwin" --set $(nix-prefetch-url --unpack --print-path --name nix-darwin-17.03 https://github.com/LnL7/nix-darwin/archive/master.tar.gz | grep nix-darwin)
ln -sfn "/nix/var/nix/profiles/per-user/$USER/darwin" "$HOME/.nix-defexpr/darwin"
```

if users have a setup like this I can add a `darwin-channel` command with update and rollback functionality like nix-channel/nix-env

cmacrae: Oh wow, that's really nice! 👌 
I'm making some changes to the script now, so I'll add and try this

yurrriq:
:+1: :100:

cmacrae:
Okay... so for some reason I'm running into this when bootstrapping now:
```
writing defaults...
ls: cannot access '/run/current-system/user/Library/LaunchAgents': No such file or directory
setting up ~/Applications...
warning: ~/Applications is a directory, skipping...
setting up /etc...
ls: cannot access '/run/current-system/Library/LaunchAgents': No such file or directory
ls: cannot access '/run/current-system/Library/LaunchDaemons': No such file or directory
configuring time...
```

Weird, because those directories are most certainly there. This is the end of the output of:
```
export NIX_PATH=darwin=$HOME/.nix-defexpr/darwin:darwin-config=$HOME/.nixpkgs/darwin-configuration.nix:$NIX_PATH
$(nix-build '<darwin>' -A system --no-out-link)/sw/bin/darwin-rebuild build
$(nix-build '<darwin>' -A system --no-out-link)/sw/bin/darwin-rebuild switch
```

Only started happening when I implemented the following in place of `curl` and `unzip`:
```
    nix-env -p "/nix/var/nix/profiles/per-user/$USER/darwin" \
	    --set $(nix-prefetch-url --unpack --print-path \
				     --name nix-darwin-17.03 \
				     https://github.com/LnL7/nix-darwin/archive/master.tar.gz \
			| grep nix-darwin)
    ln -sfn "/nix/var/nix/profiles/per-user/$USER/darwin" ~/.nix-defexpr/darwin
```

As far as I can see, everything looks good (with the repo download and the symlink).  
I've tried cleaning out my install completely several times, just to make sure there's nothing polluting the environment.  

I'll try putting the `curl` & `unzip` method back and see if the problem occurs.

cmacrae:
Hmmm... same thing now happening even with the old download method 😒

LnL7:
yeah, sounds like there's a symlink that's not right

cmacrae:
Well, I clearly messed something up 😅  
Got back to a working state. So I've added:  
- Nix daemon group/user creation support
- Introduced a flag for the above (in case users want to come back and do so)
- Use `nix-prefetch-url` to grab, unpack, and link the repository - hopefully in support for channels!

LnL7:
@cmacrae Any idea what happened? I might be able to make the activation scripts a bit more robust if I know what went wrong.

cmacrae:
@LnL7 Really not sure, I just ran the following again on its own after I was getting this failure:
```
$(nix-build '<darwin>' -A system --no-out-link)/sw/bin/darwin-rebuild build
$(nix-build '<darwin>' -A system --no-out-link)/sw/bin/darwin-rebuild switch
```
And it worked.  

Don't quote me on this, but I vaguely remember (when first trying `nix-darwin` out - before I started the boostrap) getting the same issue when first setting up. I could be mistaken though, if I do remember correctly it was one of those things where I completely disregarded it and simply re-ran my last command.  

Purely conjecture though - I guess if you want any sort of direction, it seems the linking in those scripts for launchd in particular would be where to start. As, in my testing, I'm quite sure I completely eradicated Nix from my system, and all it's symlinks (making use of `mdfind` to ensure there was nothing left over).

bootstrap.sh
LnL7: I'm not sure why it failed before when I tried it, but nix-channel also works.
```
nix-channel --add https://github.com/LnL7/nix-darwin/archive/master.tar.gz darwin
nix-channel --update
```

LnL7:
Oh, I thought the activation failed halfway through. I know what it is then.

cmacrae:
@LnL7 Oh, good - then I'm not going mad!

cmacrae:
Okay, it's good to get rid of that output 👍  But it was causing the script to exit, too.  
You can see, it would echo `configuring time...`, then it would just silently exit.  

I can probably reproduce it, if you need some help narrowing this down?

bootstrap.sh
```diff
@@ -0,0 +1,234 @@
...
+    cp "$HOME/.nix-defexpr/darwin/modules/examples/simple.nix" "$HOME/.nixpkgs/darwin-configuration.nix"
+
+    # Bootstrap build using default nix.nixPath
+    echo "Bootstrapping..."
+    export NIX_PATH=darwin=$HOME/.nix-defexpr/darwin:darwin-config=$HOME/.nixpkgs/darwin-configuration.nix:$NIX_PATH
+    $(nix-build '<darwin>' -A system --no-out-link)/sw/bin/darwin-rebuild build
+    $(nix-build '<darwin>' -A system --no-out-link)/sw/bin/darwin-rebuild switch
+
+    # Source generated bashrc
+    . /etc/static/bashrc
```
LnL7: This was causing problems in combination with `set -e`, I disabled completion for the example.

cmacrae: Ah okay, nice one 👍

cmacrae:
🎉

LnL7:
Updated the [README](https://github.com/LnL7/nix-darwin#install)

yurrriq:
:tada:

PR #31, opened 2017-06-14
Make symlink within ~/Applications if it exists instead of skipping

clhodapp:
This allows you to easily browse and spotlight-search Nix apps without having to give your entire `~/Applications` over to Nix.

### DIFF
--- a/modules/system/applications.nix
+++ b/modules/system/applications.nix
@@ -26,8 +26,10 @@ in
 
       if [ ! -e ~/Applications -o -L ~/Applications ]; then
         ln -sfn ${cfg.build.applications}/Applications ~/Applications
+      elif [ ! -e ~/Applications/Nix\ Apps -o -L ~/Applications/Nix\ Apps ]; then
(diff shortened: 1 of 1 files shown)


clhodapp:
Actually, a question here: Should this be happening at the user level or the system level? I kinda feel like we should be linking `/Applications/Nix Apps` to this Applications directory and `~/Applications)` (or `~/Applications/Nix Apps`) to the Applications directory in the user's profile.

LnL7:
Linking to `/Applications` is also possible, but then the activation script needs to be moved.

LnL7:
Sorry for the delay, Thanks!

clhodapp:
😄 No problem and thanks! I _may_ make another PR to link `~/Applications` (or `~/Applications/Nix Apps`) to applications from your user profile, `/Applications/Nix Apps` to system-level applications at some point but I think this PR is still a positive step.

LnL7:
Yeah, ideally it would link the individual apps directly to `/Applications` but that requires a little bit of logic to remove old links when a user removes a package from systemPackages.

PR #32, opened 2017-06-16
Preserve existing PATH in generated zshrc

jrolfs:
Currently sourcing the generated zshrc in my existing dotfiles breaks my `PATH`
environment variable. This change allows me to source the generated file and
preserve the existing `PATH` value.

### DIFF
--- a/modules/programs/zsh/default.nix
+++ b/modules/programs/zsh/default.nix
@@ -176,7 +176,7 @@ in
 
       setopt HIST_IGNORE_DUPS SHARE_HISTORY HIST_FCNTL_LOCK
 
-      export PATH=${config.environment.systemPath}
+      export PATH=${config.environment.systemPath}:$PATH
       ${config.system.build.setEnvironment}
       ${config.system.build.setAliases}
 


jrolfs:
Should I conditionally append this?

Also, I think in general the `zshrc` generated from the `zsh` package is a little egregious. Shouldn't it just set the needed environment variables etc. for Nix and not mess with any [other settings](https://github.com/jrolfs/nix-darwin/blob/e09264254b5ffa4417567b436383e4c0b13b02b3/modules/programs/zsh/default.nix#L173-L177)?

jrolfs:
So I realized that `/etc/static/zshrc` is really meant to stand on it's own. I was sourcing it from my already rather customized `.zshrc`. I went ahead and just added the environment variables into my configuration.

@LnL7 I would, however, love some feedback on how maintainable this is: https://github.com/jrolfs/macos/blob/master/home/.darwin.zsh#L4-L20

LnL7:
That works, however you loose some functionality like `environment.variables` .
I would recommend you load the nix-darwin zshrc first and set variables, etc... with your dot files afterwards.

Alternatively you could override the default zshrc in your configuration.nix like this, allowing you to keep support for whatever options you want to use.

```nix
{ config, lib, pkgs, ... }:
with lib;
{
  environment.etc."zshrc".text =
    let cfg = config.programs.zsh; in
    mkForce ''
      # /etc/zshrc: DO NOT EDIT -- this file has been generated automatically.
      # This file is read for interactive shells.

      bindkey -e
      # Only execute this file once per shell.
      if [ -n "$__ETC_ZSHRC_SOURCED" -o -n "$NOSYSZSHRC" ]; then return; fi
      __ETC_ZSHRC_SOURCED=1
      export PATH=${config.environment.systemPath}:''${PATH:+:$PATH}
      ${config.system.build.setEnvironment}
      ${config.system.build.setAliases}
      ${config.environment.extraInit}
      ${config.environment.interactiveShellInit}

      ${optionalString cfg.enableCompletion "autoload -U compinit && compinit"}
      ${optionalString cfg.enableBashCompletion "autoload -U bashcompinit && bashcompinit"}
      ${optionalString cfg.enableSyntaxHighlighting
        "source ${pkgs.zsh-syntax-highlighting}/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh"
      }

      # Tell zsh how to find installed completions
      for p in ''${(z)NIX_PROFILES}; do
        fpath+=($p/share/zsh/site-functions $p/share/zsh/$ZSH_VERSION/functions)
      done
    '';
}
```

jrolfs:
Wow, thanks for the response. This is super helpful!

PR #36, opened 2017-06-29
Adds support for ChunkWM

peel:
As KWM is now deprecated and ChunkWM is the thing to follow, I made myself a module for configuring it. 
The quirk however is that as of now [ChunkWM does not support custom config path](https://github.com/koekeishiya/chunkwm/issues/96), hence you either need to:
- make `/etc/{chunkwmrc,chunkwmtilingrc}` executable and symlink `/etc/{chunkwmrc,chunkwmtilingrc}` as $HOME dotfiles
- create executable wrappers in $HOME sourcing the above (as I'm using stow to manage dotfiles that's what I do for now)

Sidenote: I also do have ugly derivations for installing ChunkWM and plugins if you're interested

### DIFF
--- a/modules/services/chunkwm.nix
+++ b/modules/services/chunkwm.nix
@@ -1,0 +1,142 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.services.chunkwm;
+
+in
+
+{
+  options = {
+    services.chunkwm.enable = mkOption {
+      type = types.bool;
+      default = false;
+      description = "Whether to enable the chunkwm window manager.";
+    };
+
+    services.chunkwm.package = mkOption {
+      type = types.package;
+      default = pkgs.chunkwm-core;
+      defaultText = "pkgs.chunkwm-core";
+      description = "This option specifies the chunkwm package to use";
+    };
+
+    services.chunkwm.hotload = mkOption {
+      type = types.bool;
+      default = true;
+      description = "Whether to enable hotload";
+    };
+
+    services.chunkwm.port = mkOption {
+      type = types.int;
+      default = 3920;
+      description = "ChunkC port";
+    };
+
+    services.chunkwm.plugins.dir = mkOption {
+      type = types.path;
+      default = "/run/current-system/sw/lib/chunkwm/plugins";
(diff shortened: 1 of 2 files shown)


modules/services/chunkwm.nix
```diff
@@ -0,0 +1,142 @@
...
+
+    services.chunkwm.port = mkOption {
+      type = types.int;
+      default = 3920;
+      description = "ChunkC port";
+    };
+
+    services.chunkwm.plugins.dir = mkOption {
+      type = types.path;
+      default = "/run/current-system/sw/lib/chunkwm/plugins";
```
peel: not sure about the default though

modules/services/chunkwm.nix
```diff
@@ -0,0 +1,142 @@
...
+      description = "Chunkwm Plugins directory";
+    };
+
+    services.chunkwm.plugins.list = mkOption {
+      type = types.listOf (types.enum [ "border" "tiling" "ffm" ]);
+      default = [ "border" "tiling" "ffm" ];
```
LnL7: list types are appended by default, moving these defaults to config means that people can add plugins without loosing the defaults.

modules/services/chunkwm.nix
LnL7: nixos uses `/var/lib` for application data, but for store paths I think `share/chunkwm` makes more sense.

peel: Had some issues building chunkwm plugins and loading from share. Can we leave it for now?

LnL7: To clarify, I was talking about `/run/current-system/sw/share/chunkwm/plugins`. But it's fine if that doesn't work

peel: I'm following closely the work being done on the project. Will move those to /share once it's stable-ish.

LnL7:
Could you post the expression for chunkwm somewhere, then I can try it this out 😄 .

peel:
Sure, it's kinda ugly for now, but will sort it out as the development of the WM is going along nicely ;)
https://github.com/peel/dotfiles/tree/master/nix/.nixpkgs/pkgs/os-specific/darwin/chunkwm

PR #37, opened 2017-06-30
Adds parameter for emacs package binary

peel:
The motivation is that emacs binary is now well... remacs. Can alias it to emacs, but will impact readability when listing processes.


peel:
Closing for now. Will reopen.

PR #38, opened 2017-07-04
Adds service.emacs.exec to choose emacs binary to execute

peel:
The motivation of the change is that some forks (ie. Wilfred/remacs)
choose not to use the emacs binary. Whereas it is perfectly possible
to generate /bin/emacs in a drv, but I strongly believe it's cleaner
to have a service parameter.

### DIFF
--- a/modules/services/emacs.nix
+++ b/modules/services/emacs.nix
@@ -24,6 +24,11 @@ in
         description = "This option specifies the emacs package to use.";
       };
 
+      exec = mkOption {
+        type = types.string;
+        default = "emacs";
+        description = "Emacs command/binary to exeucte";
+      };
     };
   };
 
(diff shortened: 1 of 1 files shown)


modules/services/emacs.nix
```diff
@@ -24,18 +24,23 @@ in
         description = "This option specifies the emacs package to use.";
       };
 
+      exec = mkOption {
```
LnL7: I think a command option with the full path and a default of `"${cfg.package}/bin/emacs"` is perhaps a little nicer and more flexible.

peel: how about something like:

```nix
      command = mkOption {
        type = types.string;
        default = "bin/emacs";
        description = "Emacs command/binary to exeucte";
      };
```
and
```nix
        "${cfg.package}/${cfg.command}"
        "--daemon"
      ];
```

LnL7: n/m I think just the binary name makes more sense then.

The idea was that you'd only have to set one option. A better approach would be to add the command name to the package and do something like this.

```nix
  "${cfg.package}/${cfg.package.execName or cfg.exec}"
```

PR #39, opened 2017-07-05
Adds an idea for networking module

peel:
The idea is to follow: https://nixos.org/nixos/options.html#networking so we can
share even more configuration ideas from NixOS.

@LnL7 do you think this makes any sense?

### DIFF
--- a/modules/networking/default.nix
+++ b/modules/networking/default.nix
@@ -1,0 +1,41 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.networking;
+
+  hostName = optionalString (cfg.hostName != null) ''
+    scutil --set ComputerName "${cfg.hostName}"
+    scutil --set LocalHostName "${cfg.hostName}"
+    scutil --set HostName "${cfg.hostName}"
+  '';
+
+in
+
+{
+  options = {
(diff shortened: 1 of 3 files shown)


modules/networking/default.nix
```diff
@@ -0,0 +1,41 @@
...
+
+with lib;
+
+let
+
+  cfg = config.networking;
+
+  hostName = optionalString (cfg.hostName != null) ''
+    scutil --set ComputerName "${cfg.hostName}"
+    sudo scutil --set LocalHostName "${cfg.hostName}"
```
LnL7: No need for sudo, the `system.activationScripts` run as root. If the first command needs to run as the current user it should be included in the user activation.

LnL7:
It sounds like theres kind of a semantic difference between HostName and ComputerName/LocalHostName, perhaps those should be a different option?

peel:
Indeed there is some difference, that makes it kinda hectic, hence I don't think anyone really sets those values differently.

> Computer name: If you have file sharing turned on in Sharing preferences, users on your network can find your Mac listed by its computer name.
> Local hostname: Your computer’s local hostname, or local network name, is displayed on your local network so others on the network can connect to your Mac.

PR #50, opened 2017-10-07
Port over the NixOS fish module

cbarrett:
This PR essentially replaces this file with the nixos fish module. The only difference between the two now is nuking the reference to `environment.shells`  🎉 

The NixOS fish module has a number of nice features that we're gaining, such as more customizable initialization.

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -119,7 +119,7 @@ in
       # /etc/fish/config.fish: DO NOT EDIT -- this file has been generated automatically.
 
       # if we haven't sourced the general config, do it
-      if not set -q __fish_nixos_general_config_sourced
+      if not set -q __fish_nix_darwin_general_config_sourced
         set fish_function_path ${pkgs.fish-foreign-env}/share/fish-foreign-env/functions $fish_function_path
         fenv source /etc/fish/foreign-env/shellInit > /dev/null
         set -e fish_function_path[1]
@@ -128,11 +128,11 @@ in
 
         # and leave a note so we don't source this config section again from
         # this very shell (children will source the general config anew)
-        set -g __fish_nixos_general_config_sourced 1
+        set -g __fish_nix_darwin_general_config_sourced 1
       end
 
       # if we haven't sourced the login config, do it
-      status --is-login; and not set -q __fish_nixos_login_config_sourced
+      status --is-login; and not set -q __fish_nix_darwin_login_config_sourced
       and begin
         set fish_function_path ${pkgs.fish-foreign-env}/share/fish-foreign-env/functions $fish_function_path
         fenv source /etc/fish/foreign-env/loginShellInit > /dev/null
@@ -142,11 +142,11 @@ in
 
         # and leave a note so we don't source this config section again from
         # this very shell (children will source the general config anew)
-        set -g __fish_nixos_login_config_sourced 1
+        set -g __fish_nix_darwin_login_config_sourced 1
       end
 
       # if we haven't sourced the interactive config, do it
-      status --is-interactive; and not set -q __fish_nixos_interactive_config_sourced
+      status --is-interactive; and not set -q __fish_nix_darwin_interactive_config_sourced
       and begin
         ${fishAliases}
         
@@ -161,7 +161,7 @@ in
         # and leave a note so we don't source this config section again from
         # this very shell (children will source the general config anew,
         # allowing configuration changes in, e.g, aliases, to propagate)
-        set -g __fish_nixos_interactive_config_sourced 1
+        set -g __fish_nix_darwin_interactive_config_sourced 1
       end
     '';
 
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -1,40 +1,16 @@
-{ config, lib, pkgs, ...}:
+{ config, lib, pkgs, ... }:
 
 with lib;
 
 let
 
-  cfg = config.programs.fish;
   cfge = config.environment;
 
-  foreignEnv = pkgs.writeText "fish-foreign-env" ''
-    # TODO: environment.shellInit
-    ${config.system.build.setEnvironment.text}
-  '';
-
-  loginForeignEnv = pkgs.writeText "fish-login-foreign-env" ''
-    # TODO: environment.loginShellInit
-  '';
-
-  interactiveForeignEnv = pkgs.writeText "fish-interactive-foreign-env" ''
-    ${cfge.interactiveShellInit}
-  '';
-
-  shell = pkgs.runCommand pkgs.fish.name
-    { buildInputs = [ pkgs.makeWrapper ]; }
-    ''
-      source $stdenv/setup
-
-      mkdir -p $out/bin
-      makeWrapper ${pkgs.fish}/bin/fish $out/bin/fish
-    '';
+  cfg = config.programs.fish;
 
   fishAliases = concatStringsSep "\n" (
     mapAttrsFlatten (k: v: "alias ${k} '${v}'") cfg.shellAliases
   );
-
-  fishVariables =
-    mapAttrsToList (n: v: ''set -x ${n} "${v}"'') cfg.variables;
 
 in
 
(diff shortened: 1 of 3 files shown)


modules/programs/fish.nix
```diff
@@ -45,109 +21,163 @@ in
...
+    environment.etc."fish/foreign-env/interactiveShellInit".text = cfge.interactiveShellInit;
+    environment.etc."fish/foreign-env/extraInit".text= cfge.extraInit;
+
+    environment.etc."fish/nixos-env-preinit.fish".text = ''
+      # This happens before $__fish_datadir/config.fish sets fish_function_path, so it is currently
+      # unset. We set it and then completely erase it, leaving its configuration to $__fish_datadir/config.fish
+      set fish_function_path ${pkgs.fish-foreign-env}/share/fish-foreign-env/functions $__fish_datadir/functions
+     
+      # For "reasons not remembered," setEnvironment does not set PATH, so do it here
+      set PATH ${replaceStrings [":"] [" "] config.environment.systemPath}
```
LnL7: I think it was something related to ordering, but I guess it's actually useful in this case. How is the separator handled on nixos?

modules/programs/fish.nix
```diff
@@ -45,109 +21,163 @@ in
...
+    environment.etc."fish/nixos-env-preinit.fish".text = ''
+      # This happens before $__fish_datadir/config.fish sets fish_function_path, so it is currently
+      # unset. We set it and then completely erase it, leaving its configuration to $__fish_datadir/config.fish
+      set fish_function_path ${pkgs.fish-foreign-env}/share/fish-foreign-env/functions $__fish_datadir/functions
+     
+      # For "reasons not remembered," setEnvironment does not set PATH, so do it here
+      set PATH ${replaceStrings [":"] [" "] config.environment.systemPath} 
+
+      # source the NixOS environment config
+      fenv source ${config.system.build.setEnvironment}
```
LnL7: I inlined this in the other modules since nothing actually needed a separate file. But I can change this to a `writeText` and use the text attribute instead.

LnL7: See a8b47702751e5f53448c65ba1655747a88714445 and e8355cdd938c85b86024d22bd44ed7e3c939e17b, that should resolve most of differences with the nixos modules.

yurrriq:
Thanks for this. I’m reading the diff on my phone and it looks like you’ve done away with `variables` in the config. I use that in system config. Is there another approach now?

modules/programs/fish.nix
therealpxc: Since setting the PATH happens inside `fenv`'s temporary bash environment, the foreign-env plugin for _Oh My Fish!_ handles [converting the separator for us](https://github.com/oh-my-fish/plugin-foreign-env/blob/master/functions/fenv.apply.fish#L31). This applies just as well to nix-darwin since we're also using fenv. :-)

modules/programs/fish.nix
```diff
@@ -45,109 +21,163 @@ in
...
-      set PATH ${replaceStrings [":"] [" "] config.environment.systemPath}
-
-      ${config.system.build.setEnvironment}
-
-      fenv source ${foreignEnv}
-      ${cfg.shellInit}
-
-      ${concatStringsSep "\n" fishVariables}
+      # if we haven't sourced the general config, do it
+      if not set -q __fish_nixos_general_config_sourced
```
therealpxc: @LnL7 @cbarrett 

do we care that references to NixOS have now ended up in the module? Would it be worth adding a PR to Nixpkgs which changes these to generic names, perhaps substituting `nix` for `nixos` and makes a note that this path is depended-on in nix-darwin and potentially other declarative configuration/module systems for Nix and not just NixOS?

Is it worth it just for the name change? Is it worth it to leave a note to Nixpkgs/NixOS developers so they know to at least ping people should that change?

modules/programs/fish.nix
```diff
@@ -45,109 +21,163 @@ in
...
-    environment.pathsToLink = [ "/share/fish" ];
-
-    environment.loginShell = "${shell}/bin/fish -l";
-    environment.variables.SHELL = "${shell}/bin/fish";
+    environment.etc."fish/foreign-env/shellInit".text = cfge.shellInit;
+    environment.etc."fish/foreign-env/loginShellInit".text = cfge.loginShellInit;
+    environment.etc."fish/foreign-env/interactiveShellInit".text = cfge.interactiveShellInit;
+    environment.etc."fish/foreign-env/extraInit".text= cfge.extraInit;
+
+    environment.etc."fish/nixos-env-preinit.fish".text = ''
```
therealpxc: In the Fish package in Nixpkgs, Fish looks for this file, `/etc/fish/nixos-env-preinit.fish`. This is fine for us, but it means that we have to 'own' `/etc/fish` in `nix-darwin`. If we changed the Fish package to refer to `/etc/static/fish/nixos-env-preinit.fish` or something like that, this piece (and other pieces, if we changed them accordingly) would work even if `nix-darwin` can't or doesn't write directly to `/etc/fish`. Is that desirable? I don't think it would disrupt the way things work on NixOS.

tl;dr: Can we safely assume that `/etc/static` is where both nix-darwin and NixOS will write Fish config files to, and if so, should we change the modules (on both) to refer to those paths instead of the ones that live under `/etc`?

modules/programs/fish.nix
LnL7: Ah, interesting. I might change it to just set `environment.variables.PATH` at some point then.

therealpxc:
@yurrriq Good catch. I think maybe it would be best if we could use a global 'variables' attribute in config.environment and make sure that those variable changes are represented in `setEnvironment` or `extraInit`. (There might already be one; I haven't looked.)

I don't think that the foreign-env plugin currently applies aliases, but I think if we could enhance it to do so that would be ideal, because then we could put all of this stuff in generically as POSIX shell scripts and have `fenv` take care of all of it. Then setting variables can, for users, be independent of shell selection.

If this got through as it is now, you could set those variables with export directives inside `shellInit` or any of the other `*ShellInit` attributes in the your nix-darwin configuration.

I don't think you should have to, though. Being able to set aliases and variables directly, without worrying about the logic of when those things get sourced seems like it should be a feature. It looks like it's missing in the NixOS module, so I will seek to update that as well according to what we decide is a good idea here.

cbarrett:
Thanks for all the comments! I'll take a look at addressing them tonight hopefully, sometime this week for sure.

therealpxc:
Hey y'all. I [just submitted a PR for Fish in Nixpkgs](https://github.com/NixOS/nixpkgs/pull/30284) which makes it so that Fish installed via Nix takes care of NIxifying the environment automatically and transparently, regardless of whether you're using Fish on NixOS, Nix with GNU+Linux, Nix on macOS, or with nix-darwin.

It still needs more manual testing and probably other work (I think we should add some tests for initialization behavior as well), but it's ready for your comments. I'm using it with nix-darwin (a version similar to this PR (slightly modified and based on @LnL7 's recent updates) on this laptop and I'll be testing it on macOS without nix-darwin tonight, as well as on NixOS. I may not have a non-NixOS GNU+Linux machine online to test on until tomorrow.

cbarrett:
@therealpxc I agree that better alias support would be good. FWIW as of 2.5.0, fish seems to operate much closer to bash wrt. aliases. (e.g. typing `alias` will list all the aliases).

therealpxc:
So I hadn't got back to this, but the way this is currently set up, does it actually translates all the system-wide aliases correctly so that `fenv` is unnecessary for that part? I think I mistakenly thought the aliases in the module were a Fish-specific thing.

modules/programs/fish.nix
cbarrett: Good point. I'm down with changing the names to `nix_darwin` instead of `nixos`. How do others feel?

cbarrett:
Ah yeah, I forgot to bring back adding variables (thanks for the heads up @yurrriq)

cbarrett:
Looking into the test failure…

yurrriq:
I’ve amended my config so it’s moot for me, but @therealpxc and I have been working though reinstalling Nix and testing all this cleanly. I’ve borked my work machine now and it’s rebuilding the nix-darwin system manually (curl error 77 x a million) while at this show... here’s hoping it’s all good when I get home / by tomorrow morning..

yurrriq:
*Update*: Both my personal and work machines now have fully-functioning, fresh installs of Nix and nix-darwin and I can test out fish changes on Saturday.

therealpxc:
(If you try to test these changes and my Nixpkgs changes at the same time, your system will be broken. Eventually we will make sure that the changes in Nixpkgs are met with whatever changes are needed in nix-darwin to make things work, but we're not letting one PR hold up the other for now.)

LnL7:
@yurrriq Have you had a chance to test this with your configuration, did anything break?

yurrriq:
Last I checked it didn't break anything. FWIW I stopped using `programs.fish.variables`.

cbarrett:
@LnL7 This should be ready to go now!

yurrriq:
:tada: I’ll try this out again on my personal machine shortly.

yurrriq:
This has potentially introduced an issue on my system:
> set: Tried to change the read-only variable '_'

I'm trying to find the cause, but also doing a `nix-channel --update` now.

therealpxc:
This is a cosmetic issue and a side effect of using fenv with the
setEnvironment script. It's being addressed in our fenv patch in the
fish-foreign-env package in Nixpkgs. I think Colin recently opened a PR for
it.

cbarrett:
Correct. I've been busy with a class and have been neglecting Nix. I aim to make progress on this next week as I'm off from work.

PR #51, opened 2017-10-08
Set NetBIOS name with hostName

### LINKED ISSUE #47: Set hostname could also set NetBIOSName
In my script to set up a new Mac, I have `sudo defaults write /Library/Preferences/SystemConfiguration/com.apple.smb.server NetBIOSName -string $new_hostname`. I can't find anything saying whether this is still used in High Sierra (I assume it is?) but it does not produce an error. Could be a good thing to add at https://github.com/LnL7/nix-darwin/blob/ebc7f7b2dafea3fe4610baf2c3ef90811b193229/modules/networking/default.nix#L9-L13 ?

joseph-long:
Fixes #47

### DIFF
--- a/modules/networking/default.nix
+++ b/modules/networking/default.nix
@@ -10,6 +10,7 @@ let
     scutil --set ComputerName "${cfg.hostName}"
     scutil --set LocalHostName "${cfg.hostName}"
     scutil --set HostName "${cfg.hostName}"
+    defaults write /Library/Preferences/SystemConfiguration/com.apple.smb.server NetBIOSName -string "${cfg.hostName}"
   '';
 
 in


modules/networking/default.nix
```diff
@@ -10,6 +10,7 @@ let
     scutil --set ComputerName "${cfg.hostName}"
     scutil --set LocalHostName "${cfg.hostName}"
     scutil --set HostName "${cfg.hostName}"
+    defaults write /Library/Preferences/SystemConfiguration/com.apple.smb.server NetBIOSName -string "${cfg.hostName}"
```
LnL7: could you also expose this in the defaults module

joseph-long:
I'm worried that would create a race condition, depending on which gets applied first. Can you recommend a way to make sure there's only one set?

LnL7:
The networking option can also use defaults module instead, I just want avoiding something like what you're mentioning to happen in the future when new defaults are added.

PR #58, opened 2017-12-28
Adds attrs type for defaults. Adds additional Dock settings.

pjan:
As per the title. This adds a bunch more settings for the `Dock`, as well adds the option to pass in `attrs`, mainly there for the case when apple `Defaults` expect a `float`. I provided examples in the docs where this is used.

LMK. Additional PRs may ensue for other `defaults` currently managed in dotfiles on my end.

(sidenote: even though I'm based in 🇯🇵 , seems like we're compatriots, and you do live in my home town. Small world.)


modules/system/defaults/dock.nix
```diff
@@ -13,27 +13,79 @@ with lib;
       '';
     };
 
-    system.defaults.dock.orientation = mkOption {
-      type = types.nullOr (types.enum [ "bottom" "left" "right" ]);
+    system.defaults.dock.autohide-delay = mkOption {
+      type = types.nullOr types.attrs;
```
LnL7: I think this can be done in a nicer way with a custom type or an apply function. Currently this would require including the correct type when setting a value.

pjan: Got pointers to docs/example somewhere, so I can try to address it?

LnL7: Something like this should allow using `isType "float" x` for the values.

```nix
let
  float = mkOptionType {
    name = "float";
    description = "float";
    check = x: isInt x || (isString x && builtins.match "[0-9]*\.[0-9]+" x != null);
    merge = loc: defs: { _type = "float"; value = mergeEqualOption loc defs; };
  };
in
  # ...
```

pjan:
Updated.

I dropped the possibility to express a `float` with an int, because it could cause issues when writing the `defaults`; configuring a setting that requires a `float` with a value of e.g. `1` would the script have write the setting as an `int`.

LnL7:
Hmm, running into an issue when adding a test. The activation script ends up with `-string` for some reason.

```diff
commit 79bfe6f89563d3865bce405fc46482c0eb79a6bb (HEAD -> pr-58)
Author: Daiderd Jordan <daiderd@gmail.com>
Date:   Sun Dec 31 18:21:03 2017 +0100

    system-defaults: add test for float type

diff --git a/tests/system-defaults-write.nix b/tests/system-defaults-write.nix
index 3ba3ef5..60269f8 100644
--- a/tests/system-defaults-write.nix
+++ b/tests/system-defaults-write.nix
@@ -2,11 +2,13 @@

 {
   system.defaults.NSGlobalDomain.KeyRepeat = 1;
+  system.defaults.dock.autohide-delay = "0.12";
   system.defaults.dock.orientation = "left";

   test = ''
     echo checking defaults write in /activate-user >&2
     grep "defaults write -g 'KeyRepeat' -int 1" ${config.out}/activate-user
+    grep "defaults write com.apple.dock 'autohide-delay' -float '0.12'" ${config.out}/activate-user
     grep "defaults write com.apple.dock 'orientation' -string 'left'" ${config.out}/activate-user
   '';
 }
```

pjan:
hmpfff; stupid typo. Anyway, fixed & added the test to the PR as well. HNY

LnL7:
That was the problem, Thanks!

PR #61, opened 2018-01-02
Allow for postActivation scripts, that happen last

jwiegley:


LnL7:
Added an example here https://github.com/LnL7/nix-darwin/blob/8b10783bda1e2169b456218424b3ca61171474ef/modules/examples/hydra.nix#L79-L85

PR #62, opened 2018-01-04
Adds more NSGlobalDomain options + tests

pjan:
As per title.
- ordered them alphabetically for easier browsing
- added tests for each of them.

### DIFF
--- a/tests/system-defaults-write.nix
+++ b/tests/system-defaults-write.nix
@@ -1,13 +1,69 @@
 { config, pkgs, ... }:
 
 {
+  system.defaults.NSGlobalDomain.AppleFontSmoothing = 1;
+  system.defaults.NSGlobalDomain.AppleKeyboardUIMode = 3;
+  system.defaults.NSGlobalDomain.ApplePressAndHoldEnabled = true;
+  system.defaults.NSGlobalDomain.AppleShowAllExtensions = true;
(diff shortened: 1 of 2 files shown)


tests/system-defaults-write.nix
```diff
@@ -1,13 +1,69 @@
...
+  system.defaults.NSGlobalDomain."com.apple.trackpad.trackpadCornerClickBehavior" = 1;
+  system.defaults.NSGlobalDomain."com.apple.springing.enabled" = true;
+  system.defaults.NSGlobalDomain."com.apple.springing.delay" = "0.0";
+  system.defaults.NSGlobalDomain."com.apple.swipescrolldirection" = true;
   system.defaults.dock.autohide-delay = "0.24";
   system.defaults.dock.orientation = "left";
 
   test = ''
     echo checking defaults write in /activate-user >&2
+    grep "defaults write -g 'AppleFontSmoothing' -int 1" ${config.out}/activate-user
```
LnL7: This is probably a bit overkill, but evaluating everything with a value will help make sure nothing is broken.

pjan: this. names/capitalization are easy to get wrong, variable selection + regression testing for the future. Since they run so quickly and the output is well-defined and not in flux, I don't believe it's bad practice to cover as much ground as possible.

PR #63, opened 2018-01-04
darwin openvpn service

disassembler:


disassembler:
Rebased on master to fix merge conflicts. I've been using this for a month now with 3 different openvpn connections. I've had no issues.

disassembler:
Made change requested by @LnL7

peel:
I've been playing around with a proof of concept for wrapping pf into a networking.firewall module.
So the assumption here is that with a pf/ApplicationFirewall enabled the service should rely on GUI approvals for the firewall?

LnL7:
Could you add a basic test like https://github.com/LnL7/nix-darwin/blob/master/tests/services-skhd.nix

malob:
@domenkozar, there's been no activity on this PR in about 4 years, maybe it should just be closed?

PR #69, opened 2018-01-19
Menubar settings

pjan:
As per title.

### DIFF
--- a/modules/system/defaults-write.nix
+++ b/modules/system/defaults-write.nix
@@ -29,6 +29,7 @@ let
   screencapture = defaultsToList "com.apple.screencapture" cfg.screencapture;
   trackpad = defaultsToList "com.apple.AppleMultitouchTrackpad" cfg.trackpad;
   trackpadBluetooth = defaultsToList "com.apple.driver.AppleBluetoothMultitouch.trackpad" cfg.trackpad;
+  systemUIServer = defaultsToList "com.apple.systemuiserver" cfg.systemUIServer;
 
   mkIfAttrs = list: mkIf (any (attrs: attrs != {}) list);
 in
@@ -56,6 +57,11 @@ in
         ${concatStringsSep "\n" screencapture}
         ${concatStringsSep "\n" trackpad}
         ${concatStringsSep "\n" trackpadBluetooth}
+      '';
+
+    system.activationScripts.systemUIServer.text = mkIfAttrs [ systemUIServer ]
+      ''
+        killall SystemUIServer
       '';
 
   };
--- a/modules/system/defaults/systemUIServer.nix
+++ b/modules/system/defaults/systemUIServer.nix
@@ -1,9 +1,8 @@
-{ config, lib, pkgs, ... }:
+{ config, lib, ... }:
 
 with lib;
 
 let
-  cfg = config.system.menubar;
 
   items = {
     Airport         = "/System/Library/CoreServices/Menu Extras/AirPort.menu";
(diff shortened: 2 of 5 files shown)


pjan:
What can be done in a follow-up PR:
- enable/disable transparency in the menu
- whether to reveal IP address, hostname, OS version, etc. when clicking the clock in the menubar in the login window

cbarrett:
How does this affect (if at all) things using the public [NSStatusItem](https://developer.apple.com/documentation/appkit/nsstatusitem) api?

modules/system/menubar.nix
```diff
@@ -0,0 +1,52 @@
...
+    WWAN            = "/System/Library/CoreServices/Menu Extras/WWAN.menu";
+    iChat           = "/System/Library/CoreServices/Menu Extras/iChat.menu";
+  };
+
+in {
+  options = {
+    system.menubar.items = mkOption {
+      type = types.nullOr (types.listOf (types.enum (attrNames items)));
+      default = null;
+      description = "Items to display in the menubar";
```
LnL7: Could you add an example and preferably a description of the default value.

modules/system/menubar.nix
```diff
@@ -0,0 +1,52 @@
...
+    };
+  };
+
+  config = {
+
+    system.activationScripts.menubar.text = optionalString (!isNull cfg.items) ''
+      defaults write com.apple.systemuiserver menuExtras -array ${builtins.toString (map (k: "'" + (builtins.getAttr k items) + "'") cfg.items)}
```
LnL7: I would prefer to use the defaults module as the implementation for this.

pjan:
> How does this affect (if at all) things using the public NSStatusItem api?
It doesn't

> This looks like an option that should revert it's activation when unset.
What do you mean? What's the action point?

modules/system/menubar.nix
pjan: There's other menubar settings that can be set, not all fall under the `defaults` module, so I feel that would be more messy for end-users.

I have similar feelings with some of the existing settings in the `defaults` module; e.g. settings for "Printing" -> we now have the "show expanded dialog" option, but we could also set "automatic closing of the window when printing completed", which would then be in a different config namespace. I feel the api would be much cleaner if you would have all `printing` settings together. WDYT?

LnL7: To clarify I'm only referring to the activation scripts, not the `system.menubar` option. Modules can set options for other modules as their implementation. Something like this separates the details of defaults write from the rest of the logic.

```nix
# ...
{
  options = {
    system.menubar.items = mkOption {
      type = types.listOf (types.enum (attrValues itemPaths));
      default = [];
      example = [ "Airport" ];
    };
  };
  config = {
    system.defaults.systemuiserver.menuExtras = mkIf (cfg.items != []) (map (k: getAttr k itemPaths) cfg.items);
  };
}
```

LnL7:
Unlike services/etc, defaults don't change back to their original state when removed. To revert back you have to either rebuild with the default value mentioned in the description or use the ui.

I'm just wondering if that's ok for this.

modules/system/menubar.nix
LnL7: Does that make sense?

LnL7:
Ping

modules/system/defaults/systemUIServer.nix
```diff
@@ -0,0 +1,47 @@
...
+    WWAN            = "/System/Library/CoreServices/Menu Extras/WWAN.menu";
+    iChat           = "/System/Library/CoreServices/Menu Extras/iChat.menu";
+  };
+
+in {
+
+  options = {
+    system.defaults.systemUIServer.menuExtras = mkOption {
+      type = types.nullOr (types.listOf (types.enum (attrNames items)));
+      example = [ "Airport" "Battery" "Bleutooth" ];
```
LnL7: typo

modules/system/defaults-write.nix
```diff
@@ -58,5 +59,10 @@ in
         ${concatStringsSep "\n" trackpadBluetooth}
```
LnL7: `systemUIServer` entry is missing

LnL7:
What version of macOS are you using? Setting the defaults write seems to get overridden after restarting the SystemUIServer process.

PR #70, opened 2018-02-23
Add buildkite-agent module

rvl:
This module lets you use your Mac as a build slave for the Buildkite CI system.

The `buildkite-agent` package used will need to come from the unstable channel.

### DIFF
--- a/modules/services/buildkite-agent.nix
+++ b/modules/services/buildkite-agent.nix
@@ -1,0 +1,252 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.services.buildkite-agent;
+
+  mkHookOption = { name, description, example ? null }: {
+    inherit name;
+    value = mkOption {
+      default = null;
+      inherit description;
+      type = types.nullOr types.lines;
+    } // (if example == null then {} else { inherit example; });
+  };
+  mkHookOptions = hooks: listToAttrs (map mkHookOption hooks);
+
+  hooksDir = let
+    mkHookEntry = name: value: ''
+      cat > $out/${name} <<EOF
+      #! ${pkgs.stdenv.shell}
+      set -e
+      ${value}
+      EOF
+      chmod 755 $out/${name}
+    '';
+  in pkgs.runCommand "buildkite-agent-hooks" {} ''
+    mkdir $out
+    ${concatStringsSep "\n" (mapAttrsToList mkHookEntry (filterAttrs (n: v: v != null) cfg.hooks))}
+  '';
+
+in
+
+{
+  options = {
+    services.buildkite-agent = {
+      enable = mkEnableOption "buildkite-agent";
+
+      package = mkOption {
+        default = pkgs.buildkite-agent;
+        defaultText = "pkgs.buildkite-agent";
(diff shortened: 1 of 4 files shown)


modules/services/buildkite-agent.nix
```diff
@@ -0,0 +1,241 @@
...
+      };
+
+    environment.systemPackages = [ cfg.package ];
+
+    launchd.daemons.buildkite-agent =
+      {
+        path = cfg.runtimePackages ++ [ pkgs.coreutils cfg.package ];
+        environment = {
+          HOME = cfg.dataDir;
+          NIX_REMOTE = "daemon";
```
LnL7: This is a bit of a tricky one, but single user installs are also supported. I added `nix.useDaemon` to determine this, perhaps this should have a warning when that's the case?

modules/services/buildkite-agent.nix
```diff
@@ -0,0 +1,241 @@
...
+      ];
+    };
+  };
+
+  config = mkIf config.services.buildkite-agent.enable {
+    users.users.buildkite-agent =
+      { name = "buildkite-agent";
+        home = cfg.dataDir;
+        description = "Buildkite agent user";
+        uid = 532;
```
LnL7: Perhaps it's better to leave the ids up to the user, since we don't have any control over what's already in use.

modules/services/buildkite-agent.nix
```diff
@@ -0,0 +1,241 @@
...
+    mkdir $out
+    ${concatStringsSep "\n" (mapAttrsToList mkHookEntry (filterAttrs (n: v: v != null) cfg.hooks))}
+  '';
+
+in
+
+{
+  options = {
+    services.buildkite-agent = {
+      enable = mkEnableOption "buildkite-agent";
```
LnL7: I personally prefer to use fully qualified attributes like `services.buildkite-agent.enable` for most options because that makes searching easier,

modules/services/buildkite-agent.nix
```diff
@@ -0,0 +1,241 @@
...
+            build-path="${cfg.dataDir}/builds"
+            hooks-path="${hooksDir}"
+            ${cfg.extraConfig}
+            EOF
+
+            # Secrets exist in the buildkite-agent home directory
+            chmod 750 "${cfg.dataDir}"
+            chmod 640 "${cfg.dataDir}/buildkite-agent.cfg"
+
+            # Make /usr/bin/sw_vers accessible
```
LnL7: The `darwin.DarwinTools` package contains sw_vers.

rvl:
Thanks for the feedback, will fix it up.

rvl:
I have added a test case. Because this software requires access to the proprietary buildkite.com service, we are limited with what functionality can be tested.

On the topic of single user installs. This module would probably never work for that because it's a launchd Daemon not an Agent. Is that right?

LnL7:
A simple sanity check is fine, a full integration test would require something like the nixos test framework.

As for a single user install, nix won't work unless it's a user agent that runs as the same user. That's why I suggested a warning.

rvl:
Thanks!

PR #71, opened 2018-02-27
defaults: Add screencapture.location

cmacrae:
Very simple - I don't like cluttering my desktop 😉   
Tested and works like a charm: `defaults read com.apple.screencapture location` after customising it and rebuilding shows the new value. And, fortunately, works immediately: new screenshots go to the new specified location, without the need of a `killall SystemUIServer` (as some articles on the web would suggest is necessary).

### DIFF
--- a/modules/examples/lnl.nix
+++ b/modules/examples/lnl.nix
@@ -21,6 +21,8 @@
   system.defaults.finder.QuitMenuItem = true;
   system.defaults.finder.FXEnableExtensionChangeWarning = false;
 
+  system.defaults.screencapture.location = "/tmp";
+
   system.defaults.trackpad.Clicking = true;
   system.defaults.trackpad.TrackpadThreeFingerDrag = true;
 
--- a/modules/examples/lnl.nix
+++ b/modules/examples/lnl.nix
@@ -21,8 +21,6 @@
   system.defaults.finder.QuitMenuItem = true;
   system.defaults.finder.FXEnableExtensionChangeWarning = false;
 
-  system.defaults.screencapture.location = "/tmp";
-
   system.defaults.trackpad.Clicking = true;
   system.defaults.trackpad.TrackpadThreeFingerDrag = true;
 
--- a/modules/system/defaults/screencapture.nix
+++ b/modules/system/defaults/screencapture.nix
@@ -1,0 +1,16 @@
+{ config, lib, ... }:
+
+with lib;
+
+{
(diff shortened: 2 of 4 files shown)


modules/system/defaults/screencapture.nix
```diff
@@ -0,0 +1,16 @@
+{ config, lib, ... }:
+
+with lib;
+
+{
+  options = {
+
+    system.defaults.screencapture.location = mkOption {
+      type = types.string;
+      default = "~/Desktop";
```
LnL7: Defaults usually have a null value, in which case it's just ignored.

modules/examples/lnl.nix
```diff
@@ -21,6 +21,8 @@
   system.defaults.finder.QuitMenuItem = true;
   system.defaults.finder.FXEnableExtensionChangeWarning = false;
 
+  system.defaults.screencapture.location = "/tmp";
```
LnL7: I use this as a base for my configuration, and don't necessarily want  to set this. 😄

cmacrae: Oh... haha, sorry! I'll take this out 😉

modules/system/defaults/screencapture.nix
cmacrae: So I did initially use `null` here, but figured a default of `"~/Desktop"` was better.  
Reason being; if it's `null`, when a user decides they want to stop customising this property, and they comment out/remove their configuration for it, it will just remain the same as they had it before.  

Whereas, if it defaults to `"~/Desktop"`, when removed, it'll switch back to that - which is the default the OS comes with.  

Of course, it's up for debate if a user should expect this behaviour  - I just felt it was more natural, but happy to go with what you think is best.

LnL7: The problem is that we don't know if it still has the default value. I'm not sure how to handle this properly yet, but I don't want to somebody's settings to all get reset because of running nix-darwin with an empty config.

cmacrae: Sure, makes sense - I agree with that. I'll set a default of `null` then

LnL7: I don't really like the null values either, it makes rollback/removing values have unexpected behaviour from nix-darwin's perspective. Something like an opt-in 'strict' mode for defaults could be an easy solution for that.

cmacrae:
🎉

PR #73, opened 2018-03-03
adds skhd module

peel:
This adds [koekeishiya/skhd](https://github.com/koekeishiya/skhd) module. An alternative to already existing [khd](../tree/master/modules/services/khd) one that promises (and delivers) significant speedup. Currently we're not providing a default configuration for khd hence there is no default for skhd.

### DIFF
--- a/modules/services/skhd/default.nix
+++ b/modules/services/skhd/default.nix
@@ -30,8 +30,6 @@ in {
 
   config = mkIf cfg.enable {
 
-    security.accessibilityPrograms = [ "${cfg.package}/bin/skhd" ];
-
     environment.etc."skhdrc".text = cfg.skhdConfig;
 
     launchd.user.agents.skhd = {
--- a/modules/services/skhd/default.nix
+++ b/modules/services/skhd/default.nix
@@ -1,0 +1,49 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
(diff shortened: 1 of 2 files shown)


modules/services/skhd/default.nix
```diff
@@ -0,0 +1,49 @@
...
+      type = types.lines;
+      default = "";
+      example = "alt + shift - r   :   chunkc quit";
+      description = "Config to use for <filename>skhdrc</filename>.";
+    };
+  };
+
+  config = mkIf cfg.enable {
+
+    security.accessibilityPrograms = [ "${cfg.package}/bin/skhd" ];
```
LnL7: This was removed because of SIP.

modules/services/skhd/default.nix
```diff
@@ -0,0 +1,49 @@
...
+  options = {
+    services.skhd.enable = mkOption {
+      type = types.bool;
+      default = false;
+      description = "Whether to enable the skhd hotkey daemon.";
+    };
+
+    services.skhd.package = mkOption {
+      type = types.package;
+      example = literalExample "pkgs.skhd";
```
LnL7: Are you planning to make a pr for nixpkgs or should I push my expression?

peel: I guess eventually all the credit goes to you, so I could only be the messenger. I can do that though.

PR #75, opened 2018-04-25
buildkite-agent fixes

rvl:
Here are two fixes for buildkite-agent.

1. Port a shell quoting fix from NixOS/nixpkgs#38326
2. The daemon was starting before the hostname was set, which caused it to register itself as "localhost". This adds a fix for that.

Cheers

### DIFF
--- a/modules/services/buildkite-agent.nix
+++ b/modules/services/buildkite-agent.nix
@@ -17,7 +17,7 @@ let
 
   hooksDir = let
     mkHookEntry = name: value: ''
-      cat > $out/${name} <<EOF
+      cat > $out/${name} <<'EOF'
       #! ${pkgs.stdenv.shell}
       set -e
       ${value}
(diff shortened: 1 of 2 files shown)


modules/services/buildkite-agent.nix
```diff
@@ -17,7 +17,7 @@ let
 
   hooksDir = let
     mkHookEntry = name: value: ''
-      cat > $out/${name} <<EOF
+      cat > $out/${name} <<'EOF'
```
LnL7: Perhaps this should use `writeScript` to avoid other edge-cases like eg. `EOF`.

modules/services/buildkite-agent.nix
```diff
@@ -221,11 +221,20 @@ in
...
-          RunAtLoad = true;
           ProcessType = "Interactive";
           ThrottleInterval = 30;
 
+          # The combination of KeepAlive.NetworkState and WatchPaths
+          # will ensure that buildkite-agent is started on boot, but
+          # after networking is available (so the hostname is
+          # correct).
+          RunAtLoad = true;
+          KeepAlive.NetworkState = true;
```
LnL7: neat!

modules/services/buildkite-agent.nix
rvl: OK good point -- I have added a commit which fixes that.

PR #78, opened 2018-05-12
Extend plist serialization to support custom types

3noch:
Comments welcome.

Now you can do things like:

```nix
toPLIST {
  something = [ "a string" ];
  someNumber = types.real 3;
  someReal = types.real "3.1";
  somethingElse = types.custom "my-tag" "value";
}
```

which produces

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>someNumber</key>
	<real>3</real>
	<key>someReal</key>
	<real>3.1</real>
	<key>something</key>
	<array>
		<string>a string</string>
	</array>
	<key>somethingElse</key>
	<my-tag>value</my-tag>
</dict>
</plist>
```

### DIFF
--- a/modules/launchd/lib.nix
+++ b/modules/launchd/lib.nix
@@ -10,12 +10,12 @@ in
 
 rec {
 
-  toPLIST = x: ''
+  toPLIST = x: inPLISTDocument ("\n" + pprExpr "" x + "\n");
+
+  inPLISTDocument = x: ''
     <?xml version="1.0" encoding="UTF-8"?>
     <!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
-    <plist version="1.0">
-  '' + pprExpr "" x
-     + "\n</plist>";
+    <plist version="1.0">${x}</plist>'';
 
   pprExpr = ind: x:
     if isNull x then "" else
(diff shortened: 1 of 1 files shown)


cbarrett:
Very cool! Perhaps `toPropertyList` is a more appropriate name than all-caps? Not that familiar with the style used in nixpkgs

LnL7:
Nix has `toJSON`/`toXML` builtins and there already is `toINI` and `toYAML` in nixpkgs.

modules/launchd/lib.nix
```diff
@@ -10,48 +10,60 @@ in
...
   ]) x));
 
+  # Custom types not directly supported by nix.
+  #   Custom types are encoded as normal attrsets that contain a specially named key.
+  pprCustomType = ind: x:
+    let tag = x.${plistCustomTypeKey}; in
+    if tag == null then pprLiteral ind x.value else
+    pprTag tag ind (if isString x.value then x.value else toString x.value);
+
+  plistCustomTypeKey = "!__custom_type";  # a relatively unique attribute key to detect custom types
```
LnL7: Types in nix are usually represented as eg. `{ type = "derivation"; ... }`, maybe we could just use that. Since keys are generally capitalised so it probably won't cause issues and we could add `types.array` for special cases.

LnL7:
Sorry for the delay.

LnL7:
The PLIST helpers are in nixpkgs now so it's probably more appropriate to update it there.

PR #81, opened 2018-06-04
Init offlineimap service

periklis:
# Motivation
This PR provides a basic service module for `offlineimap`. Offlineimap can be heavily configured by cli parameters as well as through an rc file. The service module offers only basic startup options and lets the rest of the configuration to be done through an run control script `.offlineimaprc`, because account management and credentials are beyond the scope of nix-darwin afaik.

# Things done
- [x] Building with nix-darwin
- [x] Provides tests for service configuration
- [x] Smoke tests with a test confoguration

~**Disclaimer:** The tests and some local smoke tests are passing for this PR. However, i am away from one of my machines, where i make use of offlineimap. I can cover this by the end of June.~

### DIFF
--- a/modules/services/mail/offlineimap.nix
+++ b/modules/services/mail/offlineimap.nix
@@ -1,0 +1,63 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.services.offlineimap;
+in {
+
+  options.services.offlineimap = {
+    enable = mkEnableOption "Offlineimap, a software to dispose your mailbox(es) as a local Maildir(s).";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.offlineimap;
+      defaultText = "pkgs.offlineimap";
+      description = "Offlineimap derivation to use.";
+    };
+
+    path = mkOption {
+      type = types.listOf types.path;
+      default = [];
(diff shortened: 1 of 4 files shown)


modules/services/mail/offlineimap.nix
```diff
@@ -0,0 +1,63 @@
...
+    path = mkOption {
+      type = types.listOf types.path;
+      default = [];
+      example = literalExample "[ pkgs.pass pkgs.bash pkgs.notmuch ]";
+      description = "List of derivations to put in Offlineimap's path.";
+    };
+
+    startInterval = mkOption {
+      type = types.nullOr types.int;
+      default = null;
```
LnL7: Does it make sense for this to not run with some kind of default interval?

modules/services/mail/offlineimap.nix
```diff
@@ -0,0 +1,63 @@
...
+      example = literalExample "300";
+      description = "Optional key to start offlineimap services each N seconds";
+    };
+
+    runQuick = mkOption {
+      type = types.nullOr types.bool;
+      default = null;
```
LnL7: I would just make this a boolean

periklis:
Both comments were a good catch ;) Thx!

PR #83, opened 2018-06-16
synapse-bt: add synapse-bt service

marsam:
This adds [synapse-bt](https://synapse-bt.org/) service


modules/services/synapse-bt.nix
```diff
@@ -0,0 +1,74 @@
...
+    };
+  };
+
+  config = mkIf cfg.enable {
+
+    environment.systemPackages = [ cfg.package ];
+
+    launchd.user.agents.synapse-bt =
+      { path = [ cfg.package ];
+        script = ''
```
LnL7: nitpick, you can use command here

LnL7:
Could you add a basic test like https://github.com/LnL7/nix-darwin/blob/master/tests/services-skhd.nix

marsam:
Hi, sorry for the late response. I've added a basic test, but I'm not sure how to test the `configFile`

PR #84, opened 2018-06-21
[service/offlineimap]: Fix args passing to execvp

periklis:


modules/services/mail/offlineimap.nix
```diff
@@ -51,10 +51,9 @@ in {
     environment.etc."offlineimaprc".text = cfg.extraConfig;
     launchd.user.agents.offlineimap = {
       path                            = [ cfg.package ];
-      command                         = "offlineimap";
+      command                         = "${cfg.package}/bin/offlineimap -c /etc/offlineimaprc" + optionalString (cfg.runQuick) "-q";
```
LnL7: I think this needs to be `" -q"` otherwise the flag gets appended to the config file

PR #85, opened 2018-06-29
installer: Bring our own `nix-instantiate`

kirelagin:
This is needed for the case when the user already has their config on `NIX_PATH` but has no working Nix. (🤷‍♀️ Things happen!)


LnL7:
Thanks!

PR #87, opened 2018-06-29
installer and activation: Sanitise PATH

kirelagin:
This seems to work, although I didn’t really test it a lot.

### DIFF
--- a/pkgs/darwin-installer/default.nix
+++ b/pkgs/darwin-installer/default.nix
@@ -24,6 +24,9 @@ stdenv.mkDerivation {
   shellHook = ''
     set -e
 
+    orig_path="$PATH"
+    export PATH="${pkgs.openssh}/bin"  # In case nix needs it
+
     action=switch
     while [ "$#" -gt 0 ]; do
         i="$1"; shift 1
@@ -58,7 +61,7 @@ stdenv.mkDerivation {
         read -p "Would you like edit the default configuration.nix before starting? [y/n] " i
         case "$i" in
             y|Y)
-                ''${EDITOR:-nano} "$config"
+                PATH="$orig_path" ''${EDITOR:-nano} "$config"
                 ;;
         esac
     fi
--- a/modules/system/activation-scripts.nix
+++ b/modules/system/activation-scripts.nix
@@ -40,7 +40,7 @@ in
       #! ${stdenv.shell}
       set -e
       set -o pipefail
-      export PATH=${pkgs.coreutils}/bin:@out@/sw/bin:${config.environment.systemPath}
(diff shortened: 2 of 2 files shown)


pkgs/darwin-installer/default.nix
```diff
@@ -24,6 +24,9 @@ stdenv.mkDerivation {
   shellHook = ''
     set -e
 
+    orig_path="$PATH"
+    export PATH="${pkgs.openssh}/bin"  # In case nix needs it
```
LnL7: This clears path, also in what case is openssh used?

kirelagin: > This clears path

That’s the point. (The later commit add `/usr/bin` and `/bin`, but, honestly, I would prefer not to add them and have an explicit list of macOS binaries used instead.)

> in what case is openssh used?

With ssh substituters. I wonder whether Nix should be using `openssh` from nixpkgs, but currently substitution fails when `openssh` is not in `$PATH`, so, I suppose it invokes `ssh` from `$PATH`.

LnL7: I had an old tab open without the fixup commit.

An ssh substituter like remote builders? Only the nix-daemon uses ssh in that case AFAIK.

LnL7:
I didn't really think about this before but environment.systemPath also includes `/usr/local/bin` so programs installed with homebrew could potentially even influence activation and the installer.

> The later commit adds /usr/bin and /bin, but, honestly, I would prefer not to add them and have an explicit list of macOS binaries used instead.

I don't like hardcoding eg. `/usr/bin/grep` since those could in theory move. However that doesn't mean they have to be in the default PATH, using something like `PATH=/usr/bin:/bin:/usr/sbin:/sbin grep` (using an alias) also works while making the system dependency explicit.

pkgs/darwin-installer/default.nix
kirelagin: I only added this line because it was failing in my case 🤷‍♂️.
I don’t use nix-daemon, plain old single-user nix with one ssh substituter configure in addition to cache.nixos.org. It is not really a remote builder (my nix cannot offload a build to it), it is just a remote ssh-store from which it can pull some packages.

kirelagin:
> I don't like hardcoding

I was thinking about `alias grep=$(command -v grep)` before clearing `PATH`, but now I think I like your idea more, because, yes, we want to find the system `grep`, not some arbitrary Homebrew stuff are whatever wrappers the user can have on their `PATH`.

kirelagin:
Oh and we probably need to clean aliases in the beginning as well, just in case 🤔.Having something like this happen because of an alias would be insanely weird to debug.

kirelagin:
Would it work to simply use a pure `nix-shell` as an interpreter, so that it handles all the purifying for us? 🤔

PR #97, opened 2018-09-03
[services.emacs] use `fg-daemon` flag for service

peel:
As the default package is now emacs 26.1 and --daemon flag falls back to forking behaviour the recommended way is to use fg-daemon, see: https://lists.gnu.org/archive/html/emacs-devel/2016-11/msg00383.html, https://lists.gnu.org/archive/html/emacs-devel/2017-05/msg00861.html

### DIFF
--- a/modules/services/emacs.nix
+++ b/modules/services/emacs.nix
@@ -36,7 +36,7 @@ in
     launchd.user.agents.emacs = {
       serviceConfig.ProgramArguments = [
         "${cfg.package}/bin/${cfg.exec}"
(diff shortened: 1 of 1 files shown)


LnL7:
Does this flag work for the version that's in 18.03?

yurrriq:
Looks like no.

LnL7:
I'd prefer to wait with this until the 18.03 maintenance window then (~2months) in case people are using nixpgks-18.03-darwin.

PR #99, opened 2018-09-11
synergy: Add module for client and server

Enzime:


modules/services/synergy/default.nix
```diff
@@ -0,0 +1,103 @@
...
+
+      server = {
+        enable = mkOption {
+          default = false;
+          description = ''
+            Whether to enable the Synergy server (send keyboard and mouse events).
+          '';
+        };
+        configFile = mkOption {
+          default = "/etc/synergy-server.conf";
```
LnL7: Is there a reason not to manage this with nix?

modules/services/synergy/default.nix
```diff
@@ -0,0 +1,103 @@
...
+
+in
+
+{
+  options = {
+
+    services.synergy = {
+      client = {
+        enable = mkOption {
+          default = false;
```
LnL7: Missing a type, same with a bunch of other options. Most of those are probably `types.str`.

modules/services/synergy/default.nix
Enzime: Just trying to mirror the nixpkgs options

LnL7:
I fixed the nix 2.0 -> 2.1 installer issue.

PR #104, opened 2018-09-30
Add extraConfig for ssh module

luigy:


modules/programs/ssh/default.nix
```diff
@@ -81,11 +90,14 @@ in
...
-        concatStringsSep "," h.hostNames + " "
-        + (if h.publicKey != null then h.publicKey else readFile h.publicKeyFile)
-      )) + "\n";
+    environment.etc."ssh/ssh_known_hosts".text = ''
+      ${(flip (concatMapStringsSep "\n") knownHosts
+        (h: assert h.hostNames != [];
+          concatStringsSep "," h.hostNames + " "
+          + (if h.publicKey != null then h.publicKey else readFile h.publicKeyFile)
+        )) + "\n"}
+      ${cfg.extraConfig}
```
LnL7: I'm not sure how it's called in nixos, but this sounds a bit general given that it's only the known_hosts file.

luigy: I actually noticed this was missing while trying to reuse an internal module from one of my nixos machines. Here is what nixos ssh module looks like https://github.com/NixOS/nixpkgs/blob/83fd9785f65b8fd4cbaa146270f90b2a0d74b5f3/nixos/modules/programs/ssh.nix#L217

LnL7: Yeah, but that's `ssh_config` not `known_hosts`.

luigy: Whoops wrong config

luigy: moved to `ssh_config`

Enzime:
Thanks for the contribution, however I'm going to close this PR as it's quite old and I think the approach would require some rework. On macOS, `/etc/ssh/ssh_config` exists by default which means this PR in its current form would clash with the existing `/etc/ssh/ssh_config`. It also reads configuration from `/etc/ssh/ssh_config.d/*` which means users can already change SSH settings by creating files like so:

```
environment.etc."ssh/ssh_config.d/100-my-configuration.conf".text = ''
  Host *
    ForwardAgent yes
'';
```

Please open a new PR, if you're still interested in reworking this approach, possibly to write extra configurations to a file in `/etc/ssh/ssh_config.d` with the configurations, or possibly introducing a new option to forcibly override `/etc/ssh/ssh_config`.

PR #106, opened 2018-10-02
Remove Security link from README

thanegill:
Doesn't exist anymore


LnL7:
Yeah indeed, apple broke the module with SIP. 😕

PR #109, opened 2018-10-14
Do not evaluate /etc/bashrc on non-interactive shells

marsam:


modules/programs/bash/default.nix
```diff
@@ -52,6 +52,9 @@ in
       # /etc/bashrc: DO NOT EDIT -- this file has been generated automatically.
       # This file is read for interactive shells.
 
+      # If not running interactively, don't do anything
+      [[ $- != *i* ]] && return
```
LnL7: Breaking out here means the nix tools won't be in PATH over a non interactive ssh connection (eg. distributed builds) Basic things like environment variables should always be configured.

marsam: I realized it later, I've placed this early return below the basic setup https://github.com/LnL7/nix-darwin/commit/1a786eb

PR #114, opened 2018-11-07
nix module: don't disable the builders by default

zimbatm:
If nix.distributedBuilds is set then it means that /etc/nix/machines is
being managed by nix-darwin. Otherwise, leave that impure.

This will allow to fix https://github.com/nix-community/linuxkit-nix/issues/7

### DIFF
--- a/modules/nix/default.nix
+++ b/modules/nix/default.nix
@@ -41,9 +41,6 @@ let
           ''}
           trusted-users = ${toString cfg.trustedUsers}
           allowed-users = ${toString cfg.allowedUsers}
-          ${optionalString (isNix20 && !cfg.distributedBuilds) ''
-            builders =
-          ''}
           $extraOptions
           END
         '';


zimbatm:
Fair. This can also be worked around by providing the following module config:
```nix
{...}: {
nix.extraOptions = ''
   builders = "@/etc/nix/machines";
''
}
```
This will result in nix.conf having two entries but it seems to work.

LnL7:
I do mostly agree with the changes tho, `builders = @/etc/nix/machines` is the nix default so it's kind of strange to override that.

zimbatm:
It looks like Eelco introduced it in c550484c5e9d51f6844ab98fca97bf0f63f95d5f . I guess an argument could be made that on NixOS all the configuration should be pure. If the machines file was mutable it would have to be loaded from `/var/lib/nix/machines` or something like that (?)

PR #119, opened 2019-01-04
zsh: Add vendor-completions to fpath

kirelagin:
This was added to NixOS in NixOS/nixpkgs@003cd41310b5b7839eb4c402d84dc25068026c3e, but apparently never got ported to nix-darwin.

This change makes, at least, `stack` and `git-annex` completions work for me.


kirelagin:
I see, you are saying that this is merely a Debian convention. It looks like a good consistent prefix candidate, though.

PR #123, opened 2019-02-09
[WIP] fish: work around default path_helper shuffling PATH

### LINKED ISSUE #122: $PATH is broken for fish shell (and I think for bash, too)
I updated nix-darwin today and had to rollback to a version from October because my `$PATH` was broken (using fish shell).

Before (correct):
```
/Users/gape/.nix-profile/bin
/run/current-system/sw/bin
/nix/var/nix/profiles/default/bin
/usr/local/bin
/usr/bin
/bin
/usr/sbin
/sbin
```

After (broken):
```
/Users/gape/.local/bin
/Users/gape/.local/bin
/Users/gape/.nix-profile/bin
/usr/local/bin
/usr/bin
/bin
/usr/sbin
/sbin
/Users/gape/.nix-profile/bin
/run/current-system/sw/bin
/nix/var/nix/profiles/default/bin
```

The important part being that `/run/currenty-system/sw/bin` comes too late and the Nix provided binaries are not picked up correctly, resolving to `/usr/bin/git` for example. I'm also not sure where all this duplication is coming from …

I tried to get to the source of this error but was not able to figure it out exactly. What worked was to switch to an older generation and then to the newest generation again: in this case, the `$PATH` was set correctly. But when I exit fish shell and enter it again, the `$PATH` is broken.

It could be that this problem was introduced in https://github.com/LnL7/nix-darwin/commit/676ef103771aa3fc4b150290294b8ad5610d2750#diff-02a3bd02a5cdf5583b2e516a0e92d58a because the version of nix-darwin that worked for me was from 2018-10-17 and this is the major change that happened to the environment a few days later. But then maybe not because no one else has this problem?

thefloweringash:
Wrap the main fish config to undo its changes to PATH in
path_helper. It puts injected paths before other entries, which causes
nix-specific paths to appear after system paths.

Fixes #122.

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -115,10 +115,20 @@ in
 
       # clear fish_function_path so that it will be correctly set when we return to $__fish_datadir/config.fish
       set -e fish_function_path
+
+      # As part of fish's startup on darwin, it will apply the
+      # path_helper. This will unfortunately shuffle the contents of
+      # the PATH variable, that was either just set, or intended to be
+      # inherited.
+      set -g __fish_nix_darwin_preinit_path $PATH
     '';
 
     environment.etc."fish/config.fish".text = ''
       # /etc/fish/config.fish: DO NOT EDIT -- this file has been generated automatically.
+
+      # Restore the pre path_helper value of PATH
+      set -xg PATH $__fish_nix_darwin_preinit_path
+      set -e __fish_nix_darwin_preinit_path
 
       # if we haven't sourced the general config, do it
       if not set -q __fish_nix_darwin_general_config_sourced


thefloweringash:
It's a little disappointing to clobber the path_helper mechanism, but this is consitent with other shells under nix-darwin.

thefloweringash:
I think the nixpkgs initialization is inhibited in child shells, which will break this approach.

LnL7:
I'm not familiar with fish, how does this compare to the nixos module? https://github.com/NixOS/nixpkgs/blob/c941577dcb68ad413853ae744a458555b3209c21/nixos/modules/programs/fish.nix#L109-L170

thefloweringash:
The short answer is: this doesn't work. I don't know if I can fix it yet. Also I'm not familiar with fish either, first time looking at it trying to fix this!

The longer answer: it tries to work exactly the same as NixOS, but fish internally rewrites the `PATH` on every shell, so we're trying to undo that.

NixOS sets up environment once per shell tree, ignoring what kind of shell it is. This is a coarse but effective approach that means running commands directly from ssh work. Most shells are either launched with a clean environment for NixOS to setup, or launched as non-login shells.

The macOS `path_helper` isn't a problem for bash for two reasons:
  1. it only runs for login shells
  2. it runs before the nix-darwin module replaces the PATH.

(1) allows child shells to inherit paths, since they're typically non-login.
(2) allows nix-darwin to have the final say in the value of `PATH`

Neither of these are currently true in fish in nix-darwin. (1) was true, but was [reverted][], and (2) is a consequence of [nix injecting the environment early in fish's lifecycle][preinit]. I'm not sure which problems the early injection fixes. Maybe we could do environment configuration at the top of `/etc/fish/config.fish`, so as to run after the `path_helper`?

[reverted]: https://github.com/fish-shell/fish-shell/commit/c0f832a743ac9fba77443837f333e55988f3e863
[preinit]: https://github.com/NixOS/nixpkgs/commit/3f6d21bafc50d788d570a74dad957e792cea855d

thefloweringash:
I don't think this is the right approach. Sorry for noise.

PR #125, opened 2019-02-16
fonts: fix error trying to remove fonts from /System/Library/Fonts

jacereda:


LnL7:
The fonts module also has a few other problems, I'll spend some time on a rewrite and see where that gets.

LnL7:
With https://github.com/LnL7/nix-darwin/pull/126 only  disabled fonts in `/Library` are removed, could you test if that also fixes your problem.

PR #126, opened 2019-02-17
fonts: rewrite activation

### LINKED ISSUE #115: No fonts produced in the `/run/current-system/sw/share/fonts/` directory
No fonts end up in this directory, but I can see that the packages include the fonts. And then later when it is supposed to symlink these fonts it fails, e.g.:

```
resetting fonts...
updating fonts...
ln: failed to access '/run/current-system/sw/share/fonts/Arimo-Bold.ttf': No such file or directory
```

My configuration is very simple:
```
  fonts = {
    enableFontDir = true;
    fonts = [
      pkgs.iosevka-slab
      pkgs.noto-fonts
      pkgs.noto-fonts-extra
      pkgs.noto-fonts-cjk
    ];
  };
```

LnL7:
The new implementation is a bit smarter and only updates fonts that
changed.  But more importantly /run/current-system isn't used anymore
which breaks initial activation and installs the previous set of fonts
instead of the one in the new system.

Fixes #115

### DIFF
--- a/modules/fonts/default.nix
+++ b/modules/fonts/default.nix
@@ -1,57 +1,72 @@
 { config, lib, pkgs, ... }:
 
 with lib;
-with builtins;
 
 let
   cfg = config.fonts;
-  readDirsRec = path: let
-    home = readDir path;
-    list = mapAttrsToList (name: type:
-      let newPath = "${path}/${name}";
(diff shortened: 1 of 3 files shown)


jacereda:
Just tried it now, didn't get any error and the fonts installed properly.

LnL7:
Thanks for testing.

PR #128, opened 2019-02-21
users: create the home directory when creating a new user

kalbasit:


modules/users/default.nix
```diff
@@ -137,6 +137,8 @@ in
           dscl . -create '/Users/${v.name}' RealName '${v.description}'
           dscl . -create '/Users/${v.name}' NFSHomeDirectory '${v.home}'
           dscl . -create '/Users/${v.name}' UserShell '${v.shell}'
+          mkdir -p '${v.home}'
+          chown '${toString v.uid}:${toString v.gid}' '${v.home}'
```
LnL7: `/var/empty` is the home directory of multiple nixbld users and is not supposed to be writable in that case.

kalbasit: oh, that's right. I'll chown it conditionally on not `/var/empty`.

kalbasit: I've updated the PR and will mark it as ready for review!

LnL7: I would rather add an option for it instead, also take a look at `man 1 createhomedir`.

kalbasit: Done, PTAL.

PR #133, opened 2019-03-13
add mojave theme defaults

zachcoyle:


modules/system/defaults/NSGlobalDomain.nix
```diff
@@ -257,6 +257,49 @@ in {
...
+
+    system.defaults.NSGlobalDomain.AppleHighlightColor = mkOption {
+      type = types.nullOr (types.enum [
+        "0.847059 0.847059 0.862745 Graphite"
+        "1.000000 0.733333 0.721569 Red"
+        "1.000000 0.874510 0.701961 Orange"
+        "1.000000 0.937255 0.690196 Yellow"
+        "0.752941 0.964706 0.678431 Green"
+        "0.968627 0.831373 1.000000 Purple"
+        "1.000000 0.749020 0.823529 Pink"
```
LnL7: types.enum means these are the only "valid" options, is that correct?

zachcoyle: There is also the option to select from a color picker, which sets a value like `"0.549003 0.000000 0.180303 Other"`.

zachcoyle:
@LnL7 I removed the `system.defaults.NSGlobalDomain.AppleHighlightColor` setting with the freeform color picker.  I'll take a look at adding that one again when I get some time.

PR #135, opened 2019-03-25
postgresql: Add support for extraPlugins

marsam:
mostly copied from the NixOS module, tested with some plugins and seems to be working

### DIFF
--- a/modules/services/postgresql/default.nix
+++ b/modules/services/postgresql/default.nix
@@ -5,6 +5,25 @@ with lib;
 let
   cfg = config.services.postgresql;
 
+  postgresqlAndPlugins = pg:
+    if cfg.extraPlugins == [] then pg
+    else pkgs.buildEnv {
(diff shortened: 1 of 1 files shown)


modules/services/postgresql/default.nix
```diff
@@ -5,6 +5,27 @@ with lib;
...
+      name = "postgresql-and-plugins-${(builtins.parseDrvName pg.name).version}";
+      paths = [ pg pg.lib ] ++ cfg.extraPlugins;
+      # We include /bin to ensure the $out/bin directory is created which is
+      # needed because we'll be removing files from that directory in postBuild
+      # below.
+      pathsToLink = [ "/" "/bin" ];
+      buildInputs = [ pkgs.makeWrapper ];
+      postBuild =
+        ''
+          mkdir -p $out/bin
```
kalbasit: `$out/bin` is assumed to already exists per your comment above. It this necessary?

marsam: it wasn't necessary, fixed. Thanks!

Issue #143, opened 2019-05-29
Postgres databases default to SQL_ASCII


kevingriffin:
Since `initdb` has no environment to read from, it defaults to `SQL_ASCII` instead of the more common `UTF8`. A configuration option with a good default would go a long way.

PR #144, opened 2019-05-29
nix-darwin/postgresql: add character set config

kevingriffin:
`initdb` expects to read the locale or SQL_ASCII environment
variable to determine the most appropriate character set.
Instead, offer a configuration option.

Fixes https://github.com/LnL7/nix-darwin/issues/143


LnL7:
Thanks!

PR #147, opened 2019-06-24
port autossh service from nixos

eraserhd:
The options are identical, the only change is the use of `launchd.daemons` rather than the systemd configuration for starting sessions.  Well, that and `KeepAlive` is instructed to restart even on error exit, instead of just on success exit.  I'm not sure why you'd want to not restart on error.

It is working for me with [this config](https://github.com/eraserhd/dotfiles/blob/develop/macbook.nix#L23-L36).

### DIFF
--- a/modules/services/autossh.nix
+++ b/modules/services/autossh.nix
@@ -1,0 +1,106 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.services.autossh;
+
+in
+
+{
+
+  ###### interface
+
+  options = {
+
+    services.autossh = {
+
+      sessions = mkOption {
+        type = types.listOf (types.submodule {
+          options = {
+            name = mkOption {
+              type = types.string;
+              example = "socks-peer";
+              description = "Name of the local AutoSSH session";
+            };
+            user = mkOption {
+              type = types.string;
+              example = "bill";
+              description = "Name of the user the AutoSSH session should run as";
+            };
+            monitoringPort = mkOption {
+              type = types.int;
+              default = 0;
+              example = 20000;
+              description = ''
+                Port to be used by AutoSSH for peer monitoring. Note, that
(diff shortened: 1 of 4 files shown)


modules/services/autossh.nix
```diff
@@ -0,0 +1,106 @@
...
+
+      sessions = mkOption {
+        type = types.listOf (types.submodule {
+          options = {
+            name = mkOption {
+              type = types.string;
+              example = "socks-peer";
+              description = "Name of the local AutoSSH session";
+            };
+            user = mkOption {
```
LnL7: Are there usecases other then the login user? Launch agents are only started once the user logs in and run as that user instead of root or the explicitly configured user.

eraserhd:
That’s a tough one.  Everything I’m doing is attached to the user session,
but then I really use my Macbook as an expensive single-user terminal.

I can imagine wanting tunnels up before the user logs in.  Backup scripts,
for example?  If I had one of those cool Mac Pros, I’d use it more like a
server.

I thought it might be worthy of an option, but I wasn’t sure what to call
it.  `agent` = true?  `startAtLogin`?

eraserhd:
@LnL7 I've thought about this a bit, and I think using daemons instead of agents is the right answer.  My reasoning:

* Port numbers must be unique on a system level, and SSH tunnels provide an unauthenticated system-level service.  It's not possible to run one configured instance of autossh for multiple users simultaneously (because of ports), and it isn't useful anyway because any user can use the tunnel.
* The only reason I launched my own SSH tunnels as my local user was to deprivilege it.  In reality, this is bad, I should have created a separate user.
* It's definitely possible the tunnels are prerequisites for other system services to work.

If you agree, then I think this is good as-is?

LnL7:
Sure it also works like this on nixos and the user is required so root would be explicit. Just checking since agents would be easier to configure.

eraserhd:
@LnL7 oh dang, I'm trying to remember to proactively add tests to pull requests...  Test added now, commit amended.

PR #150, opened 2019-06-27
Replace etc files with links if they exist but backup first

### LINKED ISSUE #149: Allow enforcing linking etc when file exists
```
warning: not linking environment.etc."nix/nix.conf" because /etc/nix/nix.conf exists, skipping...
```

In some situations it's desireable to override etc files if they exist. For example when you install Nix and then nix-darwin, it won't take over nix.conf.

domenkozar:
Fixes #149

### DIFF
--- a/modules/system/etc.nix
+++ b/modules/system/etc.nix
@@ -48,14 +48,20 @@ in
 
       for f in $(find /etc/static/* -type l); do
         l=/etc/''${f#/etc/static/}
+        lbackup="$l.nix-darwin.bkp"
         d=''${l%/*}
         if [ ! -e "$d" ]; then
           mkdir -p "$d"
         fi
+        if [ -e "$lbackup" ]; then
+          echo "[1;31mwarning: Backup $lbackup still exists. Review/diff if it's still needed, make a backup and remove it.[0m" >&2
+        fi
         if [ -e "$l" ]; then
           if [ "$(readlink $l)" != "$f" ]; then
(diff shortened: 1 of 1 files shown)


LnL7:
I added some extra validation for nix.conf in 4bffc92, this should be relatively safe if the uninstaller restores the backup files.

domenkozar:
Thanks :)

domenkozar:
- [ ] move all backup files back to original location on uninstall

kclejeune:
@LnL7 Is this alright to merge? The installer in its current state doesn't have `darwin-rebuild` available in the path if you install nix-darwin immediately after installing nix.

domenkozar:
I've used this in production quite a few times and it works.

kclejeune:
> I've used this in production quite a few times and it works.

It looks good to me, this is pretty much the same process I did manually to get the install to work.  Does this also have the update to restore backed up files using the uninstaller?

kradalby:
I run into this with `/etc/shells` quite a bit, would this still be a candidate for merger?

kclejeune:
> I run into this with `/etc/shells` quite a bit, would this still be a candidate for merger?

This doesn't work in its current state anymore, but my work in #257 does. 

When I originally opened it, @LnL7 requested finer controls over which files are backed up, but I didn't have time to fix that at the time. I can revisit it to fix that and open the PR again.

kradalby:
@kclejeune that would be great 👍

Issue #139, opened 2019-05-05
applications are only linked for system packages, not user packages


BrianHicks:
When I do this:

```nix
{ users.users.brianhicks.packages = [ pkgs.emacsMacport ]; }
```

I expect `Emacs.app` to be available somewhere—maybe in `~/Applications/Nix Apps`. But it's not! However, if I do this:

```nix
{ environment.systemPackages = [ pkgs.emacsMacport ]; }
```

Then `~/Applications/Nix Apps/Emacs.app` gets created.

This is important to me because I manage my packages through home-manager, where I create a big ol' bundle of emacs packages all together. I'd rather not have to move all that configuration out of home-manager just to use graphical Emacs!

According to @rvolosatovs in https://github.com/rycee/home-manager/pull/702#issuecomment-489431636, it's because these lines specifically (and only) include the system packages:

https://github.com/LnL7/nix-darwin/blob/3ba909eb2dd3ddd893604f83048fa76cc2e47e74/modules/system/applications.nix#L17-L21

BrianHicks:
@LnL7 I don't have a great idea of how hard this would be to fix. Do you think it would be something someone newish (read: me) could take on? Would a patch for this behavior be welcome?

LnL7:
The main problem is that there's no link between the user running darwin-rebuild and users entries. It's probably better to link to link per user applications instead and move system packages to `/Applications`.

BrianHicks:
I'm having some trouble understanding what you're saying about the running user. Why does this matter? Shouldn't stuff end up in `/$USER_HOME/Applications` regardless of who is running it?

Regardless, is that a change you're open to? I'm happy to look into it but I would rather not waste everyone's time by making something that would be doomed before I started 😁

BrianHicks:
OK, I've taken more time to think about it today, and just want to check my understanding. I think we'd need to:

1. change https://github.com/LnL7/nix-darwin/blob/3ba909eb2dd3ddd893604f83048fa76cc2e47e74/modules/system/applications.nix to symlink into `/Applications/Nix Apps` instead of `~/Applications` or `~/Applications/Nix Apps`. 
2. Figure out some way of removing the old symlinks in `~`.
[...]

Issue #96, opened 2018-08-25
Tracking issue for switching to system-wide activation


LnL7:
- [x] use `SUDO_USER` instead of 2 separate scripts https://github.com/LnL7/nix-darwin/pull/1341
- [ ] move `launchd.user.agents` to `users.users.<user>.launchAgents`
- [x] move ~/Applications, doesn't make much sense to put this in a home directory https://github.com/LnL7/nix-darwin/pull/470
- [ ] move user-level defaults to `users.users.<user>.defaults`

yurrriq:
When you get a minute, can you briefly describe the plan for ~/Applications and `defaults write`? I'm quite curious, especially about the latter.

LnL7:
I just listed everything here that runs as the login user. For applications it's easy, that can be moved to /Applications. As for defaults, I'm not sure if that works when running as root.

emilazy:
It would be convenient for me if this distinction went away, especially if https://github.com/LnL7/nix-darwin/pull/664 lands. I assume from the mention of `$SUDO_USER` that the idea is to `sudo` the whole activation script and then `sudo` back to the normal user for anything that needs to be done per-user? That would be better for me than running the script as a user and `sudo`ing for root (because I have no `sudo` timeout), but I wonder if there are any other pitfalls. For example what about the `org.nixos.activate-system` daemon, which presumably runs outside the context of any normal user session and won't have a `$SUDO_USER`?

In general it kind of seems like nix-darwin doesn't know whether it wants to be single- or multi-user... If per-user settings were scoped appropriately and could be done for multiple accounts then this wouldn't be a consideration. But I bet macOS might make that complicated in other ways.

emilazy:
I'd be interested in taking a look at this if the question around user stuff was addressed. Should we just have configurations specify what their primary username is so the activation script can `sudo` to them?

mjm:
To me it me, as a user of both nix-darwin and home-manager, it makes sense to let nix-darwin handle system-level things and leave user-specific things to home-manager. I personally find it really confusing, for instance, that nix-darwin supports setting a bunch of defaults under `system.defaults` and some of them apply to the whole system and others are only for one user (I guess the one running `darwin-rebuild`).

home-manager already supports setting arbitrary user defaults for the user, setting up launch daemons and launch agents, and installing applications under ~/Applications. So nix-darwin can likely get away with dropping those kinds of things.

[...]

PR #153, opened 2019-07-19
System packages to applications

BrianHicks:
this moves system packages to `/Applications` and user applications to `~/Applications` or `~/Applications/Nix Apps`.

resolves one bullet of #96, #139.

### DIFF
--- a/modules/system/applications.nix
+++ b/modules/system/applications.nix
@@ -6,6 +6,17 @@ let
 
   cfg = config.system;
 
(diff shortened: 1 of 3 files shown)


BrianHicks:
@LnL7 with 76356b6, this is ready for your review anytime. 😁

BrianHicks:
just noting that while I still think this is valuable, I switched away from using Emcas. Since that was my motivation for making this change, I don't need it anymore. I'm going to close this PR. If someone else wants to move it forward, they can feel free to use my code as a starting point 😄

PR #156, opened 2019-08-22
modules: adding gitlab service

mmahut:
##### Adding the gitlab-runner service module

This module manages the gitlab-runner instances that is used to control runners within Gitlab environment. 

Sadly, there is no way to write a test, as it would be an unpure integration test, given the service won't work without connecting to gitlab.com (or internal Gitlab if any). However, we are running this service in our production environment, so I'm confided it's tested and will be for sure improved over time :)


LnL7:
All the tests except for the installer are basically unit tests. While you can't validate the service it can check some basic logic and makes sure the implementation still evaluates when refactoring, etc.

mmahut:
@LnL7 do you have any example in mind? The gitlab runner service fails to startup is it cannot reach the master node, therefore is unpure in this case.

zimbatm:
I tried to finish this PR. See https://github.com/zimbatm/nix-darwin/commit/48e8022f97eccc2461415c60083c8e8cb6fedb38

Technically it works but the shell executor doesn't load the user profile, which means Nix is not available to that user, which is really the only thing I wanted. When the service switches into the user it runs something like `su gitlab-runner bash --login`, which doesn't load the user profile for some reason.

Another small issue is that the gitlab-runner seems adamant on running `/usr/bin/git` instead of using git from the PATH, which forces the installation of the XCode tools.

In the end I gave up and just followed the official gitlab instructions instead.

PR #159, opened 2019-09-01
modules: remove usages of types.string

marsam:
nixpkgs recently properly deprecated `types.string`: https://github.com/NixOS/nixpkgs/pull/66346


LnL7:
Thanks! I did this once before in fa3f679, but looks like some new cases slipped in since then.

PR #163, opened 2019-09-15
Don't run static bashrc only for pure nix-shell

kubek2k:


LnL7:
This was added as a workaround in the first place, do you have any idea why this doesn't seem necessary on linux?

kubek2k:
> This was added as a workaround in the first place, do you have any idea why this doesn't seem necessary on linux?

I have no idea - since I don't use NixOS (or nix on any other linux). Maybe it could be a versioning issue (bash v2.7 on mac as default, as oppposed to v4 everywhere else?)

kubek2k:
@LnL7 the nix-shell man states that the `IN_NIX_SHELL` started to be set since nix 2.0. Maybe that was the cause?

hagl:
Is there a chance to get this PR merged? 
I use `programs.bash.interactiveShellInit` in my `darwin-configuration.nix` to setup a shell prompt containing the git branch, but that doesn't work in `nix-shell`s because of this early return.

modules/programs/bash/default.nix
```diff
@@ -59,7 +59,7 @@ in
       __ETC_BASHRC_SOURCED=1
 
       # Don't execute this file when running in a pure nix-shell.
-      if test -n "$IN_NIX_SHELL"; then return; fi
+      if [ "$IN_NIX_SHELL" == "pure" ]; then return; fi
```
Enzime, line 62:
```suggestion
      if [ "$IN_NIX_SHELL" = "pure" ]; then return; fi
```

Enzime (changes requested):
Thanks for the contribution, can you rebase your PR?

Enzime:
@kubek2k can you rebase the PR? I can't merge until all the tests have succeeded

Enzime (approved):
LGTM

PR #167, opened 2019-10-17
Fix path to darwin-config

lunaticare:
Original version leads to an error:
```
$(nix-build '<darwin>' -A system --no-out-link)/sw/bin/darwin-rebuild build
warning: Nix search path entry 'darwin-config=/Users/test/.nixpkgs/darwin-configuration.nix' does not exist, ignoring
error: file 'darwin-config' was not found in the Nix search path (add it using $NIX_PATH or -I), at /nix/store/z7469wjnpii0ybkfs2ny1lh4vw308n3k-darwin/darwin/default.nix:1:40
(use '--show-trace' to show detailed location information)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ sudo ln -s private/var/run /run
 # Configure the channel
 nix-channel --add https://github.com/LnL7/nix-darwin/archive/master.tar.gz darwin
 nix-channel --update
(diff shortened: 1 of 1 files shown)


tricktron:
Ah nice, had something on the back of my mind that I wanted to fix too. Great that you already took care of it :+1:

PR #172, opened 2019-10-30
[bootstrap.sh] Moves build & rebuild to after user creation

treffynnon:
If the users do not already exist then installers who already have a configuration containing

```nix
{
  services.nix-daemon.enable = false;
}
```

will not be able to complete an initial bootstrap on a clean machine. This is because the bootstrap script currently only creates users after an initial rebuild.

The users must exist before the rebuild is completed or the whole thing fails.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -174,18 +174,6 @@ install(){
       fi
     fi
 
-    # Bootstrap build using default nix.nixPath
-    echo "Bootstrapping..."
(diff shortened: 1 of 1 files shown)


treffynnon:
No reason that is really pertinent to nix-darwin. I was writing an
uninstall script and wanted a quick way to create the build users so I
could test that my script deleted them properly. My setup script doesn't
use the bootstrap script so if you plan to delete/deprecate the bootstrap
it won't be a problem for me.

PR #182, opened 2020-01-14
Remove trailing slash from path in commitIdFromGitRepo call

wildsebastian:
A change of `commitIdFromGitRepo` in nixpkgs/lib/sources.nix
resulted in the error message
"attribute '.git' missing at nixpkgs/lib/sources.nix:7:32" when
executing darwin-rebuils switch.
The commit with the change is
https://github.com/NixOS/nixpkgs/commit/c9214c394b248e1f26e45dbe1be2bd82363af3a6

Removing the trailing slash resolves the error.

### DIFF
--- a/modules/system/version.nix
+++ b/modules/system/version.nix
@@ -8,8 +8,8 @@ let
   defaultStateVersion = options.system.stateVersion.default;
 
   parseGit = path:
-    if pathExists "${path}/.git/" then rec {
-      rev = commitIdFromGitRepo "${path}/.git/";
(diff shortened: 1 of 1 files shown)


modules/system/version.nix
```diff
@@ -9,7 +9,7 @@ let
 
   parseGit = path:
     if pathExists "${path}/.git/" then rec {
```
LnL7: perhaps this one should also be removed then?

wildsebastian: Good point. I updated the PR.

PR #169, opened 2019-10-22
fonts: Adds support if Nix is on another filesystem.

tricktron:
On Catalina, the default way is to install Nix on a new volume which breaks hardlinking the font files. If that is the case I just copy them.

PR #183, opened 2020-01-25
fonts: copy instead of hardlink for Catalina support

kalbasit:
I'm not able to activate my configuration when I install any font because it's not possible to create a hard link between volumes.


kalbasit:
@LnL7 did you have time to check this change?

uosis:
This will break "if not exists" check because copied files will always have different inodes. Activation will be very slow with large number of fonts because copy will always happen.

I am just [doing rsync](https://github.com/uosis/nix-darwin/commit/233120815c5c48862d4a29716677242a1d808cf3) for now. That has its own problem of not preserving any non-nix files in /Library/Fonts, but that folder is empty on Catalina, so it doesn't matter.

tricktron:
I already tried to fix this in pull request #169 using `rsync` but still no answer from @LnL7.

PR #187, opened 2020-04-06
programs.ssh: write ssh known_hosts only if there are any set

kalbasit:
I noticed that I'm getting an empty `/etc/ssh/ssh_known_hosts`, and it seems that we were not gating its config.

### DIFF
--- a/modules/programs/ssh/default.nix
+++ b/modules/programs/ssh/default.nix
@@ -151,12 +151,13 @@ in
     services.openssh.authorizedKeysFiles = [ "%h/.ssh/authorized_keys" "/etc/ssh/authorized_keys.d/%u" ];
 
     environment.etc = authKeysFiles //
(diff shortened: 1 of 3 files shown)


LnL7:
Does this still get removed correctly when unset?

modules/programs/ssh/default.nix
```diff
@@ -73,7 +73,7 @@ in
     };
   };
 
-  config = {
+  config = mkIf (builtins.length knownHosts > 0) {
```
Enzime, line 76: Thanks for the contribution, could you rebase this PR and move this check to just affect `/etc/ssh/ssh_known_hosts`?

Enzime (changes requested):

modules/programs/ssh/default.nix
kalbasit: I'm sorry I dropped the ball on this one :( I've updated my PR and included a test.

Enzime (approved):
LGTM

PR #191, opened 2020-04-14
environment: add currentSystemPath option for /run/current-system

kalbasit:
By adding this option, it allows multiple instances of nix-darwin to setup the machine.

At work, I'm adding nix-darwin as a submodule of our module system so we can install daemons for the user. However, it's not currently possible as it conflicts with the user installed nix-darwin.

### DIFF
--- a/modules/environment/default.nix
+++ b/modules/environment/default.nix
@@ -37,6 +37,15 @@ in
       apply = x: if isList x then makeDrvBinPath x else x;
     };
 
+    environment.currentSystemPath = mkOption {
+      type = types.either types.path types.str;
+      description = ''
+        The location where to expose the nix-darwin system.
+      '';
+      default = /run/current-system;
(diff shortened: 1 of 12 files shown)


LnL7:
I'm not sure this is a good idea, many things in nix-darwin like etc are still global and thus still conflict. Also I think some options use the prefix to determine what to deactivate so changing the prefix of an existing installation would break assumptions there.

Enzime:
@kalbasit thank you for your contribution, however I'm going to close this PR. I think this is the wrong approach as `nix-darwin` runs at the system level and I don't think any computer should have multiple instances of it, I believe if you're interested in running local daemons for multiple users, I would recommend using `home-manager` which can create launchd agents via `launchd.agents`.

https://nix-community.github.io/home-manager/options.html#opt-launchd.agents

PR #176, opened 2019-11-11
Add support for yabai service

jacereda:

PR #196, opened 2020-05-06
module: add yabai service

cmacrae:
Hi again @LnL7  👋   

Here's a new service for your consideration :) I've been using this module personally for quite some time. Since you merged my yabai package PR the last week, it looks like it's now generally available in the `nixpkgs-unstable` channel - so I figured I'd get this submitted.  

This supersedes #176 

Let me know if there's any changes you want to see

### DIFF
--- a/modules/services/yabai/default.nix
+++ b/modules/services/yabai/default.nix
@@ -1,0 +1,103 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.services.yabai;
+
+  toYabaiConfig = opts:
+    concatStringsSep "\n" (mapAttrsToList
+      (p: v: "yabai -m config ${p} ${toString v}") opts);
+
+  configFile = mkIf (cfg.config != {} || cfg.extraConfig != "")
+    "${pkgs.writeScript "yabairc" (
+      (if (cfg.config != {})
+       then "${toYabaiConfig cfg.config}"
+       else "")
+      + optionalString (cfg.extraConfig != "") cfg.extraConfig)}";
+in
+
+{
+  options = with types; {
+    services.yabai.enable = mkOption {
+      type = bool;
+      default = false;
+      description = "Whether to enable the yabai window manager.";
+    };
+
+    services.yabai.package = mkOption {
+      type = path;
+      description = "The yabai package to use.";
+    };
+
+    services.yabai.enableScriptingAddition = mkOption {
+      type = bool;
+      default = false;
+      description = ''
+        Whether to enable yabai's scripting-addition.
+        SIP must be disabled for this to work.
+      '';
+    };
+
+    services.yabai.config = mkOption {
+      type = attrs;
+      default = {};
+      example = literalExample ''
+        {
+          focus_follows_mouse = "autoraise";
+          mouse_follows_focus = "off";
+          window_placement    = "second_child";
+          window_opacity      = "off";
+          top_padding         = 36;
+          bottom_padding      = 10;
+          left_padding        = 10;
+          right_padding       = 10;
+          window_gap          = 10;
+        }
+      '';
+      description = ''
+        Key/Value pairs to pass to yabai's 'config' domain, via the configuration file.
+      '';
+    };
+
+    services.yabai.extraConfig = mkOption {
+      type = str;
+      default = "";
+      example = literalExample ''
+        yabai -m rule --add app='System Preferences' manage=off
+      '';
+      description = "Extra arbitrary configuration to append to the configuration file";
+    };
+  };
+
+  config = mkMerge [
+    (mkIf (cfg.enable) {
+      environment.systemPackages = [ cfg.package ];
+
+      launchd.user.agents.yabai = {
+        serviceConfig.ProgramArguments = [ "${cfg.package}/bin/yabai" ]
+                                         ++ optionals (cfg.config != {} || cfg.extraConfig != "") [ "-c" configFile ];
+
+        serviceConfig.KeepAlive = true;
+        serviceConfig.RunAtLoad = true;
+        serviceConfig.EnvironmentVariables = {
+          PATH = "${cfg.package}/bin:${config.environment.systemPath}";
+        };
+      };
+    })
+
+    # TODO: [@cmacrae] Handle removal of yabai scripting additions
+    (mkIf (cfg.enableScriptingAddition) {
+      launchd.daemons.yabai-sa = {
+        script = ''
+          if [ ! $(${cfg.package}/bin/yabai --check-sa) ]; then
+            ${cfg.package}/bin/yabai --install-sa
+          fi
+        '';
+
+        serviceConfig.RunAtLoad = true;
+        serviceConfig.KeepAlive.SuccessfulExit = false;
+      };
+    })
+  ];
+}
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -53,6 +53,7 @@
   ./services/skhd
   ./services/synapse-bt.nix
   ./services/synergy
+  ./services/yabai
   ./programs/bash
   ./programs/fish.nix
   ./programs/gnupg.nix
--- a/release.nix
+++ b/release.nix
@@ -123,6 +123,7 @@ let
     tests.services-skhd = makeTest ./tests/services-skhd.nix;
     tests.services-synapse-bt = makeTest ./tests/services-synapse-bt.nix;
     tests.services-synergy = makeTest ./tests/services-synergy.nix;
+    tests.services-yabai = makeTest ./tests/services-yabai.nix;
     tests.system-defaults-write = makeTest ./tests/system-defaults-write.nix;
     tests.system-environment = makeTest ./tests/system-environment.nix;
     tests.system-keyboard-mapping = makeTest ./tests/system-keyboard-mapping.nix;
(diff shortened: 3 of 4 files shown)


eraserhd:
Nice!

modules/services/yabai/default.nix
```diff
@@ -0,0 +1,107 @@
...
+    services.yabai = with types; {
+      enable = mkOption {
+        type = bool;
+        default = false;
+        description = "Whether to enable the yabai window manager.";
+      };
+
+      package = mkOption {
+        type = path;
+        default = pkgs.yabai;
```
LnL7, line 25: This was only added recently to nixpkgs, this default shouldn't be included until 20.09 has been released for a while so people using the stable channel don't run into issues.

modules/services/yabai/default.nix
```diff
@@ -0,0 +1,107 @@
...
+  toYabaiConfig = opts:
+    concatStringsSep "\n" (mapAttrsToList
+      (p: v: "yabai -m config ${p} ${toString v}") opts);
+
+in
+
+{
+  options = {
+    services.yabai = with types; {
+      enable = mkOption {
```
LnL7, line 17: nit, I generally prefer to define options using their explicit attr path which makes it a bit easier to search

```suggestion
    services.yabai.enable = mkOption {
```

modules/services/yabai/default.nix
```diff
@@ -0,0 +1,107 @@
...
+    (mkIf (cfg.enable) {
+      environment.systemPackages = [ cfg.package ];
+
+      launchd.user.agents.yabai = {
+        serviceConfig.ProgramArguments = [ "${cfg.package}/bin/yabai" ]
+                                         ++ optionals (cfg.config != {} || cfg.extraConfig != "") [ "-c" "/etc/yabairc"];
+        serviceConfig.KeepAlive = true;
+        serviceConfig.RunAtLoad = true;
+        serviceConfig.EnvironmentVariables = {
+          PATH = "${cfg.package}/bin:/usr/bin:/bin:/usr/sbin:/sbin";
```
LnL7, line 80:
```suggestion
        path = [ cfg.package config.environment.systemPath ];
```

modules/services/yabai/default.nix
```diff
@@ -0,0 +1,107 @@
...
+          yabai -m rule --add app='System Preferences' manage=off
+        '';
+        description = "Extra arbitrary configuration to append to the configuration file";
+      };
+    };
+  };
+
+  config = mkMerge [
+    (mkIf (cfg.enable) {
+      environment.systemPackages = [ cfg.package ];
```
LnL7, line 72: Any point in including this in systemPackages?

modules/services/yabai/default.nix
```diff
@@ -0,0 +1,107 @@
...
+        serviceConfig.KeepAlive = true;
+        serviceConfig.RunAtLoad = true;
+        serviceConfig.EnvironmentVariables = {
+          PATH = "${cfg.package}/bin:/usr/bin:/bin:/usr/sbin:/sbin";
+        };
+      };
+    })
+
+    (mkIf (cfg.config != {} || cfg.extraConfig != "") {
+      environment.etc."yabairc".source = pkgs.writeScript "etc-yabairc" (
```
LnL7, line 86: Any point in exposing this in `/etc`?

cmacrae: Common convention that’s already adhered to in several other modules in this codebase

modules/services/yabai/default.nix
cmacrae: Simply having the ‘yabai’ binary in the user’s PATH means they can arbitrarily send messages to the socket. A generally exercised privilege when having yabai installed

modules/services/yabai/default.nix
cmacrae: Fair enough. So this option should simply be removed and default to the available package in the channels

modules/services/yabai/default.nix
cmacrae: Makes sense, I’ll make this change when I’m at a proper machine 📱

modules/services/yabai/default.nix
cmacrae: It's also necessary for `yabai` to be in the PATH when the config file is evaluated (it's just a shell script)

modules/services/yabai/default.nix
cmacrae: Also wondering how this would affect the test approach used here:  
https://github.com/LnL7/nix-darwin/pull/196/files#diff-e566c25b87e15708d8368b82febe145cR11

LnL7: Just means that the option is required. The main reason for overriding that even when there's a default is to reduce the build closure. I don't want to build all sorts of packages just to run these tests and if the package upstream breaks it shouldn't impact the test results either.

modules/services/yabai/default.nix
LnL7: Actually nevermind, while the user probably doesn't need it other services like skhd do and that currently just uses this as context.

modules/services/yabai/default.nix
LnL7: If the service supports hot reloading for example /etc can be used to update the configuration in a stable location without restarting, otherwise it's one more thing that doesn't have to be global.

modules/services/yabai/default.nix
cmacrae: Ahh, sure okay. I understand 👍 Alright, I'll remove the default.  
Thanks for explaining the test approach, too. That's a nice trick :)

modules/services/yabai/default.nix
cmacrae: Okay :) So perhaps rather than writing `/etc/yabairc` it'd just be an arbitrary path? Something like this?
```nix
{
      launchd.user.agents.yabai = {
        serviceConfig.ProgramArguments = [ "${cfg.package}/bin/yabai" ]
                                         ++ optionals (cfg.config != {} || cfg.extraConfig != "")
                                           [ "-c" pkgs.writeScript "yabairc" (
                                             (if (cfg.config != {})
                                              then "${toYabaiConfig cfg.config}"
                                              else "")
                                             + optionalString (cfg.extraConfig != "") cfg.extraConfig)];
        serviceConfig.KeepAlive = true;
        serviceConfig.EnvironmentVariables = {
          PATH = "${cfg.package}/bin:/usr/bin:/bin:/usr/sbin:/sbin";
        };
      };
}
```

modules/services/yabai/default.nix
cmacrae: Whoops! Must've resolved this in error - I'll implement the suggested change

cmacrae: The value needs to be a string, so I've implemented it like so:
```
PATH = "${cfg.package}/bin:${config.environment.systemPath}";
```

modules/services/yabai/default.nix
LnL7: Yeah, alltho would be nicer to put this in the let binding at the top and use `[ "-c" configFile ]` so it's a bit more readable

cmacrae: Good idea 👍 I've updated it like so  
Just wondering how we do the tests now? As I'd written them to grep for strings in the resulting `/etc/yabairc`, but since that path is now unknown, I'm not sure how to approach this

cmacrae: Not sure if this is really the right approach here, but I got the test passing by using `sed` to pull out the store path from the plist:
```
  test = ''
    echo >&2 "checking yabai service in ~/Library/LaunchAgents"
    grep "org.nixos.yabai" ${config.out}/user/Library/LaunchAgents/org.nixos.yabai.plist
    grep "${yabai}/bin/yabai" ${config.out}/user/Library/LaunchAgents/org.nixos.yabai.plist

    conf=`sed -En '/<string>-c<\/string>/{n; s/\s+?<\/?string>//g; p;}' ${config.out}/user/Library/LaunchAgents/org.nixos.yabai.plist`
    echo >&2 "checking config in $conf"
    grep "yabai -m config focus_follows_mouse autoraise" $conf
    grep "yabai -m rule --add app='System Preferences' manage=off" $conf
  '';
```
Tested with `nix-build release.nix -A tests.services-yabai` and it works.  
Thoughts @LnL7? Perhaps you know of a better way to do this?

cmacrae: I squashed in that change [here](https://github.com/LnL7/nix-darwin/commit/7ef533e8d1fb6ece563232b326fc5b5d5c2f9257#diff-e566c25b87e15708d8368b82febe145cR20) so you can see (can remove if there's a nicer way, of course)

LnL7: Quite a few nixos modules also have a [configFile](https://github.com/NixOS/nixpkgs/blob/release-20.03/nixos/modules/config/pulseaudio.nix#L119) option, which you could reference in the tests to access the file path. But this is also fine, the tests are more a sanity check anyway.

cmacrae: If you’re fine with this, perhaps let’s just keep it this way. If there’s interest in the future for a configFile option, I can add it :)

LnL7: Sounds good

Daniel-V1:
Should the default be set for yabai.package to pkgs.yabai?

LnL7:
See https://github.com/LnL7/nix-darwin/pull/196#discussion_r421693792

PR #197, opened 2020-05-06
module: add spacebar service

cmacrae:
Another service for your consideration :) This is to acompany this PR in NixOS/nixpkgs#86556  
Once that package is generally available in the channels, the tests included in these changes should pass

### DIFF
--- a/modules/services/spacebar/default.nix
+++ b/modules/services/spacebar/default.nix
@@ -1,0 +1,74 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.services.spacebar;
+
+  toSpacebarConfig = opts:
+    concatStringsSep "\n" (mapAttrsToList
+      (p: v: "spacebar -m config ${p} ${toString v}") opts);
+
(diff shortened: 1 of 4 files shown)


cmacrae:
I'll take the time tomorrow to clean up the code here like we went over in #196

cmacrae:
Trying to determine why these tests are failing. They pass fine on my machine with `nix-build release.nix -A tests.services-spacebar`

LnL7:
Can you rebase?

cmacrae:
How's that? Better? :)

modules/services/spacebar/default.nix
```diff
@@ -0,0 +1,76 @@
...
+      example = literalExample ''
+        echo "spacebar config loaded..."
+      '';
+      description = ''
+        Extra arbitrary configuration to append to the configuration file.
+      '';
+    };
+  };
+
+  config = mkMerge [
```
LnL7, line 60: no need for mkMerge here

cmacrae: Good catch 👍 I've addressed it in the latest revision

PR #198, opened 2020-05-09
Fix Bash Completions

Daniel-V1:
This fixes bash completion for me, though it may not be the best way.
Hopefully this isn't too crazy or someone can tell me how to do it better. I'm very new to nix. Thanks!

### DIFF
--- a/modules/programs/bash/default.nix
+++ b/modules/programs/bash/default.nix
@@ -86,9 +86,6 @@ in
             for m in "$p/etc/bash_completion.d/"*; do
               source $m
             done
-            for m in "$p/share/bash-completion/completions/"*; do
-              source $m
-            done
           done
           eval "$nullglobStatus"
           unset nullglobStatus p m
--- a/modules/programs/bash/default.nix
+++ b/modules/programs/bash/default.nix
@@ -86,6 +86,9 @@ in
             for m in "$p/etc/bash_completion.d/"*; do
               source $m
             done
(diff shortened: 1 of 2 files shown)


modules/programs/bash/default.nix
```diff
@@ -86,6 +86,9 @@ in
             for m in "$p/etc/bash_completion.d/"*; do
               source $m
             done
+            for m in "$p/share/bash-completion/completions/"*; do
```
LnL7, line 89: I had the impression this was the convention in nixpkgs, but aparently not. The new location is just sourced automatically.

https://github.com/NixOS/nixpkgs/blob/29d0e684afcc1a3f4d5a5c60a483ba4c719c73f6/nixos/modules/programs/bash/bash.nix#L15-L20

Given that I would prefer to follow the nixos approach and configure `XDG_DATA_DIRS` instead.

Daniel-V1: Alright, so I've added the XDG_DATA_DIRS and XDG_CONFIG_DIRS variables to the environment.nix file and set them up to be appended to each of the paths in NIX_PROFILES. Its seems in nixos they wind up doing the application when they are writing to PAM environment, but I'm not sure if there is a similar step for macos. Thanks for pointing me in the right direction and hopefully this looks better!

LnL7: Yeah, the environment stuff is a bit more complicated there because there's a difference between session (display manager) and shell variables IIRC.

PR #200, opened 2020-05-15
Separate yabai config and extra config by newline

shardulbee:
If a user passes both `config` and `extraConfig` to the `yabai` serivce,
the generated `yabairc` file is invalid. This is because we do not add a
newline separator when we concatenate the config string generated by
`toYabaiConfig cfg.config` with `cfg.extraConfig`.

This PR prepends a newline to `cfg.extraConfig` if it is non-empty so
that the resulting `yabairc` config is valid.

@LnL7 @cmacrae could you please take a look? Let me know if I'm missing anything important. I tested this change locally and the generated `yabairc` is now correct.

### DIFF
--- a/modules/services/yabai/default.nix
+++ b/modules/services/yabai/default.nix
@@ -14,7 +14,7 @@ let
       (if (cfg.config != {})
        then "${toYabaiConfig cfg.config}"
        else "")
-      + optionalString (cfg.extraConfig != "") cfg.extraConfig)}";
+      + optionalString (cfg.extraConfig != "") ("\n" + cfg.extraConfig + "\n"))}";
 in
 
 {
--- a/tests/services-yabai.nix
+++ b/tests/services-yabai.nix
@@ -23,5 +23,6 @@ in
     echo >&2 "checking config in $conf"
     grep "yabai -m config focus_follows_mouse autoraise" $conf
     grep "yabai -m rule --add app='System Preferences' manage=off" $conf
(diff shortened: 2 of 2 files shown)


cmacrae:
Hmm, strange - I don't see this behaviour on my system. Perhaps it's because I've been using a multi-line string for my `extraConfig`.  

I'll have a look over this tomorrow :)

shardulbee:
@cmacrae with your `extraConfig` you don't see this issue on your machine?

https://github.com/cmacrae/config/blob/403b7b9cdccbeb0b1ce76083d795206947dd7b93/modules/darwin.nix#L225-L258

cmacrae:
@shardulbee Not with my config you linked there, no :)

But, as per my edit in the above comment, it does indeed happen with a single line string.

I checked out your fork and just did some rudimentary testing, looks good.

@LnL7 Take this comment as my approval 👍

shardulbee:
@cmacrae that's interesting. This is my `extraConfig`:

`nix
extraConfig = ''
    yabai -m rule --add app="Firefox" space=^3
    yabai -m rule --add app="Slack" space=^7
    yabai -m rule --add app="Messages" space=^8
    yabai -m rule --add app="Things" space=^9
    yabai -m rule --add app="[aA]lacritty" space=^10
    yabai -m rule --add app="Dash.app" manage=off topmost=on
    yabai -m rule --add app="Finder" manage=off topmost=on
    yabai -m rule --add app="System Preferences" manage=off topmost=on
'';
```

I still have the issue I mentioned.

PR #204, opened 2020-05-27
module: add Lorri service

zachcoyle:
I've added a service module to enable the Lorri daemon, adapted from the home-manager Lorri service.  
It may still be missing something, as I'm not sure exactly what the following lines do:
https://github.com/rycee/home-manager/blob/8af92f844f6acb3957dd68072e8fc0e00d4c3eba/modules/services/lorri.nix#L41-L44

### DIFF
--- a/modules/services/lorri.nix
+++ b/modules/services/lorri.nix
@@ -1,0 +1,35 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.services.lorri;
+
+in {
+
+  options = {
+    services.lorri = {
+      enable = mkOption {
+        type = types.bool;
+        default = false;
+        description = "Whether to enable to Lorri Daemon.";
+      };
+
+      package = mkOption {
+        type = types.path;
+        default = pkgs.lorri;
+        description = "This option specifies the lorri package to use.";
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    environment.systemPackages = [ cfg.package ];
+    launchd.user.agents.lorri = {
+      command = "${cfg.package}/bin/lorri daemon";
+      serviceConfig.KeepAlive = true;
+    };
+
+  };
+}
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -42,6 +42,7 @@
   ./services/emacs.nix
   ./services/khd
   ./services/kwm
+  ./services/lorri.nix
   ./services/mail/offlineimap.nix
   ./services/mopidy.nix
   ./services/nix-daemon.nix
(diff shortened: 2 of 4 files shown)


modules/services/lorri.nix
```diff
@@ -0,0 +1,35 @@
...
+        type = types.path;
+        default = pkgs.lorri;
+        description = "This option specifies the lorri package to use.";
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    environment.systemPackages = [ cfg.package ];
+    launchd.user.agents.lorri = {
```
LnL7, line 29: From what I recall the lorri daemon runs evaluation, which means in order to evaluate expressions with `builtins.fetchTarball` or `fetchGit` it will need access to tar/git. There's a `path` option for launchd services which provides a convenient way to expose binaries.

zachcoyle:
Thanks for the clarification @LnL7

jamesottaway:
> adapted from the home-manager Lorri service

Just to throw a spanner in the works, there are a few other examples to use for reference/inspiration:

* [the NixOS module](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/development/lorri.nix)
* [`contrib/lorri.service` in the Lorri source repo](https://github.com/target/lorri/blob/master/contrib/lorri.service)
* [two other nix-darwin user's LaunchAgents](https://github.com/target/lorri/blob/master/contrib/daemon.md#run-lorri-daemon-on-macos-with-nix-using-nix-darwin), as mentioned in the Lorri source repo
* [my own `launchd.user.agents.lorri`](https://github.com/jamesottaway/nix-config/blob/master/lorri.nix), which I've been tinkering with as I learn nix-darwin

surajbarkale:
I ran into permission issues with macOS 10.15.5 and lorri running as user agent. I am launching [lorri via shell invocation](https://github.com/target/lorri/issues/96#issuecomment-588563793) so it gets the NIX_PATH correctly. I had to give "Full Disk Access" to my shell to make the daemon work with files in `~/Documents`.

It may be better to run lorri as system agent to get around this issue.

modules/services/lorri.nix
tricktron: So I guess you can just add this:
```suggestion
    launchd.user.agents.lorri = {
    path = with pkgs; [ nix gitMinimal gnutar gzip ];
```

tricktron:
@jamesottaway 
Thanks for collecting all the different configs. Unfortunately your config doesn't work for me:
My test approach:

```bash
mkdir example
lorri init
# add pkgs.cowsay to shell.nix
cd ..
cd example
cowsay
# Unknown command cowsay
```
Does that work on your machine?

jamesottaway:
@tricktron no, looks like I'm in need of some more tinkering. thanks for the heads up.

tricktron:
@LnL7 @zachcoyle @jamesottaway I made it working and opened another pull request #222.

PR #208, opened 2020-06-15
fixes zsh module shell variable assignments

pjan:
Without this, there's a collision when you define them top-level (e.g. necessary if you enable both bash & zsh, but want bash to be the default shell)

### DIFF
--- a/modules/programs/zsh/default.nix
+++ b/modules/programs/zsh/default.nix
@@ -105,8 +105,8 @@ in
 
     environment.pathsToLink = [ "/share/zsh" ];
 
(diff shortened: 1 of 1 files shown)


LnL7:
This causes priority issues when both are enabled, which should probably be the case now with Catalina.

pjan:
it is however troublesome for anyone who wants nix enabled in both bash & zsh, but doesn't want zsh to be the default shell.

LnL7:
Perhaps these defaults should just be removed alltogether.

LnL7:
I think this is obsolete now.

PR #213, opened 2020-08-03
change eval step

Kloenk:
This changes the eval step to a easier extendable version  (in my opinion)

### DIFF
--- a/eval-config.nix
+++ b/eval-config.nix
@@ -1,0 +1,35 @@
+{ lib, system ? "x86_64-darwin" }:
+
+{ configuration, inputs }:
+
+let
+  baseModules = import ./modules/module-list.nix;
+  modules = [ configuration packages ] ++ baseModules;
+
+  packages = { config, lib, pkgs, ... }: {
+    _file = ./default.nix;
+    config = {
+      _module.args.inputs = inputs;
+      _module.args.pkgs = import inputs.nixpkgs config.nixpkgs;
+      nixpkgs.system = system;
+    };
+  };
+
+  eval = lib.evalModules {
+    inherit modules;
+    args = { inherit baseModules modules; };
+    specialArgs = { modulesPath = ./modules; };
+    check = true;
+  };
+
+  # Was moved in nixpkgs #82751, so both need to be handled here until 20.03 is deprecated.
+  # https://github.com/NixOS/nixpkgs/commits/dcdd232939232d04c1132b4cc242dd3dac44be8c
+  _module = eval._module or eval.config._module;
+in
+
+{
+  inherit (_module.args) pkgs;
+  inherit (eval) options config;
+
+  system = eval.config.system.build.toplevel;
+}
--- a/eval-config.nix
+++ b/eval-config.nix
@@ -1,10 +1,40 @@
-{ lib, system ? "x86_64-darwin" }:
+#{ lib, system ? "x86_64-darwin", check ? true }:
 
-{ configuration, inputs }:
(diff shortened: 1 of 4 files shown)


Kloenk:
The tests works on my machine. maybe a problem while setting up chachix? like missing nix path?

berbiche:
I use these changes in my personal (messy) Nix configuration available [here](https://github.com/berbiche/dotfiles/tree/darwin).
I had to setup a Darwin machine for the first time yesterday with nix-darwin which is why it's so messy.

I used the following commands to build my configuration with nix-darwin and Flakes:

``` console
$ nix-env -iA nixos.nixFlakes
$ nix-shell
$ sudo nix build '.#darwinConfigurations.pc335' -vv --impure
$ nix-env -p /nix/var/nix/profiles/system --set ./result
$ ./result/activate-user
$ sudo ./result/activate
```

I had to use the `--impure` flag because I have a mixed configuration (Linux & Darwin) which required access to `builtins.currentSystem`.

The rebuild script and activation scripts would need to be updated to work with Flakes.

1. The rebuild script could accept a `--flake` argument and read the `darwinConfigurations` set from the Flake's `outputs`. `nixos-rebuild` defaults to `.#nixosConfigurations.${HOSTNAME}` when using `--flake .`. `nixos-rebuild` also supports the shorthand `--flake '.#${HOSTNAME}` to refer to `.#nixosConfigurations.${HOSTNAME}`.

2. The dependency on the `<darwin>` and `<darwin-config>` channel should be removed for the activation scripts to work.
I had to modify my `nix.nixPath` and set a "proper" `NIX_PATH` in my `shell.nix` otherwise it would fail when activating.

LnL7:
For 2. there's an option to disable the checks IIRC, as for the rest I it would be best to keep the implementation as close as possible to nixos. However there's a big difference here since nixpkgs needs to be an extra input that can be configured by the user.

berbiche:
Instead of passing the `inputs.nixpkgs` directly you can make override a Flake's inputs. See the Input schema option of https://nixos.wiki/wiki/Flakes.

Here's an example of what I had in mind:

``` nix
{
  description = "My sample nix-darwin configuration";

  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable-small";
    nix-darwin = {
      url = "github:lnl7/nix-darwin/flakes";
      inputs.nixpkgs.follow = "nixpkgs";
    };
  };

  outputs = { self, nix-darwin, nixpkgs, ... } @ args: {
    darwinConfigurations = {
      example = nix-darwin.lib.evalConfig {
        configuration = { ... }: { imports = [ ./my-config.nix ]; };
      };
    };
  };
}
```

eval-config.nix
```diff
@@ -1,10 +1,40 @@
...
+
+  eval = lib.evalModules {
+    modules = let
+      initModule = { ... }: {
+        _file = ./default.nix;
+        config = {
+          _module.args = {
+            pkgs = pkgs;
+            baseModules = darwinModules;
+          } // (if inputs != null then inputs else {});
```
berbiche, line 16: Should `inputs` be allowed to override `pkgs` and `baseModules`?

berbiche:
Relevant PR opened in https://github.com/rycee/home-manager/pull/1460

berbiche:
> Relevant PR opened in [rycee/home-manager#1460](https://github.com/rycee/home-manager/pull/1460)

PR has been merged 🚀

LnL7:
I reworked the flakes branch here https://github.com/LnL7/nix-darwin/pull/235 with proper support after taking a look at the current nixos integration. I think this probably resolves some of the goals here but if there's anything else you're missing feel free to comment there or reopen this.

PR #215, opened 2020-08-14
darwin-[un]installer: make sure all dependencies are part of the closure

basvandijk:
This makes sure the `darwin-installer` and `darwin-uninstaller` packages can be nix copied to any target and reliably be executed there.

### DIFF
--- a/pkgs/darwin-installer/default.nix
+++ b/pkgs/darwin-installer/default.nix
@@ -1,10 +1,16 @@
-{ stdenv, writeScript, nix, pkgs }:
+{ stdenv, writeScript, nix, pkgs, nix-darwin }:
 
 let
+  configuration = builtins.path {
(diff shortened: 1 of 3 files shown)


pkgs/darwin-installer/default.nix
```diff
@@ -1,10 +1,21 @@
...
 let
+  configuration = builtins.path {
+    name = "nix-darwin-installer-configuration";
+    path = ./.;
+    filter = name: _type: name != toString ./default.nix;
+  };
+
+  darwin = builtins.path {
+    name = "nix-darwin";
+    path = ../..;
```
LnL7, line 12: Any reason not to filter out .git? This repo isn't particularly big but that does make this pretty unreproducible.

LnL7: Looks good, thanks!

PR #216, opened 2020-08-16
documentation: Don't set localSystem which doesn't exist

roberth:
This is safe to remove, because it is not referenced anywhere in
nix-darwin.
It should have been discovered way earlier, but a bug in the
module system has allowed this value to be defined until
https://github.com/NixOS/nixpkgs/commit/fd75dc876586bde8cdb683a6952a41132e8db166

### DIFF
--- a/modules/documentation/default.nix
+++ b/modules/documentation/default.nix
@@ -17,7 +17,7 @@ let
     options =
       let
         scrubbedEval = evalModules {
-          modules = [ { nixpkgs.localSystem = config.nixpkgs.localSystem; } ] ++ baseModules;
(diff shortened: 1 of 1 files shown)


roberth:
This only happened with nixpkgs unstable. Users on nixpkgs 20.03 were not affected. Does that count as "tied to the release cycle" or did you have something else in mind?

LnL7:
Well I would consider unstable packages and unstable apis separate things, as an example we generally don't rename packages on unstable either without adding an alias for compatibility. As it stands there's no way to provide a transition window while still allowing rolling package updates, which is what most people use in practice.

PR #222, opened 2020-08-29
Adds Lorri module

tricktron:
I was working in parallel to this pr #204 and managed to make it work. I open a new pull request as it was easier to add my changes here.

### DIFF
--- a/modules/services/lorri.nix
+++ b/modules/services/lorri.nix
@@ -1,0 +1,59 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.services.lorri;
+in
+{
+  options = {
+    services.lorri.enable = mkOption {
+      type = types.bool;
+      default = false;
+      description = "Whether to enable the lorri service.";
+    };
+
+    services.lorri.package = mkOption {
+      type = types.path;
+      default = pkgs.lorri;
+      defaultText = "pkgs.lorri";
+      description = "This option specifies the lorri package to use.";
+    };
+
+    services.lorri.logFile = mkOption {
+      type = types.nullOr types.path;
+      default = "/var/tmp/lorri.log";
+      example =  "/var/tmp/lorri.log";
+      description = ''
+        The logfile to use for the lorri service. Alternatively
+        <command>sudo launchctl debug system/org.nixos.lorri --stderr</command>
+        can be used to stream the logs to a shell after restarting the service with
+        <command>sudo launchctl kickstart -k system/org.nixos.lorri</command>.
+      '';
+    };
+
+    services.lorri.tempDir = mkOption {
+      type = types.nullOr types.path;
+      default = null;
+      description = "The TMPDIR to use for lorri.";
+    };
+  };
+
+  config = mkIf cfg.enable {
+    environment.systemPackages = [ cfg.package ];
+    launchd.user.agents.lorri = {
+      serviceConfig = {
+        KeepAlive = true;
+        ProcessType = "Background";
+        LowPriorityIO = false;
+        StandardOutPath = cfg.logFile;
+        StandardErrorPath = cfg.logFile;
+        EnvironmentVariables = mkMerge [
+        config.nix.envVars
+        { TMPDIR = mkIf (cfg.tempDir != null) cfg.tempDir; }
+        ];
+      };
+      command = "${cfg.package}/bin/lorri daemon";
+    };
+  };
+}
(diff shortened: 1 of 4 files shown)


modules/services/lorri.nix
```diff
@@ -0,0 +1,46 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.services.lorri;
+  home = "${builtins.getEnv "HOME"}";
```
LnL7, line 7: Using getEnv isn't allowed in pure mode.

modules/services/lorri.nix
```diff
@@ -0,0 +1,46 @@
...
+          <command>sudo launchctl debug system/com.target.lorri --stderr</command>
+          can be used to stream the logs to a shell after restarting the service with
+          <command>sudo launchctl kickstart -k system/com.target.lorri</command>.
+        '';
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    environment.systemPackages = [ pkgs.lorri ];
```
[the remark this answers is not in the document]
tricktron: I only needed to add nix to make it work.

modules/services/lorri.nix
```diff
@@ -0,0 +1,46 @@
...
+    services.lorri = {
+      enable = mkOption {
+        type = types.bool;
+        default = false;
+        description = "Whether to enable the lorri service.";
+      };
+      
+      logFile = mkOption {
+        type = types.nullOr types.path;
+        default = "/var/tmp/lorri.log";
```
tricktron, line 19: What would be a good default value here?
`/var/log/lorri.log`, `/Library/Logs/lorri.log` and `/run/current-system/user/Library/lorri.log` all give permission errors. What about a log directory under `/run/current-system/user/Logs`?

tests/services-lorri.nix
```diff
@@ -0,0 +1,24 @@
+{ config, pkgs, ... }:
+
+let
+  lorri = pkgs.runCommand "lorri-0.0.0" {} "mkdir $out";
+  plistPath = "${config.out}/user/Library/LaunchAgents/org.nixos.lorri.plist";
```
tricktron, line 5: @LnL7 To make plist testing less fragile, we could use `plutil` which can lint, read and convert plist files to json. How could I integrate it into the tests here?

modules/services/lorri.nix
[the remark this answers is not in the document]
LnL7: As mentioned in that comment certain builtins depend on binaries indirectly. That's why things like git and tar are included in the nixos module, otherwise expressions that use those wouldn't work.

```
$ PATH=/nix/store/knw0knz4wkpdml3h3g2vb9fih2gvhfrx-nix-2.4pre20200721_ff314f1/bin nix-instantiate --eval -E 'builtins.fetchGit https://github.com/LnL7/nix-darwin.git'
error: error: --- SysError ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- nix-instantiate
executing 'git': No such file or directory
error: --- Error ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- nix-instantiate
program 'git' failed with exit code 1
```

modules/services/lorri.nix
LnL7: Using `StandardOutPath` doesn't do any log rotation either so it's best not to enable that by default. As for the location, yeah there's not really a great location that doesn't require extra setup to get the permissions right first. Ideally would do that part automatically for agents with logging.

tests/services-lorri.nix
LnL7: That's a great idea, we have plutil as part of `xcbuild` so that should work.

modules/services/lorri.nix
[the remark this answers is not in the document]
tricktron: Ah now I get it. Thanks for the clarification!

modules/services/lorri.nix
tricktron: Ok, I disable StandardOutPath and StandardErrorPath by default.

tests/services-lorri.nix
```diff
@@ -0,0 +1,35 @@
...
+            "exec ${builtins.unsafeDiscardStringContext pkgs.lorri}/bin/lorri daemon"
+    ],
+    "RunAtLoad": true
+  }
+  '';
+  testResult = toString (actualJson == expectedJson);
+in
+{
+  services.lorri.enable = true;
+  test = ''
```
tricktron, line 30: @LnL7 If you know a better way to write a test, let me know. Otherwise I guess this is ready to merge.

LnL7: Not especially. Checking everything is probably a bit overkill but that's fine.

tricktron: @LnL7 Ok, then I am done.

LnL7 (approved):
Thanks!

PR #223, opened 2020-09-01
Add separate options for ComputerName and LocalHostName in networking module

malob:
I usually want my Mac's "computer name" to be different from its hostname. Currently this isn't possible with `nix-darwin` since setting `networking.hostname` changes all three of the hostname, local hostname and "computer name" to the same thing.

This PR makes the following changes:

* adds the options `networking.localHostName` and `networking.computerName`,
* Refines the types of `networking.HostName` to only accept strings that are valid hostnames and uses this type for `networking.localHostName` as well,
* changes the default behavior of setting `networking.hostname` to only set the system's hostname and local hostname (if the `networking.localHostName` option isn't set separately), and
  * (This is done by having each option set the corresponding name for the system, and setting the default value of `networking.localHostName` to the value of `networking.hostname`. The idea being that I expect it to be much less common that people will want their system to have a different hostname and local hostname, and much more common to want the "computer name" to be different since it's allowed to contain spaces and arbitrary Unicode characters.)
* updates tests in `tests.networking-hostname`.

### DIFF
--- a/modules/networking/default.nix
+++ b/modules/networking/default.nix
@@ -5,14 +5,10 @@ with lib;
 let
   cfg = config.networking;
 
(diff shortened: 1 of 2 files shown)


LnL7 (approved):
Thanks

Issue #211, opened 2020-08-02
Viability for `nixpkgs.pkgs` option, similar to NixOS


jamesottaway:
I'm looking to pin my `nixpkgs` set and went looking for inspiration. I found the [`nixpkgs.pkgs` option in NixOS](https://nixos.org/nixos/options.html#nixpkgs.pkgs) and where it [sets `_module.args.pkgs`](https://github.com/NixOS/nixpkgs/blob/release-20.03/nixos/modules/misc/nixpkgs.nix#L230-L232).

I poked around `modules/nix/nixpkgs.nix` and found a [comment indicating nix-darwin uses a different approach](https://github.com/LnL7/nix-darwin/blob/master/modules/nix/nixpkgs.nix#L112). The [commit in question](https://github.com/LnL7/nix-darwin/commit/b1a264c928fdb7e674c0e9f9fe2edbe5c04f5bb2) talks about overriding the `nixpkgs` argument in [`release.nix`](https://github.com/LnL7/nix-darwin/blob/master/release.nix), hence this issue.

Is there a reason that the `_module.args.pkgs` approach is incompatible with overriding the `nixpkgs` argument in [`release.nix`](https://github.com/LnL7/nix-darwin/blob/master/release.nix)? If not I'll put a PR up for the `nixpkgs.pkgs` option, but if not I'd love to know more about the motivation behind b1a264c928fdb7e674c0e9f9fe2edbe5c04f5bb2 and how to avoid the two goals clashing.

jamesottaway:
I just saw the Flakes support so I think this issue is redundant now. Amazing work, by the way. I've already switched over and I love it: https://github.com/jamesottaway/nix-config/blob/master/flake.nix

LnL7:
Apologies for not responding.  I'm not particularly opposed to adding this but I feel like it makes much more sense to to declare this kind of stuff top down instead of overriding it at such a late point.  So unless somebody needs it for a very specific use-case I would prefer to keep this simple.

jamesottaway:
No worries about not getting back to me, I just wanted to throw the idea out into the ether while I was tinkering. I ended up wanting to pin my copy of this repo too and actually ended up doing something very similar to #231.

I totally agree about the top-down thing. I'd read a bit about Flakes before today but #235 finally got me to try it and it's definitely the right way to pin the package set.

PR #227, opened 2020-09-10
Allow the `pkgs` arg to be pinned via the `nixpkgs.pkgs` option

jamesottaway:
I didn't hear back about #211 so I had a crack at adding a `nixpkgs.pkgs` config option, similar to NixOS.

jamesottaway/nix-config@faeac44fd0dc28e72d84e9311a2880212e404ff4 is my WIP attempt to pin `nixpkgs` in my personal nix-darwin config.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -9,7 +9,6 @@ let
   packages = { config, lib, pkgs, ... }: {
     _file = ./default.nix;
     config = {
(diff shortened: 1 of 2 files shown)


jamesottaway:
Redundant. See https://github.com/LnL7/nix-darwin/issues/211#issuecomment-716139292

PR #228, opened 2020-09-11
Add option to enable sudo authentication with Touch ID

malob:
There's a one line addition to `/etc/pam.d/sudo` that allows Macs with Touch ID to use Touch ID to authenticate `sudo` requests. It's a really nice thing to have. Seem like a nice option to add to `nix-darwin`.

I wasn't quite sure where to put the option, so definitely feel free to ask me to make changes to put it somewhere that makes more sense.

I choose to use `system.patch` rather than `environment.etc` since I didn't want to require the user to have to delete the existing file potentially breaking `sudo` authentication.


jrolfs:
Does this work outside of **Terminal.app**??

malob:
Yes. I use it with Kitty terminal.

modules/system/sudo.nix
```diff
@@ -0,0 +1,15 @@
...
+let
+  cfg = config.system.sudo;
+in
+
+{
+  options = {
+    system.sudo.touchid.enable = mkEnableOption "Enable sudo authentication with Touch ID";
+  };
+
+  config = mkIf cfg.touchid.enable { system.patches = [ ./etc-pam.d-sudo.patch ]; };
```
LnL7, line 14: Sadly the idea of system.patches doesn't work as well as I initially thought. I couldn't find a way to re apply a patch or check if it's already applied. So currently this only runs once, meaning it doesn't work for files that change outside of rebuilds like system updates.

modules/system/sudo.nix
```diff
@@ -0,0 +1,15 @@
...
+
+with lib;
+
+let
+  cfg = config.system.sudo;
+in
+
+{
+  options = {
+    system.sudo.touchid.enable = mkEnableOption "Enable sudo authentication with Touch ID";
```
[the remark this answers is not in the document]
malob: Good point. Looking at NixOS options that exist, I'm now thinking that something like `security.pam.enableSudoTouchIdAuth` would be better. What do you think?

modules/system/sudo.nix
malob: Ah shoot, that will be a problem since macOS resets this file after every update. 

Given that `environment.etc` doesn't seem viable, since asking the user to delete/move the file in question seems like a recipe for trouble, I assume the only way forward here is via a custom activation script. I've got some ideas about how to implement that, but I just wanted to double check that I'm not missing some other more elegant way of implementing this option.

modules/system/sudo.nix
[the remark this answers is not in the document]
malob: Pushed a new commit moving that that name for now.

modules/system/sudo.nix
malob: Pushed a new commit with a first pass at implementing this using an activation script.

emilazy: I think it would be better to use another `sed` to delete the `pam_tid.so` line on disable, so that other changes to the `/etc` file since the `.orig` backup aren't wiped out.

A more general/scalable approach to this kind of thing could be to attach a comment containing some unique identifier to every block of text added to a system `/etc` file, and have centralized tooling for inserting and removing those blocks. This could also be used for `environment.shells` and similar options. Of course, this individual case is probably too simple to warrant infrastructure like that.

malob: > I think it would be better to use another sed to delete the pam_tid.so line on disable, so that other changes to the /etc file since the .orig backup aren't wiped out.

Good point

> A more general/scalable approach to this kind of thing could be to attach a comment containing some unique identifier to every block of text added to a system /etc file, and have centralized tooling for inserting and removing those blocks. This could also be used for environment.shells and similar options. Of course, this individual case is probably too simple to warrant infrastructure like that.

Totally agree this would be really nice to have, but don't currently have the time to play around with implementing right now.

I've update the script based on your feedback, it uses `sed` to delete the inserted like now, and plays around with the idea of using a unique identifier. Added a new thread the updated code block here: https://github.com/LnL7/nix-darwin/pull/228#discussion_r499774598

malob:
@LnL7, just wanted to check in and see if there was anything I could do to help get this merged in.

ethancedwards8:
I would also love for this to get merged as soon as possible! @LnL7

modules/security/pam.nix
```diff
@@ -0,0 +1,61 @@
...
+  mkSudoTouchIdAuthScript = isEnabled:
+  let
+    file   = "/etc/pam.d/sudo";
+    option = "security.pam.enableSudoTouchIdAuth";
+  in ''
+    ${if isEnabled then ''
+      # Enable sudo Touch ID authentication, if not already enabled
+      if ! grep 'pam_tid.so' ${file} > /dev/null; then
+        sed -i "" '2i\
+        auth       sufficient     pam_tid.so # nix-darwin: ${option}
```
esselius, line 27: The resulting `/etc/pam.d/sudo` contains accidental indention:

```
# sudo: auth account password session
  auth       sufficient     pam_tid.so # nix-darwin: security.pam.enableSudoTouchIdAuth
auth       sufficient     pam_smartcard.so
auth       required       pam_opendirectory.so
account    required       pam_permit.so
password   required       pam_deny.so
session    required       pam_permit.so
```

Still works though :D but might want to:

```suggestion
      auth       sufficient     pam_tid.so # nix-darwin: ${option}
```

malob: Good catch! Thanks :)

emilazy:
Gentle ping. Would love to see this merged so I don’t have to `sudo -e /etc/pam.d/sudo` every macOS upgrade.

ivankovnatsky:
this looks like ready and tested, any reason not to merge it?

ivankovnatsky:
maybe not the best solution, but added pam module to fix touch id under tmux: https://github.com/ivankovnatsky/nixos-config/commit/6b3c7462cb1c386ded995c685c98b66aa65a7716

rubenfonseca:
Why not merge this?

dhess:
FYI, you should change the use of `sed` in your script to `/usr/bin/sed`. If the user has a `sed` installed from Nixpkgs, it might not work with the syntax you're using.

shinzui:
Can this option also handle [pam_reattach](https://github.com/fabianishere/pam_reattach) which is also required to make sudo work in tmux?

shinzui:
Here is an example [config](https://github.com/ivankovnatsky/nixos-config/blob/6b3c7462cb1c386ded995c685c98b66aa65a7716/modules/darwin/security/pam.nix) that uses both.

malob:
@shinzui, I could imagine this module being updated to support this. Though I’d rather focus on getting the initial functionality merged before adding more complexity.

Speaking of which, @LnL7, anything I can do/change to get this merged?

dsyang:
question from a newbie to `nix` and `nix-darwin`: When I add this, it works beautifully, however if I do a `darwin-rebuild --rollback`, it appears this doesn't delete the added line from my `/etc/pam.d/sudo` file. I have to do `security.pam.enableSudoTouchIdAuth = false` and rebuild before it's removed.

Is this expected behavior?

ethancedwards8:
Can reproduce. Simply removing the option does not remove the addition,  you have to explicitly  set it to false. This probably is not expected behavior.

malob:
@dsyang, @ethancedwards8, huh that is surprising to me, you are right that this is not the expected behavior, though I'm not sure what the cause would be.

I tested this on my system (where I've got a [copy](https://github.com/malob/nixpkgs/blob/952c8bb6962d11c076008c44e72ad85d57284693/modules/darwin/security/pam.nix) of the module setup), and was not able to reproduce. Pretty sure the only difference between the two is that in the copy in my repo I'm setting `system.activationScripts.extraActivation.text` instead of `system.activationScripts.pam.text` so that I don't have to maintain an up to date copy of [modules/system/activation-scripts.nix](https://github.com/LnL7/nix-darwin/blob/master/modules/system/activation-scripts.nix). 

I don't see why that would change anything though. The `security.pam.enableSudoTouchIdAuth` option defaults to `false` so I don't know why not specifying the option as opposed to explicitly setting it to `false` would make a difference.

Any ideas?

malob:
(Just rebased on `master`.)

malob:
I just tested this module directly by changing the `nix-darwin` flake in my config to be `github:malob/nix-darwin/sudo-touchid` and still wasn't able to reproduce.

With the option not specified in my config `nix repl` shows that the option is indeed present and set to `false`,

```
nix-repl> darwinConfigurations.MaloBookProM1.config.security.pam.enableSudoTouchIdAuth
false
```

the activation script is set with the right content,

```
nix-repl> darwinConfigurations.MaloBookProM1.config.system.activationScripts.pam.text
"# PAM settings\necho >&2 \"setting up pam...\"\n# Disable sudo Touch ID authentication, if added by nix-darwin\nif grep 'security.pam.enableSudoTouchIdAuth' /etc/pam.d/sudo > /dev/null; then\n  sed -i \"\" '/security.pam.enableSudoTouchIdAuth/d' /etc/pam.d/sudo\nfi\n\n\n"
```

and after running a rebuild and switch, the relevant line from `/etc/pam.d/sudo` was removed:

```
❯ cat /etc/pam.d/sudo
# sudo: auth account password session
auth       sufficient     pam_smartcard.so
auth       required       pam_opendirectory.so
account    required       pam_permit.so
password   required       pam_deny.so
session    required       pam_permit.so
```

dsyang:
Hmm interesting. Could this be related to flakes? My setup did not use nix-flakes.

CorbanR:
@malob @dsyang I ran into the same issue where setting `security.pam.enableSudoTouchIdAuth = false` didn't remove the added text. However, it was only an issue for me when ` sed -i "" '/${option}/d' ${file}` vs ` /usr/bin/sed -i "" '/${option}/d' ${file}`. As @dhess mentions above, there is a possible  issue if the user has `gnused` installed via nix. The script will use the nix version of sed, which has different syntax than the one that comes installed at `/usr/bin/sed`

malob:
@CorbanR (cc: @dsyang, @dhess), gotcha. I've updated the the module to use `/usr/bin/sed` explicitly.

pecigonzalo:
@malob how about instead referencing `sed` from nixpkgs? That way you dont depend on the version being shipped by the OS.

malob:
@pecigonzalo, good point. I was being a little lazy since I knew the syntax wasn't exactly the same, but using GNU `sed` from `nixpkgs` does seem like the better move.

I've updated code in the PR to use GNU `sed` now and tested on my system.

pecigonzalo:
Thank you for implementing this, really looking forward to it.

dsyang:
@malob  I'm setting up a new computer today and idk what has changed but I'm not hitting the same bug anymore!

```
$> darwin-rebuild --list-generations  
   1   2022-03-09 10:38:35   
   2   2022-03-09 10:41:10   
   3   2022-03-09 10:45:53   
   4   2022-03-09 10:55:34   
   5   2022-03-09 11:00:02   (current)

$> git diff . 
diff --git a/darwin-configuration.nix b/darwin-configuration.nix
index ea78c2d..d1fab3c 100644
--- a/darwin-configuration.nix
+++ b/darwin-configuration.nix
@@ -88,7 +88,7 @@
   };
 
   # add touchid support to sudo
-  security.pam.enableSudoTouchIdAuth = true;
+  ### security.pam.enableSudoTouchIdAuth = true;
 
   # List packages installed in system profile. To search by name, run:
   # $ nix-env -qaP | grep wget


$> darwin-rebuild switch
trace: warning: literalExample is deprecated, use literalExpression instead, or use literalDocBook for a non-Nix description.
this derivation will be built:
  /nix/store/2rk0gyshrszriyn7dqn321p1g20sw7zk-darwin-system-22.05pre358993.3e644bd6248+darwin4.0000000.drv
building '/nix/store/2rk0gyshrszriyn7dqn321p1g20sw7zk-darwin-system-22.05pre358993.3e644bd6248+darwin4.0000000.drv'...
user defaults...
setting up pam...
setting up ~/Applications...
applying patches...
setting up /etc...
...

$> cat /etc/pam.d/sudo
# sudo: auth account password session
auth       sufficient     pam_smartcard.so
auth       required       pam_opendirectory.so
account    required       pam_permit.so
password   required       pam_deny.so
session    required       pam_permit.so

$> darwin-rebuild --list-generations
   1   2022-03-09 10:38:35   
   2   2022-03-09 10:41:10   
   3   2022-03-09 10:45:53   
   4   2022-03-09 10:55:34   
   5   2022-03-09 11:00:02   
   6   2022-03-09 11:05:37   (current)

$> darwin-rebuild -G 5
switching profile from version 6 to 5
user defaults...
setting up pam...
setting up ~/Applications...
applying patches...
setting up /etc...
...

$> cat /etc/pam.d/sudo
# sudo: auth account password session
auth       sufficient     pam_tid.so # nix-darwin: security.pam.enableSudoTouchIdAuth
auth       sufficient     pam_smartcard.so
auth       required       pam_opendirectory.so
account    required       pam_permit.so
password   required       pam_deny.so
session    required       pam_permit.so
```

My setup can be found at https://github.com/dsyang/nixfiles. The patch there doesn't have the new `gnused` changes though. Will try updating my patch next.

dsyang:
confirmed patching in the `gnused` changes  still work. I'm not sure what I ran into but it's gone...

CorbanR:
Tested and also works great for me!

malob:
@domenkozar, think there's a chance we could get this merged? Any changes you'd like to see before doing so?

lockejan:
What about devices without touchId? 🤔
There's no mechanism to prevent the user's from shooting themself in the foot. 
Maybe that's missing. Invalid modules would make Sudo completely unusable and force the user to boot into recovery mode.😬

ivankovnatsky:
> What about devices without touchId? thinking There's no mechanism to prevent the user's from shooting themself in the foot. Maybe that's missing. Invalid modules would make Sudo completely unusable and force the user to boot into recovery mode.grimacing

Yes, I've stumbled on something similar when configuring a new mac (but I have a bit more complex config), but you can always enable super user and edit the sudoers file as desired to fix this.

lockejan:
It's probably to unsafe in it's current state. See also https://github.com/LnL7/nix-darwin/issues/358#issuecomment-922476324

lockejan:
> I expect `pam_tid` would fail unconditionally on machines without Touch ID (and therefore fall back to password authentication, as it already does when Touch ID authentication fails). Have you tested this? I don't have any machines without Touch ID to verify this on, but I would be very surprised if the available PAM modules depend on the hardware configuration, given how macOS system volumes and updates work, and cancelling Touch ID authentication when it's offered already falls back to asking for a password.

You're probably right. At least that's the case on my machines, when my lid is closed and I use an external keyboard. However, it would be good to ensure the same goes for older machines not having any fingerprint sensors at all.

maxmouchet:
For what it's worth, it is completely transparent on macOS 12 VM. It directly asks for the password.

malob:
(Rebased on `master` to address merge conflict.)

tommyknows:
Just wanted to chime in here with a couple of points. First off, let me say that I'd love to see this merged! However, in my case I'm using `tmux`, which means I also need the [pam_reattach](https://github.com/fabianishere/pam_reattach) module to be installed and added to the `/etc/pam.d/sudo` file as well.

I think building the `pam_reattach` module & adding this as an option would _technically_ be possible, but definitely out of scope for this PR. And as I'm just getting started with Nix, I can't really provide a PR for this either. 

But it would be nice to also see an option for "additionalLines` or something. For example, I could image this to look something like:
```nix
{
  security.pam.enableSudoTouchIdAuth = true;
  security.pam.additionalLines = [
    "auth       optional       pam_reattach.so"
  ];
}
```

That would be super helpful!

DylanRJohnston-FZ:
I've extended it to handle pam_reattach https://github.com/DylanRJohnston/nixos/blob/main/common/nix-darwin/touchID.nix with pam_reattach being built as an overlay here https://github.com/DylanRJohnston/nixos/blob/main/packages/pam_reattach.nix

Enzime:
@domenkozar any chance you could take a look at this PR? I've been using it and it's been working quite well for me

I've also just tested it on my M1 Mac Mini without TouchID and it falls back to password seamlessly :)

domenkozar (approved):

malob:
❤️

modules/security/pam.nix
```diff
@@ -0,0 +1,62 @@
...
+
+let
+  cfg = config.security.pam;
+
+  # Implementation Notes
+  #
+  # We don't use `environment.etc` because this would require that the user manually delete
+  # `/etc/pam.d/sudo` which seems unwise given that applying the nix-darwin configuration requires
+  # sudo. We also can't use `system.patchs` since it only runs once, and so won't patch in the
+  # changes again after OS updates (which remove modifications to this file).
```
lilyball, line 13: What's this about `system.patches`? I haven't used it before but reading through its activation script the logic appears to be "apply the patch unless it's already applied" (where "already applied" is defined as "it can be successfully reversed"). The part that looks up the previously-applied patches is just to reverse any patches that have been removed.

Which is to say, it certainly appears as though it will attempt to repatch on every activation.

modules/security/pam.nix
```diff
@@ -0,0 +1,62 @@
...
+    file   = "/etc/pam.d/sudo";
+    option = "security.pam.enableSudoTouchIdAuth";
+    sed = "${pkgs.gnused}/bin/sed";
+  in ''
+    ${if isEnabled then ''
+      # Enable sudo Touch ID authentication, if not already enabled
+      if ! grep 'pam_tid.so' ${file} > /dev/null; then
+        ${sed} -i '2i\
+      auth       sufficient     pam_tid.so # nix-darwin: ${option}
+        ' ${file}
```
lilyball, line 29: This sed here assumes that nobody else has edited this file. If e.g. I delete the first line comment, then it will insert the `pam_tid.so` after the existing `pam_smartcard.so`. Or if I delete the comment and `pam_smartcard.so` then it will insert it after the `auto required pam_opendirectory.so`, which will make it not work at all.

modules/security/pam.nix
```diff
@@ -0,0 +1,62 @@
...
+      fi
+    '' else ''
+      # Disable sudo Touch ID authentication, if added by nix-darwin
+      if grep '${option}' ${file} > /dev/null; then
+        ${sed} -i '/${option}/d' ${file}
```
lilyball, line 34: Similarly this line here assumes that `pam_tid.so` only exists in the `auth sufficient pam_tid.so`. I don't know if there's any reason for it to exist otherwise, but I might add e.g. a comment about it that would also be deleted.

modules/security/pam.nix
```diff
@@ -0,0 +1,62 @@
...
+  # should be deleted when the option is disabled.
+  mkSudoTouchIdAuthScript = isEnabled:
+  let
+    file   = "/etc/pam.d/sudo";
+    option = "security.pam.enableSudoTouchIdAuth";
+    sed = "${pkgs.gnused}/bin/sed";
+  in ''
+    ${if isEnabled then ''
+      # Enable sudo Touch ID authentication, if not already enabled
+      if ! grep 'pam_tid.so' ${file} > /dev/null; then
```
lilyball, line 26: This grep can also be tricked. For example, if I add a comment containing the string `pam_tid.so` that will cause nix-darwin to skip actually enabling it.

Of course, with the previous `system.patches` approach any such changes would make the feature not work at all, but at least it wouldn't make potentially surprising changes to the file.

lilyball (commented):
Given the potential issues with the scenario where the user edits `/etc/pam.d/sudo` by hand, I think `system.patches` is a better approach. As I already commented, the activation script for `system.patches` appears as though it should reapply the patch on every activation, which makes me wonder why this module claims it won't.

Alternatively you could go with `environment.etc` and just add your own activation script that runs before `activationScripts.etc` and, if the option is disabled, checks if `/etc/pam.d/sudo` is a symlink to `/etc/static/pam.d/sudo` and replaces it with `/etc/pam.d/sudo.orig` instead (throwing an error if the latter doesn't exist).

I can personally confirm that using a symlink for `/etc/pam.d/sudo` works perfectly fine, I've been doing this myself for a while.

modules/security/pam.nix
maljub01: It turns out, `system.patches` is actually broken. This PR should fix it: https://github.com/LnL7/nix-darwin/pull/612

maljub01: With #612 applied to my local setup, I am currently using this configuration to enable touch ID:

```
  # Enable Enable sudo authentication with Touch ID
  system.patches = [
    (pkgs.writeText "pam.patch" ''
      --- a/etc/pam.d/sudo
      +++ b/etc/pam.d/sudo
      @@ -1,4 +1,6 @@
       # sudo: auth account password session
      +auth       optional       ${pkgs.pam-reattach}/lib/pam/pam_reattach.so  # Needed for using Touch ID within tmux
      +auth       sufficient     pam_tid.so
       auth       sufficient     pam_smartcard.so
       auth       required       pam_opendirectory.so
       account    required       pam_permit.so
    '')
  ];
```

This works pretty well and gets applied on every update. It is quite resilient and will continue to work even if I make changes to the patch (ex. changing the tmux related comment). This is because `system.patches` will reverse outdated patches before applying new ones.

modules/security/pam.nix
```diff
@@ -0,0 +1,62 @@
...
+    security.pam.enableSudoTouchIdAuth = mkEnableOption ''
+      Enable sudo authentication with Touch ID
+
+      When enabled, this option adds the following line to /etc/pam.d/sudo:
+
+          auth       sufficient     pam_tid.so
+
+      (Note that macOS resets this file when doing a system update. As such, sudo
+      authentication with Touch ID won't work after a system update until the nix-darwin
+      configuration is reapplied.)
```
winterqt, line 51: Apologies for bumping this three year old PR, but I figured this is the best place to ask this.

Isn't this not actually true due to the system activation service [being enabled by default](https://github.com/LnL7/nix-darwin/blob/eb2b9b64238349bd351561e32e260cac15db6f9a/modules/services/activate-system/default.nix#L13), which will reapply the PAM edits upon boot?

Assuming I'm correct, I think that leaving this untrue (by default) clarification in the documentation will just confuse users.

lilyball: What you linked doesn't run the entire activation script, it just runs part of it

https://github.com/LnL7/nix-darwin/blob/eb2b9b64238349bd351561e32e260cac15db6f9a/modules/services/activate-system/default.nix#L37-L39

I suppose this could have been updated to include the pam activation script too if someone thought about it. I actually didn't even realize this was here. Though at this point it's worth noting that https://github.com/LnL7/nix-darwin/pull/787 changes how all of this works.

PR #229, opened 2020-09-13
add mas module

zachcoyle:
This is very impure and has some behavior that at the very least fails the principle of least astonishment, so I'm not expecting this to necessarily get merged, but I did want to throw the concept out there. 

[mas](https://github.com/mas-cli/mas) is a tool to manage app store installations, but its [build system](https://github.com/mas-cli/mas/blob/master/script/build) is not something that looks like it's going to be easily packaged for darwin in the near future, at least purely.

I put my derivation for `mas` up in a gist here: https://gist.github.com/zachcoyle/1cb276b2b9633d9da3601608f5ba0c5d and just had that sitting in an overlay.  

Regardless, it was pretty fun learning how to build a module from scratch, and I have a better understanding of how they work now!

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -63,6 +63,7 @@
   ./programs/gnupg.nix
   ./programs/man.nix
   ./programs/info
+  ./programs/mas.nix
   ./programs/nix-index
   ./programs/ssh
   ./programs/tmux.nix
--- a/modules/programs/mas.nix
+++ b/modules/programs/mas.nix
@@ -1,0 +1,41 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let
+  cfg = config.programs.mas;
+in
+{
+  options = {
+    programs.mas.enable = mkOption {
+      type = types.bool;
+      default = false;
+      description = "Whether to enable mas.";
+    };
+
+    programs.mas.applications = mkOption {
+      type = types.listOf types.string;
+      default = [ ];
+      example = literalExample
+        ''
+          [
+          "497799835" # Xcode
+          ]
+        '';
+      description = "List of App Store applications for mas to install.";
+    };
+
+    programs.mas.package = mkOption {
+      internal = true;
+      type = types.package;
+    };
+  };
+
+  config = mkIf cfg.enable {
+    environment.systemPackages = [ cfg.package ];
+    system.activationScripts.extraUserActivation.text = ''
+      echo "setting up App Store applications..."
+      ${cfg.package}/bin/mas install ${lib.concatStringsSep " " cfg.applications}
+    '';
+
+  };
+}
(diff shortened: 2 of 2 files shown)


zachcoyle:
I was able to figure out how to package `mas` much better here: https://github.com/zachcoyle/nix-config/blob/master/pkgs/mas/default.nix and realized that combined with an activation script, this could actually work pretty well.

emilazy:
Yay, I was just looking for something like this!

Do you know how `mas install` interacts with upgrades? If it doesn't upgrade already-installed packages, maybe it'd be good to have an option to automatically run upgrades on activation too so that things are consistently the latest version rather than whatever version they happened to be when you first added or last manually upgraded them.

Also, not sure if this is intentional but it looks like the PR is missing the package definition.

zachcoyle:
@emilazy 

 `mas install` on a previously installed package just gives a warning:
```
❯ mas install 497799835
Warning: Xcode is already installed
```
But, running `mas upgrade` before mas install would ensure everything was up to date.

The package definition not being colocated with this module was intentional, but you can find it [here](https://github.com/zachcoyle/nix-config/blob/master/pkgs/mas/default.nix) currently.  It looks like 7z is capable of extracting a `.pkg` so I'm going to do a little more testing and see if `mas` could get into nixpkgs

malob:
Just wanted to drop a quick comment to mention that I've got a PR (https://github.com/LnL7/nix-darwin/pull/262) for a `homebrew` module, that looks like it's close to being merged. It generates a Brewfile, and runs `brew bundle` during activation using that Brewfile. Since Homebrew Bundle, has the ability install Mac App Store apps, using `mas`, it might provide the functionality you were looking for with this module. 

Even though there is overlap in the functionality of both modules, I don't think it's a replacement for this module though. It seems likely that there are folks who don't use Homebrew, who would appreciate the ability manage which Mac App Store apps are installed via `nix-darwin`.

zachcoyle:
`mas` is now in nixpkgs, so I'm going to be revisiting this PR

andrewcrook:
Just be aware mas currently has a lot of open issues around failure to get the correct version info for apps from the App Store. I would go as far as saying that it doesn't really work at the moment.

shyim:
I built the same by own and saw then this PR. Sad that its still not merged yet :(

lockejan:
> I built the same by own and saw then this PR. Sad that its still not merged yet :( 

Just in case you don't know, mas is already included in nix-Darwins homebrew module.✌️

Enzime:
Hey all, as `mas` is already available in the Homebrew module from #262, I'm going to close this PR. If anyone is interested, I would be interested in merging a refactor that pulls out `mas` into its own module, so that non-Homebrew users can also install apps using `mas`.

Issue #88, opened 2018-06-29
Removing defaults


kirelagin:
I am trying to add an option for the Dark Mode in Mojave.

When I enable it, it sets `NSGlobalDomain.AppleInterfaceStyle` to `Dark` and when I disable it, it simply removes the key. I am not sure how to handle this with the current `defaults-write` framework. Should I just make up an arbitrary value (e.g. `Light` and say that it is the default)?

To be honest, I don’t really know how defaults work on macOS, but I noticed that setting this key does not really switch the UI. Is this just how it works or should I look for some special command that will force it to apply the change?

LnL7:
I would love to make the defaults completely declarative, the main problem with it is determining whether a setting was defined by a previous nix-darwin configuration or the user through System Preferences, etc.

The `defaults` command basically just inspects and updates plist files. Various system services use these to store or inspect the state of settings. If you trace what happens when changing something with the System Preferences and the preferences pane of various applications you'll see that this usually just updates a plist file in ~/Library/Preferences.

eg. `plutil -convert xml1 ~/Library/Preferences/com.apple.dock.plist -o foo.plist` If you use any of the `system.defaults.dock` options you'll see those in there.

kirelagin:
When an option is set in nix-darwin, we clearly need to set it during activation. When an option is `null`, it is more complicated, but I think it would be reasonable to expect the user not to alter any options managed by nix-darwin manually, so we can just delete it. This will be consistent with options that are non-null and have a default value: if you remove it from the nix-darwin config, it will be reset to the default. Another possibility is to see whether the option was set in the previous configuration and delete it only if it was set previously but now became `null`.

> If you trace what happens when changing something with the System Preferences and the preferences pane of various applications you'll see that this usually just updates a plist file in ~/Library/Preferences

Well the second part of my question was alluding to the fact that changing the Preferences clearly does something else as well. E.g. when I change `autohide` of the Dock using `defaults`, the Dock doesn’t start/stop to hide and, furthermore, actually the checkbox in the Preferences doesn't toggle. So it seems that those preferences are used only for restoring preferences on boot and while the system is live, it uses some other mechanism.

Actually, now when I see that it is the same with Dock and likely all other Preferences, I think that it is expected that switching the Dark mode using `defaults` doesn’t work without a reboot. It would be nice to improve on this.

abathur:
[...]

PR #230, opened 2020-09-27
Add AppleAccentColor AppleHighlightColor defaults

zachcoyle:
Rewrite of PR #133 

This adds the ability to control NSGlobalDomain.AppleAccentColor and NSGlobalDomain.AppleHighlightColor. 
Following the behavior of System Preferences, if AppleAccentColor is specified, AppleHighlightColor will be set to match, unless a value is explicitly set. 

There is currently no way to set AppleHighlightColor to the system Blue, for reasons related to #88
For AppleHighlightColor, Blue is the absence of the default, but luckily for AppleAccentColor, Blue is either the absence of the default or 4.

I don't have descriptions or tests yet, but will get those added when I get a chance

### DIFF
--- a/modules/system/defaults/NSGlobalDomain.nix
+++ b/modules/system/defaults/NSGlobalDomain.nix
@@ -46,7 +46,7 @@ let
     };
     Blue = {
       accent = 4;
-      highlight = null;
+      highlight = "0.698039 0.843137 1.000000 Blue";
     };
     Purple = {
       accent = 5;
--- a/modules/system/defaults/NSGlobalDomain.nix
+++ b/modules/system/defaults/NSGlobalDomain.nix
@@ -98,6 +98,9 @@ in {
         (types.nullOr types.int);
       default = null;
       description = ''
+        Configures the color of native controls.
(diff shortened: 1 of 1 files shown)


zachcoyle:
I was able to grab the Blue value from Big Sur, and noticed that under Big Sur Blue is no longer the unset value.  The unset value is now app-specific multi-colored sidebars with blue controls, so that maybe would be a more acceptable "leave unset for the default" value?

modules/system/defaults/NSGlobalDomain.nix
```diff
@@ -27,7 +87,35 @@ in {
...
+      description = ''
+        Configures the color of native controls.
+        Catalina: The default is Blue
+        Big Sur: The default is multicolor
+      '';
+    };
+
+    system.defaults.NSGlobalDomain.AppleHighlightColor = mkOption {
+      type = types.coercedTo
+        (types.nullOr rgbDecimal)
```
zachcoyle, line 109: This should probably be updated to be able to accept either one of the named color strings or the rgb attrset for convenience

PR #235, opened 2020-10-18
flakes

LnL7:


README.md
```diff
@@ -61,6 +61,50 @@ Check out [modules/examples](https://github.com/LnL7/nix-darwin/tree/master/modu
...
+}
+```
+
+Since the installer doesn't work with flakes out of the box yet nix-darwin will need to
+be to be bootstrapped using the installer or manually.  Afterwards the flake based
+configuration can be built.  The `hostname(1)` of your system will be used to decide
+which darwin configuration is applied if it's not specified explicitly in the flake ref.
+
+```sh
+darwin-rebuild switch --flake ~/.config/darwin
```
Kloenk, line 105: What does this accept? Does it exempt flak refs? If so, should it be path:~/.config/darwin

pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -92,22 +112,48 @@ done
...
-if ! [ "$action" = build ]; then
+flakeFlags=(--experimental-features 'nix-command flakes')
+
+if [ -n "$flake" ]; then
+    if [[ $flake =~ ^(.*)\#([^\#\"]*)$ ]]; then
+       flake="${BASH_REMATCH[1]}"
+       flakeAttr="${BASH_REMATCH[2]}"
+    fi
+    if [ -z "$flakeAttr" ]; then
+      flakeAttr=$(hostname)
```
Kloenk, line 123: Will this break if I add the hostname attribute?

Kloenk (commented):
Looks good to me (but first have to get a real MacOS :-)).

I the long run we should add the example flake as an template, and add this repository to the flakes registry.

README.md
LnL7: I'm not sure about the exact terminology but but both `~/.config/darwin` and `path:$HOME/.config/darwin` work, but the latter doesn't expend ~ anymore it seems.

pkgs/nix-tools/darwin-rebuild.sh
LnL7: Not exactly sure what the question is, but this is the fallback in case the flake-ref doesn't include a hash. So these are equivalent.

```sh
darwin-rebuild switch --flake ~/.config/darwin
darwin-rebuild switch --flake ~/.config/darwin#Johns-MacBook.local
```

I did just notice that  on NixOS the nixosConfigurations prefix is always added, which makes sense so I'll fix that.

Kloenk: That was my question, did not see the fallback switch.

(Also not a fan of always prefixing it, but that's probably something to discuss at nixpkgs)

LnL7: I think it makes sense to namespace sets by their type but that makes accessing them a bit cumbersome.

PR #241, opened 2020-10-30
Add extraConfig option to Redis

awseward:
👋 Hi, I tested this out with this bit from the README:
```sh
darwin-rebuild switch -I darwin=.
```

It worked on macOS Catalina with the following in my `~/.nixpkgs/darwin-configuration.nix`:
```nix
{
  # ...

  services.redis = {
    enable = true;
    extraConfig = ''
      stop-writes-on-bgsave-error no
    '';
  };

  # ...
}
```

I'm happy to run any additional tests if necessary, but I'm fairly new to Nix so I might need a little pointing in the right direction.

### DIFF
--- a/modules/services/redis/default.nix
+++ b/modules/services/redis/default.nix
@@ -52,6 +52,12 @@ in
       default = false;
       description = "By default data is only periodically persisted to disk, enable this option to use an append-only file for improved persistence.";
     };
+
+    services.redis.extraConfig = mkOption {
+      type = types.lines;
+      default = "";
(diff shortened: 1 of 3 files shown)


LnL7:
Looks good. The tests are pretty basic you can base it on one of the other service tests. To run a test locally use eg. `nix-build release.nix -A tests.services-skhd`.

awseward:
Makes sense! Thanks for the pointer; I've gone and added a couple tests.

LnL7 (approved):
That looks perfect, thanks!

PR #242, opened 2020-10-31
add git, gnutar and gzip to lorri's path

akrmn:
When trying to use lorri, I kept getting errors similar to those mentioned by @wentasah on https://github.com/target/lorri/issues/385#issuecomment-661779666. I compared the relevant NixOS setting (shared by @curiousleo here https://github.com/target/lorri/issues/385#issuecomment-620477432) with the equivalent `darwin-nix` expression and I noticed that the one here lacked `git gnutar gzip`, so I added them.

### DIFF
--- a/tests/services-lorri.nix
+++ b/tests/services-lorri.nix
@@ -1,35 +1,27 @@
-{ config, pkgs, ... }:
+{ config, pkgs, lib, ... }:
 
 let
-  lorri = pkgs.runCommand "lorri-0.0.0" {} "mkdir $out";
   plistPath = "${config.out}/user/Library/LaunchAgents/org.nixos.lorri.plist";
-  actual = pkgs.runCommand "convert-plist-to-json" { buildInputs = [ pkgs.xcbuild ]; }
-    "plutil -convert json -o $out ${plistPath}";
-  actualJson = builtins.fromJSON (builtins.readFile "${actual.out}");
-  expectedJson = builtins.fromJSON ''
-  {
-    "EnvironmentVariables": {
-            "NIX_PATH": "${"nixpkgs="+ toString pkgs.path}",
-            "PATH": "${builtins.unsafeDiscardStringContext pkgs.nix}/bin"
-    },
-    "KeepAlive": true,
-    "Label": "org.nixos.lorri",
-    "ProcessType": "Background",
-    "ProgramArguments": [
-            "/bin/sh",
-            "-c",
-            "exec ${builtins.unsafeDiscardStringContext pkgs.lorri}/bin/lorri daemon"
-    ],
-    "RunAtLoad": true
-  }
-  '';
-  testResult = toString (actualJson == expectedJson);
+  expectedPath = "${lib.makeBinPath [config.nix.package pkgs.git pkgs.gnutar pkgs.gzip]}";
+  expectedNixPath = "${"nixpkgs="+ toString pkgs.path}";
 in
 {
   services.lorri.enable = true;
   test = ''
-    ${pkgs.xcbuild}/bin/plutil -lint ${plistPath}
-    [ ${testResult} ];
+    PATH=${lib.makeBinPath [ pkgs.xcbuild pkgs.jq ]}:$PATH
+
+    plutil -lint ${plistPath}
+    plutil -convert json -o service.json ${plistPath}
+
+    <service.json jq -e ".EnvironmentVariables.PATH     == \"${expectedPath}\""
+    <service.json jq -e ".EnvironmentVariables.NIX_PATH == \"${expectedNixPath}\""
+    <service.json jq -e ".KeepAlive                     == true"
+    <service.json jq -e ".Label                         == \"org.nixos.lorri\""
+    <service.json jq -e ".ProcessType                   == \"Background\""
+    <service.json jq -e ".ProgramArguments|length       == 3"
+    <service.json jq -e ".ProgramArguments[0]           == \"/bin/sh\""
+    <service.json jq -e ".ProgramArguments[1]           == \"-c\""
+    <service.json jq -e ".ProgramArguments[2]           == \"exec ${pkgs.lorri}/bin/lorri daemon\""
+    <service.json jq -e ".RunAtLoad                     == true"
   '';
 }
-
--- a/modules/services/lorri.nix
+++ b/modules/services/lorri.nix
@@ -32,7 +32,7 @@ in
     environment.systemPackages = [ pkgs.lorri ];
     launchd.user.agents.lorri = {
       command = with pkgs; "${lorri}/bin/lorri daemon";
-      path = with pkgs; [ nix ];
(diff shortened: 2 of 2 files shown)


LnL7:
Hmm, seems like this got lost in https://github.com/LnL7/nix-darwin/pull/222#discussion_r479796839. Looks good if you take a look at the test.

akrmn:
got it!

tests/services-lorri.nix
```diff
@@ -6,11 +6,19 @@ let
   actual = pkgs.runCommand "convert-plist-to-json" { buildInputs = [ pkgs.xcbuild ]; }
     "plutil -convert json -o $out ${plistPath}";
   actualJson = builtins.fromJSON (builtins.readFile "${actual.out}");
+  expectedPathItems = map (x: "${x}/bin")
```
akrmn, line 9: I'm not sure if this is the best way to get this done, I'm still learning nix

LnL7: Yeah, this test is is a bit overly specific.

No need to discard the string context tho since the packages will be referenced by the build anyway. To get rid of the dependencies in the test an overlay is needed in this case, but unlike eg. lorri they're pretty standard in this case so it's not a big deal. I mostly use this to avoid having to download a bunch of unnecessary stuff for the tests or in case the package in nixpkgs breaks I don't really want that to influence these tests.

```nix
{
  nixpkgs.overlays = [
    (self: super:
    {
      git = super.runCommand "git-0.0.0" {} "mkdir $out";
      gnutar = super.runCommand "gnutar-0.0.0" {} "mkdir $out";
      gzip = super.runCommand "gzip-0.0.0" {} "mkdir $out";
    })
  ];
}
```

akrmn: thanks for the feedback! I tried adding the `nixpkgs.overlays`section to the expression at [1] while removing the calls to `builtins.unsafeDiscardStringContext`, but I got an error related to `tar` [2]. I probably need to place the overlays somewhere else, but I'm a bit lost now.

[1] https://github.com/LnL7/nix-darwin/blob/ee9e1205e7892cb012a276ac158600d987cec815/tests/services-lorri.nix#L36-L42

[2]
```
builder for '/nix/store/0rz1d1d2qfy17y37gflkp0pswpp565fm-perl-5.32.0.drv' failed with exit code 1; last 4 log lines:
  unpacking sources
  unpacking source archive /nix/store/bl6b2p0aw8phljjn0s3xzyghbbzawl2r-perl-5.32.0.tar.gz
  /nix/store/9zhvxgmqwph7cd2mc9fwg9y39nziqdaz-stdenv-darwin/setup: line 847: tar: command not found
  do not know how to unpack source archive /nix/store/bl6b2p0aw8phljjn0s3xzyghbbzawl2r-perl-5.32.0.tar.gz
```

LnL7: Yeah this was more an example for reference. These dependencies specifically are very low level and used by many other things so actually replacing them breaks various things like you noticed.

akrmn: gotcha! is it fine then to keep the `builtins.unsafeDiscardStringContext` calls? is there anything else you'd like to see on this pr?

LnL7: No neither the discard context or overlays are needed.

akrmn: hm if I remove the discard context I get this error:
```
nix-repl> :b tests.services-lorri
error: the string '{
  "EnvironmentVariables": {
          "NIX_PATH": "nixpkgs=/nix/store/12pl142h03dcnphqmf1qq1v7pdv12pck-nixpkgs-21.03pre249027.f4254fba4f2/nixpkgs",
          "PATH": "/nix/store/7srl5rd0p8svg4mfwsqc118pxlwx9204-nix-2.3.8/bin:/nix/store/psvzvnzfcsdrlaibzsv7s9wp0ni96978-git-2.28.0/bin:/nix/store/56dw9pwdma8iqxx4a7ibi4l72mcm0px9-gnutar-1.32/bin:/nix/store/y64kgdcxycv5dy62jwffyyh20306xpfr-gzip-1.10/bin"
  },
  "KeepAlive": true,
  "Label": "org.nixos.lorri",
  "ProcessType": "Background",
  "ProgramArguments": [
          "/bin/sh",
          "-c",
          "exec /nix/store/pr02204rvl68n7zggz4jag8ya6jhaaml-lorri-1.2/bin/lorri daemon"
  ],
  "RunAtLoad": true
}
' is not allowed to refer to a store path (such as '!out!/nix/store/49wqxqnnp0aqz7s8ipcnd8xwhbr0bn44-gnutar-1.32.drv'), at /Users/mal/git/nix-darwin/tests/services-lorri.nix:17:18
```

LnL7: Oh I see now, this was there already. Using an attrset instead of going through fromJSON would fix that or just simply checking some specific keys in the build with eg. `jq`. But it's ok to leave that as is if you don't want to spend time on that.

akrmn: > Oh I see now, this was there already
Exactly, it was there before :)

> it's ok to leave that as is if you don't want to spend time on that.
I'm more interested in getting it right :)

> Using an attrset instead of going through fromJSON would fix that

However, I got a bit stuck trying to avoid the `fromJSON` conversion. If I use an attrset and then convert that with `builtins.toJSON` to compare with the actual value, it turns out that the `toJSON` conversion only goes one level deep.

> simply checking some specific keys in the build with eg. jq

This made me think that perhaps the way is avoiding JSON completely and working directly with the plist, but it seems there's very little support for that. The `plutil` command is fairly limited, it seems to lack a simple way to read out a key. I also tried `defaults read`, but it doesn't work with nested dictionaries. My last hope was `PlistBuddy`, which I learned about online. It seems to be part of macOS but it's absent from the PATH; it needs to be run directly as `/usr/libexec/PlistBuddy`. This one is very complete, but I'm not sure if there's a "nix way" to invoke it, and I couldn't find it under `pkgs`

LnL7: I would suggest moving the checks to the build instead like most of the other tests do, but unlike there use the plutil to check specific keys instead just using grep like the other cases do.  Checking the entire content is a bit overkill anyway.

So for example something like this:

```sh
# ...
{ 
   services.lorri.enable = true; 
   test = ''
     export PATH=${lib.makeBinPath [ pkgs.xcbuild pkgs.jq ]}
     plutil -convert json -o service.json ${service}
     [ $(jq -r '.ProgramArguments[-1]' service.json) = "exec ${pkgs.lorri}/bin/lorri daemon" ]
   ''; 
} 
```

akrmn: aha! that makes a lot of sense, thanks! I'll work on it later today

akrmn: done in d037dfa :) please let me know what you think. Thanks for the help!

LnL7: That's great, thanks!

LnL7 (approved):

PR #245, opened 2020-11-05
modules: add nextdns service

happysalada:
Adding the nextdns service.

I've tested by adding the corresponding launchd entry in my `darwin-configuration`. I don't know how to test the default.nix file.

It's pretty much taken from https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/networking/nextdns.nix
just switching systemd to launchd

### DIFF
--- a/modules/services/nextdns/default.nix
+++ b/modules/services/nextdns/default.nix
@@ -28,7 +28,7 @@ in {
 
     environment.systemPackages = [ nextdns ];
 
-    launchd.user.agents.nextdns = {
+    launchd.daemons.nextdns = {
       path = [ nextdns ];
       script = ''
         "${pkgs.nextdns}/bin/nextdns run ${escapeShellArgs cfg.arguments}";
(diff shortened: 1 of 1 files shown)


modules/services/nextdns/default.nix
```diff
@@ -0,0 +1,42 @@
...
+        description = "Additional arguments to be passed to nextdns run.";
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+
+    environment.systemPackages = [ nextdns ];
+
+    launchd.user.agents.nextdns = {
```
LnL7, line 31: Doesn't this need to run as root unless a custom port is specified?

happysalada: Yes it runs on port 53
Do I need to do something special?

When I added that part exactly to my `darwin-configuration` it worked, but if there is another way to test it, I'm interested. Maybe just like home-manager there is a way to specify a darwin repo, they have something like `home-manager -I home-manager=$HOME/Documents/Projects/home-manager switch`

LnL7: In that case this needs to be `launchd.daemons.nextdns`, agents can only run as your local user.

happysalada: thanks! fixed it.

LnL7 (approved):

PR #247, opened 2020-11-10
nextdns: add to modulelist

happysalada:
I didn't know I needed to add nextdns to a modulelist.

I kept on wondering why it wasn't working.

Sorry about that!

### DIFF
--- a/tests/services-nextdns.nix
+++ b/tests/services-nextdns.nix
@@ -1,0 +1,17 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let nextdns = pkgs.runCommand "nextdns-0.0.0" { } "mkdir $out";
+
+in {
+  services.nextdns.enable = true;
+  services.nextdns.arguments = [ "-config" "10.0.3.0/24=abcdef" ];
+
+  test = ''
+    echo >&2 "checking nextdns service in ~/Library/LaunchDaemons"
+    grep "org.nixos.nextdns" ${config.out}/Library/LaunchDaemons/org.nixos.nextdns.plist
+    grep "/bin/nextdns" ${config.out}/Library/LaunchDaemons/org.nixos.nextdns.plist
+    grep "'-config' '10.0.3.0/24=abcdef'" ${config.out}/Library/LaunchDaemons/org.nixos.nextdns.plist
+  '';
+}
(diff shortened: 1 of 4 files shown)


happysalada:
@LnL7 when you have a moment, can you check?
If you're busy, no rush.

LnL7 (commented):
Could you add a test so nothing else trivial like this fails in the future.

happysalada:
@LnL7 thanks for the feedback!

I've added a test based on the skhd test file.
I've run the test once before adding the file, and after when I tried running it again, it was just giving me the cached result. So I couldn't check on local if that test passes. Not sure how to do another nix-build with the contents of the new file.

Let me know if anything.

tests/services-nextdns.nix
```diff
@@ -0,0 +1,21 @@
...
+  nextdns = pkgs.runCommand "nextdns-0.0.0" {} "mkdir $out";
+in
+
+{
+  services.nextdns.enable = true;
+  services.nextdns.arguments = ["-config" "10.0.3.0/24=abcdef"];
+
+  test = ''
+    echo >&2 "checking nextdns service in ~/Library/LaunchAgents"
+    grep "org.nixos.nextdns" ${config.out}/user/Library/LaunchAgents/org.nixos.nextdns.plist
```
LnL7, line 15:
```suggestion
    grep "org.nixos.nextdns" ${config.out}/Library/LaunchDaemons/org.nixos.nextdns.plist
```

Since this is a demon and not an agent.

LnL7 (changes requested):

tests/services-nextdns.nix
happysalada: Thanks!

happysalada:
@LnL7 Thanks a lot for the help. The tests are passing.
Let me know if anything!

happysalada:
@LnL7 when you have a moment, if you can have another look.
No rush if you're busy.

LnL7 (approved):
Looks good, could you squash these commits?

happysalada:
@LnL7 Thanks for the help on this one!
Should be good to merge.

happysalada:
@LnL7 can you check again when you have a moment?

LnL7 (approved):

PR #252, opened 2020-11-16
emacs: add system path env var

### LINKED ISSUE #250: Wrong PATH set on emacs service
When running emacs with the service, the PATH is set to the default macos PATH (something like `/usr/bin:/bin...`)
Emacs won't be able to find any of the executables installed with nix.

I've tried the workaround suggested here 
https://github.com/LnL7/nix-darwin/issues/52

which is 
`  launchd.user.envVariables.PATH = config.environment.systemPath;`
into the `darwin-configuration.nix`

That didn't work either. I'm not sure why. but when I run `get-env PATH` in emacs I still have the same default macos PATH.

@peel maybe you have a workaround?
(no rush if you're busy)

happysalada:
Thanks for the help on this one!

Currently running the emacs daemon, will not set the path to the system path. So anything installed with nix won't be found (ripgrep, things for fancy git diffs...). This PR fixes that.

I've tested the fix by adding the same line in my `darwin-configuration.nix` file and it worked. Let me know if you want me to do something else.

I was running with the automated formatter on, so a couple of lines have changed formatting. It doesn't look like a bad thing, but let me know if it will be a problem.

closes #250 

I have question though. After using this, when I check the `getenv PATH` in emacs, I see `$HOME/.nix-profile/bin` however it looks like that  `$HOME` is hardcode and not properly replaced by it's value. I'm not sure how to fix that.

### DIFF
--- a/modules/services/emacs.nix
+++ b/modules/services/emacs.nix
@@ -6,9 +6,7 @@ let
 
   cfg = config.services.emacs;
 
-in
-
-{
+in {
   options = {
     services.emacs = {
       enable = mkOption {
@@ -23,6 +21,17 @@ in
         description = "This option specifies the emacs package to use.";
       };
 
+      additionalPath = mkOption {
+        type = types.listOf types.str;
+        default = [ ];
+        example = [ "/Users/my_user_name" ];
+        description = ''
+          This option specifies additional PATH that the emacs daemon would have.
+          Typically if you have binaries in your home directory that is what you would add your home path here.
+          One caveat is that there won't be shell variable expansion, so you can't use $HOME for example
+        '';
+      };
+
       exec = mkOption {
         type = types.str;
         default = "emacs";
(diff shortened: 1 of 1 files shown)


MaxSchlueter:
I have the same problem. The packages installed for my user are not being picked up by the emacs daemon.

As a workaround you can set `path` to the following: `(lib.replaceStrings ["$HOME"] [config.users.users.user.home] config.environment.systemPath)`

I guess one could also overwrite the value of `environment.systemPath` directly on a single user system. I don't know how to resolve this problem generally though, when there are multiple users of a system.

LnL7 (approved):
I don't think there's a way to expand variables in launchd services so handling HOME would require explicitly referencing the user's absolute homedir, which isn't correct either. This fixes the main problem and I feel like only having declarative things in PATH for a service isn't necessarily a bad thing.

Maybe you should add a note about this behaviour to the service description so it's a bit easier to discover.

happysalada:
Just thinking out loud here. Perhaps we can add an option to add to the path seen by emacs. I've added a additionalPath config option. There is a description accompanying it.
It seems it can solve both our problems, since we can just add to the path.
Curious to here your thoughts.

happysalada:
@LnL7 can you check again when you have a moment?

LnL7 (approved):
prepending path is pretty easy with the existing options but the option gives a nice place to describe the behaviour.

PR #253, opened 2020-11-17
More system.defaults

einsweniger:
This PR adds support for 
- com.apple.{desktopservices,TextEdit,TimeMachine} settings
- `-array` types from `types.listOf`

TimeMachine is disabled, because it does not seem writable (`Could not write domain /Library/Preferences/com.apple.TimeMachine; exiting`) and I don't know why that is.

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -11,6 +11,7 @@
   ./system/defaults/LaunchServices.nix
   ./system/defaults/NSGlobalDomain.nix
   ./system/defaults/GlobalPreferences.nix
+  ./system/defaults/desktopservices.nix
   ./system/defaults/dock.nix
   ./system/defaults/finder.nix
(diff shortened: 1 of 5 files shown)


modules/module-list.nix
```diff
@@ -19,6 +20,9 @@
   ./system/defaults/smb.nix
   ./system/defaults/SoftwareUpdate.nix
   ./system/defaults/spaces.nix
+  ./system/defaults/textedit.nix
+  # timemachine is not writable
+  #./system/defaults/timemachine.nix
```
LnL7, line 25: What's the reason for adding this if it can't be configured?

einsweniger: I was hoping, that somebody might know an answer to why it's not writable.
If nobody knows, I'll remove the lines and the file.
I guess we could leave the list conversion in, could be useful for other properties

PR #254, opened 2020-11-21
Add /bin/dash to shells

stephank:
I just installed the 10.15.7 upgrade, and noticed it replaced `/etc/shells` with a version that includes dash. I'm not sure which macOS release adds this, and if we should make this conditional somehow, or if this change matters at all.

### DIFF
--- a/modules/system/shells.nix
+++ b/modules/system/shells.nix
@@ -31,6 +31,7 @@ in
 
       /bin/bash
       /bin/csh
+      /bin/dash
       /bin/ksh
       /bin/sh
       /bin/tcsh


LnL7 (approved):
Thanks!

emilazy:
Does the hash of the default `/etc/shells` need updating elsewhere too so people don't have to manually remove it on first activation?

stephank:
> Does the hash of the default `/etc/shells` need updating elsewhere too so people don't have to manually remove it on first activation?

As in `knownSha256Hashes`? It appears it wasn't set for `/etc/shells`, but I can make a PR for it.

LnL7:
Yeah, I didn't add that for /etc/shells as I don't know what happens if that file becomes unreadable. Contrary to bashrc and friends it might be very hard to recover from.

PR #257, opened 2020-11-29
add backupFileExtension option to environment.nix

### LINKED ISSUE #149: Allow enforcing linking etc when file exists

kclejeune:
Introduces an option `environment.backupFileExtension` (based on the home-manager option) which, if specified, moves `/etc` files `$file` to `$file.${config.environment.backupFileExtension}` during system activation. Additionally, this updates moving files with known content hashes to keep backup filenames consistent. If the option is not specified, all behavior is unchanged.

Fixes #149 and closes #150 by updating changes from  @domenkozar to work with the existing codebase.


LnL7 (commented):
I like the idea here but would change it up a bit. One of the reservations I have with replacing existing files is that certain things can be pretty unsafe or difficult to recover from. As an example switching `/etc/sudoers` with a syntax error is quite tricky to rollback if the root user isn't enabled.

So instead of defining an extension defining an option with the known etc files that are safe to replace might be a good approach. This is very similar to the behaviour of some of the other options and also means the behaviour can be enabled by default for files like zshrc which can be moved back easily.

kclejeune:
> I like the idea here but would change it up a bit. One of the reservations I have with replacing existing files is that certain things can be pretty unsafe or difficult to recover from. As an example switching `/etc/sudoers` with a syntax error is quite tricky to rollback if the root user isn't enabled.
> 
> 
> 
> So instead of defining an extension defining an option with the known etc files that are safe to replace might be a good approach. This is very similar to the behaviour of some of the other options and also means the behaviour can be enabled by default for files like zshrc which can be moved back easily.

I can see where you're coming from with something like /etc/sudoers, but I still think the general use case for this warrants backing up as a default - what if we added a protected or excluded etc files option with sensible defaults like

`environment.protectedEtcFiles = [ /etc/sudoers ];`

and treat those cases as the exception rather than the rule?

LnL7:
Firstly making a backup regardless of everything else is indeed an obvious improvement.

As for using a blacklist I feel like that would be really hard, there are various other things like `/etc/sudoers.d/*`, `/etc/pam.d/*`, etc. fall under the same category which quickly gets out of hand. With sane defaults the only disadvantage of a whitelist are things that don't exist by default but where created by something anyway, but that seems more of an edge-case.

kclejeune:
> Firstly making a backup regardless of everything else is indeed an obvious improvement.
> 
> 
> 
> As for using a blacklist I feel like that would be really hard, there are various other things like `/etc/sudoers.d/*`, `/etc/pam.d/*`, etc. fall under the same category which quickly gets out of hand. With sane defaults the only disadvantage of a whitelist are things that don't exist by default but where created by something anyway, but that seems more of an edge-case.

This is a good point, but I'm wondering if there are any modules that actually modify these directories if they're known to be harmful? It seems counterintuitive to allow such a module in the first place, which would sort of eliminate the need for the blacklist that I described previously.

LnL7:
Yeah, but the problem is that things like `environment.etc` are freeform. The nice thing is that this is compatible with NixOS so modules that only use shared options can be used in both places but that's where assumptions can collide. NixOS is the host system so it doesn't need to play nice with anything else and it has more robust ways to rollback a broken system through the boot menu.

kclejeune:
> Yeah, but the problem is that things like `environment.etc` are freeform. The nice thing is that this is compatible with NixOS so modules that only use shared options can be used in both places but that's where assumptions can collide. NixOS is the host system so it doesn't need to play nice with anything else and it has more robust ways to rollback a broken system through the boot menu.

To regroup on this, is an `etc` whitelist the only solution that you'd be comfortable with in order to merge this?

LnL7:
I'd be fine with an option to explicitly allow replacing existing files. But unlike the whitelist I wouldn't want enable that by default, making it inferior since it would need to be set explicitly and while enabled still has the pitfalls mentioned earlier.
It's essentially the same except more granular since just like setting the flag extending the whitelist can also always be done from the user's configuration (with the same disclaimer as the flag).

```nix
{
    environment.etc."passwd".text = ''
      nobody:*:-2:-2:Unprivileged User:/var/empty:/usr/bin/false
      root:*:0:0:System Administrator:/var/root:/bin/sh
    '';
    environment.knownEtcFiles = [ "/etc/passwd" ];
}
```

kclejeune:
> I'd be fine with an option to explicitly allow replacing existing files. But unlike the whitelist I wouldn't want enable that by default, making it inferior since it would need to be set explicitly and while enabled still has the pitfalls mentioned earlier.
> It's essentially the same except more granular since just like setting the flag extending the whitelist can also always be done from the user's configuration (with the same disclaimer as the flag).

I've been a little busy recently, but I'm looking at this again and got an idea - how would you feel about adding a boolean `backup` option? It feels nicer than maintaining a separate list of files. Something like 

```nix
{
    environment.etc."passwd" = {
        backup = true;
        text = ''
          nobody:*:-2:-2:Unprivileged User:/var/empty:/usr/bin/false
          root:*:0:0:System Administrator:/var/root:/bin/sh
        '';
    };
}
```

PR #262, opened 2020-12-08
Add Homebrew Bundle module

malob:
## Overview and motivation

This module adds the ability for `nix-darwin` to manage packages/apps installed by Homebrew Bundle. This gives `nix-darwin` users the ability to not only manage packages/apps installed via Homebrew, but also apps from the Mac App Store via `mas`, and Docker containers via `whalebrew`.

While I wish I could manage installing all my packages/apps directly through Nix via `nix-darwin` there are a sizable number of cases where that currently just isn't feasible at present. I expect many other users are in the same boat. I think this module, or something like it, provides a good compromise, by allowing users to move more of their macOS system configuration into Nix while leveraging Homebrew Bundle under the hood.

Here's an example of this module in use in my current config:
https://github.com/malob/nixpkgs/blob/master/darwin/homebrew.nix

It's been working quite well in my testing so far, both on my system, and my GitHub [workflow](https://github.com/malob/nixpkgs/runs/1567099284?check_suite_focus=true#step:7:2743).

Suggestions for changes/improvements are of course welcome.
 
## Implementation details

The options,
* `homebrew.taps`,
* `homebrew.brews`,
* `homebrew.casks`,
* `homebrew.masApps`,
* `homebrew.whalebrews`,

are used to generate a Brewfile in the Nix store which defines which packages/apps Homebrew Bundle will install/upgrade. (The `homebrew.extraConfig` option is also provided to allow users to include more complex Brewfile entries.) 

The activation script runs `brew bundle` using the generated Brewfile, with the `HOMEBREW_NO_AUTO_UPDATE` environment variable set to ensure idempotence. The `homebrew.autoUpdate` option is provided for users who want to allow Homebrew to auto-update during activation.

The `homebrew.cleanup` option is provided to allow the user to enable Homebrew Bundle uninstall packages/apps that are currently installed but aren't present in the generated Brewfile. I made the default `"none"` so that new users of this module would be less likely to uninstall packages/apps accidentally.

More detailed documentation is provided in the module.

(Edit * 3: Updated to reflect changes since creating PR)


malob:
Alrighty, this seems like it's in a good state for initial review. I'll add some review comments with thoughts/questions I have.

modules/programs/brew-bundle.nix
```diff
@@ -0,0 +1,213 @@
...
+    environment.variables = mkIf cfg.enable (
+      (if cfg.setNoLockEnvvar then { HOMEBREW_BUNDLE_NO_LOCK = "1"; } else {}) //
+      (if cfg.setBrewfileEnvvar then { HOMEBREW_BUNDLE_FILE = "${brewfile}"; } else {})
+    );
+
+    system.activationScripts.brew-bundle.text = mkIf cfg.enable ''
+      # Homebrew Bundle
+      echo >&2 "Homebrew bundle..."
+      PATH=/usr/local/bin:$PATH brew update > /dev/null
+      PATH=/usr/local/bin:$PATH brew bundle ${brew-bunble-options}
```
malob, line 209: Currently during activation `brew update` is run before `brew bundle`. I set it up that way since I wanted Homebrew to update things whenever I ran `darwin-rebuild switch`. This obviously isn't idempotent, so maybe the activation script shouldn't run `brew update`, and `brew bundle` should run with the `HOMEBREW_NO_AUTO_UPDATE` environment variable set. 

Does that sound right? If so, thoughts on adding an option to enable `brew update` on activation?

malob (commented):
See my review comments for specific thoughts/questions.

I expect we'll also want test for this module. However, I'm not sure what they should be since it's my understanding that activation scripts aren't run when running tests, and most of the things I'd want to test are that things were successfully installed etc.

If there are suggestions for what should be tested. I'm happy write the test up.

modules/programs/brew-bundle.nix
kclejeune: Personally, the sluggishness of `brew update` is one of the things that drove me toward nix in the first place, so I'd like to see this be handled manually, or to at least be the default if that's an option. Similar to a flake/niv/nix-channel update, I don't want packages to change unless I specify it myself.

malob: Alrighty, I just pushed a commit that removes the invocation of `brew update`, and adds an option `noAutoUpdate` which is defaults to `true`. For the reason's you mentioned, as well as idempotence, this seems like the obvious choice. It may even be the case that it's not worth including the `noAutoUpdate` option at all and just hardcoding this behavior, but I'm inclined to leave it in for now and defer to whatever @LnL7 thinks is best when they have a chance to way in.

modules/programs/brew-bundle.nix
```diff
@@ -0,0 +1,213 @@
+# Created by: https://github.com/malob
+# Inspired by: https://github.com/lccambiaghi/nixpkgs/blob/main/modules/homebrew.nix
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.programs.brew-bundle;
```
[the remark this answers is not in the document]
eraserhd: How about `config.homebrew`?  I agree that homebrew isn't really a program, and I think the "bundle" bit is really an implementation detail.

modules/programs/brew-bundle.nix
```diff
@@ -0,0 +1,225 @@
...
+    (if cfg.noAutoUpdate then "1" else "0") +
+    " brew bundle --file='${brewfile}' --no-lock" +
+    (if cfg.cleanupType == "uninstall" || cfg.cleanupType == "zap" then " --cleanup" else "") +
+    (if cfg.cleanupType == "zap" then " --zap" else "");
+
+in
+
+{
+  options.programs.brew-bundle = {
+    enable = mkEnableOption ''
```
eraserhd, line 42: I think having an enable option could potentially trip up user configs.  It would make any module that relies on this want to set enable = true, which might confuse the user if they have enable = false; alternately if the module doesn't set enable = true, the user might be confused why it isn't working.

I'd suggest the feature is disabled if no formulae/casks/etc. are specified.

modules/programs/brew-bundle.nix
```diff
@@ -0,0 +1,213 @@
...
+      configuring your Brewfile, and installing/updating the formulas therein via
+      the <command>brew bundle</command> command, using <command>nix-darwin</command>.
+
+      Note that enabling this option does not install Homebrew. See the Homebrew website for
+      installation instructions: https://brew.sh
+    '';
+
+    cleanupType = mkOption {
```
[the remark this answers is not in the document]
eraserhd: Actually, how about just `clean`?

modules/programs/brew-bundle.nix
```diff
@@ -0,0 +1,213 @@
...
+        formula is a cask, removes all files associated with the cask. In other words,
+        <command>brew uninstall --zap</command> is run for all those formulas.
+
+        If you plan on exclusively using <command>nix-darwin</command> to manage formulas installed
+        by Homebrew, you probably want to set this option to <literal>"uninstall"</literal> or
+        <literal>"zap"</literal>.
+      '';
+    };
+
+    setNoLockEnvvar = mkOption {
```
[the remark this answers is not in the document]
eraserhd: That it is an environment variable is an implementation detail, IMHO.  And also "set" signals a kind of imperative rather than declarative intent.  I think `lockfile`.

modules/programs/brew-bundle.nix
```diff
@@ -0,0 +1,225 @@
...
+      type = types.bool;
+      default = true;
+      description = ''
+        Sets the <literal>HOMEBREW_BUNDLE_NO_LOCK</literal> enviroment variable, by adding it to
+        <option>environment.variables</option>, so that lock files aren't generated when/if you run
+        the <command>brew bundle</command> command yourself.
+      '';
+    };
+
+    setBrewfileEnvvar = mkOption {
```
eraserhd, line 99: hmmm.  For naming this, do you use this option, and if so, why?  I'm only coming up with names like `globalConfig` or something, which is super unclear.

modules/programs/brew-bundle.nix
eraserhd: Can I propose `autoUpdate = false` instead of `noAutoUpdate = true`?

modules/programs/brew-bundle.nix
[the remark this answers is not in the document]
malob: Yeah, that might make sense, though it does feel a little odd to add something new to the top-level namespace. I'll wait to hear from @LnL7 about where he thinks it should go before making a change here for now.

modules/programs/brew-bundle.nix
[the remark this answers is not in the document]
malob: So one of the heuristics I've been using when naming options for this module, is to try and stick to that names/terms that Homebrew itself uses. The idea being that existing users of Homebrew Bundle will wind the options familiar. In this case the flag for this behavior is `--cleanup`.

modules/programs/brew-bundle.nix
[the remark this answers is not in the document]
malob: Good point about using the word "set" here, and the fact that the name is probably to anchored the the implementation details.

One thing I'm trying to track with the options in this module is that some effect what happens when the activation script runs, while others (like this one) effect what happens when you run the `brew bundle` command yourself. During activation a lockfile is never generated, and so it would be nice if the option name could make it clear somehow that this  effects how `brew bundle` works when the user runs it. Maybe it's just to hard/cumbersome to communicate that in the name, and so the description will have to suffice.

modules/programs/brew-bundle.nix
malob: This is another example of me trying to stick to using the terms that Homebrew Bundle itself uses, where in this case it's the `HOMEBREW_NO_AUTO_UPDATE` environment variable. In this case though I'd be pretty happy to change it as you suggest if others agree.

modules/programs/brew-bundle.nix
malob: So I'm under the impression that having an `enable` option like this for a module if pretty idiomatic in the `Nix` ecosystem (though there are definitely exceptions), and so shouldn't trip up users. Also, a nice property of having an explicit enable option, is that it makes it easy to temporarily turn off the functionality the module provides without having to remove all the associated configuration of the various options.

Totally open to counter arguments, if you or others really feel like I'm off base here.

modules/programs/brew-bundle.nix
malob: Yeah, I do use it. My main use case is that there are times where I want to manually invoke `brew bundle` to, e.g., run a cleanup after I've installed some apps etc. manually with `brew` that I wanted to test or whatever, or update apps specified in my `Brewfile` generated by this module (without having to run a full `darwin-rebuild switch`). With this option enabled, I can do that seamlessly without having to explicitly reference the Brewfile in the store (which is annoying to do).

modules/programs/brew-bundle.nix
kclejeune: I definitely agree that an enable option is pretty idiomatic with the rest of nix and should be included, even if true by default.

modules/programs/brew-bundle.nix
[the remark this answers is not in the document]
kclejeune: I like `cleanup` in terms of sticking to the existing homebrew naming conventions.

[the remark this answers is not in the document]
malob: Done.

modules/programs/brew-bundle.nix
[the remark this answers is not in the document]
malob: Alright, I've changed this option to from `setNoLockEnvvar` to `userConfig.noLock`, which I don't love, but I think is an improvement. (Did something similar for `setBrewfileEnvbar` which is now `userConfig.brewfile`.)

modules/programs/brew-bundle.nix
malob: As mention in another comment, I renamed this option to `userConfig.brewfile`.

modules/programs/brew-bundle.nix
malob: Thinking about it more, I've updated that the option should indeed be renamed to `autoUpdate` as @eraserhd suggested, because double negatives are the worst.

modules/programs/brew-bundle.nix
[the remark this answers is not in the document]
LnL7: I don't really have an opinion on this but a homebrew namespace would make sense to me.

modules/programs/brew-bundle.nix
```diff
@@ -0,0 +1,213 @@
...
+        cask "java" unless system "/usr/libexec/java_home --failfast"
+      '';
+      description = "Extra lines to be added verbatim to the generated Brewfile.";
+    };
+  };
+
+  config = {
+    assertions = mkIf cfg.enable [
+      {
+        assertion = builtins.pathExists /usr/local/bin/brew;
```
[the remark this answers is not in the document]
LnL7: Checking this during evaluation feels like the wrong place, it means you can't build and copy the system to an other host for example. The whole idea of downloading things during activation, inherent to brew/cask, feels a bit dubious to me. However for installation a random `curl | bash` feels especially bad in that regard.

modules/programs/brew-bundle.nix
LnL7: As an added side effect it wouldn't be possible to "clear" brew formula since that would disable the brew steps altogether. I agree it's not the most intuitive pattern, but it is idiomatic and there's no great alternative given that home-brew would become a requirement otherwise.

It's always possible to add warnings/assertions for the opposite case, which doesn't make sense.

modules/programs/brew-bundle.nix
```diff
@@ -0,0 +1,225 @@
...
+        '';
+      }
+    ];
+
+    programs.brew-bundle.brews =
+      optional (cfg.masApps != {}) "mas" ++
+      optional (cfg.whalebrews != []) "whalebrew";
+
+    environment.variables = mkIf cfg.enable (
+      (if cfg.userConfig.brewfile then { HOMEBREW_BUNDLE_FILE = "${brewfile}"; } else {}) //
```
LnL7, line 214: Is there a reason to expose these? Looks kind of redundant since activation already applies the brewfile and the file itself is readonly.

modules/programs/brew-bundle.nix
```diff
@@ -0,0 +1,225 @@
...
+      (if cfg.userConfig.noLock then { HOMEBREW_BUNDLE_NO_LOCK = "1"; } else {})
```
LnL7, line 215: Where does the lockfile go if this is disabled? And as a side remark it seems like this in general doesn't support a workflow with a lockfile which feels like the better approach to me.

LnL7 (commented):
I have not used homebrew on my system for quite some time so I don't really have much feedback here. I think it's best that the people that will actually use it give their suggestions, looks good to me in general if everybody is happy with it.

modules/programs/brew-bundle.nix
```diff
@@ -0,0 +1,225 @@
...
+  brewfile = pkgs.writeText "Brewfile" (
+    (brewfileSection "Taps" "tap" cfg.taps) +
+    (brewfileSection "Brews" "brew" cfg.brews) +
+    (brewfileSection "Casks" "cask" cfg.casks) +
+    (masBrewfileSection cfg.masApps) +
+    (brewfileSection "Docker contrainers" "whalebrew" cfg.whalebrews) +
+    (if cfg.extraConfig != "" then "# Extra config\n" + cfg.extraConfig else "")
+  );
+
+  brew-bundle-command =
```
LnL7, line 31: Using a list here and `optional`/`optionals` probably makes it a bit more readable.

modules/programs/brew-bundle.nix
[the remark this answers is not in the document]
malob: Alrighty, I've moved this just to `homebrew`.

modules/programs/brew-bundle.nix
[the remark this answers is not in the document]
malob: Resolving this comment thread, and picking the conversation up about these options, in a new thread.

modules/programs/brew-bundle.nix
malob: Resolving this comment thread, and picking the conversation up about these options, in a new thread.

modules/programs/brew-bundle.nix
malob: Good call, I update this, and also made some other changes to replace other if statements with functions like `optionalString` and `optionalAttrs`.

modules/programs/brew-bundle.nix
malob: Cool, let's stick with the Nix idiom here.

modules/programs/brew-bundle.nix
malob: TLDR: I think it's handy that this module sets `HOMEBREW_BUNDLE_FILE` for me in my environment and I expect some others to also appreciate it. I'd be happy in a world where the module just set this variable in my environment when enabled, and didn't have an option, but I'm worried there may be people that have use-cases where not being able to disable setting that variable in their environment would be annoying. 

---

I like having the `HOMEBREW_BUNDLE_FILE` variable set in my environment, since sometimes I do still want to run `brew bundle` myself. 

For example, I sometimes install some packages/apps (that aren't available through Nix) using Homebrew manually to try them out, or because I only need them temporarily. With this environment variable set, I can run `brew bundle cleanup` to get a list of which things are installed on my system using Homebrew that aren't in my Brewfile. I might also want to run `brew bundle cleanup --force` to remove them.

Since I set `homebrew.cleanup` to `"zap"`, I could instead just run `darwin-rebuild switch`, but sometimes I don't want to run a full build/activation just for this. Also, I can imagine some people wanting to do similar things, but who prefer to have `homebrew.cleanup` set to `"none"`, and so the only way to do a cleanup would be to run the command manually. Without that variable set, one would have to manually specify the path to the generated Brewfile in the store which would be a pain.

The above is an example of why it's handy for this module to set `HOMEBREW_BUNDLE_FILE` in the users environment, but it isn't necessarily and argument that there should be an option for it. In my case, I basically never expect to set `homebrew.userConfig.brewfile` to `false`, so I wouldn't mine the module simply adding it to `environment.variables` when `homebrew.enable` is set to `true`. The reason I decided to included it as an option is that, I have a vague concern that it might be annoying for some people, though I don't have a great sense of why it might be. Maybe it's not uncommon for people to use multiple Brewfiles for some reason? ¯\_(ツ)_/¯

For completeness, I should acknowledge that this module doesn't *need* to have the ability to set `HOMEBREW_BUNDLE_FILE` at all. Users like me who care about it could just set it manually ourselves. My take is that setting it (and maybe also including an option to disable setting it) matches what users expect of modules like this. They way I see it, this module is about declaratively managing Homebrew Bundle. From that lens, it would seem wrong to me if, by default when I enabled this module, the `brew bundle` command didn't know about the Brewfile this module generated.

modules/programs/brew-bundle.nix
malob: The lockfile is generated in the directory where the Brewfile is located whenever `brew bundle [install]` is run. So if this variable isn't set in the users environment, and the user runs `brew bundle [install]` themselves (with `hombrew.userConfig.brewfile` set to `true`, which is the current default), they'll see a warning like:

```
Warning: Could not write to /nix/store/gyg4yzd3ds5apzkcz211l6529ndfr0jc-Brewfile.lock.json!
```

Like `homebrew.userConfig.brewfile` this is another example of an option that I basically never expect to want to disable, but since I do sometimes run `brew bundle` myself (and want it to reference the generated Brewfile in the store by default), I do like that the module sets `HOMEBREW_BUNDLE_NO_LOCK` for me. I'd be happy to kill this option, and just add the variable to `environment.variables` when the module is enabled. I included it as an option out of an abundance of caution, since just like `homebrew.userConfig.brewfile`, I was worried that for some users, having that variable set in their environment would be annoying for some use-case of theirs.

After writing this comment and the one above, I am leaning towards killing both the `userConfig` options, and instead just adding both variables to `environment.variables` whenever the module is enabled.

@eraserhd, @kclejeune, what do you guys think? Would you be annoyed if this module just added those two variables to `environment.variables`, and didn't include an option to disable it doing so? Do you know of any common Homebrew Bundle use-cases where that would be annoying?

modules/programs/brew-bundle.nix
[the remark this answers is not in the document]
malob: > Checking this during evaluation feels like the wrong place, it means you can't build and copy the system to an other host for example. 

Ah, good call, I hadn't thought of that. I assume then that a better place for this check would be in the activation script? What's the usual practice for things like this in `nix-darwin`?

How about something like this:

```bash
# Homebrew Bundle
echo >&2 "Homebrew bundle..."
if [ -f /usr/local/bin/brew ]; then
  PATH=/usr/local/bin:$PATH ${brew-bundle-command}
else
  echo -e "\e[1;31merror: Homebrew is not installed, skipping...\e[0m"
fi
```

> The whole idea of downloading things during activation, inherent to brew/cask, feels a bit dubious to me. However for installation a random curl | bash feels especially bad in that regard.

Yeah cool, happy to leave out Homebrew installation functionality.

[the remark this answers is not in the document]
malob: Added a new commit making the above change for now.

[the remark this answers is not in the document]
LnL7: Yeah, checking this in the activation snippet makes the most sense. Whether or not this should be a hard failure or just an error message like that depends a bit on preference.

modules/programs/brew-bundle.nix
LnL7: This sounds like a very specific workflow to me, which maybe shouldn't be the default. But I have no idea 😄 .

modules/programs/brew-bundle.nix
LnL7: I don't think that would complain on a single user installation and instead create an orphan lockfile in the store.

The main thing I was getting at is that I would expect the lockfile to be part of your configuration so it's actually used when switching the system. While the current approach seems to explicitly avoid it.

kclejeune: To my knowledge, brew bundle doesn't actually make use of the lockfile when installing packages since there isn't really a notion of multiple package versions like we're used to in nixpkgs. 

I think you'd be safe to have this either be generated and kept in the nix store, or to just not be generated at all.

modules/programs/brew-bundle.nix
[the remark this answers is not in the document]
malob: Great. Error message feels right to me.

modules/programs/brew-bundle.nix
malob: Alrighty, I'm going to go with removing the options, and setting both variables when the module is enabled.

modules/programs/brew-bundle.nix
malob: > To my knowledge, brew bundle doesn't actually make use of the lockfile when installing packages since there isn't really a notion of multiple package versions like we're used to in nixpkgs.

Yeah, my understanding is that the lockfiles are generated for debugging purposes.

>  I think you'd be safe to have this either be generated and kept in the nix store, or to just not be generated at all.

Given that they're just for debugging, and that the lockfile would be generated during activation (which makes putting it in the store a hassle), I definitely don't think this module should generate them itself.

> Regarding where to set the variables associated with this, I don't really see an issue with putting them directly in the module as opposed to environment.variables since I can't think of much reason to change them in the first place, but don't have a strong preference either way.

To be clear, these don't need to be set at all in the activation script, since `brew bundle` is run with the `--no-lock` and `--file` flags in the script. The only effect of the `homebrew.userConfig` options was to govern the default behavior of the `brew bundle` command, if/when invoked by the user themselves.

---

Given the conversation here so far, I do still think that it's the right call to add these variables to `environment.variables`, since I think it's intuitive that enabling a module like this would result in the relevant command (in this case `brew bundle`) being configured to be aware and play well with the configuration this module generates. However, I'm now pretty convinced that having an option for these isn't the right call, since I've updated that people are very unlikely to want to change them from their current defaults, and options should only exist if they are going to be useful to people (i.e., some people will want to change them).

As such, I've removed the `homebrew.userConfig` options, and just setup the module to add these variables to `environment.variables` when this module is enabled.

kclejeune: > > To my knowledge, brew bundle doesn't actually make use of the lockfile when installing packages since there isn't really a notion of multiple package versions like we're used to in nixpkgs.
> 
> 
> 
> Yeah, my understanding is that the lockfiles are generated for debugging purposes.
> 
> 
> 
> >  I think you'd be safe to have this either be generated and kept in the nix store, or to just not be generated at all.
> 
> 
> 
> Given that they're just for debugging, and that the lockfile would be generated during activation (which makes putting it in the store a hassle), I definitely don't think this module should generate them itself.
> 
> 
> 
> > Regarding where to set the variables associated with this, I don't really see an issue with putting them directly in the module as opposed to environment.variables since I can't think of much reason to change them in the first place, but don't have a strong preference either way.
> 
> 
> 
> To be clear, these don't need to be set at all in the activation script, since `brew bundle` is run with the `--no-lock` and `--file` flags in the script. The only effect of the `homebrew.userConfig` options was to govern the default behavior of the `brew bundle` command, if/when invoked by the user themselves.
> 
Oops, I totally overlooked this part. This makes sense. 
> 
> 
> ---
> 
> 
> 
> Given the conversation here so far, I do still think that it's the right call to add these variables to `environment.variables`, since I think it's intuitive that enabling a module like this would result in the relevant command (in this case `brew bundle`) being configured to be aware and play well with the configuration this module generates. However, I'm now pretty convinced that having an option for these isn't the right call, since I've updated that people are very unlikely to want to change them from their current defaults, and options should only exist if they are going to be useful to people (i.e., some people will want to change them).
> 
> 
> 
> As such, I've removed the `homebrew.userConfig` options, and just setup the module to add these variables to `environment.variables` when this module is enabled.

This sounds reasonable. Thanks for clarifying :)

malob: Sweet! Much appreciation for all you helpful comments/thoughts.

malob:
Ok, after the last round of changes, I think this module is now ready for primetime.

@LnL7, are there any changes/additions etc. that you'd like before you feel comfortable merging this in?

LnL7:
I feel like I didn't really get my point across in the thread so let me try to explain the type of workflow I would expect. The main concern I have here is that making the current implementation work with the lockfile would be difficult to do in a compatible way.

Even tho pinning dependencies isn't implemented (yet) the lockfile can still be used as input for the system to determine changes.

```sh
brew update
brew bundle dump
darwin-rebuild switch
```

For example, just like updating a channel of flake inputs the brew lockfile could be updated like this and the system can then verify whether the lockfile is correct or outdated during activation. This way the lockfile isn't just disabled globally and if pinned versions are implemented by homebrew at some point nothing really changes except that rollbacks also work.

The global HOMEBREW_* variables also mean that any other usecases, like a Brewfile in a repo won't work as expected, I'm not sure if this is a thing people do but keeping things contained to activation by default would be a better default I think. Your workflow can still be handled by exposing the brewfile in an option so it can be referenced in other modules

```nix
{ config, lib, pkgs, ... }:
{
    environment.variables.HOMEBREW_BUNDLE_FILE = config.homebrew.brewfile;
}
```

TheOptimist:
Is there any reason that `mas` and `whalebrew` are bundled up with `brew`? What are the implications of separating those two specific mechanics out into their own modules?

kclejeune:
> Is there any reason that `mas` and `whalebrew` are bundled up with `brew`? What are the implications of separating those two specific mechanics out into their own modules?
> 
> 
> 
> Also, forgive the intrusion. I'm just starting out on my journey with nix-darwin. I wouldn't want to take the whole of `Homebrew` just to be able to use `mas`.

This module is explicitly to introduce the ability to manage `brew bundle` which is a particular tool that exists within Homebrew for bootstrapping systems. 

Since this has support for `tap` for third party tools, `brew` for maintained formulae, `cask` for GUI applications, and `mas` for App Store applications, it should be included here. For reference, my brewfile looks like this: https://github.com/kclejeune/system/blob/master/modules/dotfiles/Brewfile

You can still use `mas` as a command line tool to install apps without using brew bundle, but there's no specific interface to achieve similarly declarative app management from something like a brewfile currently.

TheOptimist:
This is a great answer to my comment, thank you. I realized much of this (especially the tying to `brew bundle`) after writing. Then saw the uselessness of my comment (not adding anything).

modules/homebrew.nix
```diff
@@ -0,0 +1,193 @@
...
+    concatStringsSep "\n" (mapAttrsToList (name: id: ''mas "${name}", id: ${toString id}'') entries) +
+    "\n"
+  );
+
+  brewfile = pkgs.writeText "Brewfile" (
+    brewfileSection "Taps" "tap" cfg.taps +
+    brewfileSection "Brews" "brew" cfg.brews +
+    brewfileSection "Casks" "cask" cfg.casks +
+    masBrewfileSection cfg.masApps +
+    brewfileSection "Docker contrainers" "whalebrew" cfg.whalebrews +
```
probablyadev, line 23:
```suggestion
    brewfileSection "Docker containers" "whalebrew" cfg.whalebrews +
```

malob:
@LnL7, just want to flag up front that I don't feel confident I understand exactly what you were getting at. Apologies if my responses below aren't on target.

As far as I can tell, Homebrew has no plans to use `Brewfile.lock.json` as a "proper" lockfile, due to fundamental limitations of how Homebrew works. From the [PR](https://github.com/Homebrew/homebrew-bundle/pull/552) that added the ability for Homebrew Bundle to output the file:

> Starting with a caveat: we will never have a "proper" lockfile in Homebrew/homebrew-bundle that installs all your dependencies at exactly the versions you want because Homebrew doesn't work that way (and never will because e.g. pinning OpenSSL at the same version forever is a bad idea).

There could be some value in having access to the lockfile for the last successful invocation of `brew bundle [install]`, since, if a subsequent invocation fails, one could have a look at the file in the process of debugging the failure. However, I don't currently see a good way to incorporate the lockfile into this module.

From `brew bundle --help`:
> **brew bundle** will output a **Brewfile.lock.json** in the same directory as the **Brewfile** if all dependencies are installed successfully.

That poses a couple challenges:
* The Brewfile this module generates is located in the Nix store, and so during activation `brew bundle [install]` can't generate a lockfile when using the generated Brewfile, since it will try to write it to the Nix store, which it can't.
* Even if this wasn't the case, I'm don't see a good way to save the lockfile into the Nix store, given that it would be generated during activation.

This is kind of a bummer since, if `brew bundle [install]` fails during activation, there isn't a lockfile around to help with debugging, but unless I'm missing something, I don't see any good way for this module to track or work with the `Brewfile.lock.json` file.

> The global HOMEBREW_* variables also mean that any other usecases, like a Brewfile in a repo won't work as expected, I'm not sure if this is a thing people do but keeping things contained to activation by default would be a better default I think. Your workflow can still be handled by exposing the brewfile in an option so it can be referenced in other modules
> 
> ```nix
> { config, lib, pkgs, ... }:
> {
>     environment.variables.HOMEBREW_BUNDLE_FILE = config.homebrew.brewfile;
> }
> ```

It does seem to me that setting `HOMEBREW_BUNDLE_FILE` and `HOMEBREW_BUNDLE_NO_LOCK` globally, could be controversial for the reasons you mentioned. My sense is that there are two primary ways people use Brewfiles, to manage the things installed on their system for their own day-to-day use, and included with a project to make it easy to install dependencies for that project (I'm not sure how common the latter is, but I've definitely seen it).

For the former case, I do think that having these variables set leads to a more intuitive experience, and that the illustrative examples are representative of a broader class of use-cases that I expect to be fairly common. I also agree that with these variables set globally, it may be confusing/cumbersome for folks to use a Brewfile provided with a project (e.g., they'd have to run `brew bundle --file=Brewfile` rather than just `brew bundle` when in the project root, and they'd have to unset `HOMEBREW_BUNDLE_NO_LOCK` if they wanted to generate a lockfile).

Given my understanding of your concerns, here are what seem to me like the best ways to move forward:
* Don't set the variables at all in this module, add a (readonly or private?) option `homebrew.brewfile` which get's set to the path of the Brewfile in the store, and add to the description of `homebrew.enable` that some users may want to use `environment.variables` to set the variables themselves.
* Add back in options like `homebrew.global.brewfile` and `homebrew.global.noLock`, so that there's a quick way using this module for users to choose what behavior they want when running `brew bundle` commands themselves. My inclination would be for these options to be enable by default, but I'd be happy to have them be disabled by default if you disagree.

I'd prefer the latter, but would settle for the former in order to get this merged.

malob:
@LnL7, I've gone ahead and implemented the latter of the two paths forward I listed (to make things quicker if you're happy with it). Of course, if you'd really prefer the other option, or I completely misunderstood your concerns, just let me know, and I'll make the required changes.

ethancedwards8:
I would love for this to get merged. I've recently needed to install a homebrew package and I would love a nix/declarative way to do it such as this.

kclejeune:
> I would love for this to get merged. I've recently needed to install a homebrew package and I would love a nix/declarative way to do it such as this.

If you use niv or flakes to pin dependencies, you can point it to the version from this fork until it gets merged.

ethancedwards8:
> > I would love for this to get merged. I've recently needed to install a homebrew package and I would love a nix/declarative way to do it such as this.
> 
> If you use niv or flakes to pin dependencies, you can point it to the version from this fork until it gets merged.

I don't use flakes *yet*, that's on my TODO list :)

malob:
In case it's helpful, my Nix configs repo is a flake and it has and output `darwinModules` that contains and up to date version of this module: https://github.com/malob/nixpkgs/blob/master/flake.nix

LnL7:
> @LnL7, I've gone ahead and implemented the latter of the two paths forward I listed with the options disabled by default (to make things quicker if you're happy with it). Of course, if you'd really prefer the other option, or I completely misunderstood your concerns, just let me know, and I'll make the required changes.

Mostly just wanted to make sure I got the idea across. I leave the decision up to the people that will use this.

kclejeune:
> > @LnL7, I've gone ahead and implemented the latter of the two paths forward I listed with the options disabled by default (to make things quicker if you're happy with it). Of course, if you'd really prefer the other option, or I completely misunderstood your concerns, just let me know, and I'll make the required changes.
> 
> 
> 
> Mostly just wanted to make sure I got the idea across. I leave the decision up to the people that will use this.

I'm happy with this implementation - I think this looks ready to merge.

ethancedwards8:
> > > @LnL7, I've gone ahead and implemented the latter of the two paths forward I listed with the options disabled by default (to make things quicker if you're happy with it). Of course, if you'd really prefer the other option, or I completely misunderstood your concerns, just let me know, and I'll make the required changes.
> > 
> > 
> > Mostly just wanted to make sure I got the idea across. I leave the decision up to the people that will use this.
> 
> I'm happy with this implementation - I think this looks ready to merge.

I agree, I've forked this and changed my flake input to my fork. Everything's working as expected. One thing I think could improve is if homebrew isn't installed it should install it automatically.

kclejeune:
> I agree, I've forked this and changed my flake input to my fork. Everything's working as expected. One thing I think could improve is if homebrew isn't installed it should install it automatically.

We had a discussion about this (earlier in this thread) and ultimately decided that's not a good idea for a few reasons. If memory serves, this will print a warning and do nothing if `brew` isn't found in the path.

malob:
Yeah, during activation a (non-fatal) error message gets printed if Hombrew isn't installed: https://github.com/LnL7/nix-darwin/blob/2a0b9a9f60a3d3eaca57492f1a8d66082da370ea/modules/homebrew.nix#L206-L211

While I do agree it would be nice for this module to install Homebrew, this module is already pushing the boundries of the usual Nix zen, and given that I expect that installation of Homebrew would need to happen during activation, that would push things even further. Regardless, I think we should move forward with merging this initial version of the module, and leave potential additions like adding Homebrew installation to future discussions/PRs.

ethancedwards8:
> Yeah, during activation a (non-fatal) error message gets printed if Hombrew isn't installed:
> 
> https://github.com/LnL7/nix-darwin/blob/2a0b9a9f60a3d3eaca57492f1a8d66082da370ea/modules/homebrew.nix#L206-L211
> 
> While I do agree it would be nice for this module to install Homebrew, this module is already pushing the boundries of the usual Nix zen, and given that I expect that installation of Homebrew would need to happen during activation, that would push things even further. Regardless, I think we should move forward with merging this initial version of the module, and leave potential additions like adding Homebrew installation to future discussions/PRs.

Okay cool yeah, I must've missed that comment.

bphenriques:
👋  Wanted to provide feedback that it is working as intended 😄  For now I am setting my inputs as follows:
```
    darwin.url = "github:malob/nix-darwin/brew-bundle"; # Replace with lnl7/nix-darwin once https://github.com/LnL7/nix-darwin/pull/262 is merged.
    darwin.inputs.nixpkgs.follows = "nixpkgs";
```

One less thing to manage via shell scripts 🙇  It is not deterministic due to the lack of ability to pin versions but I think it is a good compromise for MacOS users.

kclejeune:
> 👋  Wanted to provide feedback that it is working as intended, thank you! 😄  For now I am setting my inputs as follows:
> 
> ```
> 
>     darwin.url = "github:malob/nix-darwin/brew-bundle"; # Replace with lnl7/nix-darwin once https://github.com/LnL7/nix-darwin/pull/262 is merged.
> 
> ```
> 
> 
> 
> One less thing to manage via shell scripts 🙇  It is not deterministic due to the lack of ability to pin versions but I think it is a good compromise for MacOS users.
> 
> 
> 
> Edit: I found out a (good?) quirk where I can include different files with different sets of `homebrew.casks` and they are all installed as if Nix is merging all the lists (e.g., importing [this](https://github.com/bphenriques/dotfiles/blob/bb292675ccc9f931538402f9aeac252e1067b755/macos/homebrew/default.nix) and [that](https://github.com/bphenriques/dotfiles/blob/bb292675ccc9f931538402f9aeac252e1067b755/macos/homebrew/personal.nix) results in all of them being installed) 🤔 This is more of a Nix question and is not specific to this forum.

This is expected behavior and is not actually unique to this module! Nix merges lists and attribute sets if they're defined in more than one location. This is helpful to design modular configurations :)

LnL7 (approved):
There hasn't really been an new feedback so merging.

frontsideair:
> Yeah, during activation a (non-fatal) error message gets printed if Hombrew isn't installed:
> 
> https://github.com/LnL7/nix-darwin/blob/2a0b9a9f60a3d3eaca57492f1a8d66082da370ea/modules/homebrew.nix#L206-L211
> 
> While I do agree it would be nice for this module to install Homebrew, this module is already pushing the boundries of the usual Nix zen, and given that I expect that installation of Homebrew would need to happen during activation, that would push things even further. Regardless, I think we should move forward with merging this initial version of the module, and leave potential additions like adding Homebrew installation to future discussions/PRs.

Are there any developments on this front? I would really like nix to transparently use homebrew instead of me, as a user, installing it beforehand.

PR #266, opened 2020-12-31
fish: generate environment natively

kevingriffin:
I found in my testing that the calls to `fenv` were slowing fish launch by more than 2x on my machine. While it's still easiest to maintain it for configuration shared with bash and zsh, a pretty good speedup can be had by generating native `set` commands at build time.

I made a few changes that might be breaking, but have added comments below for discussion.

### DIFF
--- a/modules/environment/default.nix
+++ b/modules/environment/default.nix
@@ -8,6 +8,9 @@ let
   exportVariables =
     mapAttrsToList (n: v: ''export ${n}="${v}"'') cfg.variables;
 
+  exportVariablesFish =
+    mapAttrsToList (n: v: ''set -g -x ${n} "${v}"'') cfg.variables;
+
   aliasCommands =
     mapAttrsFlatten (n: v: ''alias ${n}="${v}"'') cfg.shellAliases;
 
@@ -168,6 +171,8 @@ in
        export NIX_PROFILES="${concatStringsSep " " (reverseList cfg.profiles)}"
     '';
 
+    environment.etc."fish/foreign-env/extraInit".text = cfg.extraInit;
+
     environment.variables =
       {
         XDG_CONFIG_DIRS = map (path: path + "/etc/xdg") cfg.profiles;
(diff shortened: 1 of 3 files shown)


modules/environment/default.nix
```diff
@@ -193,6 +198,21 @@ in
...
+    system.build.setEnvironmentFish = pkgs.writeText "set-environment-fish" ''
+      # Prevent this file from being sourced by child shells.
+      set -g -x __NIX_DARWIN_SET_ENVIRONMENT_DONE 1
+
+      set -g -x PATH ${config.environment.systemPath}
+      ${concatStringsSep "\n" exportVariablesFish}
+
+      # Extra initialisation
+      set fish_function_path ${pkgs.fish-foreign-env}/share/fish-foreign-env/functions $__fish_datadir/functions
+      fenv "source /etc/fish/foreign-env/extraInit" > /dev/null
```
kevingriffin, line 210: There's a couple of ways we could avoid fenv entirely: duplicate the `extraInit` in fish or preprocess during build and output the result here. I experimented with that [here](https://github.com/kevingriffin/plugin-foreign-env/commit/9e92894586ef3e49d03fbcedfcae311e0e75fada), but I realized that making fenv lazy presents an interesting challenge with interpolation: does it get deferred? If we chose that route, it'd be possible to modify fenv to support this use case.

modules/programs/fish.nix
```diff
@@ -43,7 +43,7 @@ in
           Whether fish should use completion files provided by other packages.
         '';
       };
-      
+
```
kevingriffin, line 46: If the whitespace churn is undesirable, I can remove it or move it to a different commit.

modules/programs/fish.nix
```diff
@@ -99,18 +99,10 @@ in
 
   config = mkIf cfg.enable {
 
-    environment.etc."fish/foreign-env/shellInit".text = cfge.shellInit;
-    environment.etc."fish/foreign-env/loginShellInit".text = cfge.loginShellInit;
-    environment.etc."fish/foreign-env/interactiveShellInit".text = cfge.interactiveShellInit;
```
kevingriffin, line 104: This configuration is being set on the fish module itself, yet was being run through fenv. To me, it seems like if you're specifying init for fish specifically, it should be written in fish.

modules/environment/default.nix
```diff
@@ -193,6 +198,21 @@ in
...
 
+    system.build.setEnvironmentFish = pkgs.writeText "set-environment-fish" ''
+      # Prevent this file from being sourced by child shells.
+      set -g -x __NIX_DARWIN_SET_ENVIRONMENT_DONE 1
+
+      set -g -x PATH ${config.environment.systemPath}
+      ${concatStringsSep "\n" exportVariablesFish}
+
+      # Extra initialisation
+      set fish_function_path ${pkgs.fish-foreign-env}/share/fish-foreign-env/functions $__fish_datadir/functions
```
hauleth, line 209:
```suggestion
      set fish_function_path ${pkgs.fish-foreign-env}/share/fish-foreign-env/vendor_functions.d $__fish_datadir/functions
```

kevingriffin: Thank you! Happy to commit this suggestion. Would you mind expanding on why you selected this convention? I've definitely seen it before, but would like to hear about how it's typically used.

kevingriffin:
Closing in favor of https://github.com/LnL7/nix-darwin/pull/270

Issue #269, opened 2021-01-07
Recent changes in fish-foreign-env (Nixpkgs) break /etc/fish/nixos-env-preinit.fish


teehemkay:
Basically, `fish-foreign-env` directory layout was changed in a recent `Nixpkgs` [commit](https://github.com/NixOS/nixpkgs/commit/99bfa4bb601bfc94a72296f0b2395a4f888e5d84).
This results in breakage in the `fish` expression in `nix-darwin` which assumes a [specific directory layout](https://github.com/LnL7/nix-darwin/blob/8834df431736c3037bbef017b24c62bc4ace4645/modules/programs/fish.nix#L109).

This results in a rather annoying breakage (no `darwin-nix` environment, no `nix` even) for those using `fish` as their default shell. One has to manually relink `/nix/var/nix/profiles/system` to a previous generation to restore a working nix-darwin environment.

abathur:
This happened during the course of #107834, and https://github.com/NixOS/nixpkgs/pull/107834#issuecomment-752208683 explains that it is a breaking change, at least. I'm not fish-familiar so I don't have much insight here, but I wanted to cross-link it.

kevingriffin:
Although fixing this is straightforward, I have a pull request open to reduce the dependency on fenv in nix-darwin. This speeds up shell startup by a lot, and depending on how much we're willing to change, it could even be removed completely. If you use fish, please have a look and give me some feedback:

https://github.com/LnL7/nix-darwin/pull/266

teehemkay:
Thank you. I'll try and have a look this afternoon.

toonn:
@kevingriffin, how would #266 deal with some substitutions fish doesn't have, `${VAR:-fallback}`, for example?
Maybe [babelfish](https://github.com/NixOS/nixpkgs/pull/107834#discussion_r549615685) could be used in addition/instead?

kevingriffin:
@toonn Thank you for the feedback! In which attribute are you imagining those substitutions coming up? For the fish module of nix-darwin itself, #266 just requires you to set your configuration as fish instead of bash, and it transforms configured ENV variables into `set` statements.

Are you thinking about how to handle `extraInit` in the environment module? I'll see if I can wire up babelfish in my branch. 

I'm not sure what expectations other fish users have of this attribute, so I'd love to hear some use cases.

kevingriffin:
@toonn I've opened #270 with babelfish. Give it a try and see what you think.

[...]

PR #270, opened 2021-01-09
Native fish environment with babelfish

kevingriffin:
This is an alternative approach to https://github.com/LnL7/nix-darwin/pull/266

Adds the [babelfish](https://github.com/bouk/babelfish) package for translating the environment to fish at build time, eliminating the need for and performance cost of fenv.

I'm not sure if we want to vendor the package, get it added upsteam in nixpkgs, or some other approach. Ideally nixOS users would receive a similar change to their module as well.

Fixes https://github.com/LnL7/nix-darwin/issues/269

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -12,6 +12,29 @@ let
     mapAttrsFlatten (k: v: "alias ${k} '${v}'") cfg.shellAliases
   );
 
+  envShellInit = pkgs.writeText "shellInit" cfge.shellInit;
+
+  envLoginShellInit = pkgs.writeText "loginShellInit" cfge.loginShellInit;
+
+  envInteractiveShellInit = pkgs.writeText "interactiveShellInit" cfge.interactiveShellInit;
+
+  fenv = pkgs.fish-foreign-env or pkgs.fishPlugins.foreign-env;
+
+  sourceEnv = file:
+  if cfg.useBabelfish then
+    "source /etc/fish/${file}.fish"
+  else
+    ''
+      set fish_function_path ${fenv}/share/fish-foreign-env/functions $fish_function_path
+      fenv source /etc/fish/foreign-env/${file} > /dev/null
+      set -e fish_function_path[1]
+    '';
+
+  babelfishTranslate = path: name:
+    pkgs.runCommand "${name}.fish" {
+      nativeBuildInputs = [ cfg.babelfishPackage ];
+    } "${cfg.babelfishPackage}/bin/babelfish < ${path} > $out;";
+
 in
 
 {
@@ -27,7 +50,24 @@ in
         '';
         type = types.bool;
       };
-      
+
+      useBabelfish = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          If enabled, the configured environment will be translated to native fish using <link xlink:href="https://github.com/bouk/babelfish">babelfish</link>.
+          Otherwise, <link xlink:href="https://github.com/oh-my-fish/plugin-foreign-env">foreign-env</link> will be used.
+        '';
+      };
+
+      babelfishPackage = mkOption {
+        type = types.package;
+        description = ''
+          The babelfish package to use when useBabelfish is
+          set to true.
+        '';
+      };
+
       vendor.config.enable = mkOption {
         type = types.bool;
         default = true;
@@ -43,7 +83,7 @@ in
           Whether fish should use completion files provided by other packages.
         '';
       };
-      
+
       vendor.functions.enable = mkOption {
         type = types.bool;
         default = true;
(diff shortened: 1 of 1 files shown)


toonn:
This does seem to work fine for me. Doesn't really affect loading time of a shell started from a shell but spawning a new terminal has gone for ~8s to under 4s.

Note that I don't use POSIX variable substitutions in my environment so I can't vouch for that working.

kevingriffin:
Thanks for having a look! I prefer this approach and opened PRs on nixpkgs to [add the package](https://github.com/NixOS/nixpkgs/pull/108946) and use it in the [nixOS module](https://github.com/NixOS/nixpkgs/pull/108947).

modules/programs/fish.nix
```diff
@@ -151,12 +137,7 @@ in
       status --is-interactive; and not set -q __fish_nix_darwin_interactive_config_sourced
       and begin
         ${fishAliases}
-        
 
-        set fish_function_path ${pkgs.fish-foreign-env}/share/fish-foreign-env/functions $fish_function_path
-        fenv source /etc/fish/foreign-env/interactiveShellInit > /dev/null
-        set -e fish_function_path[1]
-
```
bouk, line 159: You're removing a bunch of the shell inits here, which will break gpg-agent for example.

I think we should translate all of the init files and save them as `environment.etc."fish/interactiveShellInit.fish"` etc. Then we also don't need `system.build.setEnvironmentFish`

bouk (changes requested):
Thank you for getting started on this, I already have something like this in my own nix config:

https://github.com/bouk/b/blob/5cfc6ec6a46d15fffef9c92788b719d07a1a950e/darwin.nix#L80-L117

For me it was a huge speed boost, shaving 3-4 seconds (!) off every new shell invocation

modules/programs/fish.nix
kevingriffin: @bouk I've tried implementing this approach. If you could have a look and give feedback, I'll port the same change to the nixOS module.

kevingriffin:
I've made changes here and on the nixOS module to make this change optional, configured on the fish module. I also committed a change to the namespace of the foreign-env package to resolve the original issue for users who don't use the new babelfish-based generation option.

modules/programs/fish.nix
```diff
@@ -99,33 +124,69 @@ in
...
+    environment = mkMerge [
+      (mkIf cfg.translateEnvironment
+      {
+        etc."fish/set-environment.fish".source = pkgs.runCommand "set-environment.fish" {
+        nativeBuildInputs = [ babelfish ];
+      } ''
+        ${babelfish}/bin/babelfish < ${config.system.build.setEnvironment} > $out;
+      '';
+
+       etc."fish/shellInit".source = pkgs.runCommand "shellInit" {
```
bouk, line 136: fish files always have .fish at the end, so I think you should add that here too.

modules/programs/fish.nix
```diff
@@ -12,6 +12,23 @@ let
...
+
+  envInteractiveShellInit = pkgs.writeText "interactiveShellInit" cfge.interactiveShellInit;
+
+  sourceEnv = useTranslated: path:
+  if useTranslated then
+    "source ${path}"
+  else
+    ''
+      set fish_function_path ${pkgs.fishPlugins.foreign-env}/share/fish-foreign-env/functions $fish_function_path
+      fenv source ${path} > /dev/null
```
hauleth, line 29: Do we still need `fenv` there as we are using `babelfish`?

kevingriffin: We don't. In the version I've done for nixOS, this requirement is removed. I'll go ahead and port it back to nix-darwin in this branch.

modules/programs/fish.nix
```diff
@@ -12,6 +12,23 @@ let
     mapAttrsFlatten (k: v: "alias ${k} '${v}'") cfg.shellAliases
   );
 
+  babelfish = pkgs.callPackage ../../pkgs/shells/fish/plugins/babelfish.nix { };
```
LnL7, line 15: I don't really want to keep package definitions in this repo.

Since this isn't guaranteed to be available in nixpkgs (eg. 20.09) the best solution is to add a package option without a default, that way it's up to the user to either introduce the version they want into their configuration or if nixpkgs is recent enough just use `pkgs.babelfish`. The later can then also the default once the oldest supported release has babelfish available.

modules/programs/fish.nix
```diff
@@ -12,6 +12,23 @@ let
...
+  envLoginShellInit = pkgs.writeText "loginShellInit" cfge.loginShellInit;
+
+  envInteractiveShellInit = pkgs.writeText "interactiveShellInit" cfge.interactiveShellInit;
+
+  sourceEnv = useTranslated: path:
+  if useTranslated then
+    "source ${path}"
+  else
+    ''
+      set fish_function_path ${pkgs.fishPlugins.foreign-env}/share/fish-foreign-env/functions $fish_function_path
```
LnL7, line 28: Similar to the babelfish package, this should still handle the case where fishPlugins isn't available in nixpkgs yet. That's what is causing the tests to fail. eg.

```nix
let
  fenv = pkgs.fish-foreign-env or pkgs.fishPlugins.foreign-env;
in
  # ...
```

Ideally an alias would be added in nixpkgs however, just renaming stuff isn't really something projects outside of the repo should have deal with without some compatibility window.

LnL7 (changes requested):
There are a few extra things to consider here compared to the NixOS module since this needs to be compatible with both 20.09 and unstable.

modules/programs/fish.nix
kevingriffin: I agree. It was convenient to spike it, but I wouldn't want to vendor an old package in nix-darwin. I'll add a package option.

modules/programs/fish.nix
kevingriffin: That's a very clean way to do it. I'll incorporate that into this branch, along with the other changes (like selecting which method to use—fenv or babelfish—) from the nixOS module and push up a new version.

kevingriffin:
@LnL7 Thanks for the review! I believe what I just pushed up addresses your comments. Let me know if there's anything else to improve.

kevingriffin:
@LnL7 The nixOS module was merged with no further changes. Is there anything else to do to make this suitable for nix-darwin?

LnL7 (approved):
With the extra compatibility changes this looks good to me. I'm not a fish user but people with more relevant knowledge should have already reviewed this on the NixOS side.

teehemkay:
There's an issue where `fish-foreign-env` has now been aliased to `throw "fish-foreign-env has been replaced with fishPlugins.foreign-env` in `Nixpkgs`.
This results in an error if `fish` is enable.

The test to bind `fenv` in the `let` expression should probably be inverted.
Maybe something like:

```
 fenv = if builtins.hasAttr "fishPlugins" pkgs then pkgs.fishPlugins.foreign-env else pkgs.fish-foreign-env;
```

modules/programs/fish.nix
```diff
@@ -12,6 +12,29 @@ let
     mapAttrsFlatten (k: v: "alias ${k} '${v}'") cfg.shellAliases
   );
 
+  envShellInit = pkgs.writeText "shellInit" cfge.shellInit;
+
+  envLoginShellInit = pkgs.writeText "loginShellInit" cfge.loginShellInit;
+
+  envInteractiveShellInit = pkgs.writeText "interactiveShellInit" cfge.interactiveShellInit;
+
+  fenv = pkgs.fish-foreign-env or pkgs.fishPlugins.foreign-env;
```
toonn, line 21: I ran into the same problem because of the redirect at `pkgs.fish-foreign-env`, looks like flipping around the packages would indeed fix it.

```suggestion
   fenv =  pkgs.fishPlugins.foreign-env or pkgs.fish-foreign-env;
```

teehemkay:
Yep. I was beeing extra cautious (check before jumping) hence the `if` expression.

Note that the issue only occurs if the user does not use `babelfish` but since this is the default...

kevingriffin:
Thank you @teehemkay and @toonn for the testing. I opened a quick PR to integrate this change:

https://github.com/LnL7/nix-darwin/pull/283

PR #275, opened 2021-01-20
modules: add mbsync service

cmacrae:
Hey @LnL7 👋 Here's a service module for [mbsync](https://isync.sourceforge.io/mbsync.html)


modules/services/mail/mbsync.nix
```diff
@@ -0,0 +1,76 @@
...
+        An optional command to run after mbsync executes successfully.
+        This is useful for running mailbox indexing tools.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+
+    launchd.user.agents.mbsync = {
+      serviceConfig.ProgramArguments = [
```
LnL7, line 64: The launchd module has support for generating scripts for more complex commands.

```suggestion
      script = ''
        # commands
      '';
```

modules/services/mail/mbsync.nix
```diff
@@ -0,0 +1,76 @@
...
+    configFile = mkOption {
+      type = types.str;
+      default = "";
+      description = ''
+        Optional configuration file to link to use instead of
+        the default file (<filename>~/.mbsyncrc</filename>).
+      '';
+    };
+
+    postExec = mkOption {
```
LnL7, line 50: How common is this usecase? Since `launchd.<name>.script` it's possible to do this without an explicit option (alltho that's a bit more implicit).

```nix
{ lib, ... }:
with lib;
{
  services.mbsync.enable = true;
  launchd.user.agents.mbsync.script = mkAfter ''
    mu index
  '';
}
```

modules/services/mail/mbsync.nix
cmacrae: Ah much nicer. I've addressed this in my latest commit

modules/services/mail/mbsync.nix
cmacrae: Personally, I'd say this is a common usecase. Since mbsync will just pull down the e-mail onto the filesystem, most users would want to invoke some sort of mail indexing utility.

I'd vouch for keeping the option :)

cmacrae:
Just working on fixing the tests ☝️

cmacrae:
Hey @LnL7 👋 Would you mind re-reviewing this?

codygman:
FWIW I was just looking for an mbsync module in nix-darwin. Not sure what the status is or if this is close to being mergeable.

PR #280, opened 2021-01-28
update fish-foreign-env package

happysalada:
Fish foreign env package has change its name on nixpkgs unstable.
nix-darwin can't be built anymore with the unstable channel.

I'm not sure what the process for updating something like this should be. Since on the latest stable channel the package still has it's old name.

Let me know if I can help with anything.


happysalada:
This is addressed in a better way in #270

PR #281, opened 2021-01-28
Use separate config option to set port of DNSmasq

hauleth:
Older approach seems to not work on macOS 11 Big Sur for some reason.


LnL7 (approved):
Thanks!

PR #283, opened 2021-01-30
fish: fix fenv reference on 20.09

kevingriffin:
Referring to the deprecated package first causes
an error to be thrown before the correct package can be selected.

Fixes issue found in https://github.com/LnL7/nix-darwin/pull/270#issuecomment-769923166


LnL7 (approved):
Thanks!

Issue #158, opened 2019-08-27
multi user install does not work in zsh (works in bash)


dmvianna:
That took me the whole day to figure out. Installation runs smoothly, and installs what's in `darwin-configuration.nix`, but when I type this in `zsh`

```
$ darwin-rebuild switch
building the system configuration...
error: file 'darwin-config' was not found in the Nix search path (add it using $NIX_PATH or -I), at /nix/store/i5gakh46n16945x7yllxwha24dy9m1gc-darwin/darwin/default.nix:1:40
(use '--show-trace' to show detailed location information)
$
```
Whereas under `bash`...
```
~ AusNet$ darwin-rebuild switch
building the system configuration...
warning: Nix search path entry '/Users/AusNet/.nix-defexpr/channels/darwin' does not exist, ignoring
Password:
warning: Nix search path entry '/Users/AusNet/.nix-defexpr/channels/darwin' does not exist, ignoring
user defaults...
setting up user launchd services...
setting up ~/Applications...
setting up /etc...
warning: not linking environment.etc."nix/nix.conf" because /etc/nix/nix.conf exists, skipping...
warning: not linking environment.etc."zprofile" because /etc/zprofile exists, skipping...
warning: not linking environment.etc."zshrc" because /etc/zshrc exists, skipping...
system defaults...
setting up launchd services...
reloading nix-daemon...
configuring networking...
$
```
Phew!

Also: could different users have different `darwin-configuration.nix`? It seems not?

jrolfs:
I believe this output is likely indicative of your problem with `zsh`:

```sh
warning: not linking environment.etc."nix/nix.conf" because /etc/nix/nix.conf exists, skipping...
warning: not linking environment.etc."zprofile" because /etc/zprofile exists, skipping...
warning: not linking environment.etc."zshrc" because /etc/zshrc exists, skipping...
```

I could be wrong as I think `/etc/zshenv` sets `$NIX_PATH` but it's worth a shot to move/backup those files and run `darwin-rebuild switch` again in `bash` to allow *nix-darwin* to properly link those init files:

```sh
sudo mv /etc/nix/nix.conf /etc/nix/nix.conf.backup-before-nix-darwin
sudo mv /etc/zprofile /etc/zprofile.backup-before-nix-darwin
sudo mv /etc/zshrc /etc/zshrc.backup-before-nix-darwin
```

In bash:
```sh
darwin-rebuild switch
```

Then you should be able to verify the links:

[...]

Issue #271, opened 2021-01-12
macOS Big Sur configuration NIX_PATH error


c4dr01d:
I follow the README instruction to install nix-darwin at normal user, after install and change the darwin-configuration.nix, I run `darwin-rebuild switch`, then generate the error of `darwin` not found. when I change to root, and check the `NIX_PATH`, I saw the `darwin-configuration.nix` in the `NIX_PATH`.
`nix-shell -p nix-info --run "nix-info -m"` result:
```
 - system: `"x86_64-darwin"`
 - host os: `Darwin 20.2.0, macOS 10.16`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.3.10`
 - channels(c4droid): `"darwin, home-manager"`
 - channels(root): `"nixpkgs-21.03pre262620.a67cfc0cdc1"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
```
normal user `echo $NIX_PATH` result:
```
nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixpkgs:/nix/var/nix/profiles/per-user/root/channels
```
root user `echo $NIX_PATH` result:
```
darwin-config=/var/root/.nixpkgs/darwin-configuration.nix:/nix/var/nix/profiles/per-user/root/channels:/var/root/.nix-defexpr/channels
```

PR #286, opened 2021-02-04
Fix sourcing nix-darwin environment in zsh

### LINKED ISSUE #151: ZSH does not export `darwin` in `NIX_PATH`
I am trying to configure `zsh` as my main shell.

So far I have it configured as:
```
environment.shells = with pkgs; [ bashInteractive zsh ];
programs.zsh.enable = true;
environment.loginShell = "/run/current-system/sw/bin/zsh";
```

I also had to do `chsh` to actually switch to `zsh`.

But `nix-darwin` doesn't appear to be working in `zsh`. The reason seems to be that `NIX_PATH` is not set correctly:

Bash/Fish:
```
$ echo $NIX_PATH
darwin-config=/Users/araga/.nixpkgs/darwin-configuration.nix:/nix/var/nix/profiles/per-user/root/channels:/Users/araga/.nix-defexpr/channels
```

ZSH:
```
% echo $NIX_PATH
nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixpkgs:/nix/var/nix/profiles/per-user/root/channels
```

Setting `nix.nixPath` in `darwin-configuration.nix` doesn't help :(

`/etc/zshenv` does contain everything that is necessary:
```
if [ -n "$__ETC_ZSHENV_SOURCED" ]; then return; fi
export __ETC_ZSHENV_SOURCED=1

# Don't execute this file when running in a pure nix-shell.
if test -n "$IN_NIX_SHELL"; then return; fi

if [ -z "$__NIX_DARWIN_SET_ENVIRONMENT_DONE" ]; then
  . /nix/store/hz8ya45g8s21hi9vwvwi2g0krmxl8i5q-set-environment
fi
```

And when I source `/nix/store/hz8ya45g8s21hi9vwvwi2g0krmxl8i5q-set-environment` directly and manually it sets `NIX_PATH` as expected.

This (running in a new shell) suggests that this file has been sourced:
```
% echo $__ETC_ZSHENV_SOURCED
1
echo $__NIX_DARWIN_SET_ENVIRONMENT_DONE
1
```

However, `NIX_PATH` is not set. It looks like it has been reset, or as if `__NIX_DARWIN_SET_ENVIRONMENT_DONE` has been set _before_ sourcing `/etc/zshenv` happened for the 1st time...

Am I doing anything wrong?

starcraft66:
As discussed in the comments of #158, the environment set by `nix-darwin` for `zsh` is clobbered by the one set by Nix itself due to `zsh`'s load order. This PR moves the environment sourcing to a more convenient location in the load order. Fixes #151  and #158.

### DIFF
--- a/modules/programs/zsh/default.nix
+++ b/modules/programs/zsh/default.nix
@@ -117,10 +117,6 @@ in
       # Don't execute this file when running in a pure nix-shell.
       if test -n "$IN_NIX_SHELL"; then return; fi
 
-      if [ -z "$__NIX_DARWIN_SET_ENVIRONMENT_DONE" ]; then
-        . ${config.system.build.setEnvironment}
-      fi
-
       ${cfg.shellInit}
 
       # Read system-wide modifications.
@@ -155,6 +151,13 @@ in
       # Only execute this file once per shell.
       if [ -n "$__ETC_ZSHRC_SOURCED" -o -n "$NOSYSZSHRC" ]; then return; fi
       __ETC_ZSHRC_SOURCED=1
+
+      # Don't execute this file when running in a pure nix-shell.
+      if test -n "$IN_NIX_SHELL"; then return; fi
+
+      if [ -z "$__NIX_DARWIN_SET_ENVIRONMENT_DONE" ]; then
+        . ${config.system.build.setEnvironment}
+      fi
 
       # history defaults
       SAVEHIST=2000
--- a/tests/programs-zsh.nix
+++ b/tests/programs-zsh.nix
@@ -20,13 +20,13 @@
      echo >&2 "checking for share/zsh in /sw"
      test -e ${config.out}/sw/share/zsh
 
-     echo >&2 "checking setEnvironment in /etc/zshenv"
-     fgrep '. ${config.system.build.setEnvironment}' ${config.out}/etc/zshenv
      echo >&2 "checking nix-shell return /etc/zshenv"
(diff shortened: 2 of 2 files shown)


c4dr01d:
This PR can solve my problem #271

starcraft66:
This PR is really just a work-around for an issue in the darwin daemon installation of Nix itself. The NIX_PATH load order has already been fixed twice in Nix but for some reason, the fix hasn't been backported to the master branch. I'm leaving this open in case it can help anyone but it seems to break a test I haven't been able to look into.

LnL7:
I created https://github.com/LnL7/nix-darwin/compare/installer-link-etc-files as a way to resolve this in a nicer way, just haven't verified that this restores everything properly after the uninstaller.

indiv0:
Does anyone have a recommendation on how to fix this problem at the moment? I'm completely new to nix, nix-darwin, etc.

It sounds like I can manually modify `/etc/zshrc` after installation with the change from this PR, but I think that goes against Nix's philosophy. Is there a way for me to apply this PR (or your linked PR @LnL7) locally?

the-destro:
@LnL7 do you need assistance verifying that the uninstaller on your branch works the way you want? I'm willing to help

Enzime:
Closing this as it should no longer be an issue as `programs.zsh.enable` defaults to `true` as of #1134 and we fully manage `/etc/zshrc` as of #687 which means that `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh` shouldn't get sourced anymore as the same behaviour is already implemented in `nix-darwin`

PR #288, opened 2021-02-05
fish: fix fenv function path

r-burns:
Necessary as of https://github.com/NixOS/nixpkgs/commit/d94921db12d9c9e2d3e9c2e6e25e50f03f789ecb

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -20,12 +20,17 @@ let
 
   fenv = pkgs.fishPlugins.foreign-env or pkgs.fish-foreign-env;
 
+  # fishPlugins.foreign-env and fish-foreign-env have different function paths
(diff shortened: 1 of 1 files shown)


r-burns:
Hmm I wasn't aware of https://github.com/LnL7/nix-darwin/pull/270 when I opened this, but that didn't address this specific issue. I'm on nixpkgs master, and unless I use this fix, fish tries to use the old incorrect fenv function path and fails to set up my environment properly. Maybe anyone still using fenv (not babelfish) hadn't encountered this issue yet, because https://github.com/LnL7/nix-darwin/pull/283 needed to be addressed first?

LnL7 (approved):
This is pretty ugly, but I guess there's not really a nice solution for this.

PR #289, opened 2021-02-09
adds gitlab-runner

angerman:
This can be used similar to the linux service (in fact the module is almost identical to the linux one) with
```nix
  users.knownGroups = [ "gitlab-runner" ];
  users.knownUsers = [ "gitlab-runner" ];

  services.gitlab-runner = {
    enable = true;
    extraPackages = [ pkgs.nix ];
    services = {
      "aarch64-darwin-1" = {
        registrationConfigFile = pkgs.writeText "gitlab-runner-registration" ''
        CI_SERVER_URL=<gitlab server>
        REGISTRATION_TOKEN=<token from gitlab>
        '';
  	executor = "shell";
  	tagList = [ "aarch64-darwin" "nix" ];
      };
    };
  };
```

now, whether or not using a shell executor is a smart idea is a question one needs to be aware of when doing this.

One thing to note: it requires
```nix
  networking.hostName = ...
```
to be set, *but* this fails with
```
configuring networking...
SCPreferencesSetLocalHostName() failed: Invalid argument
```
:-/

### DIFF
--- a/modules/services/gitlab-runner.nix
+++ b/modules/services/gitlab-runner.nix
@@ -526,23 +526,23 @@ in
 
     users.users.gitlab-runner =
       { name = "gitlab-runner";
-        uid = mkDefault 532;
-        gid = mkDefault config.users.groups.gitlab-runner.gid;
+        # uid = mkDefault 532;
+        # gid = mkDefault config.users.groups.gitlab-runner.gid;
         home = mkDefault "/var/lib/gitlab-runner";
         shell = "/bin/bash";
         description = "Gitlab agent user";
       };
     users.groups.gitlab-runner =
       { name = "gitlab-runner";
-        gid = mkDefault 532;
+        # gid = mkDefault 532;
         description = "Gitlab agent user group";
       };
 
 
-    system.activationScripts.preActivation.text = let user = config.users.users.gitlab-runner; in ''
-      mkdir -p '${user.home}'
-      chown ${toString user.uid}:${toString user.gid} '${user.home}'
-    '';
+    # system.activationScripts.preActivation.text = let user = config.users.users.gitlab-runner; in ''
+    #  mkdir -p '${user.home}'
+    #  chown ${toString user.uid}:${toString user.gid} '${user.home}'
+    #'';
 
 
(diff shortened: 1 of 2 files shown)


modules/services/gitlab-runner.nix
```diff
@@ -0,0 +1,643 @@
...
+          };
+        };
+      });
+    };
+  };
+  config = mkIf cfg.enable {
+
+    users.users.gitlab-runner =
+      { name = "gitlab-runner";
+        uid = mkDefault 532;
```
LnL7, line 529: I think a uid from a different range would be better, regular users get ids starting with 501 which could result in conflicts quite quickly.

modules/services/gitlab-runner.nix
```diff
@@ -0,0 +1,643 @@
...
+
+        script = ''
+            ${configureScript}/bin/gitlab-runner-configure && ${startScript}/bin/gitlab-runner-start
+        '';
+
+        serviceConfig = {
+            ProcessType = "Interactive";
+            ThrottleInterval = 30;
+
+          StandardOutPath = "/var/lib/gitlab-runner/out.log";
```
LnL7, line 577: Note that launchd doesn't do any logfile rotation, unless there's a very specific reason it's probably better to use syslog by default.

modules/services/gitlab-runner.nix
```diff
@@ -0,0 +1,643 @@
...
+        description = "Gitlab agent user";
+      };
+    users.groups.gitlab-runner =
+      { name = "gitlab-runner";
+        gid = mkDefault 532;
+        description = "Gitlab agent user group";
+      };
+
+
+    system.activationScripts.preActivation.text = let user = config.users.users.gitlab-runner; in ''
```
LnL7, line 542: The launchd module should probably handle this automatically but the `preActivation`, etc. are intended purely for users. Using priorities in a configuration would break the module which isn't ideal.

LnL7 (commented):
I only glanced over the parts that I'd expect to be the same on nixos, so mostly just smaller remarks.

bgamari:
What is the status of this? GHC is using this branch for our CI infrastructure; it would be great if this could be upstreamed.

domenkozar (approved):

PR #290, opened 2021-02-10
hercules-ci-agent: init

### LINKED ISSUE #291: meta.maintainers for modules
The module system supports a `meta` attribute in the module syntax that adds to the `meta.<...>` options, but I don't see that feature used in nix-darwin. I could open a PR based on NixOS's solution if you think it makes sense to use this feature.

roberth:
Source files originate from the hercules-ci-agent repository and I will make sure to keep them in sync, bidirectionally.

The module is split into two files to make maintenance of the common parts with NixOS easier.

Unlike the module that existed in the hercules-ci fork for a long time, this module reuses the hercules-ci-agent package from Nixpkgs, so it will require little maintenance in nix-darwin itself.

UPDATE: force-pushed to retry unrelated github actions failure
UPDATE 2: still fails. It seems to be a nixUnstable bug, which is unrelated to this pr.

Closes #291

### DIFF
--- a/modules/services/hercules-ci-agent/default.nix
+++ b/modules/services/hercules-ci-agent/default.nix
@@ -8,6 +8,10 @@ in
 {
   imports = [ ./common.nix ];
 
+  meta.maintainers = [
+    lib.maintainers.roberth or "roberth"
+  ];
+
   options.services.hercules-ci-agent = {
 
     logFile = mkOption {
--- a/modules/services/hercules-ci-agent/default.nix
+++ b/modules/services/hercules-ci-agent/default.nix
@@ -69,5 +69,13 @@ in
       name = "_hercules-ci-agent";
       description = "System group for the Hercules CI Agent";
     };
+
+    services.hercules-ci-agent.settings.labels = {
+      darwin.label = config.system.darwinLabel;
+      darwin.revision = config.system.darwinRevision;
+      darwin.version = config.system.darwinVersion;
+      darwin.nix.useDaemon = config.nix.useDaemon;
+      darwin.nix.useSandbox = config.nix.useSandbox;
+    };
   };
 }
--- a/modules/services/hercules-ci-agent/default.nix
+++ b/modules/services/hercules-ci-agent/default.nix
@@ -1,0 +1,67 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let
+  cfg = config.services.hercules-ci-agent;
+  user = config.users.users.hercules-ci-agent;
+in
+{
+  imports = [ ./common.nix ];
+
+  options.services.hercules-ci-agent = {
+
+    logFile = mkOption {
+      type = types.path;
+      default = "/var/log/hercules-ci-agent.log";
+      description = "Stdout and sterr of hercules-ci-agent process.";
+    };
+  };
+
+  config = mkIf cfg.enable {
+    launchd.daemons.hercules-ci-agent = {
+      script = "exec ${cfg.package}/bin/hercules-ci-agent --config ${cfg.tomlFile}";
+
+      path = [ config.nix.package ];
+      environment = {
+        NIX_SSL_CERT_FILE = "${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt";
+      };
+
+      serviceConfig.KeepAlive = true;
+      serviceConfig.RunAtLoad = true;
+      serviceConfig.StandardErrorPath = cfg.logFile;
+      serviceConfig.StandardOutPath = cfg.logFile;
+      serviceConfig.GroupName = "hercules-ci-agent";
+      serviceConfig.UserName = "hercules-ci-agent";
+      serviceConfig.WorkingDirectory = user.home;
+      serviceConfig.WatchPaths = [
+        cfg.settings.staticSecretsDirectory
+      ];
+    };
+
+    system.activationScripts.preActivation.text = ''
+      touch '${cfg.logFile}'
+      chown ${toString user.uid}:${toString user.gid} '${cfg.logFile}'
+    '';
+
+    # Trusted user allows simplified configuration and better performance
+    # when operating in a cluster.
+    nix.trustedUsers = [ "hercules-ci-agent" ];
+    services.hercules-ci-agent.settings.nixUserIsTrusted = true;
+
+    users.knownGroups = [ "hercules-ci-agent" ];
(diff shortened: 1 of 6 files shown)


modules/services/hercules-ci-agent/default.nix
```diff
@@ -0,0 +1,67 @@
...
+    # Trusted user allows simplified configuration and better performance
+    # when operating in a cluster.
+    nix.trustedUsers = [ "hercules-ci-agent" ];
+    services.hercules-ci-agent.settings.nixUserIsTrusted = true;
+
+    users.knownGroups = [ "hercules-ci-agent" ];
+    users.knownUsers = [ "hercules-ci-agent" ];
+
+    users.users.hercules-ci-agent = {
+      uid = mkDefault 532;
```
LnL7, line 55: I think a uid from a different range would be better, regular users get ids starting with 501 which could result in conflicts quite quickly.

dhess:
Also, will this not have the same issue with System Update that `nixbld*` / `_nixbld*` appears to?

LnL7:
That depends on what the actual trigger for that problem is. Unlike that case most service users like this should have a homedir, which might result in the problem not occurring. If it also causes problems then that migration bug is even more crazy than I thought.

modules/services/hercules-ci-agent/default.nix
roberth: This follows the example set by the existing services `hydra` and `ofborg`.
I've looked through the launchd-related man pages but couldn't find any recommendations and my googling wasn't successful either. Do you have a suggestion for a range?

LnL7: Hmm, I guess I made the assumption that there are most likely only a couple regular users on a machine used for CI.
From what I understand of the nixbld discussion role accounts (kind of hidden unix users?) have to be in the 200-300 range, of which the build users will already take up a decent chunk depending on the size of the machine.

roberth: I'm ok with whatever scheme you decide.
I think you're referencing this PR https://github.com/NixOS/nix/pull/4532 which suggests to use the 300-400 range.
Perhaps counting down (https://github.com/NixOS/nix/pull/4532#issuecomment-774791756), so what about 399?

roberth: `gid` has the check `check = t: isInt t && t > 501;`. I think we should loosen that to `>= 300`

modules/services/hercules-ci-agent/default.nix
```diff
@@ -0,0 +1,69 @@
...
+    services.hercules-ci-agent.settings.nixUserIsTrusted = true;
+
+    users.knownGroups = [ "hercules-ci-agent" ];
+    users.knownUsers = [ "hercules-ci-agent" ];
+
+    users.users.hercules-ci-agent = {
+      uid = mkDefault 399;
+      gid = mkDefault config.users.groups.hercules-ci-agent.gid;
+      home = mkDefault cfg.settings.baseDirectory;
+      name = "_hercules-ci-agent";
```
roberth, line 58: This seems to have caused the removal of the old `hercules-ci-agent` user but not the creation of the `_hercules-ci-agent` user.

roberth: users/default.nix is confusing attribute keys and names `ConfigName` and `Name` below.

This breaks `users.users.<name>.name`. (`<name>` corresponds to `ConfigName` and `name` corresponds to `Name` below)

```
  # isCreated : [ConfigName] -> ConfigName -> Bool
  isCreated = listOfKnown: name: elem name listOfKnown;
  # isDeleted : Map *ConfigName *Config -> Name -> Bool
  isDeleted = attrs: name: ! elem name (mapAttrsToList (n: v: v.name) attrs);

  # gids : Map GIDString Name
  gids = mapAttrsToList (n: toGID) (filterAttrs (n: v: isCreated cfg.knownGroups /* expecting ConfigName but actual is Name */ v.name) cfg.groups);
  # uids : Map UIDString Name
  uids = mapAttrsToList (n: toUID) (filterAttrs (n: v: isCreated cfg.knownUsers /* expecting ConfigName but actual is Name */ v.name) cfg.users);

  # createdGroups : [GroupConfig]
  createdGroups = mapAttrsToList (n: v: cfg.groups."${v}") cfg.gids;
  # createdGroups : [UserConfig]
  createdUsers = mapAttrsToList (n: v: cfg.users."${v}") cfg.uids;
  # deletedGroups : [GroupConfigName]
  deletedGroups = filter (n: isDeleted cfg.groups n) cfg.knownGroups;
  # deletedUsers : [UserConfigName]
  deletedUsers = filter (n: isDeleted cfg.users n) cfg.knownUsers;
```

roberth: I guess I was hoping this could support renames. I'll stick with delete and create for the migration.

roberth: I've turned this finding into an issue https://github.com/LnL7/nix-darwin/issues/294

roberth: This PR does not depend on a fix for it.

roberth: Done.

roberth:
I've changed the user to a have a lower uid and an underscore prefix, as hinted by the previously referenced discussion  https://github.com/NixOS/nix/pull/4532. Is there anything else I should change?

roberth:
@LnL7 I've implemented the change. Do you think this can be merged?

roberth:
I've added `meta.maintainers` with the recommendation to use `maintainers.alice or "alice"` instead of the usual Nixpkgs way. This wasn't strictly necessary for building nix-darwin configurations, but allows to build on top `maintainers` even when the inevitable happens.

roberth:
Is there anything I can do?

domenkozar:
@roberth is this ready to go?

roberth:
> @roberth is this ready to go?

Has been for a long time, but let me update and test this first.

roberth:
The stuff about system users is already ready to go afaic.
I think this pr got a bit lost because of uncertainty around that change, if I may speculate :)

domenkozar (approved):

domenkozar:
Oh I can't merge because @lnl7 is tagged as a reviewer

roberth:
Closing while syncing up and testing.

roberth:
@domenkozar I've opened a new PR
 - #586

PR #292, opened 2021-02-14
Fix gnupg-agent being kept alive

MaxSchlueter:
The agent is restarted automatically after it has been launched
successfully, this creates a lot of noise in the logs. This change
reloads the agent only if it has crashed, i.e. returns with a non-zero
exit code.


LnL7 (approved):
Thanks!

PR #301, opened 2021-02-26
Support the Apple Silicon Homebrew directory

matchai:
Hey folks! 👋 

This is a small change allowing Homebrew to be used if installed in the new installation directory for Apple Silicon machines, as detailed in the Homebrew 3.0.0 changelogs: https://brew.sh/2021/02/05/homebrew-3.0.0/

PR #303, opened 2021-03-05
Support the Apple Silicon Homebrew directory

weihua-lu:
matchai's PR fix https://github.com/LnL7/nix-darwin/pull/301 would still fail on silicon brew due to path issue, made a quick fix based on that one.

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -204,8 +204,10 @@ in
     system.activationScripts.homebrew.text = mkIf cfg.enable ''
       # Homebrew Bundle
       echo >&2 "Homebrew bundle..."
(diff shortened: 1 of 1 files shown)


LnL7 (commented):
Perhaps #303 is a nicer alternative, for which the default can be replaced with a state version bump at some point? The disadvantage is that this will need to be configured explicitly for now.

malob:
Yeah, my guess is that doing something like #304 makes the most sense here.

malob:
Actually, my guess is that figuring out exactly what the "right" thing to do if any with #304, might take a little more discussion. Given that this PR fixes the current issue, and it's easy to make further changes later, I'd suggest this PR be merged now so that there's a fix available for this issue sooner rather than later.

malob:
@LnL7, just a heads up that this PR can likely be closed since #304 was merged.

PR #304, opened 2021-03-06
homebrew: make brew path configurable

aaschmid:
[x] Tested using `darwin-rebuild switch -I darwin=$(pwd)` using nix daemon
[ ] Update documentation
[ ] write an automated testcase?
...

Reason:
I am currently building a nix package for `brew` which is working in its first alpha version for installing my used "Casks".
(I am happy to share if someone is interested even though I am not sure getting it done the "nix way" 100% - means without non-deterministic side-effects).

I am happy for every comment / feedback :-)

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -52,6 +52,14 @@ in
       '';
     };
 
+    brewPrefix = mkOption {
+      type = types.str;
+      default = "/usr/local/bin";
+      description = ''
+        Customize path prefix where executable of <command>brew</command> is searched for.
+      '';
+    };
+
     cleanup = mkOption {
       type = types.enum [ "none" "uninstall" "zap" ];
       default = "none";
@@ -204,8 +212,8 @@ in
     system.activationScripts.homebrew.text = mkIf cfg.enable ''
       # Homebrew Bundle
       echo >&2 "Homebrew bundle..."
-      if [ -f /usr/local/bin/brew ]; then
-        PATH=/usr/local/bin:$PATH ${brew-bundle-command}
+      if [ -f "${cfg.brewPrefix}/brew" ]; then
+        PATH="${cfg.brewPrefix}":$PATH ${brew-bundle-command}
       else
         echo -e "\e[1;31merror: Homebrew is not installed, skipping...\e[0m" >&2
       fi


aaschmid:
@malob - any thoughts?

malob:
> I am currently building a nix package for brew which is working in its first alpha version for installing my used "Casks".

I'd definitely be interested to see this!

There's something that doesn't currently feel quite right to me about this. For solving the particular issue of the different paths on Intel and Apple Silicon based Macs, #303 is nice and simple, and requires no configuration. In terms of adding support for a Nix package for `brew`, my sense is that an option like `homebrew.package` would fit better with the Nix zen. The thing I'm most confused about is how to reconcile two. Maybe with another option like `homebrew.useNixPackage` that would default to `false`, but could be set to `true` to enable setting the a `brew` package using `homebrew.package`?

aaschmid:
I would also be fine with `homebrew.package` actually thought about this as well. I also agree that this fits better with Nix zen. 

However, I am unsure that the default way of configuring `brew` for `nix` will not be using a separate package as `brew` needs a lot of adjustments and workarounds to be able to run from a readonly path (see below). And that would be my decision driver of this:
* Default will be `brew` package some day if I finish the `brew` package most likely with workarounds -> `homebrew.package`
* Default will be a standard `brew` installation as described on https://docs.brew.sh/Installation -> `homebrew.brewPath` ([Untar anyway](https://docs.brew.sh/Installation#untar-anywhere) already allows arbitrary locations)

Bringing #303 and #304 together could be easy by having the default with `homebrew.brewPath` unset does scanning the default installation locations, while overriding this behaviour with setting it to a dedicated path. Or is this thought to easy?

In my nix package for `brew` I currently had to create one quite huge workaround because `brew` itself downloads and persists a lot of things into it's own directory which I hackily symlinked to something `.local/share/homebrew/...` while installing. I feel quite dirty but I haven't had a better idea yet. On the other side for my use cases this is quite OK as I only install `casks` using brew and that is now done automatically via `darwin-rebuild switch` :-)

aaschmid:
Side note: [here](https://github.com/aaschmid/nix-darwin/blob/brew.pkgs/pkgs/brew.nix) is my current version of a possible nix package for `brew`. Any comments welcome @malob

aaschmid:
@malob any thoughts to my comment above (https://github.com/LnL7/nix-darwin/pull/304#issuecomment-806118845)?

modules/homebrew.nix
```diff
@@ -204,8 +212,8 @@ in
     system.activationScripts.homebrew.text = mkIf cfg.enable ''
       # Homebrew Bundle
       echo >&2 "Homebrew bundle..."
-      if [ -f /usr/local/bin/brew ]; then
-        PATH=/usr/local/bin:$PATH ${brew-bundle-command}
+      if [ -f ${cfg.brewPath} ]; then
+        PATH=''$(dirname ${cfg.brewPath}):$PATH ${brew-bundle-command}
```
LnL7, line 216: Since the dirname is needed here I'd prefer a brewPrefix or package option instead of the full path to the binary.

LnL7 (commented):
Sorry for the delay

modules/homebrew.nix
aaschmid: done

aaschmid:
Thanks!

rpearce:
You're a legend for this PR. Thank you for adding this

PR #305, opened 2021-03-12
spotifyd: add spotifyd service

marsam:
Hi:
I added a [`spotifyd`](https://github.com/spotifyd/spotifyd) service


LnL7 (approved):
Thanks!

PR #306, opened 2021-03-21
Add an option to override the Nix used by darwin-rebuild.

dhess:
This is useful in particular when you want to use a stable version of Nix for the system Nix, but an unstable version (e.g., with flakes support) for darwin-rebuild.

We've encountered quite a few problems lately with nixUnstable and remote builders for CI (Hydra & Hercules CI), so we've reverted to nixStable on these builders. However, I've also gotten very accustomed to managing them with nix-darwin's flake support, so I don't want to lose that in the process.

This commit feels like a bit of a hack, but it was the most straightforward way I could come up with for supporting this scenario. It's backward compatible and shouldn't affect any existing installations.

### DIFF
--- a/modules/nix/nix-darwin.nix
+++ b/modules/nix/nix-darwin.nix
@@ -5,7 +5,9 @@ with lib;
 let
   inherit (pkgs) stdenv;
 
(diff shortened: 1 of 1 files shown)


dhess:
Closing this. It's not particularly useful anymore now that Nix stable has Flakes support.

Issue #307, opened 2021-03-21
darwin-rebuild switch --flake is broken by new nix


ahmedelgabri:
This is what I get when I try to run `darwin-rebuild switch --flake`

```
$ darwin-rebuild switch --flake .
error: 'info' is not a recognised command
Try 'nix --help' for more information.
```

A quick investigation seems like it fails here on this line https://github.com/LnL7/nix-darwin/blob/67da67bb7a843d48c8cdae6770656c5333e75de0/pkgs/nix-tools/darwin-rebuild.sh#L129

```
$ nix flake info --json
error: 'info' is not a recognised command
Try 'nix --help' for more information.
```

This is an issue on nixpkgs that might be related https://github.com/NixOS/nixpkgs/issues/116938

I think the command changed to `nix flakes metadata` instead. `nix flakes --help` & `nix flakes metadata --help` confirms this.

MattPColeman:
Have you found a workaround for this? Just ran into it on a new laptop, can't install my flake dotfiles 😭

hardselius:
@MattPColeman I have a fix pending merge. You can use my fork in the meantime.

MattPColeman:
> @MattPColeman I have a fix pending merge. You can use my fork in the meantime.

hero

Enzime:
Fixed by #308

PR #308, opened 2021-03-21
Fix nix flake info error

hardselius:
Fixes

```
error: 'info' is not a recognised command'
```

Issue: #307

### DIFF
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -126,7 +126,7 @@ if [ -n "$flake" ]; then
 fi
 
 if [ -n "$flake" ]; then
-    flake=$(nix "${flakeFlags[@]}" flake info --json "${extraBuildFlags[@]}" "${extraLockFlags[@]}" -- "$flake" | jq -r .url)
+    flake=$(nix "${flakeFlags[@]}" flake metadata --json "${extraBuildFlags[@]}" "${extraLockFlags[@]}" -- "$flake" | jq -r .url)
 fi
 
 if [ "$action" != build ] && [ -z "$flake" ]; then
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -126,7 +126,7 @@ if [ -n "$flake" ]; then
 fi
 
 if [ -n "$flake" ]; then
-    if nix flake metadata &>/dev/null; then
+    if nix "${flakeFlags[@]}" flake metadata --version &>/dev/null; then
         cmd=metadata
     else
         cmd=info
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -126,7 +126,13 @@ if [ -n "$flake" ]; then
 fi
 
 if [ -n "$flake" ]; then
-    flake=$(nix "${flakeFlags[@]}" flake metadata --json "${extraBuildFlags[@]}" "${extraLockFlags[@]}" -- "$flake" | jq -r .url)
+    if nix flake metadata &>/dev/null; then
+        cmd=metadata
+    else
+        cmd=info
+    fi
+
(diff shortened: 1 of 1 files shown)


eraserhd:
This worked for me.

LnL7 (commented):
I disagree a bit with the approach taken in nixpkgs, not handling the previous case at all anymore. There are various cases where nix might not be upgraded yet pinned inputs, rollback, etc.

At least until it's ok to assume that 2.4pre is out of rotation for most people both cases should be handled.

hardselius:
@LnL7 Yup. I can get behind that. My shell script fu is so-so, but I can probably take a stab at it. I'll gladly accept pointers and suggestions. The approach would be something along the lines of trying the `metadata` sub command and fall back on `info` if it fails?

pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -126,7 +126,7 @@ if [ -n "$flake" ]; then
 fi
 
 if [ -n "$flake" ]; then
-    flake=$(nix "${flakeFlags[@]}" flake info --json "${extraBuildFlags[@]}" "${extraLockFlags[@]}" -- "$flake" | jq -r .url)
+    flake=$(nix "${flakeFlags[@]}" flake metadata --json "${extraBuildFlags[@]}" "${extraLockFlags[@]}" -- "$flake" | jq -r .url)
```
LnL7, line 129: Something kind of like this will probably work.

```suggestion
    if nix flake metadata --help >/dev/null; then
        cmd=metadata
    else
        cmd=info
    fi
    flake=$(nix "${flakeFlags[@]}" flake "$cmd" --json "${extraBuildFlags[@]}" "${extraLockFlags[@]}" -- "$flake" | jq 
```

hardselius: Alright. Pushed. I tested my fork and it looks fine.

hardselius:
Alright. Failed due to rate limiting. I don't think I can trigger a manual re-run.

By the way, I think you can get around the rate limiting issue with something like

```
uses: cachix/install-nix-action@v12
with:
  extra_nix_config: |
    access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}
```

pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -126,7 +126,13 @@ if [ -n "$flake" ]; then
 fi
 
 if [ -n "$flake" ]; then
-    flake=$(nix "${flakeFlags[@]}" flake info --json "${extraBuildFlags[@]}" "${extraLockFlags[@]}" -- "$flake" | jq -r .url)
+    if nix flake metadata &>/dev/null; then
```
emilazy, line 129: This doesn’t work unless you have `experimental-features = nix-command flakes` in your `nix.conf`.
```suggestion
    if nix "${flakeFlags[@]}" flake metadata &>/dev/null; then
```

emilazy: Also… this actually only works if you're in a directory that is actually a flake itself. How about `nix "${flakeFlags[@]}" flake metadata --version &>/dev/null`?

hardselius: Excellent! Thanks!

hardselius: @emilazy I just tested it with your proposed changes in my own dotfiles build pipeline. I'm moving `nix.conf` before switching. It works, whereas it didn't before I added your change. Thanks for pointing that out!

malob:
Looking forward to this landing soon :)

snoord:
Can confirm that the changes in this PR work for me.

```bash
$ nix --version
nix (Nix) 2.4pre20210317_8a5203d
```

antifuchs (approved):
Can confirm that darwin-rebuild works with this fix in place, on `nix (Nix) 2.4pre20201102_550e11f`

happysalada:
Is there anything that I can do to help move this PR forward? No worries if there is nothing to do.

LnL7 (approved):

PR #309, opened 2021-03-23
darwin-rebuild: use `hostname -s` for flakeAttr

emilazy:
Previously this would fail to strip the domain if you set a different FQDN in the macOS interface or with `networking.hostName`.

Another alternative would be to use the full domain unconditionally if it doesn’t end in `.local`, but that’d require at least fixing the quoting.


emilazy:
CI failure seems spurious. Although I admittedly didn’t actually test this.

emilazy:
Rebased in time for Christmas! (No pressure to review/merge over the holidays, of course.)

domenkozar:
#340

PR #317, opened 2021-04-09
Pass system to darwinSystem rather than eval-config.

dhess:
This allows us to specify what kind of darwinSystem we want to build, rather than determining it at evaluation time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ A minimal example of using an existing configuration.nix:
 
   outputs = { self, darwin, nixpkgs }: {
     darwinConfigurations."Johns-MacBook" = darwin.lib.darwinSystem {
+      system = "x86_64-darwin";
       modules = [ ./configuration.nix ];
     };
   };
@@ -90,6 +91,7 @@ accessible as an argument, similar to pkgs and lib inside the configuration.
 
 ```nix
 darwin.lib.darwinSystem {
+  system = "x86_64-darwin";
   modules = [ ... ];
   inputs = { inherit darwin dotfiles nixpkgs; };
 }
--- a/default.nix
+++ b/default.nix
@@ -6,9 +6,10 @@
 }:
 
 let
-  evalConfig = import ./eval-config.nix { inherit lib system; };
+  evalConfig = import ./eval-config.nix { inherit lib; };
 
   eval = evalConfig {
+    inherit system;
     modules = [ configuration ];
     inputs = { inherit nixpkgs; };
   };
--- a/eval-config.nix
+++ b/eval-config.nix
@@ -1,9 +1,10 @@
-{ lib, system ? builtins.currentSystem or "x86_64-darwin" }:
+{ lib }:
 
-{ modules
+{ system ? builtins.currentSystem or "x86_64-darwin"
+, modules
 , inputs
 , baseModules ? import ./modules/module-list.nix
-, specialArgs ? {}
+, specialArgs ? { }
 }@args:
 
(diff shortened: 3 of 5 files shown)


bphenriques:
This should fix #319

dhess:
FYI, I'm not quite sure why the build failed. I've been using this successfully in my fork since I submitted this PR.

domenkozar:
This is a good example how https://github.com/NixOS/nix/issues/3843 would simplify flakes and remove the need to hardcode system like in this case.

domenkozar:
@dhess could you rebase and check if tests pass?

dhess:
Done.

domenkozar:
I've added some design thoughts to https://github.com/NixOS/nix/issues/3843#issuecomment-915177908

domenkozar:
```
error: anonymous function at /nix/store/nb2kapr1yf12ra2x3jmdbm2lwfh9c1pr-source/eval-config.nix:3:1 called with unexpected argument 'system'

       at /nix/store/nb2kapr1yf12ra2x3jmdbm2lwfh9c1pr-source/flake.nix:10:58:

            9|
           10|       darwinSystem = { modules, inputs ? {}, ... }@args: self.lib.evalConfig (args // {
             |                                                          ^
           11|         inputs = { inherit nixpkgs; darwin = self; } // inputs;
```

dhess:
I'll try to take a look at this over the weekend.

emilazy:
This PR worked perfectly for me, thank you! I think @domenkozar's error must be related to issues with specifying this fork in `flake.nix`:

* the excerpted code `darwinSystem = { modules, inputs ? {}, ... }@args:` doesn't appear in the PR branch's `flake.nix`;
* the `/nix/store/nb2kapr1yf12ra2x3jmdbm2lwfh9c1pr-source` source path matches what I get for the current upstream nix-darwin flake, whereas this PR has the source path `/nix/store/b8igm25p67844z4gylbmskndi7r8r607-source` (you can check with `nix flake metadata …`).

The error looks what you'd get passing `system` to the `darwinSystem` from current nix-darwin.

dhess:
Good catch.

I'm not sure how to fix this, or whether it's even worth bothering, given that it won't be a problem once it's merged. I'll wait to hear from @LnL7, in any case.

emilazy:
FWIW, `darwin.url = "github:hackworthltd/nix-darwin/darwinSystem";` was how I successfully tested this PR with flakes.

LnL7:
Ah I see, now the error makes sense. It seems like the `nix registry` override that the tests use doesn't have precedence anymore over the input defined in the flake.

LnL7:
@dhess I can push a fix for the tests if you allow edits from maintainers.

dhess:
> @dhess I can push a fix for the tests if you allow edits from maintainers.

Sorry for being dense, but how do I do that?

LnL7:
There should be a checkbox in the sidebar. I would expect it to be checked by default however, maybe I my push attempt was incorrect?

dhess:
I don't see that option. Perhaps it's because the branch is from a repo owned by an org and not an individual?

LnL7:
Weird, guess it must be something like that. Well you can just cherry-pick it from https://github.com/LnL7/nix-darwin/commits/flakes-test then. 😄

dhess:
Or you could just merge your branch :) I don't care either way, whatever you prefer.

dhess:
Excellent, thanks!

PR #322, opened 2021-04-23
Bundle with /opt/homebrew if brew exists in there

### LINKED ISSUE #321: Homebrew on apple silicon (aarch64) installs to /opt/homebrew
From https://docs.brew.sh/Installation, it sounds like apple silicon homebrew installs will always go into `/opt/homebrew`; that means the path check in https://github.com/LnL7/nix-darwin/blob/3b28c4675a4c6a1fa2a1448e8e9abcfb4cb5c770/modules/homebrew.nix#L207-L211 will never pick it up.

I tried adding a symlink from /usr/local/bin/brew -> /opt/homebrew/bin/brew, and that made the plugin attempt to bundle, but casks failed to install due to the different HOMEBREW_CELLAR settings:

```
Warning: your HOMEBREW_PREFIX is set to /usr/local but HOMEBREW_CELLAR is set
to /opt/homebrew/Cellar. Your current HOMEBREW_CELLAR location will stop
you being able to use all the binary packages (bottles) Homebrew provides. We
recommend you move your HOMEBREW_CELLAR to /usr/local/Cellar which will get you
access to all bottles."
Error: Download failed on Cask 'monodraw' with message: No such file or directory @ dir_s_rmdir - /usr/local/var/homebrew/locks/bb28a6f9dfb330ea41ea6e5184164548dd38b1cb2eebedbe873a141f4445bdaf--Monodraw-b107.zip.incomplete.lock
```

I'm guessing we just have to call the brew program in either /opt/homebrew/bin or /usr/local/bin.

antifuchs:
This fixes #321 - by using /opt/homebrew's brew binary first, we default to the right thing on Apple Silicon (and help anyone who may have used the same work-around as I did).

With this change, I can build&activate my system configuration with homebrew on an M1 macbook air.

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -204,7 +204,9 @@ in
     system.activationScripts.homebrew.text = mkIf cfg.enable ''
       # Homebrew Bundle
       echo >&2 "Homebrew bundle..."
-      if [ -f /usr/local/bin/brew ]; then
+      if [ -x /opt/homebrew/bin/brew ]; then
+        PATH=/opt/homebrew/bin:$PATH ${brew-bundle-command}
+      elif [ -f /usr/local/bin/brew ]; then
         PATH=/usr/local/bin:$PATH ${brew-bundle-command}
       else
         echo -e "\e[1;31merror: Homebrew is not installed, skipping...\e[0m" >&2


rubas:
Thanks. Just ran into the same issue.

antifuchs:
Note & heads-up though: If you attempt to install formulas (not casks or mas apps) in currently-unpatched nix, this will fail to work due to an architecture mismatch. Because nix is running as an x86-64 program, anything launched by it will also be running in x86-64 mode (especially universal binaries), so we may have to use `arch -arm64` in the "using arm64 homebrew" case. Not sure that's the right choice when we finally get Apple Silicon support in nix proper, though.

purcell:
Thanks from me too - this is helpful. Despite the caveat, I think this should be safe to merge, since it at least makes `cask`/`mas` apps work on M1 machines.

purcell:
Hmm, #304 covers this case too.

antifuchs:
Oh! #304 seems to be quite good too - I think that should definitely go in (and I missed #303, which I'd missed when I searched for issues/PRs). I think when #304 goes in, this PR will have to change to adjust for the custom-settable dir (or we should default it to the right value per-platform), but either way, I think we ought to support the different dirs (:

purcell:
#304 is merged now, btw, so this PR could theoretically be fixed up to better detect the defaults.

malob:
Agree with @purcell. My take is that for this PR to still be relevant, the default value for `homebrew.brewPrefix` should be set based on the architecture. I suggest changing this line: https://github.com/LnL7/nix-darwin/blob/5851d9613edf8b2279746c7e5b9faac55ff17e8a/modules/homebrew.nix#L57

To something like:
```
default = if (pkgs.stdenv.hostPlatform.darwinArch == "aarch64") then "/opt/homebrew/bin" else "/usr/local/bin";
```

jmatsushita (approved):
LGTM

purcell:
I filed #418 as a replacement for this PR, using approximately the code suggested by @malob.

LnL7:
Closing this since #418 was merged.

PR #332, opened 2021-05-12
change inner double quotes to single quotes

jfchevrette:
The double quotes conflicted with the outer ones resulting in invalid instruction being given to the user

Before
```
$ echo runtprivate/var/run | sudo tee -a /etc/synthetic.conf
```

After
```
$ echo 'run	private/var/run' | sudo tee -a /etc/synthetic.conf
```


LnL7 (approved):
Thanks!

PR #341, opened 2021-07-10
work around multi-user nix zsh evaluation order

ylh:
Presented for consideration is a solution less intrusive than either #286 or 34e672b. I've got a branch of my own that's just this commit plus the channel added by the installer pointing to that branch, and I can confirm that it does a fine job fixing environment clobbering by `nix-daemon.sh` in the case of a multi-user install.


Enzime:
Closing this as it should no longer be an issue as `programs.zsh.enable` defaults to `true` as of #1134 and we fully manage `/etc/zshrc` as of #687 which means that `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh` shouldn't get sourced anymore as the same behaviour is already implemented in `nix-darwin`

PR #352, opened 2021-08-08
fish: remove `promptInit` in favor of `interactiveShellInit`

kidonng:
Derived from nix-community/home-manager#2231

`promptInit` does nothing special and can already be implemented via `interactiveShellInit`, so I suggest just keep the latter.

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -44,6 +44,14 @@ in
 
 {
 
+  imports = [
+    (mkRemovedOptionModule [ "programs" "fish" "promptInit" ] ''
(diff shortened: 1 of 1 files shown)


emilazy:
Thanks for this. While I agree that it seems like this option is probably pointless, it's still present upstream in NixOS, and in our `zsh` module too. I realize that you probably don't want to go to the effort of trying to get this deprecated in NixOS too, especially if you don't use it, but I'm inclined to err on the side of keeping configuration compatible between NixOS and nix-darwin in the absence of a strong reason to diverge.

PR #353, opened 2021-08-09
fish: simplify config guards

kidonng:
Derived from nix-community/home-manager#2230

There's no need to use separate variables for each section. We only use one variable for such usage in nix-darwin's [Zsh module](https://github.com/LnL7/nix-darwin/blob/007d700e644ac588ad6668e6439950a5b6e2ff64/modules/programs/zsh/default.nix#L112-L115).

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -187,43 +187,27 @@ in
         etc."fish/config.fish".text = ''
         # /etc/fish/config.fish: DO NOT EDIT -- this file has been generated automatically.
 
-        # if we haven't sourced the general config, do it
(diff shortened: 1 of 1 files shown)


emilazy:
It looks like this logic is based on the upstream NixOS module. I believe this change might break things if you run a login shell inside a non-login shell, or if you have an interactive shell nested inside a non-interactive shell? I'm not sure about the Bash module, but the Zsh one seems to effectively match this logic, by writing separate `zprofile` and `zshrc` files with their own guards. Fish doesn't have separate configuration files for these two cases, so I believe the logic here better matches what's going on for Zsh.

PR #357, opened 2021-08-24
defaults: magic mouse option

eraserhd:
This allows setting the magic mouse to "one button" or "two button" mode.

There are other options, but I'm not familiar with them and don't use them.  Also, maybe `magicmouse` is not the right name?


LnL7 (approved):
Thanks!

PR #379, opened 2021-11-14
zsh: _nix collision from nix-zsh-completions on >=2.4pre

### LINKED ISSUE #373: Completions for nix will conflict with Nix 2.4
Nix 2.4 defines its own zsh completions. When I used https://github.com/NixOS/nixpkgs/pull/144197's `pkgs.nix` as my `nix.package`, I started getting errors during the build:

```
error: collision between `/nix/store/j5i4bhqa2dh8wk98l0g2csnmz63l2k9y-nix-2.4/share/zsh/site-functions/_nix' and `/nix/store/bbz7556b8yk668y1dpyma3qcwhrswf0d-nix-zsh-completions-0.4.4/share/zsh/site-functions/_nix'
```

I believe the problem is that enabling the `programs.zsh.enableCompletion` option does two things:

https://github.com/LnL7/nix-darwin/blob/44da835ac40dab5fd231298b59d83487382d2fab/modules/programs/zsh/default.nix#L103

1. adding `pkgs.nix-zsh-completions` to system packages
2. run `autoload -U compinit && compinit` in /etc/zshrc

I would suggest either splitting the option into two (both defaulting to true, to be consistent with the current behavior) or removing the first behavior.

As a workaround, I'm looking into whether I can rely on home-manager's zsh setup to call compinit instead (and disable this flag completely), worst case overriding `nix-zsh-completions` to an almost-empty package.

Update: I had something else that was wrong in nix-darwin setup, apparently just disabling the option here makes things okay (home-manager's zsh will call compinit anyway)

antifuchs:
This ports over a conditional from nixpkgs, which prevents a collision between zsh completion files when nix >=2.4pre (and consequently, 2.4 release) is in use: Remove the _nix definitions from nix-zsh-completions.

This PR fixes #373.

### DIFF
--- a/modules/programs/zsh/default.nix
+++ b/modules/programs/zsh/default.nix
@@ -96,12 +96,24 @@ in
   };
 
   config = mkIf cfg.enable {
-
     environment.systemPackages =
-      [ # Include zsh package
+      let
+        completions =
+          if lib.versionAtLeast (lib.getVersion config.nix.package) "2.4pre"
+          then
+            pkgs.nix-zsh-completions.overrideAttrs
+              (oldAttrs: {
+                installPhase = oldAttrs.installPhase + ''
+                  rm $out/share/zsh/site-functions/_nix
+                '';
+              })
+          else pkgs.nix-zsh-completions;
+      in
+      [
+        # Include zsh package
         pkgs.zsh
-      ] ++ optional cfg.enableCompletion pkgs.nix-zsh-completions
-        ++ optional cfg.enableSyntaxHighlighting pkgs.zsh-syntax-highlighting;
+      ] ++ optional cfg.enableCompletion completions
+      ++ optional cfg.enableSyntaxHighlighting pkgs.zsh-syntax-highlighting;
 
     environment.pathsToLink = [ "/share/zsh" ];
 


kubukoz (approved):

dhess:
I don't understand why, but this doesn't work for me. For the record, I'm not using `home-manager` in the particular config that I'm using to test it.

antifuchs:
So I just tried this somewhat harder than before & it fails for me, too. I'll convert this to a draft, as more work is definitely needed.

antifuchs:
Updated this with a patch to nix-zsh-completions that actually results in a non-installed `_nix` file. I'm not sure why setting a `postInstall` on a derivation that has an `installHook` doesn't seem to do anything, but that's what it looks like: I had to append to the `installHook`, and now things no longer conflict (:

dhess:
Yay, it works! Thank you!

LnL7 (commented):
The proper approach to handle these type of known conflicts is to give the package a priority. eg.

```nix
self: super: {
  nix-zsh-completions = super.nix-zsh-completions // {
    meta = super.nix-zsh-completions.meta // {
      priority = 6;  # prevent collision with nixUnstable
    };
  };
}
```

antifuchs:
Huh, ok. I guess I should port *that* over to nixpkgs.

antifuchs:
Filed https://github.com/NixOS/nixpkgs/pull/146030

antifuchs:
The fix has been merged into nixpkgs. Should land on unstable soon.

LnL7:
Great!

Issue #360, opened 2021-09-17
`system.activationScripts` missing `deps` attribute


montchr:
I've noticed, via https://github.com/ryantm/agenix/issues/60, that nix-darwin's `system.activationScripts` option does not include a declaration for `deps` while NixOS seems to expect it for each activation script (with a default value of an empty list).

From https://github.com/NixOS/nixpkgs/blob/8dd8bd8be74879f9f7919b16a4cb5ab2a75f18e5/nixos/modules/system/activation/activation-script.nix#L31-L43 :

```nix
  scriptType = with types;
    let scriptOptions =
      { deps = mkOption
          { type = types.listOf types.str;
            default = [ ];
            description = "List of dependencies. The script will run after these.";
          };
        text = mkOption
          { type = types.lines;
            description = "The content of the script.";
          };
      };
    in either str (submodule { options = scriptOptions; });
```

Because of this, I think, some modules (like the one provided by agenix, linked above) which provide a definition for an activation script with NixOS compatibility will fail with nix-darwin.

Disclaimer: I am new-ish to Nix and can not yet consider myself a NixOS user so I apologize if I've misunderstood the issue.

montchr:
Closing since this is specific to systemd and not a "bug" here.

PR #381, opened 2021-11-22
security: add agenix module

considerate:
This PR adapts the [agenix NixOS module](https://github.com/ryantm/agenix/blob/main/modules/age.nix) to nix-darwin's `system.activationScripts` structure and makes it available and configurable at `config.age.{secrets,sshKeyPaths}`.

### DIFF
--- a/modules/security/age/default.nix
+++ b/modules/security/age/default.nix
@@ -1,0 +1,117 @@
+# adapted from https://github.com/ryantm/agenix/blob/53aa91b4170da35a96fab1577c9a34bc0da44e27/modules/age.nix
+{ config, options, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.age;
+
+  # we need at least rage 0.5.0 to support ssh keys
+  inherit (pkgs) rage;
+  ageBin = "${rage}/bin/rage";
+
+  users = config.users.users;
+
+  identities = builtins.concatStringsSep " " (map (path: "-i ${path}") cfg.sshKeyPaths);
+  installSecret = secretType: ''
+    echo "decrypting ${secretType.file} to ${secretType.path}..."
+    TMP_FILE="${secretType.path}.tmp"
+    mkdir -p $(dirname ${secretType.path})
+    (
+      umask u=r,g=,o=
+      ${ageBin} --decrypt ${identities} -o "$TMP_FILE" "${secretType.file}"
+    )
+    chmod ${secretType.mode} "$TMP_FILE"
+    chown ${secretType.owner}:${secretType.group} "$TMP_FILE"
+    mv -f "$TMP_FILE" '${secretType.path}'
+  '';
+
+  isRootSecret = st: (st.owner == "root" || st.owner == "0") && (st.group == "root" || st.group == "0");
+  isNotRootSecret = st: !(isRootSecret st);
+
+  rootOwnedSecrets = builtins.filter isRootSecret (builtins.attrValues cfg.secrets);
+  installRootOwnedSecrets = builtins.concatStringsSep "\n" ([ "echo '[agenix] decrypting root secrets...'" ] ++ (map installSecret rootOwnedSecrets));
+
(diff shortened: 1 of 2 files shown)


montchr:
Awesome! I'll have to check this out. Seems like it might help address the issue I was running into with https://github.com/ryantm/agenix/issues/60 (and also sort of #360 in this repo).

considerate:
@montchr I think it should solve the issues you linked, at least the usage of the non-existent option `deps` is now gone. Please let me know of your experience trying this out.

montchr:
Yeah, it does resolve those issues 😄 Thanks!

Though I may need to find something other than `agenix` for secret management. When `rage` tried to decrypt my test file I got a strange error – "Error: identity file ... contains non-identity data on line 1". So I'm not sure that agenix is suited for my current approach (GPG controlling ssh-agent with the private key stored on a Yubikey).

considerate:
> Yeah, it does resolve those issues :smile: Thanks!

Great, thanks for letting me know.

> I got a strange error – "Error: identity file ... contains non-identity data on line 1".

I think this might be because you haven't added your machine's public key to the secrets.nix file or because you haven't added your machine's private to to sshKeyPaths in the agenix config.

modules/security/age/default.nix
```diff
@@ -0,0 +1,117 @@
...
+        default = "/run/secrets/${config.name}";
+        description = ''
+          Path where the decrypted secret is installed.
+        '';
+      };
+      mode = mkOption {
+        type = types.str;
+        default = "0400";
+        description = ''
+          Permissions mode of the in octal.
```
montchr, line 64:
```suggestion
          Permissions mode of the decrypted secret in octal format.
```

montchr (commented):
I've got this working in my system configuration. I opted to ditch my gpg-agent/yubikey setup to simplify things a bit, so I just created a regular old SSH key stored locally on this machine.

One suggestion (other than a wording change):

`agenix` has [recently added support for non-symlinked secrets](https://github.com/ryantm/agenix/pull/27/files) – ideally, a new `age` module for nix-darwin would support that too. Right now, that's not the case:

```
error: The option `age.secrets.testSecret.symlink' does not exist. Definition values:
       - In `/nix/store/1lki7ldadihkcspd2xz4ssa7mawxiqyp-source/users/profiles/secrets': false
(use '--show-trace' to show detailed location information)
```

I'd be happy to try my hand at implementing it when I get some time in the next few days.

happysalada:
Really interested in this, friendly ping!

montchr:
My personal take on this after some thought:

Although I used this module a few months ago, I've since moved onto using a couple in-progress PRs on https://github.com/ryantm/agenix (see below). I'm also a fan of https://github.com/cole-h/agenix-cli/ which provides a somewhat simpler UX. 

I'd recommend testing one of the alternative implementations whose scope is clearly defined and will be more likely to be maintained in the long run. There have already been quite a few significant changes to agenix which aren't reflected here, and there's two candidates for darwin support directly in agenix as https://github.com/ryantm/agenix/pull/79 (out of date, possibly abandoned?) and https://github.com/ryantm/agenix/pull/107 (I opened this one based off the work in the other PR to reconcile it with the more recent additions to agenix).

Is there something that a nix-darwin module for agenix would provide that a module within agenix itself cannot provide? Or, perhaps, is there a way this darwin module could be reshaped to provide something that agenix does not provide (aside from working darwin support)?

considerate:
@montchr I think implementing this in agenix directly makes sense. I added this PR because I didn't know whether I could create a module upstream that cleanly implements both the nixos and the nix-darwin modules. It seems that https://github.com/ryantm/agenix/pull/107 is getting some traction so I'll close this PR in the hopes that the upstream PR will get merged in due time.

> Is there something that a nix-darwin module for agenix would provide that a module within agenix itself cannot provide?

I don't think so. The main benefit of implementing it here would be that it wouldn't make the upstream repo as complicated and it's more likely that nix-darwin specific features get implemented more frequently. That said, I think (https://github.com/ryantm/agenix/pull/107) shows that the conditional features are quite contained and most of the functionality can be nicely shared between the platforms.

PR #384, opened 2021-11-28
Service Yabai: Use Root User for scripting addition

andrew-womeldorf:
[The yabai scripting addition must be loaded by a root user](https://github.com/koekeishiya/yabai/wiki/Installing-yabai-(latest-release)#macos-big-sur---automatically-load-scripting-addition-on-startup). Not running as root causes the service to exit (`1`) and restart after 10 seconds, and each restart causes this awful flash on the screen.


andrew-womeldorf:
Actually - hold on this. This doesn't actually solve the problem.

d12frosted:
Any ideas on how to solve this? Just run into the same problem on a fresh install.

PR #388, opened 2021-12-02
fix: change user doc hint for synthetic.conf to output literal tabs

virchau13:
This PR makes sure that the hint that the installer gives you when `/run` doesn't exist actually outputs a literal tab to `/etc/synthetic.conf`. The code before appended the literal string`run\tprivate/var/run` into `/etc/synthetic.conf`, which doesn't work.

### DIFF
--- a/modules/system/checks.nix
+++ b/modules/system/checks.nix
@@ -27,7 +27,7 @@ let
         echo "Create a symlink to /var/run with:" >&2
         if test -e /etc/synthetic.conf; then
             echo >&2
-            echo "$ echo 'run\tprivate/var/run' | sudo tee -a /etc/synthetic.conf" >&2
+            echo "$ printf 'run\tprivate/var/run\n' | sudo tee -a /etc/synthetic.conf" >&2
             echo "$ /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -B" >&2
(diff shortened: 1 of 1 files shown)


modules/system/checks.nix
```diff
@@ -27,7 +27,7 @@ let
         echo "Create a symlink to /var/run with:" >&2
         if test -e /etc/synthetic.conf; then
             echo >&2
-            echo "$ echo 'run\tprivate/var/run' | sudo tee -a /etc/synthetic.conf" >&2
+            echo "$ printf 'run\tprivate/var/run\n' | sudo tee -a /etc/synthetic.conf" >&2
             echo "$ /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -B" >&2
```
virchau13, line 31: This flag (`-B`) doesn't exist anymore in `apfs.util`, so I got confused as to what to do with this command, when the solution was to just ignore this command. Should this line be removed, or is this relevant on older versions of macOS?

LnL7: Yeah the flags changed at some point so the installer tries both, this message was added before the flag changed.

emilazy:
Thank you for this; looks like https://github.com/LnL7/nix-darwin/pull/453 did an equivalent change.

Issue #336, opened 2021-06-17
Relative `private/var/run` path


kidonng:
I noticed there are some code and instructions using relative `private/var/run` instead of the full `/private/var/run`, the most conspicuous example being the first command under the "Manual Install" section of the README: `sudo ln -s private/var/run /run`. From what I can tell, this won't work unless the user is at `/` or `/System/Volumes/Data`. Is this a mistake or intended?

Also, pardon me for being dumb here, how is the `$PWD` set so the code can run correctly?

LnL7:
Symlinks can be relative, what a relative symlink points to depends on it's location not your working directory. So in this case it's equivalent, If you take a look at the root layout on macOS you'll see that builtin symlinks are created the same way.

```
lrwxr-xr-x   1 root admin   11 Dec  5  2019 etc -> private/etc
lrwxr-xr-x   1 root admin   11 Dec  5  2019 tmp -> private/tmp
lrwxr-xr-x   1 root admin   11 Dec  5  2019 var -> private/var
```

kidonng:
Ohh you are right, I was being dumb and thinking about the relative links will be resolved to absolute path.

Still, I think the `sudo ln -s private/var/run /run` command will fail after 10.15, right? I think a note in the README is still beneficial.

LnL7:
Ah indeed, the installer has been updated to handle this but the manual install section doesn't mention any of the new needed steps.

PR #389, opened 2021-12-06
Add manual install command to create symlink after 10.15 Catalina

kidonng:
Resolve https://github.com/LnL7/nix-darwin/issues/336#issuecomment-868720301


LnL7 (approved):
Thanks!

Issue #374, opened 2021-11-06
config.nixpkgs.overlays incorrectly disallows functors


lilyball:
nixpkgs doesn't seem to care if an overlay is a function or a functor. But nix-darwin declares `config.nixpkgs.overlays` such that it uses `builtins.isFunction` instead of `lib.isFunction`. The latter supports functors while the former does not.

emilazy:
Seems to have been fixed by https://github.com/LnL7/nix-darwin/pull/394.

PR #394, opened 2021-12-07
nixpkgs: Allow functors.

jsoo1:
Replace builtins.isFunction with lib.isFunction, which allows functors.

#374 

Not sure this is exactly what is required here, I am not 100% on the nixpkgs functor internals.

### DIFF
--- a/modules/nix/nixpkgs.nix
+++ b/modules/nix/nixpkgs.nix
@@ -4,10 +4,10 @@ with lib;
 
 let
   isConfig = x:
-    builtins.isAttrs x || builtins.isFunction x;
+    builtins.isAttrs x || lib.isFunction x;
 
   optCall = f: x:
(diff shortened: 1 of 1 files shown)


LnL7 (approved):
Using functors sounds like a pretty weird usecase but looks good. 😄

jsoo1:
> Using functors sounds like a pretty weird usecase but looks good. 😄

I don't understand it either, but I wanted to address the issue.

lilyball:
My use-case for functors here was being able to define a nixpkgs overlay that I could then filter out when cross-compiling nixpkgs for Rosetta. Equality comparisons don't work on functions, but they do work for sets and therefore functors. I ultimately had to modify the overlay itself to change its behavior based on `super.stdenv.hostPlatform.system`, but I imagine there may be other times in which functors would be helpful.

Issue #392, opened 2021-12-07
Evaluation failure using flakes and using nixpkgs release-20.03


jsoo1:
If I use the 20.03 release, there is an evaluation error when building this system:
```nix
{
  inputs = {
    nixpkgs.url = "git+https://github.com/nixos/nixpkgs?ref=release-20.03";
    darwin.url = "github:LnL7/nix-darwin";
  };
  outputs = { self, darwin, nixpkgs, ... }: {
    darwinConfigurations.my-hostname = let system = "x86_64-darwin"; in
      darwin.lib.darwinSystem {
        inherit system;
        modules = [{ nixpkgs = import nixpkgs { inherit system; }; }];
      };
  };
}
```

```json
{
  "nodes": {
    "darwin": {
      "inputs": {
        "nixpkgs": "nixpkgs"
      },
      "locked": {
        "lastModified": 1634994402,
        "narHash": "sha256-xmlCVVOYGpZoxgOqsDOVF0B0ASrnbNGVAEzID9qh2xo=",
        "owner": "LnL7",
        "repo": "nix-darwin",
        "rev": "44da835ac40dab5fd231298b59d83487382d2fab",
        "type": "github"
      },
      "original": {
        "owner": "LnL7",
        "repo": "nix-darwin",
        "type": "github"
      }
    },
    "nixpkgs": {
      "locked": {
        "lastModified": 1602411953,
        "narHash": "sha256-gbupmxRpoQZqL5NBQCJN2GI5G7XDEHHHYKhVwEj5+Ps=",
        "owner": "LnL7",
        "repo": "nixpkgs",
        "rev": "f780534ea2d0c12e62607ff254b6b45f46653f7a",
        "type": "github"
      },
      "original": {
        "id": "nixpkgs",
        "type": "indirect"
      }
    },
    "nixpkgs_2": {
      "locked": {
        "lastModified": 1625876338,
        "narHash": "sha256-pys6cebTmrZ7FHT4JUAFl4wk2IBff9aDE3hD6BcPYho=",
        "ref": "release-20.03",
        "rev": "eb73405ecceb1dc505b7cbbd234f8f94165e2696",
        "revCount": 216226,
        "type": "git",
        "url": "https://github.com/nixos/nixpkgs"
      },
      "original": {
        "ref": "release-20.03",
        "type": "git",
        "url": "https://github.com/nixos/nixpkgs"
      }
    },
    "root": {
      "inputs": {
        "darwin": "darwin",
[...]

PR #397, opened 2021-12-08
eval-config: Add check parameter for compatibility with nixpkgs.

jsoo1:
This allows more passthrough parameters in compatibility with nixpkgs. (See #392)


LnL7 (approved):
Could you rebase so the tests are green?

PR #402, opened 2021-12-13
fix dependence on apfs.util exit code

### LINKED ISSUE #378: error: Directory /run does not exist, aborting activation Create a symlink to /var/run with:
```
shauryas-mbp% ./result/sw/bin/darwin-rebuild switch --flake .#shaunsingh-laptop
Password:
error: Directory /run does not exist, aborting activation
error: Directory /run does not exist, aborting activation
Create a symlink to /var/run with:

$ echo 'run\tprivate/var/run' | sudo tee -a /etc/synthetic.conf
$ /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -B
$ /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -t

The current contents of /etc/synthetic.conf is:

    run\tprivate/var/run
    nix

shauryas-mbp% 
```

Upon running `echo 'run\tprivate/var/run' | sudo tee -a /etc/synthetic.conf` then `/System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -t`, then `./result/sw/bin/darwin-rebuild switch --flake .#shaunsingh-laptop` and the error is still there

### LINKED ISSUE #401: apfs.util step failing on Big Sur+?
I tried a clean install of my macOS bootstrap the other day and noticed a "warning: failed to execute apfs.util" in the output. I think this has probably been failing since whenever 8834df431736c3037bbef017b24c62bc4ace4645 was released, but I'm a little surprised not to see any other reports of it. 

(though #378 and similar may be a byproduct of this failing at install time and not getting noticed?)

The return codes from `apfs.util -t` aren't good for normal shell semantics. I tested this a little on the machine that had the failed install and see things like:

```console
$ /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -t; echo $?
failed to stitch firmlinks and/or create synthetics for root volume (c002) ...
252
$ sudo mv /etc/synthetic.conf.bak /etc/synthetic.conf
Password:
$ /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -t; echo $?
253
```

abathur:
apfs.util (or at least it's newer -t flag) has bad exit code semantics
and will return non-zero codes for both success and errors.

Closes #401 and most-likely fixes #378 as well

### DIFF
--- a/pkgs/darwin-installer/configuration.nix
+++ b/pkgs/darwin-installer/configuration.nix
@@ -43,15 +43,20 @@ with lib;
       case "$i" in
           y|Y)
               if ! grep -q '^run\b' /etc/synthetic.conf 2>/dev/null; then
-                  echo "setting up /etc/synthetic.conf..."
+                  echo "setting up /run via /etc/synthetic.conf..."
                   echo -e "run\tprivate/var/run" | sudo tee -a /etc/synthetic.conf >/dev/null
-                  /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -B &>/dev/null \
-                      || /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -t &>/dev/null \
-                      || echo "warning: failed to execute apfs.util"
+                  /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -B &>/dev/null || true
+                  /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -t &>/dev/null || true
+                  if ! test -L /run; then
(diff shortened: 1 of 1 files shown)


domenkozar:
cc @LnL7

abathur:
@LnL7 I saw some notification emails from failed test runs and it looks like you fixed up some actions issues a bit ago, so I rebased against master and force-pushed.

LnL7 (approved):
Perfect, I'm a bit confused why the checks don't seem to be running against the merge commit. I would expect rebasing to not be required.

Issue #403, opened 2021-12-17
flakes: nixChannels checks are run when using flakes


jsoo1:
If you install using flakes only and want to switch to a new configuration, system checks will fail.

Enzime:
As of #1026, if you want to disable channels, you can use:

```
nix.channel.enable = false;
```

PR #404, opened 2021-12-17
flakes: Do not verify channels when using flakes.

jsoo1:
#403

### DIFF
--- a/modules/system/checks.nix
+++ b/modules/system/checks.nix
@@ -205,6 +205,12 @@ in
       description = "Whether to run the NIX_PATH validation checks.";
     };
 
+    system.checks.verifyNixChannels = mkOption {
+      type = types.bool;
+      default = true;
+      description = "Whether to run the nix-channels validation checks.";
+    };
+
     system.checks.text = mkOption {
       internal = true;
       type = types.lines;
@@ -222,7 +228,7 @@ in
       (mkIf (!config.nix.useDaemon) singleUser)
       nixStore
       (mkIf (config.nix.gc.automatic && config.nix.gc.user == null) nixGarbageCollector)
-      nixChannels
+      (mkIf cfg.verifyNixChannels nixChannels)
       nixInstaller
       (mkIf cfg.verifyNixPath nixPath)
     ];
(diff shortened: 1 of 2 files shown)


LnL7 (dismissed):
I think this check is still relevant regardless of how you build your system, this check only fires if the channels both exist and point to an unexpected location.

The goal here was to attempt to notice when someone has used `sudo nix-channel --update` which (if not still the case) could mess up your local channel references.

jsoo1:
It's been a while and I had to back track from using flakes. But if memory serves, if the system has no channels at all, the check will fail and just not activate the system at all.

jsoo1:
In any case, when using flakes, nix-darwin won't even refer to a channel, correct?

socksy:
After a full night's sleep I solved it! :)

I was doing 
```bash
sudo rm -r ~/.nix-channels
nix-channel --update
```

which would error out when I did `darwin-rebuild switch --flake ".#"` with a `symlink does not point your users channels, aborting activation`.

I solved it last night by just doing `sudo darwin-rebuild switch --flake ".#"`, which was not ideal. Now I looked closer, and saw that the second time around I was actually getting the error that `~/.nix-defexpr/channels` was pointing at the wrong thing. The solution simply was to 

```bash
sudo rm -r  ~/.nix-defexpr/channels
nix-channel --update
```

and now the darwin-rebuild works again

yangm97:
LGTM. Can we get this merged?

Enzime:
@domenkozar just ran into this issue, any chance you can take a look at this PR?

domenkozar:
@jsoo1 could you rebase?

jsoo1:
Sure. It's been a bit since I thought about this one. One sec...

Enzime (approved):
LGTM

emilazy:
Note that https://github.com/LnL7/nix-darwin/pull/723 ended up effectively reverting this, due to https://github.com/LnL7/nix-darwin/pull/723#discussion_r1257442026. Hopefully if people start having this problem again we can figure out what's causing it or else re-disable the check.

PR #409, opened 2021-12-20
eval-config.nix: Fix args warning

### LINKED ISSUE #405: `evalModules` now warns when passing `args`
Nixpkgs now warns when using the `args` (or `check`) parameter to `lib.evalModules` (see https://github.com/NixOS/nixpkgs/pull/148315). nix-darwin is using `args` in `eval-config.nix` and in `modules/documentation/default.nix`. It should avoid this and pass args using `_module.args` instead. The various commits in the aforementioned PR can probably be used as a template for the changes in nix-darwin.

roberth:
Resolves
```
trace: warning: The args argument to evalModules is deprecated. Please set config._module.args instead.
```

`evalModules { args }` and `config._module.args` are equivalent and have been available for years.

Fixes #405

### DIFF
--- a/eval-config.nix
+++ b/eval-config.nix
@@ -8,10 +8,12 @@
 }@args:
 
 let
-  inputsModule = {
+  argsModule = {
     _file = ./eval-config.nix;
     config = {
-      _module.args.inputs = inputs;
+      _module.args = {
+        inherit baseModules inputs modules;
+      };
     };
   };
 
@@ -33,8 +35,7 @@ let
   });
 
   eval = libExtended.evalModules (builtins.removeAttrs args [ "inputs" "system" ] // {
-    modules = modules ++ [ inputsModule pkgsModule ] ++ baseModules;
-    args = { inherit baseModules modules; };
+    modules = modules ++ [ argsModule pkgsModule ] ++ baseModules;
     specialArgs = { modulesPath = builtins.toString ./modules; } // specialArgs;
   });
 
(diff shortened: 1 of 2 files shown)


jsoo1:
I think this fixes #405.

How can I make #392 work with this (nixpkgs v 20.xx)? (Sorry, I know very little about the evalModules machinery).

jacobfoard:
I applied this to my local config and it did remove one of the warning lines, but looks like the second one is coming from https://github.com/LnL7/nix-darwin/blob/master/modules/documentation/default.nix#L21 as pointed out in #405. 

I don't know the nix internals well enough to be able to offer a patch that works for it, but 🤞 it looks like the same amount of effort.

lilyball:
While this code is being touched, what is the `builtins.removeAttrs args [ "inputs" "system" ]` bit about? Surely if someone actually passes `baseModules` to `evalConfig` that will produce an error as `evalModules` doesn't take that arg? And the other attributes in `args` are either removed or already being overridden.

lilyball:
Also it might be nice to toss in `check` support. https://github.com/LnL7/nix-darwin/pull/397 adds the arg declaration, but passing that through (with the aforementioned `builtins.removeAttrs args [ "inputs" "system" ]`) will produce a warning, but if you remove that and stick the value in the `argsModule` then that should work.

LnL7 (dismissed):
Thanks! Based on the comment I assume will still work for the previous releases.
Could you rebase so the tests pass?

roberth:
Yes, this is backwards compatible.

PR #414, opened 2021-12-24
defaults: add options for common defaults in the `com.apple.universalaccess` domain

montchr:
See https://github.com/montchr/dotfield/commit/66b90fc02417bc9a719d354f1c6aad0f1bc0eee6 for successful implementation.

## Added

- Add an option to disable transparency in system UI (especially noticeable in the menu bar).
	- **source:** https://github.com/mathiasbynens/dotfiles/blob/66ba9b3cc0ca1b29f04b8e39f84e5b034fdb24b6/.macos#L28-L29
	- **personal note:** I recently updated to Monterey and found that my menu bar appeared to be light even when I had dark mode enabled. Turns out that turning this setting on "fixes" that behavior.
- Add an option to enable screen zoom while holding the ctrl key and scrolling. Also added the related option for controlling whether the zoomed area will follow the cursor.
	- **source:** https://github.com/mathiasbynens/dotfiles/blob/66ba9b3cc0ca1b29f04b8e39f84e5b034fdb24b6/.macos#L148-L152

## Changed

- Apply formatting to `modules/system/defaults-write.nix` via `nixpkgs-fmt`. The formatting seemed necessary since one of the lists for `mkIfAttrs` was getting quite long and I had to add another value to it.
- Sort some variables alphabetically in `modules/system/defaults-write.nix`.

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -21,6 +21,7 @@
   ./system/defaults/SoftwareUpdate.nix
   ./system/defaults/spaces.nix
   ./system/defaults/trackpad.nix
+  ./system/defaults/universalaccess.nix
   ./system/etc.nix
   ./system/keyboard.nix
   ./system/launchd.nix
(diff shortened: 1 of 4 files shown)


LnL7 (approved):
Thanks!

PR #421, opened 2022-01-11
move installer into a separate file

domenkozar:
This allows using the installer script independently of `<user-darwin-config>` that isn't working with flakes.

cc @LnL7

### DIFF
--- a/pkgs/darwin-installer/configuration.nix
+++ b/pkgs/darwin-installer/configuration.nix
@@ -3,63 +3,8 @@
 with lib;
 
 {
-  imports = [ <user-darwin-config> ];
+  imports = [ <user-darwin-config> ./installer.nix ];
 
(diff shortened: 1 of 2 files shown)


LnL7 (approved):
What's your idea for the next steps? The regular installer kind of does 2 things currently:

1. Generate your initial configuration, for flakes perhaps some kind of flake project init would make sense?
2. Run a few extra setup steps, there's not a whole lot that has such overhead it couldn't be included in regular activation. The thing here are files in `/etc` if instead of the content hashes that this requires I'd also be fine with a whitelist/blacklist of known safe files to touch.

domenkozar:
I think the initial configuration should be a thing of the documentation and separate from the installer.

Extra steps could just be part of the regular activation to make things easier!

PR #426, opened 2022-01-18
Fix homebrew.brewPrefix default value on Apple Silicon

malob:
Turns out that #418, has a small bug:`pkgs.stdenv.hostPlatform.darwinArch` returns `"arm64"` on `aarch64-darwin` systems not `"aarch64"`.

Instead of changing to code to check whether the value is `"arm64"`, I've changed the predicate to use `pkgs.stdenv.hostPlatform.isAarch64`.

I've tested on my own system and this works properly for both Intel and Apple Silicon Macs now.

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -54,7 +54,7 @@ in
 
     brewPrefix = mkOption {
       type = types.str;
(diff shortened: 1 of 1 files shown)


malob:
@LnL7, would be cool to get this merged quickly. (Sorry for not catching this when I review the recent PR adding this functionality.)

LnL7 (approved):

PR #429, opened 2022-01-23
Add top-level `lib` option based on NixOS

montchr:
I'm working on https://github.com/divnix/digga/pull/438. The project expects a top level `lib` option in system configuration, which is declared in NixOS, but not in nix-darwin.

https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/misc/lib.nix

Description from the option declaration:

> This option allows modules to define helper functions, constants, etc.

### DIFF
--- a/modules/misc/lib.nix
+++ b/modules/misc/lib.nix
@@ -1,0 +1,15 @@
+{ lib, ... }:
+
+{
+  options = {
+    lib = lib.mkOption {
+      default = { };
+
+      type = lib.types.attrsOf lib.types.attrs;
+
+      description = ''
+        This option allows modules to define helper functions, constants, etc.
+      '';
+    };
+  };
+}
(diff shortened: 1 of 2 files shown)


montchr:
Related general question: when I open a pull request on nix-darwin adding a new option or changing something that should require a documentation update, is there any command I need to run locally to regenerate the docs, or is that handled automatically in CI?

montchr:
https://github.com/divnix/digga/pull/438 is just about done (all checks passed for the first time). Is there anything I can do to help move this PR along?

LnL7 (approved):

LnL7:
> Related general question: when I open a pull request on nix-darwin adding a new option or changing something that should require a documentation update, is there any command I need to run locally to regenerate the docs, or is that handled automatically in CI?

The local manpages are always rebuilt, the online documentation used to be manual and thus was quite out of date. However someone recently setup automation for this https://github.com/LnL7/nix-darwin/pull/449.

PR #438, opened 2022-02-16
Add Cachix Agent 

domenkozar:


modules/services/cachix-agent.nix
```diff
@@ -0,0 +1,76 @@
...
+
+    logFile = mkOption {
+      type = types.nullOr types.path;
+      default = "/var/log/cachix-agent.log";
+      description = "Absolute path to log all stderr and stdout";
+    };
+  };
+
+  config = mkIf cfg.enable {
+    launchd.daemons.cachix-agent = {
```
LnL7, line 52: launchd daemons run as root by default, that's intended for this service?

LnL7 (approved):

modules/services/cachix-agent.nix
domenkozar: Yeah, it has to run as root.

Issue #367, opened 2021-10-20
Replace `literalExample` with `literalExpression`/`literalDocBook`


lilyball:
Current nixpkgs-unstable now issues warnings on `literalExample`. https://github.com/NixOS/nixpkgs/pull/136909 introduced this along with the replacements `literalExpression` and `literalDocBook`. This means that every `darwin-rebuild build` now warns me about the use of `literalExample`, which I can't fix myself as it's in nix-darwin.

home-manager switched over to `literalExpression`/`literalDocBook` already and added [a compatibility shim](https://github.com/nix-community/home-manager/pull/2371/commits/468c4611396f6ca3be5eb54f1c864e34cc19dfe2). nix-darwin could adopt the same approach.

jab:
I just installed latest nix-darwin in a fresh environment and have been getting lots of
```console
warning: literalExample is deprecated, use literalExpression instead, or use literalDocBook for a non-Nix description
```
Googling this warning along with "nix-darwin" gave this issue as the top result. But this was closed almost a year ago.

Should this be reopened, or is this tracked elsewhere? And is there something users can do in the meantime to silence this (alarmingly-red) warning which seems to appear on every `darwin-rebuild switch` (making it harder for users to notice warnings that they actually can and should address in their own configs)?

Thanks for maintaining nix-darwin!

lilyball:
@jab That sounds like https://github.com/LnL7/nix-darwin/issues/504, which has been fixed.

jab:
No longer seeing this after `nix-channel --update && darwin-rebuild switch`, thanks @lilyball for letting me know it was fixed!

PR #439, opened 2022-02-21
Use literalExpression instead of literalExample in cachix-agent

xuanduc987:
Because nixpkgs warns `literalExample has been deprecated, please use literalExpression instead`

## Related
- https://github.com/LnL7/nix-darwin/pull/438
- https://github.com/LnL7/nix-darwin/issues/367


domenkozar:
Fixed in master.

PR #447, opened 2022-03-14
add options for configuring activity monitor

dsyang:
This PR enables nix-darwin to configure some activity monitor functionality. These options were part of my osx-setup script that I'm porting to nix and mostly taken from https://github.com/joeyhoer/starter/blob/master/apps/activity-monitor.sh.  

Specifically these options enable:
- Configuring which processes to show in the main CPU view
- Configuring the dock icon.
- Configuring the sort column and order of the main window.


LnL7 (approved):
Thanks!

PR #448, opened 2022-03-16
fix: move fpath autocomplete setup from zshrc to zshenv

virchau13:
This moves the code for setting up `fpath` for zsh's autocomplete from /etc/zshrc to /etc/zshenv, which keeps compatibility with NixOS.

I got very confused why my setup had autocomplete on NixOS but not on Darwin despite sharing most of the same Nix code. It turns out that I had `setopt no_global_rcs` in my `~/.zshenv` because I didn't want the default NixOS prompt, and in this prevented autocomplete in nix-darwin due to the fact that the autocomplete-fpath code was in /etc/zshrc, not /etc/zshenv. I think it makes more sense to put the autocomplete-fpath code in /etc/zshenv, since it's expected of practically all shell sessions, and needs to be manually enabled with `compinit` anyway.

### DIFF
--- a/modules/programs/zsh/default.nix
+++ b/modules/programs/zsh/default.nix
@@ -121,6 +121,11 @@ in
         . ${config.system.build.setEnvironment}
       fi
 
+      # Tell zsh how to find installed completions
+      for p in ''${(z)NIX_PROFILES}; do
+        fpath+=($p/share/zsh/site-functions $p/share/zsh/$ZSH_VERSION/functions $p/share/zsh/vendor-completions)
+      done
+
       ${cfg.shellInit}
 
       # Read system-wide modifications.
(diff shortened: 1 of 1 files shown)


LnL7 (approved):
From what I remember about the rc files `zshenv` is loaded everywhere so strictly speaking things related to completion and the prompt don't belong there. But if this is what happens on NixOS I'm fine with moving it.

FYI The zsh module on both platforms should have a specific option for the prompt configuration which can be overridden.

montchr:
> From what I remember about the rc files `zshenv` is loaded everywhere so strictly speaking things related to completion and the prompt don't belong there. But if this is what happens on NixOS I'm fine with moving it.

This struck me as odd too (though I'm not an expert). But it is indeed the same on NixOS:

https://github.com/NixOS/nixpkgs/blob/2627bb20f913f1b3a982758383f7dd3192c7f47f/nixos/modules/programs/zsh/zsh.nix#L185-L188

virchau13:
> From what I remember about the rc files `zshenv` is loaded everywhere so strictly speaking things related to completion and the prompt don't belong there.

I mean, I guess, but I don't think I've ever seen anyone actually use a zsh script for anything other than interactive purposes.

> FYI The zsh module on both platforms should have a specific option for the prompt configuration which can be overridden.

Yeah, but I like to keep my own ~/.zshrc, lol.

Enzime:
This change has already been merged in #1066

PR #452, opened 2022-03-31
add mouse scaling system setting

psm14:
Adds support for the `com.apple.mouse.scaling` GlobalPreferences setting. This is [useful for disabling mouse acceleration](https://devhints.io/macos-mouse-acceleration).


domenkozar:
The option type should really be part of nixpkgs repo.

domenkozar (approved):

PR #453, opened 2022-04-15
Fix synthetic.conf instructions

### LINKED ISSUE #451: error: Directory /run does not exist, aborting activation
I am sorry if I am doing something wrong, but I am getting the below error when using with flakes.
BigSur 11.6.4
While I did not try the below presented instructions, I was wondering if this should happen in the first place.
Similar issue : https://github.com/LnL7/nix-darwin/issues/378
image
Followup Question: Why do we need to symlink (synthetic) `/private/var/run` to `/run`, instead of directly using `/private/var/run` ?

mtoohey31:
Closes #451? Waiting to hear back from the OP there, but I think this should resolve it; it did for me.

This PR adds comments to the two `apfs.util` commands to explain on which versions of MacOS each should be used. This should clear up some confusion. Also, I switched the command for adding the necessary line to synthetic.conf to use printf instead because on any shell other than zsh, echo is not a builtin, and `echo '...\t...'` doesn't expand the `\t`, so we end up with a literal `\t` in synthetic.conf.

### DIFF
--- a/modules/system/checks.nix
+++ b/modules/system/checks.nix
@@ -27,9 +27,9 @@ let
         echo "Create a symlink to /var/run with:" >&2
         if test -e /etc/synthetic.conf; then
             echo >&2
-            echo "$ echo 'run\tprivate/var/run' | sudo tee -a /etc/synthetic.conf" >&2
-            echo "$ /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -B" >&2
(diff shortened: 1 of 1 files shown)


LnL7 (approved):
Looks like this message was indeed update incorrectly.
https://github.com/LnL7/nix-darwin/blob/16c9343b250c8a2c6186de702a034e05b6d0d248/pkgs/darwin-installer/installer.nix#L42

PR #210, opened 2020-08-02
feat: keyboard module improvements

eqyiel:
Hi @LnL7, thanks for your work on `nix-darwin`!

I wrote this so that I could do more flexible key bindings without resorting to Karabiner (which requries a kernel extension, which is a whole different kind of problem https://github.com/LnL7/nix-darwin/pull/91).

I'm using this module with this configuration:

```system.keyboard.enableKeyMapping = true;
system.keyboard.mappings = [
  {
    productId = 273;
    vendorId = 2131;
    mappings = {
      # Bind Caps Lock to Left Function for Realforce 87u.
      "Keyboard Caps Lock" = "Keyboard Left Function (fn)";
    };
  }

  {
    productId = 638;
    vendorId = 1452;
    mappings = {
      # For the built-in MacBook keyboard, change the modifiers to match a
      # traditional keyboard layout.
      "Keyboard Caps Lock" = "Keyboard Left Function (fn)";
      "Keyboard Left Alt" = "Keyboard Left GUI";
      "Keyboard Left Function (fn)" = "Keyboard Left Control";
      "Keyboard Left GUI" = "Keyboard Left Alt";
      "Keyboard Right Alt" = "Keyboard Right Control";
      "Keyboard Right GUI" = "Keyboard Right Alt";
    };
  }
];
```

Despite my best efforts there is a change in behaviour, and I changed it because I think it was wrong.

The `HIDKeyboardModifierMappingSrc` attribute set by `system.keyboard.nonUS.remapTilde` [was `30064771172`](https://github.com/LnL7/nix-darwin/blob/47fe6db9c9cd22c4ba57df3281f98f454880bea7/modules/system/keyboard.nix#L54), which is `Keyboard Non-US \ and |`, not `Keyboard Non-US # and ~`.

I changed it to reflect the name (it now remaps `"Keyboard Non-US # and ~" → "Keyboard Grave Accent and Tilde"` rather than `"Keyboard Non-US \ and |" → "Keyboard Grave Accent and Tilde"`).

I'm pretty sure that apart from that, there are no breaking changes.  If the old behaviour is preferred,  it's possible to get it back like this:

```
system.keyboard.enableKeyMapping = true;
system.keyboard.mappings = {
  "Keyboard Non-US \ and |" = "Keyboard Grave Accent and Tilde";
};
```

[...]

PR #455, opened 2022-04-17
feat: add serviceConfig.LaunchEvents configuration option

eqyiel:
This is required if you want to register an "xpc event stream handler". Check out the README for https://github.com/snosrap/xpc_set_event_stream_handler for more info.

I also did this in https://github.com/LnL7/nix-darwin/pull/210 before giving up and making it a local module. This is the only part that has to be upstreamed and I don't want to keep rebasing 😛 so if it's OK, please let me add this!

### DIFF
--- a/modules/launchd/launchd.nix
+++ b/modules/launchd/launchd.nix
@@ -708,6 +708,31 @@ with lib;
       });
     };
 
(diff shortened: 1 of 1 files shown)


LnL7 (approved):
Thanks!

PR #463, opened 2022-05-08
defaults: add options for dragOnGesture feature

2gn:
Enable ( or disable ) NSWindowShouldDragOnGesture option without executing the following command:
``` shell
defaults write -g NSWindowShouldDragOnGesture -bool true   
```
for more information about this feature:
https://mmazzarolo.com/blog/2022-04-16-drag-window-by-clicking-anywhere-on-macos/


domenkozar:
CI is failing

2gn:
I don't know why the error happens. In my environment, it's working so far. And as far as I can see, the diff of the commit looks fine.

modules/system/defaults/NSGlobalDomain.nix
```diff
@@ -216,6 +216,14 @@ in {
       '';
     };
 
+    system.defaults.NSGlobalDomain.NSWindowShouldDragOnGesture = mkOption {
+      type = types.nullOr types.bool;
+      default = null;
+      description = ''
+      	Whether to enable moving window by holding anywhere on it like on Linux. The default is null.
```
malob, line 223:
```suggestion
      	Whether to enable moving window by holding anywhere on it like on Linux. The default is false.
```

When listing the default value for `defaults` options in `nix-darwin` the convention is to specify the default value of the key in macOS rather than the default value of the `nix-darwin` option (since it’s always `null`).

2gn:
why was it cancelled...?

2gn:
oops, I think I made a mistake.

modules/system/defaults/NSGlobalDomain.nix
2gn: > When listing the default value for `defaults` options in `nix-darwin` the convention is to specify the default value of the key in macOS rather than the default value of the `nix-darwin` option (since it’s always `null`).


I noticed this message now, sorry. And I didn't know that I should write `false` instaed of `null`. Thank you.

2gn:
I did something wrong again
I'm sorry if I pinged someone.

lockejan:
@2gn can you please add some basic tests. There are already some for other options of this module.

2gn:
I'm not sure what is happening here but I'm sure that I've ruined the git tree, and I wonder if I could delete this pull request and open another. May I?

lockejan:
Sure. You can create a new branch from the current master and cherry-pick your commits from this one.

2gn:
> Sure. You can create a new branch from the current master and cherry-pick your commits from this one.
Ok, thanks! I'll do it.

PR #464, opened 2022-05-09
Add NSWindowShouldDragOnGesture defaults option

purcell:


purcell:
Lol, duplicate of #463, should have looked first!

Issue #356, opened 2021-08-22
system.extraDependencies


antifuchs:
This nixos attribute is really useful for warming the cache in a flake-based system configuration that has `apps` defined: By adding the apps' derivations to the store, you get them pre-built (and referenced in the system closure so they don't get gc'ed) without having them clutter up the $PATH.

Here's a super minimal definition that I have in my system:

```nix
{ lib, config, ... }:
let
  cfg = config.nixos-compat;
in
{
  options = with lib; {
    nixos-compat.system.extraDependencies = mkOption {
      description = "Extra derivations to build & include in the
        system closure that are otherwise unreferenced.";
      type = types.listOf types.path;
      default = [ ];
    };
  };

  config = {
    environment.etc."nix/extra-dependencies.list" = {
      text = (toString cfg.system.extraDependencies);
    };
  };
}
```

This lets me add the following module to any darwin mods in my flake definition:

```nix
      prebuilt-flake-apps = system: { lib, ... }: {
        _file = "~/home#prebuilt-flake-apps";
        nixos-compat.system.extraDependencies =
          lib.mapAttrsToList (_n: app: app.program)
            (lib.filterAttrs (_n: app: app.type == "app") self.apps.${system});
      };
```

PR #468, opened 2022-06-06
services/postgresql: update to the latest upstream nixos module

lf-:
I want a newish postgres config to work on my Mac.

This contains a little hack (putting the check-config derivation into
the start script as a comment) to deal with not having #356. Also,
it does not implement the stuff that requires a postStart script.

It's otherwise identical to the upstream module.

I changed the service to use utf8 always because the charset option got deleted in the upstream module. It seems to work and I can connect to it.

### DIFF
--- a/modules/services/postgresql/default.nix
+++ b/modules/services/postgresql/default.nix
@@ -3,69 +3,198 @@
 with lib;
 
 let
+
   cfg = config.services.postgresql;
 
-  postgresqlAndPlugins = pg:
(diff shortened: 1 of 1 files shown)


domenkozar (approved):

lf-:
Yes I used it every day with this module so it definitely works. no recollection as to how the auth was set up, it was done 6 months ago, and i don't have the computer anymore.

PR #226, opened 2020-09-05
Disable taking control of ~/Applications folder

berbiche:
Programs like Steam add applications to ~/Applications and such.

This PR disables linking `~/Applications` to nix-darwin applications in the `/nix/store` and makes nix-darwin use a subfolder within `~/Applications`.

See https://github.com/rycee/home-manager/issues/1341#issuecomment-687286866

PR #470, opened 2022-06-26
Disable taking control of ~/Applications folder MkII

toonn:
This PR supercedes #226.

So far application bundles were always linked to `~/Applications` or
`~/Applications/Nix Apps` if the former was an existing directory.

In nix-community/home-manager#1341 a conflict was found with suggested
new Home Manager behavior, where applications installed through Home
Manager would end up in `~/Applications/Home Manager Apps`. This was in
an attempt to make them discoverable through Spotlight but further
investigation suggest Spotlight does not pick up symlinked apps (details
in the issue).

However, there are other programs that expect to be able to write to
`~/Applications` so taking over the directory is unfortunate.

PR #226 dropped linking `~/Applications` and instead made sure the
directory exists so we can always link in `~/Applications/Nix Apps`.

After further discussion in #macos:nixos.org we came to the conclusion
that we shouldn't link applications to a user directory at all. Since we
manage packages for multiple users, application bundles should go in
`/Applications`.

Because previous code will likely leave a symlink at
`~/Applications/{,Nix Apps}`, which will become dangling once the path
it links to is garbage collected from the store we test to see if a link
exists and it conforms to the path we're expecting and if it does remove
it.

**Note**: This is a draft because I haven't had time to test it.

### DIFF
--- a/modules/system/applications.nix
+++ b/modules/system/applications.nix
@@ -24,12 +24,12 @@ in
       # Set up applications.
       echo "setting up ~/Applications..." >&2
 
-      if [ ! -e ~/Applications -o -L ~/Applications ]; then
-        ln -sfn ${cfg.build.applications}/Applications ~/Applications
-      elif [ ! -e ~/Applications/Nix\ Apps -o -L ~/Applications/Nix\ Apps ]; then
+      mkdir -p ~/Applications
+
+      if [ ! -e ~/Applications/Nix\ Apps -o -L ~/Applications/Nix\ Apps ]; then
         ln -sfn ${cfg.build.applications}/Applications ~/Applications/Nix\ Apps
       else
-        echo "warning: ~/Applications and ~/Applications/Nix Apps are directories, skipping App linking..." >&2
+        echo "warning: ~/Applications/Nix Apps is not owned by nix-darwin, skipping App linking..." >&2
       fi
     '';
 
--- a/modules/system/applications.nix
+++ b/modules/system/applications.nix
@@ -24,20 +24,21 @@ in
       # Set up applications.
       echo "setting up /Applications/Nix Apps..." >&2
 
+      ourLink () {
+        local link
+        link=$(readlink "$1")
+        [ -L "$1" ] && [ "''${link#*-}" = 'system-applications/Applications' ]
+      }
+
       # Clean up for links created at the old location in HOME
-      if [ -L ~/Applications
-           -a $(readlink ~/Applications | grep --quiet
-                 '/nix/store/.*-system-applications/Applications')
-         ]
+      if ourLink ~/Applications; then
         rm ~/Applications
-      elif [ -L '~/Applications/Nix Apps'
-             -a $(readlink '~/Applications/Nix Apps' | grep --quiet
-                   '/nix/store/.*-system-applications/Applications')
-           ]
-        rm '~/Applications/Nix Apps'
+      elif ourLink ~/Applications/'Nix Apps'; then
+        rm ~/Applications/'Nix Apps'
       fi
 
-      if [ ! -e '/Applications/Nix Apps' -o -L '/Applications/Nix Apps' ]; then
+      if [ ! -e '/Applications/Nix Apps' ] \
+         || ourLink '/Applications/Nix Apps'; then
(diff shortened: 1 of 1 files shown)


modules/system/applications.nix
```diff
@@ -22,14 +22,26 @@ in
...
+
+      ourLink () {
+        local link
+        link=$(readlink "$1")
+        [ -L "$1" ] && [ "${link#*-}" = 'system-applications/Applications' ]
+      }
+
+      # Clean up for links created at the old location in HOME
+      if ourLink ~/Applications; then
+        rm ~/Applications
```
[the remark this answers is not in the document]
toonn: Only if it's a link left over from the previous setup, @ggPeti.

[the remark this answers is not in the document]
toonn: That's what the `ourLink` function's supposed to check for.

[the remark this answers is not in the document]
toonn: It's actually explained in the OP and the commit message:

> Because previous code will likely leave a symlink at
~/Applications/{,Nix Apps}, which will become dangling once the path
it links to is garbage collected from the store we test to see if a link
exists and it conforms to the path we're expecting and if it does remove
it.

modules/system/applications.nix
```diff
@@ -22,14 +22,26 @@ in
...
+
+      ourLink () {
+        local link
+        link=$(readlink "$1")
+        [ -L "$1" ] && [ "''${link#*-}" = 'system-applications/Applications' ]
+      }
+
+      # Clean up for links created at the old location in HOME
+      if ourLink ~/Applications; then
+        rm ~/Applications
```
supermarin, line 35: Think you're missing a `mkdir ~/Applications` with the current user as the owner? i.e. after deleting the symlink there'll be no parent ~/Applications folder to symlink into. And we don't want to have a root owned one either

toonn: No, it's not up to nix-darwin to create `~/Applications`. It's a directory that the user either creates for themselves or sometimes a program they run does it for them.
We don't need the directory anyway if we use `/Applications`. And since nix-darwin manages the system and all the users, it's unexpected behavior to create `~/Applications` just for the user running `darwin-rebuild`.

supermarin: Oh sorry, I've missed the fact you're using /Applications vs ~/Applications in this patch. That sounds good.

supermarin (dismissed):
PR looks good to me, haven't tested locally

supermarin:
@toonn can this be converted out of a Draft?

toonn:
I still haven't had time to test it. Until someone does I think this should stay a draft.

bbenne10:
I just had a chance to check this out (after discussing with @toonn in #macos:nixos.org and can confirm that this seems to be working as intended. `/Applications/Nix Apps` is created with the installed app bundle symlinked into it. `~/Applications` was in my case removed entirely.

modules/system/applications.nix
```diff
@@ -22,14 +22,26 @@ in
 
     system.activationScripts.applications.text = ''
       # Set up applications.
-      echo "setting up ~/Applications..." >&2
+      echo "setting up /Applications/Nix Apps..." >&2
+
+      ourLink () {
+        local link
+        link=$(readlink "$1")
+        [ -L "$1" ] && [ "''${link#*-}" = 'system-applications/Applications' ]
```
LnL7, line 30: Not sure this is the best condition, but not sure either what would be given that a link to the store isn't enough. Given that the name now has significance a comment in the corresponding expression would be a good idea.

LnL7 (approved):
Haven't validated this either but the logic looks good to me.

modules/system/applications.nix
toonn: Which name do you mean, the suffix `system-applications/Applications`, in case someone happens to create a link at `~/Applications` for a (store) path ending with that suffix or something?

supermarin:
@LnL7 can we merge? @bbenne10 confirmed it's working

firestack (approved):
I've pulled this onto my machine when looking for a fix and can confirm it's working 👍🏻

Enzime:
@domenkozar any chance you can take a look at this PR?

PR #487, opened 2022-08-11
Link /Applications to ~/Applications/Nix\ Darwin

supermarin:
Step 1: try to resolve conflicts between nix-darwin and home-manager regarding linking in ~/Applications.
Currently, [home-manager has disabled linking to ~/Applications](https://github.com/nix-community/home-manager/blob/db00b39a9abec04245486a01b236b8d9734c9ad0/modules/targets/darwin/linkapps.nix) because nix-darwin is behaving inconsistently and sometimes creating a race:
- If there's no ~/Applications folder on the system, nix-darwin will symlink it's apps directly to ~/Applications. This is bad because home-manager can't symlink under ~/Applications/Home Manager Apps
- If there is ~/Applications folder already on the system, nix-darwin will symlink into ~/Applications/Nix Apps.

This PR ensures that there's an ~/Applications folder, and always uses ~/Applications/Nix Apps for symlinks.

Related issue: https://github.com/nix-community/home-manager/issues/1341

### DIFF
--- a/modules/system/applications.nix
+++ b/modules/system/applications.nix
@@ -23,14 +23,15 @@ in
     system.activationScripts.applications.text = ''
       # Set up applications.
       echo "setting up ~/Applications..." >&2
+      mkdir -p ~/Applications
 
-      if [ ! -e ~/Applications -o -L ~/Applications ]; then
-        ln -sfn ${cfg.build.applications}/Applications ~/Applications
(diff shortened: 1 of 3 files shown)


toonn:
I opened #470 for this, takes care of pre-existing ~/Applications symlinks managed by nix-darwin and such too, only needs testing.

supermarin:
Thanks @toonn

PR #488, opened 2022-08-12
fix zsh eating output without new line ending

cuichenli:
caught the issue described in here https://github.com/NixOS/nixpkgs/issues/30121 and feeling it is worth to patch a fix here. 

basically the issues is that when you run a command with output missing the new line ending, it will not be displayed.

```
root@nixos> echo -n "abc\ndef"
abc
```

something like this ☝️

### DIFF
--- a/modules/programs/zsh/default.nix
+++ b/modules/programs/zsh/default.nix
@@ -54,7 +54,7 @@ in
 
     programs.zsh.promptInit = mkOption {
       type = types.lines;
-      default = "autoload -U promptinit && promptinit && prompt walters";
+      default = "autoload -U promptinit && promptinit && prompt walters && setopt prompt_sp";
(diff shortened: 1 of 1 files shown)


emilazy (dismissed):
Identical to the NixOS fix, LGTM. Sorry this PR was forgotten for such a long time. Can you rebase on current `master`?

cuichenli:
thanks @emilazy , i have merged the latest master branch into this fix branch.

emilazy (commented):
Thanks for the quick response! Would you mind doing `git reset --hard 5fea32493127cb4ba6231fffc08033047a0d04b7 && git pull --rebase upstream master` instead of a merge? We'd generally like to avoid having merges from `master` in the commit history outside perhaps of long-lived branches.

cuichenli:
sure, updated.

emilazy (approved):
Thank you!

PR #489, opened 2022-08-12
[github] Use macos-12

supermarin:
It might be beneficial to set up a matrix ([docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-multi-dimension-matrix)) and run all major supported macOS versions in parallel.

Reason: macos-10.15 is [being deprecated](https://github.com/actions/runner-images/issues/5583) and will be fully unsupported by Aug 30 2022.


supermarin:
@LnL7

domenkozar (approved):

PR #490, opened 2022-08-14
Bring `nix` module back in sync with the NixOS module (as much as possible)

### LINKED ISSUE #433: Reflect `nix` option changes from NixOS
https://github.com/NixOS/nixpkgs/commit/4a9d9928dc01421b878cfe4a16961c1e4862d846 changed a number of options in the `nix` module. For the sake of consistency, nix-darwin should mirror those changes and keep up with NixOS.

malob:
The `nix` module in `nix-darwin` has fallen pretty far out of sync with its NixOS counterpart. This PR is an attempt to bring them back in sync (up to the 22.05 release version), to the extent possible. I've tested these changes on my personal machine, and haven't experienced any issues. I suggest reviewing the changes by going through the commits one at a time, since it's essentially impossible to clearly tell what has changed by looking at the overall diff. 

With these changes the diff between the `nix` module in `nix-darwin` and the one in NixOS is relatively small, and very easy to parse. Additionally, now all options from the `nix` module in NixOS are present in `nix-darwin` with the exception of a small few where no equivalent option makes sense, and the `nix.optimise` and `nix.sshServe` submodules, which have not been implemented in `nix-darwin` yet.

Summary of significant changes:

* Moved applicable options to `nix.settings` submodule.
  * Key commits: f729a09, a00b383, 5786c07.
  * Old options, e.g., `nix.binaryCaches` (now `nix.settings.substituters`), will continue to work with a warning that tells the user the option as been renamed.
  * `nix.settings` is freeform, so settings can be configured for which no option is explicitly defined. For example, `nix.settings.experimental-features = [ "nix-command" "flakes" ]`, works even though that option isn't defined in the `nix.settings` submodule.
  * Now checks that the generated `nix.conf` file is valid (with a new option to turn that off).
  * **Implementation of the submodule requires Nix >= 2.2**.
* Removed the ability for `nix.package` to be a path.
  * Key commit: 5f14136
  * This was a carry over from when `nix.profile` was deprecated in favor of `nix.package`.
  * **It's _possible_ that this is a breaking change**, but I don't actually think it is. There seem to be many places in the codebase where `nix.package` is assumed to be a package.
  * Without this change, the `nix.settings` submodule above would need substantial changes from the one in NixOS.
* Updated `nix.buildMachines` to be a proper submodule like in NixOS.
  * Key commit: c3bdd6d
  * I think this should be completely backwards compatible and just work, but it would be helpful if someone could review/test this.
* Updated definition/implementation of `nix.nixPath` to more closely match the NixOS module
  * Key commits: 7648c9b, and ffc8ec5
  * The first commit shouldn't effect the functionality of the option, but was just done to make the diff with the NixOS module cleaner.
  * The second commit removes `$HOME/.nix-defexpr/channels` from the defaults for this option, and instead adds it to `NIX_PATH` as part of the shell's init script like NixOS does.
  * **This is the change that I think is most likely to have unforeseen side effects, so careful review by others would be greatly appreciated.**
* Moved build user configuration options definition/implementations into `nix` module (like NixOS)
  * Key commit: 7e74c1c
  * I was very careful when making this change, and confident that this will not change the values of the relevant `users.*` options, with the exception of the `gid` assigned to the `_nixbld*` users.
  * Previously `users.users._nixbld*.gid` was being set to `300`, even though the ID for the group itself (`users.groups.nixbld.gid`) was set to `30000`. I checked on my machine what the `gid` of `_nixbld*` users was, and it was `30000`. As such, I changed to code to set `users.users._nixbld*.gid` as `30000`.
* Updated the options related to CPU/IO priority/limits
  * Key commit: 08edc0e
  * These might be the most opinionated changes due to there being essentially no overlap between the relevant `systemd` and `launchd` configuration options. I tried my best to do something that made sense.
  * Most notable change is the removal of the setting for changing the process' `nice` level, in favor of a new option that sets the `ProcessType` option, along with a change from the default value of that `launchd` option from `Interactive` to `Standard`.

(Fixes #433)


malob:
Hmmm, wasn't able to tell at a quick glance with the (non-Flake) install workflow is failing, but my guess is it's due to 4df41e83e2a2b6de95ab94f4396571d359f61fa5.

domenkozar:
cc @LnL7

malob:
(Rebased to get CI updates.)

starcraft66:
Thanks for opening this much-needed PR! Hope it gets merged.

malob:
Fixed failing Test CI. 

The issue was that one of the installer checks was checking the `$NIX_PATH` environment variable and was assuming that  `$HOME/.nix-defexpr/channels` would be present, but this is no longer the case now that it is only added to `$NIX_PATH` when the shell is initialized if the directory is present. 

This [diff](https://github.com/LnL7/nix-darwin/compare/3fed47abf9e943ca3a6550f28e656212e81ad321..796a72f0fe2c8df0e37f0febda3b6aad3b351a68) shows the relevant changes.

domenkozar (approved):

srid:
Why was `nix.gc.dates` removed, when the nixos module still has it? https://github.com/NixOS/nixpkgs/blob/0fc9fca9c8d43edd79d33fea0dd8409d7c4580f4/nixos/modules/services/misc/nix-gc.nix#L23

There is no `interval` in the nixos module either.

Was this an oversight?

malob:
@srid, as far as I can tell `nix.gc.dates` was never present in `nix-darwin`, it's always had `nix.gc.interval` instead. I don't know the reasons for this.

With my changes here, I just used `mkRemovedOptionModule` to provide a helpful message for folks who might be trying to port over their NixOS configs. I think I did that for basically all `nix.*` options present in NixOS but not `nix-darwin`.

While I didn't get around to it in this PR, I'm sure a PR would be welcome that added that removed `nix.gc.interval` in favor of `nix.gc.dates`, if it didn't result in a loss of functionality, or adding `nix.gc.dates` in addition to `nix.gc.interval` if they could be made to play well together.

Issue #469, opened 2022-06-17
unable to override Homebrew appdir


sellout:
I’ve added the following to my darwin-configuration.nix
```nix
{
  homebrew.extraConfig = ''
    cask_args appdir: "/Applications/Homebrew Apps"
  '';
}
```
but everything in `homebrew.casks` still gets installed directly in /Applications. Modifying the brewfile to put `cask_args` at the beginning causes the correct directory to be used.

I see two solutions – either having `extraConfig` prepended instead of appended, or adding explicit `homebrew.appdir` (and `homebrew.fontdir` etc.) options to the module.

sellout:
Just wanted to say thanks to @AlexOwl @malob for this work! The Homebrew integration has been great. And thanks to @domenkozar for keeping nix-darwin moving forward!

PR #493, opened 2022-08-19
Fix brewfile behavior to allow cask_args, etc

AlexOwl:
Fix of the issue #469


malob (approved):
Thanks for catching this! I'm happy with this change as a quick way to resolve the issue.

(FYI, I'm currently working on an overhaul to the module that will add the ability to set `cask_args` directly, and provide a way to directly specify arguments etc. for individual `taps`, `brews`, and `casks`.)

domenkozar (approved):

PR #496, opened 2022-08-21
modules/homebrew: add global.githubToken

colin-arnott-xero:


malob:
As a general rule, I don't think it's a good idea to add options for people to include plain text credentials like this. Of course people can choose to do so themselves, but I don't think we should encourage it by providing and option.

colin-arnott-xero:
I guess this lined up with things like nixos' `networking.wireless.networks.<name>.psk` in my head, and iiuc this is the only way to provide that token to homebrew.

Is there any better suggestion to make this configuration declarative? I am running on a single user system thus secrets in the store are of less concern than secrets in nix/flake source, thus I do a `"{builtins.readFile /path/to/token_file}`.

colin-arnott-xero:
yeah, I have that setup currently (just the `environment.variables`, no `password), but I have been wanting something like the systemd secrets that nixops and company have. but it is not baked into nix or nixpkgs, and definitely nowhere near darwin compatible. I will have a look at your 1password setup, as I would love the ability to embed secret references in my flake and store.

that said, does this tangent not justify adding parameters like this one? I would be happy to add a warning like `networking.wireless.networks.<name>.psk`.

malob:
So for `networking.wireless.networks.<name>.psk` in NixOS, the module uses the value provided to that option when it generates the content of the `wpa_supplicant.conf` file. So there really isn't an alternative for the user, if they want Nix to generate that configuration file for them, and they want to include the pre-shared key in that configuration file.

In the case of the `homebrew` module in `nix-darwin`, the only benefit that adding the `homebrew.global.githubToken` option would provide (unless I'm confused about something here) is saving a few keystrokes, i.e, you be able to write,

```nix
{ homebrew.global.githubToken = "<token>"; }
```

instead of needing to write,

```nix
{ environment.variables.HOMEBREW_GITHUB_API_TOKEN = "<token>"; }
```

I'm definitely in favor of adding new options in cases where, even though it would be possible to achieve the same thing with existing options, a new option would make achieving the same thing a lot more succinct or intuitive (or it would be helpful for making a feature, or some functionality, more discoverable to users). 

This doesn't seem to me to be one of those cases though since, it's just substituting setting one option with a slightly shorter one, in addition to the fact that, folks who want this functionality will very likely already know how to accomplish it using `environment.variables`, that I think that adding an option like this will make it more likely that people who don't really know what they are doing add secrets to their config when the probably shouldn't (even with a warning in the description), and that there's also a small (but non-trivial) cost to adding new options in terms of adding more clutter.

MikaelElkiaer:
In case anyone finds this useful.

An alternative to handling auth within nix-darwin is to simply pre-fetch sources beforehand.
E.g.: `brew fetch ...`.

PR #498, opened 2022-08-23
Add ability to configure all known Brewfile line arguments with `homebrew` module

malob:
Figuring out what arguments can be used with `cask_arg`, `tap`, `brew`, and `cask` entries in a Brewfile, and what they do, is kind of a pain in the butt. The documentation in the Homebrew Bundle repo readme provides some examples, but they aren't very clear, and they're far from comprehensive. The `brew` manpage, and `brew bundle --help` offer a few more clues, but aren't that helpful.

Being annoyed by this, I decided to jump into the actual code for the `bundle` command to figure it out, and revamp the `homebrew` module based on my findings.

This PR reimplements the `homebrew.{taps,brews,casks}` options as lists of submodules (and adds a new option `homebrew.caskArgs`), with well documented options for all available arguments for these Brewfile entry types, while preserving backwards compatibility for the previous implementation.


modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
+          If you prefer this behavior you many also want to consider enabling
+          ${mkDocOptionLink "homebrew.global.noAutoUpdate"}.
+        '';
+      };
+      upgrade = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Whether to enable Homebrew to upgrade outdated formulae and Mac App Store apps during
+          <command>nix-darwin</command> system activation. The default is <literal>false</literal>
```
lockejan, line 112: I'm a bit unsure about  brews ability to upgrade `Mac App Store apps`. Are you implicitly referring to mas here?

malob: Well `brew bundle` can initiate upgrading of apps installed from the Mac App Store that are listed in the Brewfile with `mas` entries. When the `--no-upgrade` flag is passed to `brew bundle` it doesn't ask `mas` to upgrade the apps listed in the Brewfile that are outdated.

lockejan (commented):
Great job on the documentation. I don't have much opinion on the actual implementation, other than it's really nice to see that nix-darwin is going to provide an even richer interface for homebrew soon. Thanks for your effort.

I could build my config against your branch without problems, besides the usual mas flakiness (sometimes updates fail without obvious reason).

modules/homebrew.nix
lockejan: Oh, nice. TIL.

AlexOwl:
I would like to have homebrew global and per user configuration (especially for casks)
That's can be implemented with boolean config option like that will add proper "appdir" arg (/Applications,  ~/Applications)

malob:
> I would like to have homebrew global and per user configuration (especially for casks)
That's can be implemented with boolean config option like that will add proper "appdir" arg (/Applications, ~/Applications)

@AlexOwl, can you say a little more about what you are thinking here, and maybe provide code a example of the type of thing you'd like to be able to write in your config?

modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
+          Whether to enable Homebrew to automatically use the Brewfile that this module generates in
+          the Nix store, when you manually invoke <command>brew bundle</command>.
+
+          Implementation note: when enabled, this option sets the
+          <literal>HOMEBREW_BUNDLE_FILE</literal> environment variable to the path of the Brewfile
+          that this module generates in the Nix store, by adding it to
+          ${mkDocOptionLink "environment.variables"}.
+        '';
+      };
+      noAutoUpdate = mkOption {
```
toonn, line 146: Are we sure the double negative here won't be confusing? Consider flipping the default and dropping the "no."

modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
+
+          You may want to consider enabling this option if you have
+          ${mkDocOptionLink "homebrew.onActivation.upgrade"} enabled, and
+          ${mkDocOptionLink "homebrew.onActivation.update"} disabled, if you want to ensure that
+          your installed formulae will only be upgraded during <command>nix-darwin</command> system
+          activation, after you've explicitly run <command>brew update</command>.
+
+          Implementation note: when enabled, this option sets the
+          <literal>HOMEBREW_NO_AUTO_UPDATE</literal> environment variable, by adding it to
+          ${mkDocOptionLink "environment.variables"}.
```
toonn, line 163:
```suggestion
      autoUpdate = mkOption {
        type = types.bool;
        default = true;
        description = ''
          Whether to allow Homebrew to automatically update itself and all formulae
          when you manually invoke commands like <command>brew install</command>,
          <command>brew upgrade</command>, <command>brew tap</command>, and
          <command>brew bundle [install]</command>.

          You may want to consider disabling this option if you have
          ${mkDocOptionLink "homebrew.onActivation.upgrade"} enabled, and
          ${mkDocOptionLink "homebrew.onActivation.update"} disabled, if you want to ensure that
          your installed formulae will only be upgraded during <command>nix-darwin</command> system
          activation, after you've explicitly run <command>brew update</command>.

          Implementation note: when disabled, this option sets the
          <literal>HOMEBREW_NO_AUTO_UPDATE</literal> environment variable, by adding it to
          ${mkDocOptionLink "environment.variables"}.
```

modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
+        description = ''
+          Whether to disable lockfile generation when you manually invoke
+          <command>brew bundle [install]</command>. This is often desirable when
+          ${mkDocOptionLink "homebrew.global.brewfile"} is enabled, since
+          <command>brew bundle [install]</command> will try to write the lockfile in the Nix store,
+          and complain that it can't (though the command will run successfully regardless).
+
+          Implementation note: when enabled, this option sets the
+          <literal>HOMEBREW_BUNDLE_NO_LOCK</literal> environment variable, by adding it to
+          ${mkDocOptionLink "environment.variables"}.
```
toonn, line 178:
```suggestion
      lockfile = mkOption {
        type = types.bool;
        default = true;
        description = ''
          Whether to enable lockfile generation when you manually invoke
          <command>brew bundle [install]</command>. This is often undesirable when
          ${mkDocOptionLink "homebrew.global.brewfile"} is enabled, since
          <command>brew bundle [install]</command> will try to write the lockfile in the Nix store,
          and complain that it can't (though the command will run successfully regardless).

          Implementation note: when disabled, this option sets the
          <literal>HOMEBREW_BUNDLE_NO_LOCK</literal> environment variable, by adding it to
          ${mkDocOptionLink "environment.variables"}.
```

modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
+      };
+
+      homebrewEnvironmentVariables = mkInternalOption { type = types.attrs; };
+    };
+
+    config = {
+      homebrewEnvironmentVariables = {
+        HOMEBREW_BUNDLE_FILE = mkIf config.brewfile "${brewfileFile}";
+        HOMEBREW_BUNDLE_NO_LOCK = mkIf config.noLock "1";
+        HOMEBREW_NO_AUTO_UPDATE = mkIf config.noAutoUpdate "1";
```
toonn, line 189:
```suggestion
        HOMEBREW_BUNDLE_NO_LOCK = mkIf (!config.lockfile) "1";
        HOMEBREW_NO_AUTO_UPDATE = mkIf (!config.autoUpdate) "1";
```

Accompanying changes for earlier suggestions.

modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
+        brewfileLine =
+          "tap ${sCfg.name}"
+          + optionalString (sCfg ? clone_target) ", ${sCfg.clone_target}"
+          + optionalString (sCfg ? force_auto_update)
+            ", force_auto_update: ${sCfg.force_auto_update}";
+      };
+  };
+
+  # Sourced from https://docs.brew.sh/Manpage#global-cask-options
+  # and valid value for `HOMEBREW_CASK_OPTS`.
```
toonn, line 238:
```suggestion
  # and valid values for `HOMEBREW_CASK_OPTS`.
```

At least this reads as if it should be plural.

modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
+
+          Homebrew's default is <literal>false</literal>.
+        '';
+      };
+      no_quarantine = mkNullOrBoolOption {
+        description = "Whether to disable quarantining of downloads.";
+      };
+      no_binaries = mkNullOrBoolOption {
+        description = "Whether to disable linking of helper executables.";
+      };
```
toonn, line 360:
```suggestion
      quarantine = mkNullOrBoolOption {
        description = "Whether to enable quarantining of downloads.";
      };
      binaries = mkNullOrBoolOption {
        description = "Whether to enable linking of helper executables.";
      };
```

modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
+          Whether to run <command>brew services start</command> for the formula and register it to
+          launch at login (or boot).
+
+          Homebrew's default is <literal>false</literal>.
+        '';
+      };
+      link = mkNullOrBoolOption {
+        description = ''
+          Whether to link the formula to the Homebrew prefix. When this option is
+          <literal>null</literal>, Homebrew will use it's default behavior which is to link the
```
toonn, line 420:
```suggestion
          <literal>null</literal>, Homebrew will use its default behavior which is to link the
```

modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
+      let
+        sCfg = mkProcessedSubmodConfig config;
+        sCfgSubset = removeAttrs sCfg [ "name" "restart_service" ];
+      in
+      {
+        brewfileLine =
+          "brew ${sCfg.name}"
+          + optionalString (sCfgSubset != { }) ", ${mkBrewfileLineOptionsListString sCfgSubset}"
+          # We need to handle the `restart_service` option seperately since it can be either bool
+          # or the string value ":changed" the Brewfile.
```
toonn, line 439:
```suggestion
          # or the string value "changed".
```

Not sure this is correct but the sentence as-is seems broken.

modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
+  ];
+
   options.homebrew = {
     enable = mkEnableOption ''
-      configuring your Brewfile, and installing/updating the formulas therein via
-      the <command>brew bundle</command> command, using <command>nix-darwin</command>.
-
-      Note that enabling this option does not install Homebrew. See the Homebrew website for
-      installation instructions: https://brew.sh
+      <command>nix-darwin</command> to manage installing/updating/upgrading Homebrew taps, formulae,
```
toonn, line 497: This sentence seems to start a bit non sequitur, "nix-darwin to manage..."

modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
-        <command>brew bundle [install]</command> with the <command>--cleanup --zap</command>
-        flags. This uninstalls all formulas not listed in the generated Brewfile, and if the
-        formula is a cask, removes all files associated with that cask. In other words,
-        <command>brew uninstall --zap</command> is run for all those formulas.
-
-        If you plan on exclusively using <command>nix-darwin</command> to manage formulas installed
-        by Homebrew, you probably want to set this option to <literal>"uninstall"</literal> or
-        <literal>"zap"</literal>.
+        Options for configure the behavior of the <command>brew bundle</command> command that
+        <command>nix-darwin</command> runs during system activation.
```
toonn, line 541:
```suggestion
        Options for configuring the behavior of the <command>brew bundle</command> command that
        <command>nix-darwin</command> runs during system activation.
```

toonn (approved):
LGTM, some nitpicks.

You pushed some commits while I was reviewing, hope all the suggestions still make sense.

malob:
> You pushed some commits while I was reviewing, hope all the suggestions still make sense.

@toonn, I didn't add any new commits, just made a couple small tweaks.

Thanks for the nits, I'm looking them over now.

modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
+
+  mkProcessedSubmodConfig = attrs: mapAttrs (_: mkBrewfileLineValueString)
+    (filterAttrsRecursive (n: v: n != "_module" && n != "brewfileLine" && v != null) attrs);
+
+
+  # Submodules -------------------------------------------------------------------------------------
+  # Option values and descriptions for Brewfile entries are sourced/derived from:
+  #   * `brew` manpage: docs.brew.sh/Manpage
+  #   * `brew bundle` source files:
+  #     * github.com/Homebrew/homebrew-bundle/blob/master/lib/bundle/dsl.rb
```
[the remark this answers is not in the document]
malob: Good call, will do.

modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
+          This option manages what happens to formulae installed by Homebrew, that aren't present in
+          the Brewfile generated by this module, during <command>nix-darwin</command> system
+          activation.
+
+          When set to <literal>"none"</literal> (the default), formulae not present in the generated
+          Brewfile are left installed.
+
+          When set to <literal>"uninstall"</literal>, <command>nix-darwin</command> invokes
+          <command>brew bundle [install]</command> with the <command>--cleanup</command> flag. This
+          uninstalls all formulae not listed in generate Brewfile, i.e.,
```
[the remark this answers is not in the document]
malob: Will fix.

modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
+      brewBundleCmd = mkInternalOption { type = types.str; };
+    };
+
+    config = {
+      brewBundleCmd = concatStringsSep " " (
+        optional (!config.update) "HOMEBREW_NO_AUTO_UPDATE=1"
+        ++ [ "brew bundle --file='${brewfileFile}' --no-lock" ]
+        ++ optional (!config.upgrade) "--no-upgrade"
+        ++ optional (config.cleanup == "uninstall" || config.cleanup == "zap") "--cleanup"
+        ++ optional (config.cleanup == "zap") "--zap"
```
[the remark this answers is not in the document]
malob: Agreed, will change.

modules/homebrew.nix
malob: Yup, thanks for catching this.

modules/homebrew.nix
malob: All the option names here for the `homebrew.caskArgs` are exact matches for valid arguments for the `cask_args` entry of a Brewfile (and the `args` argument for `casks` entries). For example, enabling those two options will create the following entry in the Brewfile:

```ruby
cask_args no_binaries: true, no_quarantine: false
# or
casks "google-chrome", args: { no_binaries: true, no_quarantine: false }
```

I think it's best to have the option names here match what ends up in the Brewfile. Additionally I don't like the idea of having to add special cases in the code to process these two options.

modules/homebrew.nix
malob: This is already fixed in the final diff.

modules/homebrew.nix
malob: Thanks for catching this, will fix.

modules/homebrew.nix
malob: `mkEnableOption` creates the description like this: `"Whether to enable ${<the argument to mkEnableOption>}."`.

modules/homebrew.nix
malob: 👍, will fix.

modules/homebrew.nix
toonn: That's unfortunate.  There's both `--[no-]{binaries,quarantine}` options so I figured both ways might be valid. The inverted boolean flag is an unnecessarily confusing anti-pattern.

malob: So I will confess that I'm not completely sure how `brew bundle` is doing things under the hood here, but I'm pretty sure what it's doing is pretty naive. My guess is that, for binary arguments provided to `cask_args`, if they are set to `true`, the corresponding flag is added to the `brew install --cask` command that it runs, and if set they are to `false` it just leave off those flags.

malob: I'm also inferring this from the Homebrew manpages,

> HOMEBREW_CASK_OPTS 
> Append these options to all cask commands. All --*dir options, --language, --require-sha, --no-quarantine and --no-binaries are supported.

and the fact that, by default, quarantining and linking of binaries is enabled.

modules/homebrew.nix
```diff
@@ -1,219 +1,751 @@
...
+  globalOptions = { config, ... }: {
+    options = {
+      brewfile = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Whether to enable Homebrew to automatically use the Brewfile that this module generates in
+          the Nix store, when you manually invoke <command>brew bundle</command>.
+
+          Implementation note: when enabled, this option sets the
```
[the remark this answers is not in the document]
malob: Yeah, I had flagged this in the `homebrew.global.noLock` description, which used to say:

```
          Whether to disable lockfile generation when you manually invoke
          <command>brew bundle [install]</command>. This is often desirable when
          ${mkDocOptionLink "homebrew.global.brewfile"} is enabled, since
          <command>brew bundle [install]</command> will try to write the lockfile in the Nix store,
          and complain that it can't (though the command will run successfully regardless).
```

Your comment here did bring to my attention that folks might not see this, and end up enabling `homebrew.global.brewfile` without also enabling `homebrew.global.noLock` when they would have wanted to. In fact, it seems unlikely that folks would want to enable the former without also enabling the latter.

As a result, I've changed the description of `homebrew.global.brewfile` to include,
```
          [...]
          Enabling this option will change the default value of
          ${mkDocOptionLink "homebrew.global.noLock"} to <literal>true</literal> since,
          <command>brew bundle [install]</command> will try to write the lockfile in the Nix store,
          and complain that it can't (though the command will run successfully regardless).
          [...]
```

and the description of `homebrew.global.noLock` to,
```
          [...]
          This option will default to <literal>true</literal> if
          ${mkDocOptionLink "homebrew.global.brewfile"} is enabled, since
          <command>brew bundle [install]</command> will try to write the lockfile in the Nix store,
          and complain that it can't (though the command will run successfully regardless).
          [...]
```

and added the following the the configuration of the `homebrew.global` submodule,
```
      # See description of `homebrew.global.{brewfile,noLock}` options for context.
      noLock = mkIf config.brewfile (mkDefault true);
```

along with a note in the changelog about this.

modules/homebrew.nix
malob: I'm torn about this. Here's a brain dump of my current thinking:

* When I'm designing am module, I want it's default behavior to be what most people would expect, e.g., not cause surprising or unexpected changes to how their system works.
* I like to add boolean options as toggles to enable some additional/different behavior/functionality/setting that folks would want to explicitly opt-in to, where the default is that it's off (false), and so the user setting the option to true is them saying "I want this additional behavior/functionality/setting enabled".
* That said, I'm definitely not dogmatic about this. I have no love for double negatives, and so I sometimes make an exception from the above when it's needlessly confusing.
* In this particular case I'm currently leaning against removing the double negative because:
  * I like that, for folks that are familiar with Homebrew, the option names line up with the with names of the environment variables they'd be familiar with and be looking to set.
  * The `homebrew.global.noLock` option has existed since the module was created. I'm  aware that I could use `mkChangedOptionModule` to rename the option while preserving backwards compatibility with the current option, but I don't love that folks would have to notice that they should swap the value when renaming the option to address the warning they'll see when trying to build their config after this update. I guess I could use `doRename` directly, and provide a custom warning message, but I have a reflex against doing that since it will clutter up the module a little more. (Maybe I should just get over that thought ¯\_(ツ)_/¯.)

malob: That said, I don't feel very strongly about this, and so if most folks think this is an obvious improvement, I'd be happy to defer to the wisdom of the crowd and make the changes.

malob:
@toonn, alrighty, I've now addressed the majority of your review comments, and left replies on the remainder. Back to you :)

modules/homebrew.nix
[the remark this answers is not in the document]
toonn: This seems sensible to me 👍

modules/homebrew.nix
toonn: I'm leaving this up to your discretion. The default doesn't have to change just because the name changes though.

**My thoughts on the topic**:

I don't feel like modules should follow UI choices of CLIs, they're fundamentally different interfaces. Adding `--no-lock` unambiguously means "don't lock," setting `noLock = true` does not share this clear-cut interpretation.

Another way to accomodate experienced users and new users alike is to offer both options and asserting they are each others' opposite if both are set. Not elegant but it covers both concerns, inexperienced people  aren't forced into boolean contortions and experienced people don't have to unlearn their domain knowledge.

modules/homebrew.nix
toonn: This is up to you again. It's a bit of a tug of war between simpler UX on one hand and simple implementation on the other.

toonn (approved):
Everything that's left is fairly subjective and shouldn't stand in the way of this being accepted.

modules/homebrew.nix
malob: > It's a bit of a tug of war between simpler UX on one hand and simple implementation on the other.

Totally.

Thanks, for the feedback. In this case I'd like to leave these as is for the reasons I mentioned above related to uncertainty about how `brew bundle` works under the hood, and therefore wanting the line in the Brewfile to always include the `no_` prefixed version of the options, and not wanting to add custom processing in the submodule to rename them while still using the `no_` prefix in the generated brewfileLine for the submodule.

modules/homebrew.nix
malob: I'm feeling more convinced that this change should be made. I'll play around with this today.

modules/homebrew.nix
```diff
@@ -1,219 +1,780 @@
...
+          activation, after you've explicitly run <command>brew update</command>.
+
+          Implementation note: when disabled, this option sets the
+          <literal>HOMEBREW_NO_AUTO_UPDATE</literal> environment variable, by adding it to
+          ${mkDocOptionLink "environment.variables"}.
+        '';
+      };
+      lockfiles = mkOption {
+        type = types.bool;
+        default = true;
```
toonn, line 180: Maybe we should make clear that the default depends on the brewfile option?
```suggestion
        default = !brewfile;
```

modules/homebrew.nix
```diff
@@ -1,219 +1,780 @@
...
+        description = ''
+          Whether to enable Homebrew to generate lockfiles when you manually invoke
+          <command>brew bundle [install]</command>.
+
+          This option will default to <literal>false</literal> if
+          ${mkDocOptionLink "homebrew.global.brewfile"} is enabled since, with that option enabled,
+          <command>brew bundle [install]</command> will default to using the Brewline that this
```
toonn, line 187:
```suggestion
          <command>brew bundle [install]</command> will default to using the Brewfile that this
```

modules/homebrew.nix
```diff
@@ -1,219 +1,780 @@
...
+        default = false;
+        description = ''
+          Whether to enable Homebrew to auto-update itself and all formulae during
+          <command>nix-darwin</command> system activation. The default is <literal>false</literal>
+          so that repeated invocations of <command>darwin-rebuild switch</command> are idempotent.
+
+          Note that Homebrew auto-updates when it's been more then 5 minutes since it last updated.
+
+          If you prefer this behavior you may want to consider enabling
+          ${mkDocOptionLink "homebrew.global.noAutoUpdate"}.
```
toonn, line 106:
```suggestion
          If you do not prefer this behavior you may want to consider disabling
          ${mkDocOptionLink "homebrew.global.autoUpdate"}.
```

I think what we want to say here is to turn the global option off if we never want brew to automatically update, right?

toonn (commented):
One `noAutoUpdate` survived the change : )

modules/homebrew.nix
malob: Thanks for catching this! I actually did this twice. Will fix.

modules/homebrew.nix
malob: Another great catch. Fixed.

modules/homebrew.nix
malob: Yeah, that does seem better. I had considered doing this, but didn't since I was under the impression for some reason that using values from `config` when setting option defaults would cause infinite recursion issues.

I've now changed this to,
```
        default = !config.brewfile;
        defaultText = literalExpression "!config.homebrew.global.brewfile";
```

and removed,
```
      lockfiles = mkIf config.brewfile (mkDefault true);
```

malob:
@domenkozar, I believe this is now ready to be merged.

domenkozar (approved):

PR #499, opened 2022-08-24
Karabiner-Elements

Enzime:
PR adding Karabiner-Elements to nixpkgs: https://github.com/NixOS/nixpkgs/pull/188129

### DIFF
--- a/modules/launchd/launchd.nix
+++ b/modules/launchd/launchd.nix
@@ -194,6 +194,21 @@ with lib;
             '';
           };
 
+          Crashed = mkOption {
+            type = types.nullOr types.bool;
+            default = null;
+            description = ''
+              If true, the the job will be restarted as long as it exited due to a signal which is typically
+              associated with a crash (SIGILL, SIGSEGV, etc.). If false, the job will be restarted in the
+              inverse condition.
+            '';
+          };
+
+          AfterInitialDemand = mkOption {
+            type = types.nullOr types.bool;
+            default = null;
+          };
+
         };
       }));
       default = null;
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -47,6 +47,7 @@
   ./services/dnsmasq.nix
   ./services/emacs.nix
   ./services/gitlab-runner.nix
+  ./services/karabiner-elements
   ./services/khd
   ./services/kwm
   ./services/lorri.nix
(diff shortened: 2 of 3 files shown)


Enzime:
My PR to add Karabiner-Elements to Nixpkgs has been merged now :)

Enzime:
@domenkozar @toonn if you get a chance, can you take a look at this PR?

domenkozar (approved):

toonn:
Thank you for this. I personally use Karabiner-Elements so this is a welcome addition : )
I tried setting this up for an older macOS but it looks like Karabiner-Elements 12.10.0's installer package doesn't have the same structure and launchd units. Are you familiar with how the older versions worked?

Enzime:
@toonn which version of macOS are you running? I'm not too familiar but I'm happy to have a look into it in the future, but I have a feeling that the launchd units and setup might be quite different and require a new module.

toonn:
It's High Sierra 10.13.6, I don't think launchd changed much or anything. It's more that Karabiner-Elements used to have fewer services, they probably had to add stuff to work with the new signature system. Last compatible version of Karabiner-Elements is 12.10.0.

DavSanchez:
Hello all,

I have enabled Karabiner-Elements using the nix-darwin option, disabled it later, and both Karabiner-Elements and Karabiner-EventViewer applications are still there after switching and restarting. The uninstall option present in the Karabiner-Elements GUI does nothing, and I'm not sure what can I do to actually remove the application from my system.

Do you have any advice for this? Thanks!

Enzime:
@DavSanchez The only thing that should be left over is the folder `/Applications/.Nix-Karabiner` which only contains the system extension, which you can safely delete. If you close Karabiner-Elements, it’ll still exist in the Nix Store but there shouldn’t be any symlinks to it in `/Applications` or `~/Applications`

DavSanchez:
Then I don't know what happened in my case, but surely the GUI applications were still accessible from the launchpad even after having restarted after switching 🤷‍♂️ 

In any case, it seems like directly deleting everything from Karabiner with `find` did solve it.

toonn:
I think macOS resolves any symlinks when putting things in the Dock, Launchpad is probably similar. So as long as the applications aren't GC'ed from the store they'll probably keep appearing there.

DavSanchez:
Ahh that could be the case. I think I tried performing GC as well but of this I'm really not sure. Thank you very much to all!

Enzime:
@DavSanchez it wouldn't get garbage collected if you didn't delete the profiles that include Karabiner-Elements, on NixOS that would be `sudo nix-collect-garbage -d`, not sure if you would need extra flags on macOS.

PR #508, opened 2022-08-29
docs: add missing environment defaults

### LINKED ISSUE #395: "Configuration Options" page not mentioning `environment.variables.EDITOR` and `PAGER` being set
Took me a while to find where my `EDITOR` was set to `nano`. 

`man 5 configuration.nix` and [Configuration Options page](https://daiderd.com/nix-darwin/manual/index.html#opt-environment.variables) is saying that the default value is empty, but `EDITOR` and `PAGER` was set by default:
https://github.com/LnL7/nix-darwin/blob/44da835ac40dab5fd231298b59d83487382d2fab/modules/environment/default.nix#L175-L176

Maybe update the manual or remove the default value?

lockejan:
Actual environment defaults and those that are shown in the options-page differ.

This PR adds the ones mentioned in #395 .

However, there are even more env-defaults, namely XDG_CONFIG_DIRS and XDG_DATA_DIRS, see https://github.com/LnL7/nix-darwin/blob/44da835ac40dab5fd231298b59d83487382d2fab/modules/environment/default.nix#L173-L174

But as those are not static I'd rather omit them at this point.

closes #395


lockejan:
after figuring out defaultText, I wonder if this should be the preferred way to present it @malob

malob:
This is something I've often been a little annoyed by with modules in the Nix ecosystem. Where the particular annoyance is that there are some, what I'll call, core options (like `environment.variables`, that are set to values "by default" in the module they are declared in (and sometimes in other modules as well) to some value (i.e, before the user adds their on modules with additional configuration they will always have the same predictable values), and the only reliable way to figure out what their "default" will be is to look rummage through the code base and figure it out.

I think something like the suggested change here is likely an improvement. My main concern would be that it's kind of brittle, in that, if a change is made to `nix-darwin` that would add something extra to `environment.variables` "by default", the person making the change (or a reviewer) would have to notice this and update the `defaultText` for the option accordingly. In this world it seems like it's only a matter of time before someone makes a change and that latter step is missed, resulting in `defaultText` no longer accurately representing the "default" value of the option.

I'm currently leaning towards thinking that adding `defaultText` with the risk it becoming  inaccurate at some point is still better than the status quo, where it's obviously wrong, but I'm not confident about that. I expect peoples intuitions here to vary.

lockejan:
Can we merge @LnL7 or @domenkozar ?

domenkozar:
I'd add a comment at both locations that there's another place to update the values.

lockejan:
> I'd add a comment at both locations that there's another place to update the values.

I'm not sure if I understood you correctly.
Does my last commit aligns with what you were referring to? @domenkozar

domenkozar:
Yeah, but I'd say something like `WARNING: XDG vars are also set up in this file above`

lockejan:
> Yeah, but I'd say something like `WARNING: XDG vars are also set up in this file above`

Sorry for the ping-pong, but maybe there's a misunderstanding here or I'm just confused about the motivation for your request. 🙈😄

Just to be clear here, my initial additions are only documentation and aren't changing any behaviour.
The entries in XDG vars are the result of concatenating "/etc/xdg" and "/share" to each entry in NIX_PROFILES which are defined under `environment.profile` in the same file.

There is currently no other place where XDG vars are set. 
That been said, can you please give me some more context why you think an additional comment is necessary?

domenkozar:
I took a proper look at this one, why don't we just set the correct default?

lockejan:
Replaced the expression with the actual result. 
This way it's clear what is being set, as long as the list in `environment.profile` doesn't change.

lockejan:
How about merging this @domenkozar ? (:

emilazy (changes requested):
The argument to `literalExpression` should be a valid Nix expression, so please add the surrounding `{ ... }` and make sure that quotes and semicolons are in the correct places. I agree with @domenkozar that there should be a comment by the actual definition reminding people to update the option definition if the default is changed.

LnL7:
I'm not sure if hardcoding the defaults is a great idea, actually evaluating the defaults with something like `darwin-option` instead would ensure things stay in sync. The current implementation doesn't interpet the `mkDefault` values correctly however.

```
$ darwin-option environment.variables
Value:
{ EDITOR = "nano"; NIX_PATH = "darwin-config=$HOME/.nixpkgs/darwin-configuration.nix:/nix/var/nix/profiles/per-user/root/channels"; NIX_SSL_CERT_FILE = "/etc/ssl/certs/ca-certificates.crt"; PAGER = "less -R"; TERMINFO_DIRS = "$HOME/.nix-profile/share/terminfo:/run/current-system/sw/share/terminfo:/nix/var/nix/profiles/default/share/terminfo:/usr/share/terminfo"; XDG_CONFIG_DIRS = "$HOME/.nix-profile/etc/xdg:/run/current-system/sw/etc/xdg:/nix/var/nix/profiles/default/etc/xdg"; XDG_DATA_DIRS = "$HOME/.nix-profile/share:/run/current-system/sw/share:/nix/var/nix/profiles/default/share"; }

Default:
{ }
```

Issue #481, opened 2022-07-24
darwin-rebuild should use `sudo -H` instead of `sudo` to work around Nix warning


lilyball:
Nix 2.10 has started printing a warning whenever the current user doesn't own `$HOME`, which is the case on macOS when running `sudo nix` because macOS `sudo` doesn't reset `$HOME` by default. `darwin-rebuild` is currently running `sudo nix-env -p "$profile" --set "$systemConfig"`. This command shouldn't care about `$HOME`[^1] and yet it now triggers a warning. It would be nice to avoid the warning.

[^1]: In theory `$HOME` can affect it through per-user `nix.conf` files but that seems rather niche.

Issue #477, opened 2022-07-17
Nix complains about $HOME being retained by sudo


Sciencentistguy:
Ever since this (https://github.com/NixOS/nix/pull/6676) change in nix, it checks that $HOME is owned by the current user. 

When using darwin-rebuild, this check is tripped, causing a warning to be printed:
```
warning: $HOME ('/Users/jamie') is not owned by you, falling back to the one defined in the 'passwd' file.
```

As far as I can tell, this occurs during this invokation of nix-env:
https://github.com/LnL7/nix-darwin/blob/54a24f042f93c79f5679f133faddedec61955cf2/pkgs/nix-tools/darwin-rebuild.sh#L181

This happens because `nix-env` is executing as UID 0, however `$HOME` still points to `/Users/jamie` (sudo on macos does not clear `$HOME`)

This does not appear to be actually causing any issues, but thought I'd report it here anyway, in case this is fixable from this end.

joscha:
Ah, thanks for tracking this down. I was searching high and low where this was suddenly coming from! I've not noticed any bad side effects either.

FlakM:
I think I've got into some troubles due to that change. My account is not present in /etc/passwd file (I think is manages like NSS on linux) I get this warning and all applications seem to think that my home directory is located in `/var/empty` is this possible?

supermarin:
Setting users.users.USER.home doesn't fix this for me. Just noticed this warning today after bumping nixpkgs. Nix-darwin with home-manager as a module.

kkharji:
I still face this issue, 

```
warning: $HOME ('/Users/kkharji') is not owned by you, falling back to the one defined in the 'passwd' file ('/var/root')
```

 Using nix flake, info 
 
 ```
 ├───nix-darwin: github:lnl7/nix-darwin/6349b99bc2b96ded34d068a88c7c5ced406b7f7f
│   └───nixpkgs follows input 'nixpkgs'
 ```

PR #509, opened 2022-08-29
Add `-H` to `sudo nix-env [...]` calls in `darwin-rebuild.sh` to deal with warning

malob:
After NixOS/nix#6676, the following warning is displayed when running `darwin-rebuild switch`:
> warning: $HOME ('/Users/jamie') is not owned by you, falling back to the one defined in the 'passwd' file.

This happens because, on macOS `sudo` does not set `$HOME` by default, the `-H` flag needs to be added for it to do so.

Fixes: #481
Fixes: #477

### DIFF
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -160,7 +160,7 @@ fi
 
 if [ "$action" = list ] || [ "$action" = rollback ]; then
   if [ "$USER" != root ] && [ ! -w $(dirname "$profile") ]; then
-    sudo nix-env -p "$profile" "${extraProfileFlags[@]}"
+    sudo -H nix-env -p "$profile" "${extraProfileFlags[@]}"
   else
     nix-env -p "$profile" "${extraProfileFlags[@]}"
   fi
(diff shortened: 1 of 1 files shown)


malob:
@lilyball, seems to know more about this than I do, so tagging her here in case she has context etc. to add.

domenkozar (approved):

lilyball:
It might be nice to add `-H` to the `sudo "$systemConfig/activate"` line as well. A quick skim of the activate script doesn't show any references to `$HOME` but it's still probably a good idea for consistency, especially as someone could write a custom activation script that references this variable or invokes a command that itself consults this variable, and system activation should probably try to avoid any dependency on the initiating user.

lilyball:
@malob BTW I absolutely love all the recent improvements you've been making here!

malob:
> It might be nice to add `-H` to the `sudo "$systemConfig/activate"` line as well. A quick skim of the activate script doesn't show any references to `$HOME` but it's still probably a good idea for consistency, especially as someone could write a custom activation script that references this variable or invokes a command that itself consults this variable, and system activation should probably try to avoid any dependency on the initiating user.

@lilyball, I don't feel knowledgeable enough about this at the moment to propose this change myself, but if you do, I think you should totally create a PR to propose the change.

zowoq:
> It might be nice to add `-H` to the `sudo "$systemConfig/activate"` line as well.

Opened a PR for this: https://github.com/LnL7/nix-darwin/pull/783

Issue #454, opened 2022-04-15
Supporting Builtin Float?


mtoohey31:
As of [a while ago](https://github.com/NixOS/nix/pull/762), Nix has builtin floats. Would you be open to a PR that refactors everything that uses [this custom float type](https://github.com/LnL7/nix-darwin/blob/4fdbb8168f61d31d3f90bb0d07f48de709c4fe79/modules/system/defaults-write.nix#L8) to use the builtin float type?

mtoohey31:
Actually, I suppose for backwards compatability we should refactor so that the custom float type accepts builtin float, as well as all the types it currently accepts.

beetcb:
Yeah, a little bit frustrating that the `float`(in mannul) is actually `string`.
I think we can start with improving the documentation first @LnL7 

<img width="596" alt="Screen Shot 2022-08-10 at 11 04 32" src="https://user-images.githubusercontent.com/63141491/183801613-9a6067a7-5628-409b-bab9-3599a074db37.png">

PR #510, opened 2022-08-29
Transition to using native floats for option types

malob:
This PR transitions all options that use strings to represent float values to use builtin floats. 

This change is not backwards compatible, since using string values for those options will produce and error, but to ease the transition, I've create a `floatWithDeprecationError` type, that will print a helpful error message informing users of the change.

Fixes: #454

### DIFF
--- a/modules/nix/default.nix
+++ b/modules/nix/default.nix
@@ -37,7 +37,7 @@ let
         if v == null then ""
         else if isInt v then toString v
         else if isBool v then boolToString v
-        else if isFloat v then floatToString v
+        else if isFloat v then strings.floatToString v
         else if isList v then toString v
         else if isDerivation v then toString v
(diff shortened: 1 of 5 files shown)


domenkozar (approved):

joelbu:
Hey @malob, I know this is late, but is it possible that system.defaults.".GlobalPreferences"."com.apple.mouse.scaling" got lost in this transition? I couldn't turn off mouse acceleration because my float wasn't accepted. I'm still a nix noob, so I'm pretty unsure of ... basically everything, but I see a regex for a stringy float type in here https://github.com/LnL7/nix-darwin/blob/master/modules/system/defaults/GlobalPreferences.nix and indeed "-1" works for me.

malob:
Looks like those options were merged in after the changes I made in this PR: #452

Issue #415, opened 2021-12-27
Documentation: switch freenode->libera?


jsoo1:
The readme mentions freenode.net as the primary irc network for contact. It seems like #nix-darwin on freenode is quite empty. Is the libera #nix-darwin the new place to go?

LnL7:
I'm not sure there are still that many people from the nix community in general on irc since moving away from freenode. The macOS channel on the official matrix is probably a better option.

jsoo1:
Ok That's `#macos:nixos.org`?

LnL7:
Yeah, I think so.

PR #512, opened 2022-08-29
Update readme to point to Matrix instead of IRC

malob:
Fixes: #415

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,5 +211,5 @@ darwin-rebuild switch -I darwin=.
 ```
 
 Also feel free to contact me if you have questions,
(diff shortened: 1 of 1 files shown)


domenkozar:
It's best to avoid names as we have a couple of hundred people around there

malob:
> It's best to avoid names as we have a couple of hundred people around there

@domenkozar, is this what you are referring to: `@daiderd:matrix.org`? 

That's just LnL's user on Matrix. What do you think we should change that line to? I don't have strong opinions here, I'm just trying to clear out issues that are easy to resolve.

PR #520, opened 2022-09-06
Add an `--update` CLI arg to `darwin-switch`

yvan-sraka:
This mimics the behaviour of [`nixos-rebuild switch --upgrade`](https://nixos.org/manual/nixos/stable/index.html#sec-upgrading), running `nix-channel --update darwin; nixos-rebuild switch`.


pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -112,6 +116,10 @@ done
 
 if [ -z "$action" ]; then showSyntax; fi
 
+if [ "$updateChannels" = true ] ; then
+  nix-channel --update darwin
```
lockejan, line 120: Hmm, that would fail in a lot of cases, where people use nix as a deamon service and therefore root owns it.

lockejan:
Can you provide some tests to verify that this works as intended?

pkgs/nix-tools/darwin-rebuild.sh
yvan-sraka: You're right… is it reasonable to try to catch all that cases?

yvan-sraka:
> Can you provide some tests to verify that this works as intended?

Yes! Indeed.

pkgs/nix-tools/darwin-rebuild.sh
lockejan: You could test if the nix-channel exists and if nix runs as a daemon.
However the latter needs sudo, to inspect the running services.

(There's also the possibility that people have nix-darwin as a user and a root channel, without knowing it. I guess that's currently a bug in the installer. At least to my understanding.)

dstcruz: Oh?! I think I have the scenario described here: root channel and nixpkgs, user nix-darwin. If that's "wrong", how can I clean that up?

yvan-sraka: Maybe that command line argument is a good occasion to throw a user warning/advice if it detects a "wrong" configuration?

yvan-sraka:
Supersede by https://github.com/LnL7/nix-darwin/pull/1014

PR #523, opened 2022-09-09
fix: default to root-channel for nix-darwin and nixpkgs

### LINKED ISSUE #522: Use root-channels per default
Currently the installer defaults to install nix-darwin as a user-channel.
This made sense while nix favoured single-user installs under macOS.
Since quite some time nix defaults to a multi-user install under macOS, which results in nixpkgs being registered as a root-channel.
The `nix-darwin`-installer registers nix as a daemon service per default and this daemon consequently belongs to root as well.

There is also quite some confusion around this during beginners, because they often end up having nixpkgs as a root and user-channel.
It's a reoccurring topic in the [matrix-channel](https://matrix.to/#/#macos:nixos.org), see for example [this](https://matrix.to/#/!lheuhImcToQZYTQTuI:nixos.org/$DKZ1m8iAMNop2ldZZpcR5oDc4saDhi7oWyDPZJlGkN8?via=nixos.org&via=matrix.org&via=nixos.dev) or [this](https://matrix.to/#/!lheuhImcToQZYTQTuI:nixos.org/$cX1q9rrM7j7Qb_prH0oU5d2CshOKtDs4ySm-YKfC1t8?via=nixos.org&via=matrix.org&via=nixos.dev) post.

We discussed the topic a few times on matrix ([most recently](https://matrix.to/#/!lheuhImcToQZYTQTuI:nixos.org/$7lLCy1CM4PN3X57EfRWpnckqS1ZO7IZ9ASE7MHVa1jg?via=nixos.org&via=matrix.org&via=nixos.dev)) and came to the conclusion that there's no reason to keep it the old way.
Therefore the installer and all affected parts of the repo should be updated to default to `sudo -i nix-channel ...` to interact with roots channels.

lockejan:
* closes #522

Running all tests locally passed.

There are still some files containing user-channel interaction:
https://github.com/LnL7/nix-darwin/blob/master/modules/services/gitlab-runner.nix#L282-L283
https://github.com/LnL7/nix-darwin/blob/master/modules/services/ofborg/default.nix#L26-L27
https://github.com/LnL7/nix-darwin/blob/master/modules/examples/ofborg.nix#L11-L12
Wasn't sure if those should be moved to roots channels as well.

https://github.com/LnL7/nix-darwin/blob/master/modules/checks.nix#L104
This seems to me to deliberately interact with user-channels and therefore should remain untouched.

The motivation for this change has been roughly described in #522


lockejan:
CI is currently failing for the install action. Guess I overlooked something. Won't have time for the next days, but maybe someone else is able to spot it in the meantime.

lockejan:
Ok, well, had some minutes to look into it. Should be good to go now, I guess? What do you think? Anything missing? @toonn

toonn (dismissed):
LGTM

I don't have experience with GitLab runners so I'm not sure whether that channel setup should change. It seems to be running in a Docker container based on Alpine? So I'm not sure there is root access there or whether matters for a project mainly used on macOS.

I think the OfBorg examples can be updated to use root channels too. Otherwise activation of the configuration would only work for users who happen to have specified a user channel for OfBorg?

The checks seem fine as well. The few remaining tests using user channels could still be changed or maybe it's handy for tests not to require root permissions at all and everything should be changed back?

lockejan:
> I don't have experience with GitLab runners so I'm not sure whether that channel setup should change. It seems to be running in a Docker container based on Alpine? So I'm not sure there is root access there or whether matters for a project mainly used on macOS.
> 
> I think the OfBorg examples can be updated to use root channels too. Otherwise activation of the configuration would only work for users who happen to have specified a user channel for OfBorg?

Sounds reasonable to me. I'll update the OfBorgs examples and leave the GitLab Runner as it is.

lockejan:
> The checks seem fine as well. The few remaining tests using user channels could still be changed or maybe it's handy for tests not to require root permissions at all and everything should be changed back?

Ah, I didn't really get how they work until now. I'd pledge for changing them to root, but they are too dumb IMO. I'm thinking towards evaluating the present links and echoing a warning for still present nix-darwin entries in user-channels.

LnL7:
It's very much intentional that root channels are not used. With OS updates apple regularly nukes everything under `/root` so IMHO that's a bad place to manage channels on macOS, especially by default.

lockejan:
> It's very much intentional that root channels are not used. With OS updates apple regularly nukes everything under `/root` so IMHO that's a bad place to manage channels on macOS, especially by default.

I see your point, that defaulting to root-channels could possibly gets us in trouble in the future (depending on apple). 
However, the nix-installer creates a root-channel for nixpkgs. 
What would be a better solution then? 
Removing all root-channels including the ones coming from the nix-installer and only using user-channels?
In this case the nix-installer should consequently never create root-channels.
Anyway, the current situation feels inconsistent (to me) and makes it harder for beginners as roughly described in #522 .

lockejan:
I rebased and added the remaining stuff in case we decide to complete the change to root channels.
Don't expect this to be merged soon, but I wanted to get this of my table...

toonn:
@LnL7, the problem with using user channels is that only the user that happened to install nix-darwin will be able to run darwin-rebuild. Other users won't have the channel set up at all.

We either need a single location for the channels, and so far that location has been root's channels, since they are inherited by all users. Or we need to install channels for all users. (Potentially limited to administrator accounts or something.)

Maybe we can set up a LaunchAgent (or LaunchDaemon?) that checks and potentially restores root channels on every boot?

LnL7:
That's true, but the installer already makes assumptions about the current user as the generated/used system configuration is placed under ~/.config of the "admin" user that did the initial installation.

For multi-user context "default" channel makes sense but there's no reason that needs to be located under a home directory, pointing nix directly to a profile with `nix-env -p /nix/var/nix/profiles/default ...` works AFAIK. In comparison fakes does away with this concept entirely making everything user/project bound with defaults essentially hardcoded in nix itself.

toonn:
We could do what NixOS does and put the configuration in `/etc`, that way all admin users can manage the system.

Specifying a profile explicitly may work but it's not very user-friendly, if it's hard-coded in darwin-rebuild how would you specify a different one for a single run? And while Flakes may be a better solution they're still experimental.

I think this merits further consideration.

PR #525, opened 2022-09-10
skhd: add more options

IvarWithoutBones:
This adds more configuration options to the `services.skhd` module, before it only had an option to generate a config file from a string which wasn't the user experience.

Usage now looks something like this:
```nix
  services.skhd = {
    enable = true;

    keybindings = {
      "alt - h" = "yabai -m window --focus west";
      "alt - j" = "yabai -m window --focus south";
      "alt - k" = "yabai -m window --focus north";
      "alt - l" = "yabai -m window --focus east";
    };
  };
```

The `keybindings` are validated against a regex which will complain if the entry is not valid, and because these are defined in an attribute set you also cannot define the same keybinding twice. This validation is very useful because skhd will just not do anything if one of the options in the configuration file is not valid.

### DIFF
--- a/modules/services/skhd/default.nix
+++ b/modules/services/skhd/default.nix
@@ -4,42 +4,101 @@ with lib;
 
 let
   cfg = config.services.skhd;
+
+  validateKeybinding = binding:
+    let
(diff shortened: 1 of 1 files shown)


gshpychka:
@LnL7 can you take a look?

PR #529, opened 2022-09-18
Update nixpkgs-darwin version in README for flakes

motevets:
Using the example as is was previously written lead to the error:

    error: cannot coerce null to a string

       at /nix/store/v05pr9aaswlpxxpilmvmkx7qi44gsrfj-source/pkgs/stdenv/generic/make-derivation.nix:192:34:

          191|         // (lib.optionalAttrs (!(attrs ? name) && attrs ? pname && attrs ? version)) {
          192|           name = "${attrs.pname}-${attrs.version}";
             |                                  ^
          193|         } // (lib.optionalAttrs (stdenv.hostPlatform != stdenv.buildPlatform && !dontAddHostSuffix && (attrs ? name || (attrs ? pname && attrs ? version)))) {

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ A minimal example of using an existing configuration.nix:
   description = "John's darwin system";
 
   inputs = {
-    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-20.09-darwin";
+    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-21.11-darwin";
     darwin.url = "github:lnl7/nix-darwin/master";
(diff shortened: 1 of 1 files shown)


lockejan:
Good catch. However, 21.11 is EOL. See also [nixos-infra-status](https://status.nixos.org/). We'd rather have 22.05 instead. (:

domenkozar:
@motevets could you update to 22.05?

LnL7:
The default locked nixpkgs flake input should probably also be updated, if I'm not mistaken that's what the flake install test will use.

lockejan:
Superseded by https://github.com/LnL7/nix-darwin/pull/568

PR #533, opened 2022-09-20
environment: support installing terminfo files

Enzime:


domenkozar (approved):

lilyball:
Please see #543, this PR breaks existing functionality as it does not preserve the system-supplied terminfo.

PR #534, opened 2022-09-25
Add comma for clarity

zwhitchcox:
It was kind of hard to understand this...it sounds like "out of the box, yet nix-darwin..." so, just adding a comma for clarity

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ darwin.lib.darwinSystem {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


lockejan:
While you are at it, you could also remove the duplicate "to be" (:

zwhitchcox:
Done @lockejan :)

LnL7 (approved):
Thanks, the test is only because the base is outdated.

PR #545, opened 2022-10-12
nix: Remove readOnlyStore option as it has no effect

amarshall:


emilazy (approved):
Thank you for this. It would be nice if we could have a proper read-only store on macOS, but we don't, and given that the NixOS option was moved into `boot.*` there's no reason to keep this around for compatibility. I will happily merge if you rebase on latest `master`.

Enzime (approved):
LGTM

PR #552, opened 2022-10-19
Fix manual by escaping `&quot;` as well

Ma27:
See https://github.com/NixOS/nixpkgs/issues/196651 for context.

__NOTE__: untested because I haven't found a quick way to build the manual without building a nix-darwin configuration, so it would be cool if someone could do a sanity-check before merging!

### DIFF
--- a/doc/manual/options-to-docbook.xsl
+++ b/doc/manual/options-to-docbook.xsl
@@ -19,7 +19,15 @@
       <title>Configuration Options</title>
       <variablelist xml:id="configuration-variable-list">
         <xsl:for-each select="attrs">
-          <xsl:variable name="id" select="concat('opt-', str:replace(str:replace(str:replace(str:replace(attr[@name = 'name']/string/@value, '*', '_'), '&lt;', '_'), '>', '_'), '?', '_'))" />
+        <xsl:variable name="id" select="
+            concat(
+                'opt-',
+                translate(
+                    attr[@name = 'name']/string/@value,
+                    '*&lt; >[]:&quot;',
+                    '________'
+                )
+            )" />
           <varlistentry>
             <term xlink:href="#{$id}">
               <xsl:attribute name="xml:id"><xsl:value-of select="$id"/></xsl:attribute>


joerdav:
This fixed it for me :+1:

vrinek:
I tried the following to install `nix-darwin`:

```
nix-build https://github.com/LnL7/nix-darwin/archive/master.tar.gz -A installer
./result/bin/darwin-installer
# got the "manual-combined.xml fails to validate" error

nix-build https://github.com/Ma27/nix-darwin/archive/fix-manual.tar.gz -A installer
./result/bin/darwin-installer
# worked with no errors
```

Thank you for the fix.

firestack:
This fixed my system build, thank you @Ma27! 

+1 for merging, I'd approve if I understood docbook at all 😅.

domenkozar (approved):

domenkozar:
would be great to run tests on acitons to prevent such failures in the future

Ma27:
@domenkozar the problem is, the breaking change stems from nixpkgs and I'm not a mac user (and wasn't even aware that it has a custom xsl for docs in here), so we'd have to run manual tests against these repos as well when changing the internals of the module system or the documentation system in nixpkgs. I'm not sure how feasible that is considering that this will make the test-matrix even more complex and a workaround for affected users was provided fairly quickly, so I'd consider this kind of an exception.
Also, we'd have to discuss against which projects to check considering that this is not an official NixOS project or in nix-community (though I acknowledge that it's fairly popular).

But still, if someone has a sensible suggestion on how to achieve that, I'd be open to discuss about that.

bestlem:
This issue seems to be back.

I built a new flake from https://github.com/vic/mk-darwin-system

nix flake check gives

`
error: builder for '/nix/store/2cx1w76kj1lpmn99ambkddxbknqmy62k-darwin-manual-combined.drv' failed with exit code 3;
       last 10 log lines:
       > manual-combined.xml:14: element variablelist: Relax-NG validity error : Did not expect element variablelist there
       >     10         <para>Nix modules for darwin.</para>
       >     11   </preface>
       >     12      <chapter xml:id="sec-options">
       >     13   <title>Configuration Options</title>
       >     14    <variablelist xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink" xml:id="configuration-variable-list" xml:base="generated/options-db.xml"><varlistentry><term xlink:href="#opt-documentation.enable" xml:id="opt-documentation.enable"><option>documentation.enable</option></term><listitem><para>Whether to install documentation of packages from
       >     15    <option>environment.systemPackages</option> into the generated system path.</para><para>See "Multiple-output packages" chapter in the nixpkgs manual for more info.
       >     16     </para><para><emphasis>Type:</emphasis> boolean</para><para><emphasis>Default:</emphasis> <literal>
       >
       > manual-combined.xml fails to validate
       For full logs, run 'nix log /nix/store/2cx1w76kj1lpmn99ambkddxbknqmy62k-darwin-manual-combined.drv'.
error: 1 dependencies of derivation '/nix/store/p0b2f0lxf9v4qh4riwizyq8vyclsb0qc-darwin-manpages.drv' failed to build
error: 1 dependencies of derivation '/nix/store/vrg15pap2pyq6vccpd5mhnhazab585i6-darwin-manual-html.drv' failed to build
error: 1 dependencies of derivation '/nix/store/3r4dmmgm67bya7kl6ycw7dswx4hq30mn-system-applications.drv' failed to build
error: 1 dependencies of derivation '/nix/store/gcmm9x5zk9c8vyl9dcwfy60g80rk9x0q-darwin-system-23.05.20230219.a1291d0+darwin4.02d2551.drv' failed to build
`

The error loois like line 15 not replacing " by &quot;  

I tried to search here and in nix and can't find the source

PR #557, opened 2022-10-24
Custom defaults writes

### LINKED ISSUE #517: Custom defaults/plist writes?
From what I can see, there is no way to define custom values to be written with `defaults`.
Though that would be very useful. Some time ago I've been using ansible to provision Macs and created a [long list](https://github.com/7mind/mac-configurator/blob/master/configs/config-settings.yml) of various customisations, not all of them are currently supported by `nix-darwin`.

pshirshov:
Fixes #517

Allows to specify arbitrary options to configure so custom software using `defaults` and system options not yet supported in `nix-darwin` can be configured 

```
  system.defaults = {
    CustomUserPreferences = {
      "NSGlobalDomain" = { "TISRomanSwitchState" = 1; };
      "com.apple.Safari" = {
        "com.apple.Safari.ContentPageGroupIdentifier.WebKit2DeveloperExtrasEnabled" =
          true;
      };
    };
}
```

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -14,6 +14,7 @@
   ./system/defaults/LaunchServices.nix
   ./system/defaults/NSGlobalDomain.nix
   ./system/defaults/GlobalPreferences.nix
+  ./system/defaults/CustomPreferences.nix
   ./system/defaults/dock.nix
   ./system/defaults/finder.nix
   ./system/defaults/screencapture.nix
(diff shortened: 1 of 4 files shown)


pshirshov:
@domenkozar , @n3utrino

domenkozar (approved):

pshirshov:
Even with the approval I can't merge. Could you merge?..

pshirshov:
Awesome, thanks.

dinko-pehar:
@pshirshov Thank you for adding this.

I'm just wondering, what is the difference between `CustomUserPreferences` and `CustomSystemPreferences` ? How are they different when writing the settings using the `defaults` command ? I assume one is for user and one is for general, but not sure 🤷

yacinehmito:
@pshirshov @domenkozar

> I'm just wondering, what is the difference between CustomUserPreferences and CustomSystemPreferences ? How are they different when writing the settings using the defaults command ? I assume one is for user and one is for general, but not sure 🤷

I am wondering the same. I looked at the source and did not see any meaningful differences between the two.

Samasaur1:
Looks like the difference is whether the `defaults` command is run as the user or as root. In practice I'm not exactly sure how the two vary.

PR #566, opened 2022-11-14
feat: add system.screencapture module

bphenriques:
Create the screencapture directory if provided similar to the `users.users.${username}.createHome`. It will do nothing if the folder already exists. Note that if the directory does not exist, screencapture will not work, hence this feature.

Notes:
- Considered asserting if the location is not present but discarded (as we can avoid user intervention).
- Not using `system.defaults.screencapture` as the attributes inside are used to generate a script that sets the keys..

I am using this inside my flakes [here](https://github.com/bphenriques/dotfiles/blob/master/modules/darwin/system/screencapture/default.nix) with `system.defaults.screencapture.location` set to `~/Pictures/screenshots`. As with everything Nix, this is a convenience way to having to create the folder automatically.

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -31,6 +31,7 @@
   ./system/keyboard.nix
   ./system/launchd.nix
   ./system/patches.nix
+  ./system/screencapture.nix
   ./system/shells.nix
   ./system/version.nix
   ./time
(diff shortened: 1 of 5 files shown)


bphenriques:
Using `~` is not great as it creates a folder named `~` where the script is activated. I ended-up using absolute paths.

I am closing this, but we can improve by:
- Verify that the path is absolute
- Create the folder if it does not exist.

I may open a proper pull-request If I resume this.

PR #568, opened 2022-11-17
update flake example in readme and flake.lock

lockejan:
Supersedes #529


lockejan:
I'm not sure why updating the flake.lock resulted in replacing LnL7 with NixOS🤔
Does that look fine to you? @LnL7

domenkozar (approved):

PR #570, opened 2022-11-20
Don't create result links when building with flakes

2xsaiko:
Like when building without flakes, unless running `darwin-rebuild build`, there shouldn't be a result link left over in the current directory after rebuilding. Instead, this changes it to create a temporary directory and place the link in there, similar to nixos-rebuild.


domenkozar:
This should be fixed on master

PR #310, opened 2021-03-23
darwin-rebuild: use --no-link for flakes too

emilazy:
This appears to have been special‐cased because of the `readlink -f result`, but `nix path-info` gives us the same information. (Theoretically there might be an annoying race condition here where Nix refreshes the flake registry between the `build` and `path-info` errors out, but I think the consequences should be fairly harmless. Also not sure if `${extraLockFlags[@]}` should be passed along here.)

PR #578, opened 2022-12-08
Rebuild flake no out link, like #310 but with latest master

saccarosium:
Hi,
this PR is basecly like #310 but with latest master.


emilazy:
FWIW the reason I didn't get around to rebasing the original PR is because I wanted to fix the race condition with garbage collection by keeping the link in a temporary directory like `nixos-rebuild` does. That would be a bit better overall, although of course given how long I procrastinated on doing it and the fact that this race condition already exists when not using flakes this is still an improvement on the status quo by itself.

domenkozar (approved):

PR #585, opened 2022-12-21
Update mopidy service to allow passing extensions

bbenne10:
This allows bundling mopidy with extensions, as is done on nixOS.
(In fact, much of this is cribbed directly from nixpkgs with only small modification)

Additionally, this "purifies" the configuration handling of mopidy.
Rather than reading from `~/.config/mopidy/mopidy.conf`, this will force the configuration into the store.

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -46,6 +46,7 @@
   ./services/buildkite-agent.nix
   ./services/chunkwm.nix
   ./services/cachix-agent.nix
+  ./services/davmail.nix
   ./services/dnsmasq.nix
   ./services/emacs.nix
   ./services/gitlab-runner.nix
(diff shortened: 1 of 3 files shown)


bbenne10:
I have no idea what I've done to cause these tests to recurse that deeply, but this is reproducible on my end. I'll spend some time later trying to figure out where I've gone wrong.

bbenne10:
I think that the recursion was caused my a local open of `pkgs` in my commit. I've removed that and fixed another problem with `lib.mdDoc` not existing. I think that this is ready for review.

bbenne10:
Because these changes have gone unreviewed and I have other changes I'd like to open a PR for (and GitHub does not allow you to update the source branch for a PR), I'm closing this PR and opening a new one with a better source branch. I'll open a new PR with a new branch, but the same changeset.

PR #587, opened 2022-12-23
math-symbols-input: Add package

ryanorendorff:
This module adds the [Math Symbols Input](https://github.com/knrafto/mathsymbolsinput) input method. The input method can then be enabled, with custom inputs, using the following 

```
  services.math-symbols-input = {
    enable = true;
    symbols = {
      "eyeroll" = "🙄"; # insert by typing \eyeroll and it will be replaced with 🙄
      "xd" = "😆";
      "emdash" = "—";
      "zany" = "🤪";
      "shrug" = "🤷";
      "wink" = "😉";
      "oops" = "😅";
      "rofl" = "🤣";
      "check" = "✓";
      "fp" = "🤦";
      "+1" = "👍";
      "lol" = "😂";
      "r" = "→";
      "skull" = "💀";
    };
  };
```

I am unfamiliar with how to write modules for nix darwin, so any help would be greatly appreciated! I have tested this on my system using an `imports` list in my configuration, but I am not sure of the best way to integrate this.

There is an included python program for handling the custom symbols (`generate-commands-plist`) since the symbols are encoded in a plist; python has a plist module in the standard library so that seemed like the simplest way to generate the configuration. I could not otherwise find a good way to access the `plutil` command found on macOS.

### DIFF
--- a/modules/services/math-symbols-input/commands-to-plist.nix
+++ b/modules/services/math-symbols-input/commands-to-plist.nix
@@ -1,0 +1,29 @@
+{ stdenv, callPackage, python3Packages, writers }:
+commands:
+
+let
+
+  custom-commmands-file = callPackage ./custom-commands.nix { };
+
+  generate-commands-plist =
+    python3Packages.callPackage ./generate-commands-plist {
+      inherit (writers) writePython3Bin;
+    };
+  math-symbols-input = callPackage ./package { };
+
+in stdenv.mkDerivation {
+  name = "commands-plist";
+
+  phases = [ "installPhase" ];
+
+  installPhase = ''
+    mkdir -p $out
+
+    # Call the generate_commands_plist python cli to convert the default
+    # list of replacements and a custom list of replacements to a plist file.
+    ${generate-commands-plist}/bin/generate_commands_plist \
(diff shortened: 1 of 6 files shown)


ryanorendorff:
Originally the generated plist file was symlinked into place, but macOS will not refresh its cache of the file. Now it will write the plist using `defaults`.

The current code has some oddities that I'd love some advice on how to fix.

1. The code assumes the user using `~` and `$SUDO_USER`. I am not sure this is a good assumption, I have not seen it elsewhere in the repo.
2. It seems that using `system.activationScripts.preActivation` is the wrong way to implement the installation script based on the comments near the definition of `system.activationScripts.preActivation`. What would be better?
3. This is in the `services` folder, but maybe it would make more sense elsewhere? Maybe an `inputMethods` folder? I have not hooked this module into the rest of the module system yet in case there is a better spot for this module.
4. Adding the input method files does not add it to the users Input Sources in the Keyboard preference pane.

knrafto:
Regarding 4: in the [latest release](https://github.com/knrafto/MathSymbolsInput/releases/tag/v1.3) I've factored out the logic in the installer that enables/selects the input method into a standalone `activate` script, partially so that scripted installs like this are easier.

ryanorendorff:
Bumped to v1.3

PR #588, opened 2022-12-24
Add nix-optimise module

mschuwalow:
Adds a module to run `nix-store --optimise` as a launchdaemon .

### DIFF
--- a/modules/nix/nix-optimise.nix
+++ b/modules/nix/nix-optimise.nix
@@ -11,12 +11,12 @@ in
       enable = mkOption {
         default = false;
         type = types.bool;
-        description = lib.mdDoc "Automatically run the nix store optimiser at a specific time.";
+        description = "Automatically run the nix store optimiser at a specific time.";
       };
       periodSeconds = mkOption {
         default = 60 * 60 * 8;
         type = types.int;
-        description = lib.mdDoc "Interval between runs of the optimizer in seconds.";
+        description = "Interval between runs of the optimizer in seconds.";
       };
     };
   };
--- a/modules/nix/nix-optimise.nix
+++ b/modules/nix/nix-optimise.nix
@@ -1,0 +1,30 @@
+{ config, lib, ...}:
+with lib;
+
+let
+  cfg = config.nix.optimise;
+in
+
+{
+  options = {
+    nix.optimise = {
+      enable = mkOption {
+        default = false;
+        type = types.bool;
+        description = lib.mdDoc "Automatically run the nix store optimiser at a specific time.";
+      };
+      periodSeconds = mkOption {
+        default = 60 * 60 * 8;
+        type = types.int;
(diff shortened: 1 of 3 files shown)


modules/nix/nix-optimise.nix
```diff
@@ -0,0 +1,30 @@
...
+with lib;
+
+let
+  cfg = config.nix.optimise;
+in
+
+{
+  options = {
+    nix.optimise = {
+      enable = mkOption {
```
emilazy, line 11: Please rename this to `automatic` to match [the upstream NixOS module](https://github.com/NixOS/nixpkgs/blob/29f96cc801de8e2efd80f76ca80a389f548d3595/nixos/modules/services/misc/nix-optimise.nix#L10), so that people can share configuration between their nix-darwin and NixOS systems.

modules/nix/nix-optimise.nix
```diff
@@ -0,0 +1,30 @@
...
+        default = false;
+        type = types.bool;
+        description = "Automatically run the nix store optimiser at a specific time.";
+      };
+      periodSeconds = mkOption {
+        default = 60 * 60 * 8;
+        type = types.int;
+        description = "Interval between runs of the optimizer in seconds.";
+      };
```
emilazy, line 20: The upstream NixOS module uses absolute time rather than a recurring period to schedule this. Unless there's some reason to diverge here, we should match that behaviour and its 03:45 time; perhaps rename this to `intervals` (to match `nix.gc.interval`), use it as `StartCalendarInterval` in the launchd daemon, and set the default to `[{ Hour = 3; Minute = 45; }]`. (You can set the type to `types.listOf types.attrs`; it's too much of a pain to try fishing out the exact correct type from the `launchd` options as things stand, unfortunately.)

A potential issue here is that, although launchd will run timers missed due to system sleep on wake, it will not do so if the computer was turned entirely off during the scheduled time. The same issue applies already to `nix.gc.interval` and `StartInterval` in general, though, and using a list allows people to set multiple times per day if they are worried about that. You could add a note in the description about this if you want, but you don't have to.

(Additional, very minor nit: we should say "optimiser" here to match the setting name, as Nix uses standard British English spelling.)

emilazy (changes requested):
Thank you for your contribution! I'm sorry it took over half a year for a review, and I understand if you're not interested in updating this now. However, I'd be happy to merge this with a few changes. You should also be able to re-add the `mdDoc` markers now.

modules/nix/nix-optimise.nix
emilazy: I just saw https://github.com/LnL7/nix-darwin/pull/716 - I'd like to amend my suggestion to naming it `interval` for consistency, and if that PR is merged before this one, it should be easier to specify the correct type.

Enzime:
Superseded by #915

PR #594, opened 2023-01-09
Add option to sort folders first in Finder

al3xtjames:
This change adds com.apple.finder _FXSortFoldersFirst to system.defaults.finder, which keeps folders on top when sorting by name in Finder.


Enzime:
Closing this PR as this option has been added in #1057

PR #595, opened 2023-01-09
sketchybar: init

### LINKED ISSUE #581: Add sketchybar
Sketchybar (https://github.com/FelixKratz/SketchyBar) is another panel for macOS but is more configurable than spacebar. It would be great to have this included, it's already in `nixpkgs`.

azuwis:
Fix #581

### DIFF
--- a/modules/services/sketchybar/default.nix
+++ b/modules/services/sketchybar/default.nix
@@ -1,0 +1,59 @@
+{ config, lib, pkgs, ... }:
+
+let
+  inherit (lib) literalExpression maintainers mdDoc mkEnableOption mkIf mkPackageOptionMD mkOption optionals types;
+
+  cfg = config.services.sketchybar;
+
+  configFile = pkgs.writeScript "sketchybarrc" cfg.config;
+in
+
+{
+
+  meta.maintainers = [
+    maintainers.azuwis or "azuwis"
+  ];
+
+  options.services.sketchybar = {
+    enable = mkEnableOption (mdDoc "sketchybar");
+
+    package = mkPackageOptionMD pkgs "sketchybar" { };
+
+    extraPackages = mkOption {
+      type = types.listOf types.package;
+      default = [ ];
+      example = literalExpression "[ pkgs.jq ]";
+      description = mdDoc ''
+        Extra packages to add to PATH.
+      '';
+    };
+
+    config = mkOption {
+      type = types.lines;
+      default = "";
+      example = ''
+        sketchybar --bar height=24
+        sketchybar --update
+        echo "sketchybar configuration loaded.."
+      '';
+      description = mdDoc ''
+        Contents of sketchybar's configuration file. If empty (the default), the configuration file won't be managed.
+
+        See [documentation](https://felixkratz.github.io/SketchyBar/)
+        and [example](https://github.com/FelixKratz/SketchyBar/blob/master/sketchybarrc).
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    environment.systemPackages = [ cfg.package ];
+
+    launchd.user.agents.sketchybar = {
+      path = [ cfg.package ] ++ cfg.extraPackages ++ [ config.environment.systemPath ];
+      serviceConfig.ProgramArguments = [ "${cfg.package}/bin/sketchybar" ]
+        ++ optionals (cfg.config != "") [ "--config" "${configFile}" ];
+      serviceConfig.KeepAlive = true;
+      serviceConfig.RunAtLoad = true;
+    };
+  };
+}
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -71,6 +71,7 @@
   ./services/postgresql
   ./services/privoxy
   ./services/redis
+  ./services/sketchybar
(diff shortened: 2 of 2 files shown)


bphenriques:
I wish I saw this before! I did something on my setup as well: https://github.com/bphenriques/dotfiles/blob/master/darwin/modules/services/sketchybar/default.nix

In particular, feel free to cherry-pick the part where it adds to `PATH` additional dependencies (e.g., `jq`) and how to debug (the path is hardcoded but if it is fine for me). It is only focused on binaries themselves as I manage the config using `home-manager` (which I may change).

fishhead108:
Bump

emilazy:
Happy to take a look at this, but this will need rebasing for CI to pass and the documentation needs converting to `lib.mdDoc` (https://nixos.org/manual/nixpkgs/unstable/#sec-contributing-markup) at least.

modules/services/sketchybar/default.nix
```diff
@@ -0,0 +1,66 @@
...
+    environment.systemPackages = [ cfg.package ];
+
+    launchd.user.agents.sketchybar = {
+      path = [ cfg.package ] ++ cfg.extraPackages ++ [ config.environment.systemPath ];
+      serviceConfig.ProgramArguments = [ "${cfg.package}/bin/sketchybar" ];
+      serviceConfig.KeepAlive = true;
+      serviceConfig.RunAtLoad = true;
+      serviceConfig.EnvironmentVariables = {
+        XDG_CONFIG_HOME = mkIf (cfg.config != "") "${configHome}";
+      };
```
Enzime, line 63: This feels like it should live in a `home-manager` module

azuwis: Last time I checked, sketchybar does not provide command line arg to change config file path, but use XDG_CONFIG_HOME env.

Now it have `--config` option, I'll update the module.

modules/services/sketchybar/default.nix
```diff
@@ -0,0 +1,59 @@
...
+    config = mkOption {
+      type = types.lines;
+      default = "";
+      example = literalExpression ''
+        sketchybar --bar height=24
+        sketchybar --update
+        echo "sketchybar configuration loaded.."
+      '';
+      description = mdDoc ''
+        Content of configuration file. If empty(default), configuration file won't be managed.
```
Enzime, line 40:
```suggestion
        Contents of sketchybar's configuration file. If empty (the default), the configuration file won't be managed.
```

Enzime (changes requested):

modules/services/sketchybar/default.nix
```diff
@@ -0,0 +1,59 @@
...
+      description = mdDoc ''
+        Extra packages to add to PATH.
+      '';
+      example = literalExpression "[ pkgs.jq ]";
+    };
+
+    config = mkOption {
+      type = types.lines;
+      default = "";
+      example = literalExpression ''
```
emilazy, line 34: `literalExpression` is wrong here, should just be a plain string

modules/services/sketchybar/default.nix
```diff
@@ -0,0 +1,59 @@
...
+
+  config = mkIf cfg.enable {
+    environment.systemPackages = [ cfg.package ];
+
+    launchd.user.agents.sketchybar = {
+      path = [ cfg.package ] ++ cfg.extraPackages ++ [ config.environment.systemPath ];
+      serviceConfig.ProgramArguments = [ "${cfg.package}/bin/sketchybar" ]
+        ++ optionals (cfg.config != "") [ "--config" "${configFile}" ];
+      serviceConfig.KeepAlive = true;
+      serviceConfig.RunAtLoad = true;
```
bphenriques, line 56: We'll need to set `EnvironmentVariables`.`PATH` so that we can include `jq` and other binaries being used in the config. 

See:
- https://github.com/bphenriques/dotfiles/blob/master/darwin/modules/services/sketchybar/default.nix#L56C9-L56C29 
- Usage: https://github.com/bphenriques/dotfiles/blob/master/darwin/default.nix#L86

bphenriques: Example of a sketchybar config that requires additional binaries in the `PATH`: https://github.com/bphenriques/dotfiles/blob/master/home/config/sketchybar/config/plugins/bluetooth.sh#L5

The plugin won't work without `jq` in the `PATH`

Enzime: The `PATH` environment variable gets set here already: https://github.com/LnL7/nix-darwin/pull/595/files#diff-c768f292b1281301e751d244aedf4a9a1bf53b978f81a4a914989bdc042244c1R52

bphenriques: Nice 👍 I very much prefer that approach

modules/services/sketchybar/default.nix
azuwis: Done in latest push.

modules/services/sketchybar/default.nix
azuwis: Done in latest push.

Enzime (approved):
Thanks for your contribution, LGTM

PR #599, opened 2023-01-21
Requires a list, update the example to match

n8henrie:


emilazy:
Thank you for the contribution! This is definitely an improvement over the status quo but since it has conflicts with `master` and https://github.com/LnL7/nix-darwin/pull/716 would make this default valid, which matches what launchd accepts better, I'm going to close it in favour of that.

PR #600, opened 2023-01-21
fix: add '://' to built flake uri

### LINKED ISSUE #597: Latest commit in darwin-rebuild.sh broke local git repo reference for flakes
I think there is a regression caused by: 87b9d090ad39b25b2400029c64825fc2a8868943
It's related to `darwin-rebuild switch` command called with a flake argument, where flake refers to a local git repo by an absolute path.

What worked before:
```bash
$ darwin-rebuild switch --flake "git+file://$HOME/Code/github.com/konradmalik/dotfiles#$(hostname)"
```

Now returns an error indicating that there something weird going on with protocol/filepath/scheme parsing:
```bash
path '/Users/konrad/Code/github.com/konradmalik/dotfiles/git+file/Users/konrad/Code/github.com/konradmalik/dotfiles' does not contain a 'flake.nix', searching up
error: getting status of '/Users/konrad/Code/github.com/konradmalik/dotfiles/git+file/Users/konrad/Code/github.com/konradmalik/dotfiles': No such file or directory
```
Of course that same syntax works well with both `nixos-rebuild` and `home-manager rebuild`.

This works fine though:
```bash
$ darwin-rebuild switch --flake ".#$(hostname)"
```

konradmalik:
Fixes #597. Tested on my mac.

Problem - from that same RFC that is quoted:
```
For example, matching the above expression to

      http://www.ics.uci.edu/pub/ietf/uri/#Related

   results in the following subexpression matches:

      $1 = http:
      $2 = http
      $3 = //www.ics.uci.edu
      $4 = www.ics.uci.edu
      $5 = /pub/ietf/uri/
      $6 = <undefined>
      $7 = <undefined>
      $8 = #Related
      $9 = Related
```
and when building the URI, we concatenated:
```
flake=${group 2}${group 4}${group 5}${group 6}
```
Which results (for the URL example in question) in:
```
flake=httpwww.ics.uci.edu/pub/ietf/url/
```

Thus, this MR adds the missing `://`. 

I could've also just used group 1 instead of 2 and group 3 instead of 4, but I think the current way is more clear and readable and there are no problems with naming (how to name `scheme` with `:` included? and how to name `authority` with `//` included?).

EDIT: Actually we need to use groups instead of hardcoding `://` due to local relative paths.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -114,3 +114,7 @@ jobs:
         darwin-rebuild build \
           --flake /tmp/test-nix-darwin-submodules?submodules=1#simple \
           --override-input darwin .
+        # Should also succeed
+        darwin-rebuild build \
+          --flake git+file:///tmp/test-nix-darwin-submodules?submodules=1#simple \
+          --override-input darwin .
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -123,13 +123,13 @@ flakeFlags=(--extra-experimental-features 'nix-command flakes')
 if [ -n "$flake" ]; then
     # Offical regex from https://www.rfc-editor.org/rfc/rfc3986#appendix-B
     if [[ "${flake}" =~ ^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))? ]]; then
-       scheme=${BASH_REMATCH[2]}
-       authority=${BASH_REMATCH[4]}
-       path=${BASH_REMATCH[5]}
+       scheme=${BASH_REMATCH[2]} # eg. http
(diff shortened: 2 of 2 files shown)


brettmilford:
This fixes #597 for me when using the scheme `git+https://`.

reckenrode:
This also fixes #628 for me. I am able to build my flake at `github:reckenrode/nixos-configs`.

n8henrie:
Thank you!

konradmalik:
@domenkozar - mentioning you since you seem to be active on this repo recently regarding merging.
Can we merge this?
Ps. The test above needs to be rerun, seems like it crashed due to a transient network error. I don't think I can rerun it myself.

domenkozar:
Could you rebase?

konradmalik:
> Could you rebase?

done

domenkozar (approved):

PR #601, opened 2023-01-21
ssh: add extraHostNames option

montchr:
https://github.com/NixOS/nixpkgs/blob/cdead16a444a3e5de7bc9b0af8e198b11bb01804/nixos/modules/programs/ssh.nix

### DIFF
--- a/modules/programs/ssh/default.nix
+++ b/modules/programs/ssh/default.nix
@@ -7,6 +7,12 @@ let
 
   knownHosts = (attrValues cfg.knownHosts);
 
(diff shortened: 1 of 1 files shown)


luizribeiro:
It would be great to have this landed for my configs. I took a look at your code and it all looks reasonable since it's mostly the code from upstream. @domenkozar, would this be something you could help us get approve?

emilazy (approved):
Thank you for this great change. I'm happy to merge this if you rebase on `master` (you may need to translate some configuration to Markdown; you can copy upstream from NixOS in that case). Adding it to the test would be great too, but I won't hold this up for that.

PR #604, opened 2023-02-03
services.github-runners: init

DavHau:
The runners work fine, but there is still work left to do in order to make the runners secure, therefore opening this as a draft.
I do not intent to continue working on this, as I went with a remote builder based setup instead by now.

Whoever is interested in moving this forward, please feel free to fork and take it on from here.

In order to make github runners secure for use with public github projects, two things are important:

### 1. The github runner user must not be able to read the `tokenFile` and copies of it made during the setup
...as this would allow an attacker to open a PR, changing the workflow logic and extract the authentication token.

By porting the module from nixos (systemd) to nix-darwin, several mechanism which protected the tokenFile were removed:
  - [adding the token file and the copy of it to `InaccessiblePaths`](https://github.com/NixOS/nixpkgs/blob/d8cf9821d6b648ec11bba9f1484631a7cb5d344e/nixos/modules/services/continuous-integration/github-runner/service.nix#L194-L199)
  - running the setup logic as root via `ExecStartPre`, while the `ExecStart` will run as unprivileged user

### 2. It must be impossible to persist state between runs
On public projects, github protects workflow secrets from outside contributors by not supplying PR runs with secrets.
But if an attacker manages to persist processes or files between runs, it becomes possible to influence future runs on that same machine to reveal secrets.

The only secure mode of operation is if the entire state is wiped between runs. Without namespaces this is hard to guarantee, but at least we should try to come as close as possible to that ideal behavior.

The systemd version of that service created temporary users and directories for each run. I'm not sure if we can to something similar in darwin.

It seems to me that the safest mode of operation on a macos would be to utilize virtualization tools like tart. I stitched together a quite hacky version of the github-runners module utilizing tart here: https://github.com/holochain/holochain-infra/tree/3ad70877dc0e0ba69900375f6241284142bc4d2f/modules/modules.darwin.github-runners-tart
Though there is no tart module on nix-darwin, therefore this setup isn't entirely declarative

### Things to do
In my opinion, those things still need to be worked on:
- [ ] Generate individual users for each runner (currently they all share the `github-runner` user)
- [ ] Ensure that the runner users doesn't have access to the token (do setup in separate root service?)

@domenkozar pinging you as you indicated interest in this

### DIFF
--- a/modules/services/github-runners/default.nix
+++ b/modules/services/github-runners/default.nix
@@ -1,0 +1,58 @@
+{ config
+, pkgs
+, lib
+, ...
+}@args:
+
+with lib;
+
+let
+  cfg = config.services.github-runners;
+
+in
+
+{
+  options.services.github-runners = mkOption {
+    default = {};
+    type = with types; attrsOf (submodule { options = import ./github-runner/options.nix (args // {
+      # services.github-runners.${name}.name doesn't have a default; it falls back to ${name} below.
+      includeNameDefault = false;
+    }); });
+    example = {
+      runner1 = {
+        enable = true;
+        url = "https://github.com/owner/repo";
+        name = "runner1";
+        tokenFile = "/secrets/token1";
+      };
+
+      runner2 = {
+        enable = true;
+        url = "https://github.com/owner/repo";
+        name = "runner2";
+        tokenFile = "/secrets/token2";
+      };
+    };
+    description = lib.mdDoc ''
+      Multiple GitHub Runners.
+    '';
+  };
+
+  config = {
+    systemd.services = flip mapAttrs' cfg (n: v:
+      let
+        svcName = "github-runner-${n}";
+      in
+        nameValuePair svcName
+        (import ./github-runner/service.nix (args // {
+          inherit svcName;
+          cfg = v // {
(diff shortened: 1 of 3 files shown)


Enzime:
@DavHau how about adding an option that the user understands that this is insecure and should not be run on public repositories, as an alternative to securing this service before it gets merged? I would be interested in running this in a VM and am willing to accept the risks

modules/services/github-runners/default.nix
```diff
@@ -0,0 +1,69 @@
...
+    description = lib.mdDoc ''
+      GitHub Actions runners.
+
+      Note: GitHub recommends using self-hosted runners with private repositories only. Learn more here:
+      [About self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners).
+    '';
+  };
+
+  config = {
+    # create a launchd service for each gtihub runner
```
domenkozar, line 40:
```suggestion
    # create a launchd service for each github runner
```

will:
I've been using this for a couple of months now and it works great, and has been very useful. I'd be in favor of merging as-is for what my opinion is worth

Enzime:
Superseded by #859

PR #605, opened 2023-02-07
feat(system.defaults.dock): add springboard-columns and springboard-rows

nouun:
Add `springboard-columns` and `springboard-rows` to set Launchpad columns and rows.

Currently it requires rebooting or running `killall Dock`; what would be the best way to proceed before marking this as ready? Should I add a note in the description, add a check if the option is enabled and append `killall Dock` to the activation script, or just do nothing?

### DIFF
--- a/modules/system/defaults/dock.nix
+++ b/modules/system/defaults/dock.nix
@@ -255,5 +255,22 @@ in {
       '';
     };
 
+    system.defaults.dock.springboard-columns = mkOption {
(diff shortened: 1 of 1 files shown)


nouun:
As there are other options which also require `killall Dock` and this hasn't been implemented already I will go ahead and mark this as ready.

nouun:
I've just pushed a change which will automatically run `killall Dock` if there are any dock defaults set as most of them require the dock to be restarted.

PR #606, opened 2023-02-14
security.pam: Add pam_reattach support

sestrella:
This PR enables sudo authentication via Touch ID inside tmux as it's been described in the following [article](https://www.hein.dev/blog/2020/01/using-touchid-tmux-pam_reattach/).


maljub01:
I believe that with #612 we should be able to switch from using `sed` to using `system.patches` for managing `/etc/pam.d/sudo`, which would be a much cleaner and nicer approach.

This PR could then simply use something like this:

```
system.patches = [
  (let
    enablePamReattach = cfg.enableSudoTouchIdAuth && cfg.enablePamReattach;
    newLineCount = 5 + (if enablePamReattach then 1 else 0);
  in
  pkgs.writeText "pam.patch" (''
    --- a/etc/pam.d/sudo
    +++ b/etc/pam.d/sudo
    @@ -1,4 +1,${newLineCount} @@
     # sudo: auth account password session
  '' + (if enablePamReattach then ''
    +auth       optional       ${pkgs.pam-reattach}/lib/pam/pam_reattach.so
  '' else "") + ''
    +auth       sufficient     pam_tid.so
     auth       sufficient     pam_smartcard.so
     auth       required       pam_opendirectory.so
     account    required       pam_permit.so
  ''))
];
```

modules/security/pam.nix
```diff
@@ -15,48 +15,86 @@ let
...
+      enableTmuxTouchIdSupport = mkEnableOption ''
+        Enable sudo authentication with Touch ID inside tmux
+
+        When enabled, this option adds the following line to /etc/pam.d/sudo:
+
+            auth       optional     pam_reattach.so
+
+        (Note that macOS resets this file when doing a system update. As such,
+        sudo authentication with Touch ID inside tmux won't work after a system
+        update until the nix-darwin configuration is reapplied.)
```
maljub01, line 81: While this is primarily useful for tmux (and therefore I agree that tmux support should be mentioned in the description/documentation), I don't think the option itself should have "tmux" in the name:

```suggestion
      enablePamReattach = mkEnableOption ''
        Enable sudo authentication with Touch ID within programs running in the background session by reattaching to the authenticating user's per-session bootstrap namespace.
        
        This allows users to make use of the pam_tid module (Touch ID) from within tmux.

        When enabled, this option adds the following line to /etc/pam.d/sudo:

            auth       optional     pam_reattach.so

        (Note that macOS resets this file when doing a system update. As such,
        sudo authentication with Touch ID inside tmux won't work after a system
        update until the nix-darwin configuration is reapplied.)
```

sestrella:
@maljub01 thank you for the code review, I'm switching this PR back to draft in order to try it out with the latest changes in #612

vamega:
@maljub01 I ended up trying your suggestion now that #612 has been merged.
That didn't work exactly, since the multiline strings end up screwing with the patch formatting.

I ended up with this:

```
config =
let
  anyEnabled = cfg.enableSudoTouchIdAuth || cfg.enablePamReattach;
in
{
  environment.systemPackages = optional anyEnabled pkgs.pam-reattach;

  environment.pathsToLink = optional anyEnabled "/lib/pam";

  system.patches =
    optional anyEnabled
      (
        let
          newLineCount = toString (
            4 + (count (x: x) [ cfg.enableSudoTouchIdAuth cfg.enablePamReattach ])
          );
          reattachLine = (
            if cfg.enablePamReattach
            then "\n+auth       optional       ${pkgs.pam-reattach}/lib/pam/pam_reattach.so"
            else ""
          );
          touchIdLine = (
            if cfg.enableSudoTouchIdAuth
            then "\n+auth       sufficient     pam_tid.so"
            else ""
          );
        in
        pkgs.writeText "pam.patch" (
          ''
            --- a/etc/pam.d/sudo
            +++ b/etc/pam.d/sudo
            @@ -1,4 +1,${newLineCount} @@
             # sudo: auth account password session${reattachLine}${touchIdLine}
             auth       sufficient     pam_smartcard.so
             auth       required       pam_opendirectory.so
             account    required       pam_permit.so
          ''
        )
      );
};
```

I've pushed that to my fork. You can see it at this commit: https://github.com/vamega/nix-darwin/commit/cc493f3628386336e46c2b60c79319604d26b0bd

On looking at it further, I think removing the complexity of the string interpolation for creating the patches makes it easier to read and reduces the needs for the test.

With straightforward patches, then the tests are simply exercising the `system.patches` functionality.

vamega:
@sestrella Why did you add the package to the `systemPackages` and add to the `pathsToLink`.
Things seem to work fine without those changes from what I can see.

maljub01:
@vamega I agree. The code you shared looks much better and is more readable than what I had.

sestrella:
@vamega I don't recall the main reason, but I believe I was attempting to avoid locking myself out of sudo during a 'pam-reattach' upgrade; however, pointing directly to the store shouldn't be an issue because different versions of 'pam-reattach' should be able to coexist. BTW, I believe your solution is way more complete than the one presented here, so if you want to submit a PR I would be more than happy to close this one

vamega:
Sure I'll submit a CR, didn't want to step on any toes since I mostly combined your work and @maljub01.

PR #611, opened 2023-03-02
feat(nix-gc): support multiple intervals

steveej:
this is useful to run the GC more than once per hour.

cc @DavHau

### DIFF
--- a/modules/services/nix-gc/default.nix
+++ b/modules/services/nix-gc/default.nix
@@ -40,6 +40,11 @@ in
         description = "The time interval at which the garbage collector will run.";
       };
 
+      intervals = mkOption {
+        type = types.listOf types.attrs;
+        description = "Multiple intervals at which the garbage collector will run.";
+      };
+
       options = mkOption {
         default = "";
         example = "--max-freed $((64 * 1024**3))";
@@ -63,7 +68,7 @@ in
       command = "${config.nix.package}/bin/nix-collect-garbage ${cfg.options}";
       environment.NIX_REMOTE = optionalString config.nix.useDaemon "daemon";
       serviceConfig.RunAtLoad = false;
-      serviceConfig.StartCalendarInterval = [ cfg.interval ];
+      serviceConfig.StartCalendarInterval = cfg.intervals or [ cfg.interval ];
       serviceConfig.UserName = cfg.user;
     };
 
--- a/modules/services/nix-gc/default.nix
+++ b/modules/services/nix-gc/default.nix
@@ -42,6 +42,7 @@ in
 
       intervals = mkOption {
         type = types.listOf types.attrs;
+        default = [cfg.interval];
         description = "Multiple intervals at which the garbage collector will run.";
       };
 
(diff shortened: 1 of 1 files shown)


modules/services/nix-gc/default.nix
```diff
@@ -40,6 +40,11 @@ in
         description = "The time interval at which the garbage collector will run.";
       };
 
+      intervals = mkOption {
+        type = types.listOf types.attrs;
+        description = "Multiple intervals at which the garbage collector will run.";
+      };
```
DavHau, line 46:
```suggestion
      intervals = mkOption {
        type = types.listOf types.attrs;
        default = [cfg.interval];
        description = "Multiple intervals at which the garbage collector will run.";
      };
```

modules/services/nix-gc/default.nix
```diff
@@ -63,7 +68,7 @@ in
       command = "${config.nix.package}/bin/nix-collect-garbage ${cfg.options}";
       environment.NIX_REMOTE = optionalString config.nix.useDaemon "daemon";
       serviceConfig.RunAtLoad = false;
-      serviceConfig.StartCalendarInterval = [ cfg.interval ];
+      serviceConfig.StartCalendarInterval = cfg.intervals or [ cfg.interval ];
```
DavHau, line 71:
```suggestion
      serviceConfig.StartCalendarInterval = cfg.intervals;
```

modules/services/nix-gc/default.nix
DavHau: That should fix the failing test

DavHau (approved):

modules/services/nix-gc/default.nix
```diff
@@ -40,6 +40,12 @@ in
         description = "The time interval at which the garbage collector will run.";
       };
 
+      intervals = mkOption {
+        type = types.listOf types.attrs;
+        default = [cfg.interval];
```
LnL7, line 45: I don't remember exactly what the behaviour for this is but I seem to recall that defaults is generally only used for static values and a dynamic value that dynamic default values are defined using `mkDefault` in the config so the priority of the value is lowered and gets replaced when the user specifies a value as expected.

```nix
{
  options = {
    # ...
    intervals = mkOption {
      type = types.listOf types.attrs;
      default = [];
    };
  };
  config = mkIf cfg.automatic {
    # ...
    intervals = mkDefault [cfg.interval];
  };
}
```

steveeJ: i just noticed that i haven't pushed a local change to the PR branch. seems like i squashed too so i opted for force-pushing. the behavior i had locally is equivalent to what you've described. PTAL

emilazy:
Thank you for this. I think that https://github.com/LnL7/nix-darwin/pull/716 achieves the same end goal without breaking backwards compatibility, so I'm going to close this in favour of that PR. (Sorry; I know it's frustrating to have a PR in limbo get closed like this.)

PR #612, opened 2023-03-07
Fix system.patches

maljub01:
Currently, `system.patches` doesn't work because it will attempt to first detect if the patch has already been applied by checking if it can be applied in reverse.

However, when that happens, `patch` detects that the supplied patch is incorrectly reversed and attempts to ask the user if they want to "Ignore -R":

```
Unreversed (or previously applied) patch detected!  Ignore -R? [y]
```

Because the output is piped to `/dev/null` the user will basically see nothing and `darwin-rebuild switch` will hang until the user presses "Enter" (possibly to check if the terminal is frozen). At which point, patch will ignore the --reverse and exit successfully, preventing the patch from being applied at all.

This change fixes that bug by using `--force` when running the `patch` command with `--reverse`. This tells patch that we know what we're doing and prevents it from prompting the user if they want to ignore the `--reverse` argument. Note that this is unnecessary when using `--forward` because that already instructs `patch` to ignore patches it thinks are reversed or already applied.

### DIFF
--- a/modules/system/patches.nix
+++ b/modules/system/patches.nix
@@ -56,14 +56,14 @@ in
 
       for f in $(ls /run/current-system/patches 2> /dev/null); do
           if test ! -e "${config.system.build.patches}/patches/$f"; then
-              patch --reverse --backup -d / -p1 < "/run/current-system/patches/$f" || true
+              patch --force --reverse --backup -d / -p1 < "/run/current-system/patches/$f" || true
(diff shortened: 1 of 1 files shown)


LnL7 (approved):
Thanks! I also noticed this recently, but I think this used to work so I wonder since this is activation if `patch` from macOS is used and something changed there.

PR #616, opened 2023-03-14
feat: support writing arrays to system defaults

### LINKED ISSUE #590: Writing arrays to system defaults
Currently, the defaults keys with array (and dict) types cannot be written with `system.defaults.CustomUserPreferences` per https://github.com/LnL7/nix-darwin/blob/3db1d870b04b13411f56ab1a50cd32b001f56433/modules/system/defaults-write.nix#L11:L15

Would it make sense to implement a way to write arrays as well? Sound simple enough, or am I missing something? I might be able to throw together a PR.

gshpychka:
Fixes #590


LnL7 (approved):

LnL7:
Thanks!

PR #624, opened 2023-03-28
fix(modules/fonts): ignore repeated fonts

lovesegfault:
Without this, attempts to build the fonts dir with repeated font
packages in `fonts.fonts` will yield:

```
ln: failed to create symbolic link '/nix/store/6im9rm87nxc82nqbv350hfp2w7ja1z47-fonts/Library/Fonts/IBMPlexSansThai-Thin.otf': File exists
```


vzctl:
Encountered the same problem. Tested the fix - works great, thanks for the PR!

domenkozar (approved):

PR #189, opened 2020-04-12
Add support for configuring system shortcuts

kirelagin:
So, basically, this is a module that, once enabled, will start managing system keyboard shortcuts. I had to make a choice of what to do with the shortcuts that are unset in the config and the options I was considering were: do not touch, disable, reset to system default. I went with the latter one, but, I guess, this is open for discussion.

Naturally, the activation script writes directly to the plist file, so for the changes to get picked up one needs to re-login. (Note that using `defaults` would not help either.) I tried all sane options, but could not find a workaround. _I guess I’m writing an XPC sniffer now..._

Lastly, finding out the ids of the shortcuts is very boring, so I implemented only the ones that I care about. Help with this is appreciated.

Resolves #185.

PR #636, opened 2023-04-19
Add support for configuring system shortcuts

### LINKED ISSUE #185: Manage keyboard shortcuts
Would it be a good idea to add a module for configuring system (or not only system) keyboard shortcuts, such as switching keyboard input language?

I figured out how to do it, so I may contribute it, if there is interest.

theutz:
This just rebases the content from #189 on `master` so it can be merged. @domenkozar @kirelagin 

Resolves #185 
Resolves #189

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -33,6 +33,7 @@
   ./system/launchd.nix
   ./system/patches.nix
   ./system/shells.nix
+  ./system/shortcuts.nix
   ./system/version.nix
   ./time
   ./networking
--- a/modules/system/activation-scripts.nix
+++ b/modules/system/activation-scripts.nix
@@ -108,6 +108,8 @@ in
       ${cfg.activationScripts.userLaunchd.text}
       ${cfg.activationScripts.homebrew.text}
 
+      ${cfg.activationScripts.shortcuts.text}
+
       ${cfg.activationScripts.postUserActivation.text}
 
       exit $_status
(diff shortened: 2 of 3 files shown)


jmatsushita:
Hey @domenkozar any chance we could merge this? https://github.com/LnL7/nix-darwin/pull/189#issuecomment-1354741186

jmatsushita:
I got a deprecation warning when using the PR branch:
```
trace: warning: literalExample is deprecated, use literalExpression instead, or use literalDocBook for a non-Nix description.
```

jmatsushita:
After testing this, I'm not sure it's ready for merging...  I'm trying this:
```nix
  system.keyboard.shortcuts = {
    enable = true;
    spotlight.search.enable = false;
  };
```
But unfortunately, this doesn't seem to work for me. I do see the `/Users/jun/Library/Preferences/com.apple.symbolichotkeys.plist` XML file with the correct values, but even after logging out, the shortcut settings aren't enabled, and the file stays uncompressed.

If I compress it with `plutil -convert binary1 ~/Library/Preferences/com.apple.symbolichotkeys.plist` and use the activation trick [here](https://zameermanji.com/blog/2021/6/8/applying-com-apple-symbolichotkeys-changes-instantaneously/) it also doesn't work. Whereas the approach described to use `defaults write com.apple.symbolichotkeys.plist ...` followed by the `activateSettings -u` works flawlessly.

UPDATE: after diffing the XML generated by the 2 methods, I see that setting `enable` to false is not sufficient, and the first parameter of the array needs to be set somteimes to an ascii code, and sometimes to `65535`.

PR #638, opened 2023-04-23
feat(screensaver): add support of askForPassword and askForPasswordDelay

aerialls:
Add support of two new macOS parameters.

- com.apple.screensaver.askForPassword
- com.apple.screensaver.askForPasswordDelay

https://developer.apple.com/documentation/devicemanagement/screensaver


aerialls:
Not sure why the `install-flake` test is falling. I'll be happy to fix it if I can have some hints if it's linked or not and how to fix.

LnL7 (approved):
Thanks!

PR #641, opened 2023-04-28
Update ci and introduce matrix builds against stable and unstable

lockejan:
This PR introduces matrix builds to run tests against latest stable and unstable channel.
While adding matrix build I've also changed all remaining references from 22.05 to 22.11.
This way we can see at first glance if a PR would break on stable or unstable right away.
I've also updated the referenced actions in the test-workflow to the most recent version.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
   build:
     runs-on: macos-12
     steps:
-    - uses: actions/checkout@v2
-    - uses: cachix/install-nix-action@v17
+    - uses: actions/checkout@v3
+    - uses: cachix/install-nix-action@v20
     - run: |
         nix build ${{ github.event.client_payload.args }} -vL
--- a/.github/workflows/debug.yml
+++ b/.github/workflows/debug.yml
@@ -12,8 +12,8 @@ jobs:
   debug:
     runs-on: macos-12
     steps:
-    - uses: actions/checkout@v2
-    - uses: cachix/install-nix-action@v17
+    - uses: actions/checkout@v3
+    - uses: cachix/install-nix-action@v20
     - run: |
         nix-channel --add https://nixos.org/channels/nixpkgs-unstable nixpkgs
         nix-channel --update
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,33 +2,74 @@ name: "Test"
 on:
   pull_request:
   push:
+
 jobs:
   tests:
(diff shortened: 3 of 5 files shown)


lockejan:
@domenkozar does that makes sense to you? (:

domenkozar:
For this one we need @LnL7 since required checks will need to be changed

domenkozar (approved):

LnL7:
I didn't use unstable in the tests because it used to happen sometimes that the channel updates while only linux builds are cached. As well as the general fact that darwin packages are a bit less reliable. In those cases there's nothing that can be done in this project other than wait for a channel bump which is rather annoying.

Is there a way to set this up so these are extra builds that don't affect the overall pr status for example?

LnL7:
Nothing super specific, but things like the manpages build pull in packages that can be broader than you might expect.
I haven't really kept up with the state of unstable and hydra infrastructure recently. If you think it's fine I set both to required and we can see how things go.

lockejan:
> Is there a way to set this up so these are extra builds that don't affect the overall pr status for example?

We could register on a different event with the unstable tests if that's better, [when a PR has been merged](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges) for example.

lockejan:
> Hmm, won't that suggestion completely hide those jobs?

At least on the PR, yes. But they would be indicated on the commit in master after the merge has happened, but I see now that's not what we want here.

> Ideally they would show up in the checks below, but not affect the green checkmark on the list of pull requests. But if that's not possible I'm fine with leaving it as is.

We have two options here, to my understanding, as this isn't a built-in feature for matrix jobs yet:
1. We split stable and unstable in different jobs. However, this will come at the cost of having to duplicate all steps.
2. Manually implementing logic so that unstable checks don't affect the overall status, by manipulating `set-output`. 

With the first approach you wouldn't have to change anything as the new job isn't part of the required checks.
But the code duplication bugs me, so I'd favour the second approach, but it would take more time to implement this.
I'll try to hack something together this weekend.

lockejan:
I ditched matrix-jobs and introduced single jobs instead. 
To better distinguish each one, I slightly changed their names.
Not sure if the required check will then updated accordingly in repo settings.
Anyway, you should be able to only mark the checks/jobs against stable as required now, I guess. @LnL7 

Sidenote: the job naming could probably be improved.

LnL7 (approved):
Thanks!

PR #643, opened 2023-04-30
feat: defaults write com.apple.menuextra.clock

hraban:
Expose menubar clock preferences through nix-darwin.

Works locally (apple silicon, Mac 13.3.1).

## Questions

1. Does anyone know what the difference is between
```
defaults write com.apple.menuextra.clock ShowDate -int 1
```
and
```
defaults write com.apple.menuextra.clock ShowDate -int 2
```
?
2. Is this the expected way to organise the nix-darwin namespace (`system.defaults.clock.xyz`) vs the actual namespace (`com.apple.menuextra.clock.xyz`)?

Edit: Technically, from the name "clock" it's not clear that this is about the clock in the menubar. That's an argument against the hierarchy I chose.

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -16,6 +16,7 @@
   ./system/defaults/NSGlobalDomain.nix
   ./system/defaults/GlobalPreferences.nix
   ./system/defaults/CustomPreferences.nix
+  ./system/defaults/clock.nix
   ./system/defaults/dock.nix
   ./system/defaults/finder.nix
   ./system/defaults/screencapture.nix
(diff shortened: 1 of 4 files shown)


hraban:
`menuExtraClock` sounds good.

Re the other point: I like the idea of still at least exposing the lower level defaults write operations. When apple changes their defaults write schema, you might otherwise end up having to bend over backwards to provide a translation layer from the old API. See e.g. https://github.com/tech-otaku/menu-bar-clock/blob/main/menu-bar-clock.sh which is a valiant effort for the menu clock's api change in the latest mac os.

LnL7 (approved):
Thanks

Issue #644, opened 2023-05-01
Unable to install nix-darwin on aarch64-darwin, following the flake instructions


bryanhonof:
After running `./result/sw/bin/darwin-rebuild switch --flake .` I get the following output.
```shell
bryanhonof@Bryans-MBP meta % ./result/sw/bin/darwin-rebuild switch --flake .
warning: unknown setting 'bash-prompt-prefix'
warning: Git tree '/Users/bryanhonof/Desktop/meta' is dirty
building the system configuration...
warning: unknown setting 'bash-prompt-prefix'
warning: Git tree '/Users/bryanhonof/Desktop/meta' is dirty
warning: unknown setting 'bash-prompt-prefix'
user defaults...
setting up user launchd services...
shell-init: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
setting up /Applications/Nix Apps...
job-working-directory: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
setting up pam...
job-working-directory: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
applying patches...
job-working-directory: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
setting up /etc...
job-working-directory: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
job-working-directory: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
error: not linking environment.etc."bashrc" because /etc/bashrc already exists, skipping...
existing file has unknown content 6757a6969501994e2bf161ffa39a934f27f85664c901cdc0c4fcd8b2be7dba23, move and activate again to apply
job-working-directory: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
error: not linking environment.etc."nix/nix.conf" because /etc/nix/nix.conf already exists, skipping...
existing file has unknown content f83c4a7e56c00dc9be977b00ee051432b37d179b2247bb742ee3423159d99353, move and activate again to apply
job-working-directory: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
job-working-directory: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
job-working-directory: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
job-working-directory: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
job-working-directory: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
[...]

PR #645, opened 2023-05-01
Update flake install section (root)

bryanhonof:
See #644 for rationale.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,12 +106,17 @@ darwin.lib.darwinSystem {
 ```
 
 Since the installer doesn't work with flakes out of the box yet, nix-darwin will need to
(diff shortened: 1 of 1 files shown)


LnL7:
Requiring special privileges for building expressions is not expected, `darwin-rebuild` uses sudo internally for the specific part that needs privileges.

bryanhonof:
That's odd. If I try without everlasting privileges, I get the word from #644. Maybe NixOS/nix#6291 is related?

Enzime:
I don't believe `darwin-rebuild` should ever need to be run with `sudo` which speaks to a deeper issue, I'll reopen #644 for any users who run into a similar issue

PR #647, opened 2023-05-02
feat(services.skhd): improve skhd performance by setting SHELL to dash

ajaxbits:
Ensure fast performance by setting the skhd shell to `dash`<sup>[[1]](https://wiki.archlinux.org/title/Dash)</sup>.

I was experiencing very slow performance from `skhd` (moving between windows time ~= 1 sec), but I tried the solution [outlined in this GitHub issue](https://github.com/koekeishiya/skhd/issues/42#issuecomment-401880327) in the `skhd` repo, and it significantly improved performance (no noticeable delay).

Tested on Macbook Pro, M1 Max, OS v.12.6.5 (Monterey).

### DIFF
--- a/modules/services/skhd/default.nix
+++ b/modules/services/skhd/default.nix
@@ -34,6 +34,7 @@ in
 
     launchd.user.agents.skhd = {
       path = [ config.environment.systemPath ];
+      environment."SHELL" = "${pkgs.dash}/bin/dash";
 
       serviceConfig.ProgramArguments = [ "${cfg.package}/bin/skhd" ]
         ++ optionals (cfg.skhdConfig != "") [ "-c" "/etc/skhdrc" ];


jhenahan:
One issue I've run into with this approach: applications launched using `skhd` inherit the `SHELL`. For example, when launching Emacs normally I get my login shell, but if I launch it with `skhd` then I get `dash`.

emilazy:
I'm a little hesitant to merge this as-is for the reason @jhenahan gave. Are you using Fish? Enabling Babelfish should speed up shell startup significantly (if you're using Home Manager you'll need to use the unstable branch to get it to use Babelfish as well):

```nix
{
  programs.fish.useBabelfish = true;
  programs.fish.babelfishPackage = pkgs.babelfish;
}
```

I'd like to set this by default at some point if possible, since the startup time difference is so huge.

ajaxbits:
Absolutley @emilazy, that makes a lot of sense. I'd be happy to close this PR if it's not the direction the project should take here.

And I hadn't heard about the `bablefish` startup time speedup! I'll try that right now, thanks

emilazy:
Yeah, I think this would at least have to be a toggle option to merge, and it might not be worth it as this can be overridden in the user configuration if they're using a particularly slow shell. Ideally skhd would implement https://github.com/koekeishiya/skhd/issues/277 and then we could expose an option for that.

emilazy:
Closing this for now, but if skhd adds a shell option then I'd be happy to review a PR for that.

PR #650, opened 2023-05-05
nix: Fix registry extra attrs not being applied

amarshall:
This was

    mkDefault { } // filterAttrs () x

which is interpreted as

    (mkDefault { }) // (filterAttrs () x)

but the intention is

    mkDefault ({ } // filterAttrs () x)

Resulting in lastModified, rev, etc. not being included. This is essentially just bringing this clause up-to-date with the one from NixOS.


LnL7 (approved):
Thanks

PR #662, opened 2023-05-17
Add option to pam module to use pam_reattach

vamega:
This PR enables sudo authentication via Touch ID inside tmux as it's been described in the following [article](https://www.hein.dev/blog/2020/01/using-touchid-tmux-pam_reattach/).

I've tested it on my M1 Macbook Air, and went through a MacOS upgrade with this change.
As expected the changes to `/etc/pam.d/sudoers` was lost on the upgrade, but a quick `darwin-rebuild switch` later, things were working correctly once more.

Closes #606, and mostly derived from the code of @sestrella and @maljub01

### DIFF
--- a/modules/security/pam.nix
+++ b/modules/security/pam.nix
@@ -1,62 +1,91 @@
-{ config, lib, pkgs, ... }:
-
-with lib;
-
-let
+{ config
+, lib
+, pkgs
+, ...
+}:
+with lib; let
   cfg = config.security.pam;
-
-  # Implementation Notes
-  #
-  # We don't use `environment.etc` because this would require that the user manually delete
-  # `/etc/pam.d/sudo` which seems unwise given that applying the nix-darwin configuration requires
-  # sudo. We also can't use `system.patchs` since it only runs once, and so won't patch in the
-  # changes again after OS updates (which remove modifications to this file).
-  #
-  # As such, we resort to line addition/deletion in place using `sed`. We add a comment to the
-  # added line that includes the name of the option, to make it easier to identify the line that
-  # should be deleted when the option is disabled.
-  mkSudoTouchIdAuthScript = isEnabled:
-  let
-    file   = "/etc/pam.d/sudo";
-    option = "security.pam.enableSudoTouchIdAuth";
-    sed = "${pkgs.gnused}/bin/sed";
-  in ''
-    ${if isEnabled then ''
-      # Enable sudo Touch ID authentication, if not already enabled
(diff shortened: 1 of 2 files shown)


LnL7:
Given that there are multiple permutations for the same file, I'm not sure how well the patch module will handle switching between variants rather than just applying or rolling back the same patch. Does that work without issues?

vamega:
I just tested every state transition (manually, but if someone knows how this could be automated please let me know), and every transition works with the latest change I pushed.

My original attempt had a mistake, where the patch for only pam-reattach, was trying to add the touchid module, the discrepancy with the line count in the patch file caused the patch to not apply.

With the current patch, every state transition works correctly. The patches module handles switching between variants just fine it seems. Table for transitions below.

|  original tid value | original reattach value | new tid value  |  new reattach value  | success |
|---|---|---|---|---|
| false| false | false| true| y|
  | false| false | true| false| y|
  | false| false | true| true| y|
  | false| true | true| true| y|
  | false| true | false| false| y|
  | false| true | true| false| y|
  | true| false | true| true| y|
  | true| false | false| false| y|
  | true| false | true| false| y|
  | true| true | false | true| y|
  | true| true | false| false| y|
  | true| true | true| false| y|

maljub01:
The way the patch module is implemented, it will undo all "removed" patches first before applying any new ones. That's why all the tests were successful.

This works across generations too so if we were to modify the patch file later on, it will still work fine when everyone upgrades, because it will have a copy of the old version of the patch and remove it first before applying the new version

I think it should be pretty robust actually, as long as we (and the end user) don't try to juggle multiple patches for the same file at the same time.

vamega:
I've now been running this for about a month on two different Macbooks (one on Apple Silicon, the other is an intel Macbook).

I've upgraded both to Ventura since using this, and outside of having to activate my profile again to reapply the patches I've run into no issues.

I think this is in a place where it's worth merging.

stringang:
any progress?

shinzui:
Can we please merge this?

emilazy:
Thank you for your work on this, and I do think we should have this feature, but unfortunately patches are quite fragile and compose badly, so I’m not too comfortable merging this approach, especially with the tendency for system upgrades to overwrite this file. I’m going to close this with the hopes of getting an approach like https://github.com/LnL7/nix-darwin/pull/787 or https://github.com/LnL7/nix-darwin/pull/591 merged soon. I’m sorry that it took so long to get around to this.

PR #665, opened 2023-05-18
Make font handling match NixOS

quentinmit:
Previously, nix-darwin would look in the `share/fonts` directory in each of the packages mentioned in the `fonts.fonts` option. Not all fonts packaged in nixpkgs place their fonts in this directory. It looks like this logic recently changed to just link every file in every package, whether it's a font or not. This PR changes the handling of `fonts.fonts` to match NixOS's behavior, which is to use `find` to find all .ttf, .ttc, and .otf files in any of the passed-in paths. This allows using `fonts.fonts` with more font packages.


modules/fonts/default.nix
```diff
@@ -33,13 +33,12 @@ in
...
-      { paths = cfg.fonts; preferLocalBuild = true; }
+      { preferLocalBuild = true; }
       ''
         mkdir -p $out/Library/Fonts
-        for path in $paths; do
-            find -L $path/share/fonts -type f -print0 | while IFS= read -rd "" f; do
-                ln -sf "$f" $out/Library/Fonts
-            done
+        font_regexp='.*\.\(ttf\|ttc\|otf\)'
+        find -L ${toString cfg.fonts} -regex "$font_regexp" -type f -print0 | while IFS= read -rd "" f; do
```
sielicki, line 40: This prevents non-fonts from being iterated-over in the activationScripts, correct?

sielicki (approved):
LGTM.

emilazy (approved):
Seems reasonable. The NixOS behaviour includes other font types, but I don't think macOS supports any of them.

emilazy:
Can you rebase this against latest `master`? It can't be merged because the checks added between your commit and now haven't run.

quentinmit:
> Can you rebase this against latest `master`? It can't be merged because the checks added between your commit and now haven't run.

Apologies for the delay, I just saw this comment. I've merge in the current `master`.

emilazy:
Thanks!

PR #672, opened 2023-05-26
Add `AppleScrollerPagingBehavior` option

sxyazi:
The option location: System Settings - Appearance - Click in the scroll bar to


tests/system-defaults-write.nix
```diff
@@ -10,6 +10,7 @@
   system.defaults.NSGlobalDomain.ApplePressAndHoldEnabled = true;
   system.defaults.NSGlobalDomain.AppleShowAllExtensions = true;
   system.defaults.NSGlobalDomain.AppleShowScrollBars = "Always";
+  system.defaults.NSGlobalDomain.AppleScrollerPagingBehavior = true;
```
LnL7, line 13: It's probably enough to only include options with more special values in the test, there's not really anything that can go wrong with a boolean option I think. 😄

LnL7 (approved):

PR #675, opened 2023-05-29
buildkite-agent: fix launchd daemon environment

rvem:
Add missing 'NIX_SSL_CERT_FILE'

### DIFF
--- a/modules/services/buildkite-agents.nix
+++ b/modules/services/buildkite-agents.nix
@@ -228,7 +228,8 @@ in
       { path = cfg.runtimePackages ++ [ cfg.package pkgs.coreutils pkgs.darwin.DarwinTools ];
         environment = {
           HOME = cfg.dataDir;
-        }// (if config.nix.useDaemon then { NIX_REMOTE = "daemon"; } else {});
+          inherit (config.environment.variables) NIX_SSL_CERT_FILE;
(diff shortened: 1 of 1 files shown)


modules/services/buildkite-agents.nix
```diff
@@ -228,7 +228,8 @@ in
       { path = cfg.runtimePackages ++ [ cfg.package pkgs.coreutils pkgs.darwin.DarwinTools ];
         environment = {
           HOME = cfg.dataDir;
-        }// (if config.nix.useDaemon then { NIX_REMOTE = "daemon"; } else {});
+          NIX_SSL_CERT_FILE = "${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt";
```
Enzime, line 231: I think it would be preferable to inherit it

```suggestion
          inherit (config.environment.variables) NIX_SSL_CERT_FILE;
```

Enzime (changes requested):

modules/services/buildkite-agents.nix
LnL7: That's indeed a nice way to avoid the duplication, looks good otherwise. Launchd as an isolated environment so any service that needs tls connections will need this in it's environment.

rvem: Somehow missed this thread, updated my PR

LnL7 (approved):

PR #676, opened 2023-05-29
Use LocalHostName in darwin-rebuild with a flake

carlthome:
When running `darwin-rebuild switch --flake flake-uri` on other networks (like eduroam), I'm getting issues  since my hostname isn't actually what I expect in my flake config. By using the LocalHostName it won't change suddenly when being on other Wi-Fi:s.

Maybe there's a setting for this I've missed but otherwise I think this could be a reasonable change, if you agree.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,11 +86,10 @@ If you don't have an existing `configuration.nix`, you can run the following com
 mkdir -p ~/.config/nix-darwin
 cd ~/.config/nix-darwin
 nix flake init -t nix-darwin
+sed -i '' "s/simple/$(scutil --get LocalHostName)/" flake.nix
 ```
 
-Make sure to replace all occurrences of `simple` with your hostname which you can find by running `scutil --get LocalHostName`.
-
-> NOTE: Make sure to change `nixpkgs.hostPlatform` to `aarch64-darwin` if you are using Apple Silicon.
+Make sure to change `nixpkgs.hostPlatform` to `aarch64-darwin` if you are using Apple Silicon.
 
 </details>
 
@@ -120,7 +119,7 @@ Add the following to `flake.nix` in the same folder as `configuration.nix`:
 
 Make sure to replace `Johns-MacBook` with your hostname which you can find by running `scutil --get LocalHostName`.
 
-> NOTE: Make sure to set `nixpkgs.hostPlatform` in your `configuration.nix` to either `x86_64-darwin` (Intel) or `aarch64-darwin` (Apple Silicon).
+Make sure to set `nixpkgs.hostPlatform` in your `configuration.nix` to either `x86_64-darwin` (Intel) or `aarch64-darwin` (Apple Silicon).
 
 </details>
 
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ cd ~/.config/nix-darwin
 nix flake init -t nix-darwin
 ```
 
-Make sure to replace all occurrences of `simple` with your short hostname which you can find by running `hostname -s`.
+Make sure to replace all occurrences of `simple` with your hostname which you can find by running `scutil --get LocalHostName`.
 
 > NOTE: Make sure to change `nixpkgs.hostPlatform` to `aarch64-darwin` if you are using Apple Silicon.
 
@@ -118,7 +118,7 @@ Add the following to `flake.nix` in the same folder as `configuration.nix`:
 }
 ```
 
-Make sure to replace `Johns-MacBook` with your short hostname which you can find by running `hostname -s`.
+Make sure to replace `Johns-MacBook` with your hostname which you can find by running `scutil --get LocalHostName`.
 
 > NOTE: Make sure to set `nixpkgs.hostPlatform` in your `configuration.nix` to either `x86_64-darwin` (Intel) or `aarch64-darwin` (Apple Silicon).
 
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -132,7 +132,7 @@ if [ -n "$flake" ]; then
        flakeAttr=${fragment}
     fi
     if [ -z "$flakeAttr" ]; then
-      flakeAttr=$(hostname -s)
+      flakeAttr=$(scutil --get LocalHostName)
     fi
     flakeAttr=darwinConfigurations.${flakeAttr}
 fi


Enzime:
Alternative solution is setting `networking.hostName`:

https://daiderd.com/nix-darwin/manual/index.html#opt-networking.hostName

sielicki (approved):

LnL7 (commented):
I like the idea, but I've noticed with my devices quite often that bonjour hostnames end up having conflicts. As a result it also gets renamed so not sure which one is better to use as a result. eg.

```
Johns-MacBook-Pro-2
```

bestlem:
When my mac gets renamed in Bonjour I would want nix to reset the name to its normal version.

in the many times this has happened this has never been the wrong thing to do as the renaming is a bug in the router or as it seems Mac only a bug in macOS.

Enzime (changes requested):
I think we should use `Computer Name` instead, as this is the name displayed in System Preferences and it seems a lot less likely to change than Hostname or Local Hostname

emilazy:
That might have unforeseen consequences due to its freeform nature:

```console
emily@yuyuko ~> scutil --get ComputerName
幽々子
```

Probably best to stick with something Unix-y for consistency with NixOS, shell prompts, etc.? (but I'm agnostic about `hostname -s` vs. `LocalHostName`: they can both break strangely, although it seems like the former is easier to avoid.)

purcell:
I submitted #772 without seeing this first, sorry: in my case, VPNs were causing `hostname -s` to misreport the hostname, and `scutil --get LocalHostName` was the best fix for me.

purcell:
Despite [the documentation of `networking.localHostName`](https://github.com/LnL7/nix-darwin/blob/511177ffe8226c78c9cf6a92a7b5f2df3684956b/modules/networking/default.nix#L55C5-L55C29), the value of `scutil --get LocalHostName` doesn't include the `.local` for me. `scutil --get HostName` reports an error because I haven't set that. I've only ever set the computer name in the MacOS preferences.

`ComputerName` might indeed be an even more appropriate choice — I read @emilazy's concerns above, and if you set the computer name to a unicode string then that does of course get returned as `ComputerName` by `scutil`, but it's unclear where `LocalHostName` comes from in that case. I tried changing my computer name to an emoji and saw this (no change in local host name) fwiw:
Screenshot 2023-09-10 at 09 21 10

I'd suggest that consistency with NixOS or problems with shell prompts aren't issues here, and that `ComputerName` would be the most straightforward choice, because of aligning with the only way of setting the computer name in the MacOS prefs.

emilazy:
In my experience macOS tries to set the local hostname intelligently from the computer name by default, e.g. "Foo's MacBook Pro" becoming `Foos-MacBook-Pro`, CJK text getting automatically romanized(!), etc.; it's possible this has changed in recent versions but I'd be surprised. Not sure about the emoji case; if you manually set the local hostname before then that could override it.

purcell:
Yup. In any case, I can confirm that `hostname -s` is prone to unpredictable output which make the current behaviour useless, so it's be good to either merge this in its current form, or with ComputerName.

Enzime:
How about using `ComputerName` by default and if that fails, falling back to `LocalHostName`?

purcell:
> How about using `ComputerName` by default and if that fails then falling back to `LocalHostName`?

It's not even clear that it could fail and need a fallback, but I agree `ComputerName` would be the best default. The scope of this string is just the name of the configuration attribute after all, so it's hard to see what could go wrong.

carlthome:
>It's not even clear that it could fail and need a fallback

When would it fail?

emilazy:
I think the main issue with `ComputerName` is just that I would expect it to break the majority of existing flake configurations, including with stock OS installs (since their names have spaces in them) and the example in the README. That feels painful to me. `LocalHostName` should usually match the Unix hostname so the scope for possible breaking changes is much smaller.

emilazy (dismissed):
Let's land this. Neither the current nor the PR solution seem ideal - they can both break sometimes - but we've had multiple people ask for and PR this so it seems worth a try. If we get complaints about failures due to clashing Bonjour hostnames (which is something I've run into before, so for me personally this PR is probably a regression) we can consider reverting or even the `ComputerName` approach.

emilazy:
Looks like I can't actually merge because of @Enzime's review

Enzime (changes requested):
This PR should update the README as well

purcell:
> When would it fail?

Exactly! I can't think why it would. If it reliably prints a string that the user considers a fair version of their computer name, that's good enough I think.

purcell:
> If we get complaints about failures due to clashing Bonjour hostnames (which is something I've run into before, so for me personally this PR is probably a regression) we can consider reverting or even the ComputerName approach.

I think this is why the consensus appear to be converging on ComputerName.

emilazy:
I think breaking ~literally everyone's flake setup is too high a cost to bear, and the `ComputerName` is, notably, never used as a unique identifier for a machine on a network etc. (that's precisely why the others can potentially change or have clashes).

Enzime:
I’m not on a computer at the moment, but the last time I tested if you try to use unicode as a key for a Nix attrset, the compiler error is quite unclear (replaces unicode characters with question marks).

My personal preference is definitely still `ComputerName` and possibly if it contains unicode then falling back to `LocalHostName` (or falling back if whenever `ComputerName` doesn’t exist)

emilazy:
I'm pretty sure the default, out-of-the-box computer names include Unicode (`"…’s MacBook Pro"`)

purcell:
Fair...
```
error: 'git+file:///Users/steve/.nixpkgs#darwinConfigurations.💪.system' is not a valid URL
```

purcell:
Thank you for disabusing me of my assumption that Nix would generally handle the unicode attribute name sanely.

carlthome:
>This PR should update the README as well

Not sure what in the README should be updated. Can you clarify?

Enzime:
This line needs updating:

https://github.com/LnL7/nix-darwin/blob/80bb201f4925cdda5a7a3c7b1900fb26bb2af2e8/README.md?plain=1#L121

Also could you remove this line and add a command that does the replacing i.e.

```
sed 's/simple/$(scutil ...)' flake.nix
```

https://github.com/LnL7/nix-darwin/blob/80bb201f4925cdda5a7a3c7b1900fb26bb2af2e8/README.md?plain=1#L91

Enzime (approved):
LGTM

PR #679, opened 2023-06-01
Document font overriding behaviour

dbaynard:
This just adds the description —  #624 allows font packages with conflicting font file names to coexist, and through the use of `ln -f` selects the latter.

(I've had a patch to apply the `ln -sf` change for a while, but I didn't realize others were hitting this issue, too — I'll work on pushing my improvements up stream.)


LnL7 (approved):
To me failing on duplicates makes more sense than inconsistent behaviour, but at least it's described now. Thanks!

PR #681, opened 2023-06-06
system/checks: support systems with auto-allocate-uids enabled

cole-h:


cole-h:
Now that I think about it, it might be better to just allow disabling this check altogether rather than doing heuristics that may not hold.

domenkozar (approved):

bestlem:
I think the default has to be don't do the check.

The reason is how do you turn the check off - that is not in the documentation.

bestlem:
The install I was doing was install nix and then nix-darwin and only then look at config.

cole-h:
You'll want to follow the instructions for installing nix-darwin from here: https://github.com/LnL7/nix-darwin#flakes-experimental

Specifically, you create your nix-darwin config, then you build it on your darwin machine, and then you use the `darwin-rebuild` script in that built darwin configuration to complete the switch.

PR #684, opened 2023-06-07
nix/linux-builder: init

Enzime:
This PR adds `darwin.builder` as a `launchd` service :tada: 

Depends on https://github.com/NixOS/nixpkgs/pull/236389 as it expects `darwin.builder` to run on port 31022 by default

### DIFF
--- a/modules/nix/linux-builder.nix
+++ b/modules/nix/linux-builder.nix
@@ -1,0 +1,95 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  inherit (pkgs) stdenv;
+
+  cfg = config.nix.linux-builder;
+
+  builderWithOverrides = cfg.package.override {
+    modules = [ cfg.modules ];
+  };
+in
+
+{
+  options.nix.linux-builder = {
+    enable = mkEnableOption (lib.mdDoc "Linux builder");
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.darwin.linux-builder;
+      defaultText = "pkgs.darwin.linux-builder";
+      description = lib.mdDoc ''
+        This option specifies the Linux builder to use.
+      '';
+    };
+
+    modules = mkOption {
+      type = types.listOf types.anything;
+      default = [ ];
+      example = literalExpression ''
+        [
+          ({ config, ... }:
+
+          {
+            virtualisation.darwin-builder.hostPort = 22;
+          })
+        ]
+      '';
+      description = lib.mdDoc ''
+        This option specifies extra NixOS modules and configuration for the builder. You should first run the Linux builder
+        without changing this option otherwise you may not be able to build the Linux builder.
+      '';
+    };
+  };
+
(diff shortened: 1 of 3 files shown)


modules/nix/linux-builder.nix
```diff
@@ -0,0 +1,78 @@
...
+      };
+      serviceConfig = {
+        ProgramArguments = [
+          "/bin/sh" "-c"
+          "/bin/wait4path /nix/store &amp;&amp; exec ${cfg.package}/bin/create-builder"
+        ];
+        KeepAlive = true;
+        RunAtLoad = true;
+        WorkingDirectory = "/var/lib/darwin-builder";
+        StandardOutPath = "/var/log/darwin-builder.log";
```
LnL7, line 53: Launchd doesn't have any log rotation so I would avoid enabling this by default in modules, a use can always enable logs for debugging in the local configuration. Or dynamically attach the log output as described here https://github.com/LnL7/nix-darwin/blob/master/modules/services/nix-daemon.nix#L28-L31.

Maybe the description of the Standard*Path options should also be extended with some information about that.

modules/nix/linux-builder.nix
```diff
@@ -0,0 +1,78 @@
...
+
+    nix.distributedBuilds = true;
+
+    nix.buildMachines = [{
+      hostName = "linux-builder";
+      sshUser = "builder";
+      sshKey = "/etc/nix/builder_ed25519";
+      system = "${stdenv.hostPlatform.uname.processor}-linux";
+      supportedFeatures = [ "kvm" "benchmark" "big-parallel" ];
+      publicHostKey = "c3NoLWVkMjU1MTkgQUFBQUMzTnphQzFsWkRJMU5URTVBQUFBSUpCV2N4Yi9CbGFxdDFhdU90RStGOFFVV3JVb3RpQzVxQkorVXVFV2RWQ2Igcm9vdEBuaXhvcwo=";
```
LnL7, line 73: Is this completely static? Would be better to retrieve this information from an attribute of the builder package if possible.

LnL7 (commented):
This builder package sounds quite magical, very cool stuff!

modules/nix/linux-builder.nix
Enzime: Removed :+1:

modules/nix/linux-builder.nix
Enzime: The host key is currently completely static, the host keys are files in the Nixpkgs tree and are not easily overridable, this value comes from the Nixpkgs documentation:

https://github.com/NixOS/nixpkgs/blob/master/doc/builders/special/darwin-builder.section.md?plain=1#L54

lovesegfault:
@LnL7 @domenkozar Since the upstream PR has landed in nixpkgs, I think this is ready for review/merge :)

Very excited about this! Thanks for pushing it forward @Enzime :)

domenkozar (approved):

PR #686, opened 2023-06-10
nix: Add build machines protocol option.

CathalMullan:
This add the `protocol` option to the Nix build machines config.

[Simple copy and paste of the NixOS config option](https://github.com/NixOS/nixpkgs/blob/nixos-23.05/nixos/modules/services/misc/nix-daemon.nix#L231).

### DIFF
--- a/modules/nix/default.nix
+++ b/modules/nix/default.nix
@@ -227,6 +227,19 @@ in
                 The hostname of the build machine.
               '';
             };
+            protocol = mkOption {
+              type = types.enum [ null "ssh" "ssh-ng" ];
+              default = "ssh";
+              example = "ssh-ng";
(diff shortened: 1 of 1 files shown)


modules/nix/default.nix
```diff
@@ -227,6 +227,19 @@ in
                 The hostname of the build machine.
               '';
             };
+            protocol = mkOption {
+              type = types.enum [ null "ssh" "ssh-ng" ];
+              default = "ssh";
```
LnL7, line 232: Given that `null` doesn't add any `://` scheme prefix that seems like a better option for backwards compatibility to me. But guess it's fine since that's also that the nixos module uses.

LnL7 (approved):

PR #687, opened 2023-06-11
Install nix-darwin on flake-based systems

### LINKED ISSUE #398: flake: Add apps for {un,}installer?
If the defaultApp of the flake were `darwin-installer`, installation with flakes would be as easy as `nix run github:LnL7/nix-darwin`.

Enzime:
- flake: add packages.darwin-uninstaller
- flake: add packages.darwin-{option,rebuild}
- Allow flaky installation with darwin-rebuild
- readme: update with new flaky instructions
- installer: match flaky installation logic
- doc: store a copy of known files
- etc: fail if we can't add a file
- etc: use .before-nix-darwin instead of .orig

Closes #398 
Closes #399

Tested installing with `darwin-installer` for a channels based system and `darwin-rebuild` for a flake based system on `aarch64-darwin`


sielicki (approved):

Enzime:
@domenkozar can you take a look at this PR? The tests are failing due to an unrelated issue (https://github.com/NixOS/nixpkgs/pull/237557) which looks like it'll be a pretty big change that'll require its own PR

domenkozar:
I can't merge PRs that fail CI, so we'll have to address it.

.github/workflows/test.yml
```diff
@@ -119,28 +119,16 @@ jobs:
...
-        nix-shell -A installer
-        nix-shell -A installer.check
-    - name: Build simple flake configuration
-      run: |
-        nix build ./modules/examples/flake#darwinConfigurations.simple.system --override-input darwin .
-    - name: Activate derivation of simple flake build
-      run: |
-        ./result/sw/bin/darwin-rebuild switch --flake ./modules/examples/flake#simple --override-input darwin .
-    - name: Rebuild and activate simple flake, but this time using nix-darwins flake interface
+        nix build ./modules/examples/flake#darwinConfigurations.simple.system --override-input darwin . --override-input nixpkgs nixpkgs/${{ env.CURRENT_STABLE_CHANNEL }}
```
LnL7, line 126: Wouldn't it make more sense to expose the the darwin tools package directly instead of building an entire dummy system as the main usecase here is bootstrapping an install?

LnL7:
@domenkozar I think the merge is only blocked by a missing review, while the pr status is red only the stable checks are marked as required. (at least that was my intent with [#641](https://github.com/LnL7/nix-darwin/pull/641))

README.md
```diff
@@ -7,12 +7,14 @@
...
 
 This project aims to bring the convenience of a declarative system approach to macOS.
-Nix-darwin is built up around [Nixpkgs](https://github.com/NixOS/nixpkgs), quite similar to [NixOS](https://nixos.org/).
+nix-darwin is built up around [Nixpkgs](https://github.com/NixOS/nixpkgs), quite similar to [NixOS](https://nixos.org/).
 
 ## Install
 
 To install nix-darwin, a working installation of [Nix](https://github.com/NixOS/nix#installation) is required.
 
+> NOTE: Using `darwin-installer` is no longer necessary on flake based systems.
```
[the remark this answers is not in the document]
Enzime: So far, the only thing I've moved to regular activation is creating `/run`.

The things I didn't move to regular activation:

- Copying example configuration to `~/.nixpkgs/darwin-configuration.nix`
- Sourcing the static versions of `/etc/bashrc` and `/etc/zshrc` as `programs.bash.enable` defaults to `true` and `programs.zsh.enable = true;` is uncommented in both `simple.nix` and the example `flake.nix`
- Adding `darwin` to `nix-channel`
- Setting `nix.configureBuildUsers = true;` to migrate old style users to new style users, as there's a check that ensures it's set when old style users are present.

.github/workflows/test.yml
Enzime: Replaced with `nix run .#darwin-rebuild` :+1:

README.md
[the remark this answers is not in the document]
LnL7: This used to be more but more modules have been added since the initial version so that's great.

I think the main functional change in that case is the fact that the known hashes for zshrc/bashrc will always need to be correct since there's no installer that edits the existing files anymore. In that case perhaps it's best to also remove those steps from the installer so that both installation paths work the same.

pkgs/nix-tools/default.nix
```diff
@@ -0,0 +1,7 @@
+{ darwin }:
```
LnL7, line 1: Instead of passing a whole system to nix-tools it really only needs references to default path locations, this way we don't need a complete system to build the packages.

```suggestion
{ pkgs
, profile ? "/nix/var/nix/profiles/system"
, nixPackage ? "/nix/var/nix/profiles/default"
, defaultPath ? "/usr/local/bin:/usr/bin:/usr/sbin:/bin:/sbin"
}:

let
  extraPath = lib.makeBinPath [ nixPackage pkgs.coreutils pkgs.jq pkgs.git ];
in

{
  darwin-option = pkgs.writeProgram "darwin-option"
    {
      inherit profile;
      inherit (pkgs.stdenv) shell;
    }
    ./darwin-option.sh;

  darwin-rebuild = pkgs.writeProgram "darwin-rebuild"
    {
      inherit profile;
      inherit (pkgs.stdenv) shell;
      path = "${extraPath}:${defaultPath}";
    }
    ./darwin-rebuild.sh;
}
```

Enzime: Updated :+1:

README.md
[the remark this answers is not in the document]
Enzime: I've removed the logic in `darwin-installer` to update `/etc/bashrc` and `/etc/zshrc`, as well as removed the `nix.configureBuildUsers = true;`.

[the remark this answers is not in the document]
Enzime: @LnL7 this PR is ready to review now :smile: 

I've updated the known hashes for `/etc/nix/nix.conf` and `/etc/{bash,zsh}rc`, I've also added a copy of known files as reference for future.

I've also changed the `etc` activation script to fail if any of the files don't match the known hash and need to be replaced manually.

modules/nix/default.nix
```diff
@@ -674,9 +674,12 @@ in
...
     # Not in NixOS module
     environment.etc."nix/nix.conf".knownSha256Hashes = [
-      "7c2d80499b39256b03ee9abd3d6258343718306aca8d472c26ac32c9b0949093"  # nix installer
+      "7c2d80499b39256b03ee9abd3d6258343718306aca8d472c26ac32c9b0949093"  # official Nix installer
       "19299897fa312d9d32b3c968c2872dd143085aa727140cec51f57c59083e93b9"
       "c4ecc3d541c163c8fcc954ccae6b8cab28c973dc283fea5995c69aaabcdf785f"
+      "ef78f401a9b5a42fd15e967c50da384f99ec62f9dbc66ea38f1390b46b63e1ff"  # official Nix installer 2.0
+      "c06b0c6080dd1d62e61a30cfad100c0cfed2d3bcd378e296632dc3b28b31dc69"  # official Nix installer as of 2.0.1
+      "ff08c12813680da98c4240328f828647b67a65ba7aa89c022bd8072cba862cf1"  # official Nix installer as of 2.4
     ];
```
emilazy, line 683: Would be good to add the hashes from https://github.com/DeterminateSystems/nix-installer if none of these match.

Enzime: The issue with adding hashes from Determinate Systems' Nix installer is that they have version numbers in their `nix.conf` so I suspect the hashes will go out of date very quickly, and I'm also not sure how static the contents of `nix.conf` is with their installer.

emilazy: I see, that's a shame :( Maybe we could coordinate with them to get it to be more, uh, deterministic? Not trying to block the PR on this though.

flake.nix
```diff
@@ -29,19 +35,32 @@
       path = ./modules/examples/flake;
       description = "nix flake init -t nix-darwin";
     };
+  } //
+  flake-utils.lib.eachSystem [ "x86_64-darwin" "aarch64-darwin" ] (system:
```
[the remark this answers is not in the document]
Enzime: Updated :+1:

Enzime:
@LnL7 @domenkozar can you take a look at this PR when you have a chance?

modules/system/flake-overrides.nix
```diff
@@ -11,6 +11,9 @@ in
     system.checks.verifyNixPath = mkDefault false;
     system.checks.verifyNixChannels = mkDefault false;
 
+    # Still overridable with lib.mkForce
+    nix.nixPath = lib.mkDefault [ ];
```
LnL7, line 15: Won't this make specifying `nix.nixPath` mandatory for flake configurations?
I think wether a system uses `<nixpkgs>` as a channel or not is completely separate from if the configuration itself happens to be built using flakes.

LnL7 (dismissed):
A few small comments, but looks good otherwise.

modules/system/flake-overrides.nix
Enzime: I'll remove this from this PR, the reason why I added it in the first place was primarily because `NIX_PATH` contained `darwin-configuration=` which was not relevant when using flakes. I think it would be good in future to have some more granular options for controlling the values added to `nix.nixPath`.

.github/workflows/test.yml
```diff
@@ -47,15 +41,23 @@ jobs:
...
     - name: Install nix-darwin and test
       run: |
         export NIX_PATH=$HOME/.nix-defexpr/channels
+
+        # We run nix-darwin twice to test that it can create darwin-configuration correctly for us
+        # but we expect it to fail setting up /etc/nix/nix.conf
+        nix-shell -A installer || true
+
+        nixConfHash=$(shasum -a 256 /etc/nix/nix.conf | cut -d ' ' -f 1)
+        /usr/bin/sed -i.bak \
```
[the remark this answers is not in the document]
Enzime: The reason it's necessary in CI is because `nix.conf` contains the GitHub access token

modules/system/flake-overrides.nix
Enzime: Removed :+1:

LnL7: Ah, luckily nix.nixPath can also be an attribute set so I think it shouldn't be that hard to conditionally add the configuration path.

.github/workflows/test.yml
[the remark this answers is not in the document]
LnL7: Makes the testing setup a bit messy, but the way those currently run isn't ideal anyway. With a nicer setup a lot more things during activation could be validated properly.

LnL7 (approved):

.github/workflows/test.yml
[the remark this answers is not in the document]
Enzime: I'm happy to investigate this further in another PR, any things that come to mind that we should test for?

README.md
```diff
@@ -73,15 +75,41 @@ Configuration lives in `~/.nixpkgs/darwin-configuration.nix`. Check out
 There is also preliminary support for building your configuration using a [flake](https://nixos.wiki/wiki/Flakes).  This
 is mostly based on the flake support that was added to NixOS.
 
-A minimal example of using an existing configuration.nix:
+<details>
+<summary>
+
+Getting started from scratch
+
+</summary>
```
emilazy, line 83: Please collapse this into a single line; it renders as a `<p>` on GitHub and adds a weird line break.

README.md
```diff
@@ -73,15 +75,41 @@ Configuration lives in `~/.nixpkgs/darwin-configuration.nix`. Check out
...
+> NOTE: Make sure to change `system` to `aarch64-darwin` if you are using Apple Silicon.
+
+</details>
+
+<details>
+<summary>
+
+Migrating from an existing `configuration.nix`
+
+</summary>
```
emilazy, line 102: (as above)

README.md
```diff
@@ -73,15 +75,41 @@ Configuration lives in `~/.nixpkgs/darwin-configuration.nix`. Check out
...
+
+If you don't have an existing `configuration.nix`, you can run the following commands to generate a basic `flake.nix` inside `~/.config/nix-darwin`:
+
+```bash
+mkdir -p ~/.config/nix-darwin
+cd ~/.config/nix-darwin
+nix flake init -t nix-darwin
+```
+
+> NOTE: Make sure to change `system` to `aarch64-darwin` if you are using Apple Silicon.
```
emilazy, line 93: This should reference `nixpkgs.hostPlatform` now.

README.md
```diff
@@ -92,6 +120,22 @@ A minimal example of using an existing configuration.nix:
...
+
+```bash
+nix run nix-darwin#darwin-rebuild -- switch --flake ~/.config/nix-darwin#Johns-MacBook
+```
+
+After installing, you can run `darwin-rebuild` directly:
+
+```bash
+darwin-rebuild switch --flake ~/.config/nix-darwin#Johns-MacBook
+```
```
emilazy, line 135: These make sense if the reader expanded the migration section, but not for the from-scratch one. I suggest adding wording to the from-scratch section telling the reader to replace `simple` with their system's hostname, as is best practice, and omitting the `#...` here.

README.md
```diff
@@ -92,6 +120,22 @@ A minimal example of using an existing configuration.nix:
...
+
+> NOTE: The `hostname(1)` of your system will be used to decide
+> which darwin configuration is applied if it's not specified explicitly in the flake ref.
```
emilazy, line 138: Maybe say `hostname -s` instead:

```console
emily@yuyuko ~> hostname 
yuyuko.emily.moe
emily@yuyuko ~> hostname -s
yuyuko
```

The latter is used by `darwin-rebuild`.

flake.nix
```diff
@@ -34,6 +36,12 @@
           });
     };
 
+    overlays.default = final: prev: {
+      inherit (prev.callPackage ./pkgs/nix-tools { }) darwin-rebuild darwin-option;
+
+      darwin-uninstaller = prev.callPackage ./pkgs/darwin-uninstaller { nix-darwin = self; };
```
emilazy, line 42: The uninstaller should be updated to also remove the `run` line from `synthetic.conf` and poke APFS appropriately.

It also completely removed `/etc/nix/nix.conf`, which is not great. It should restore the pre-nix-darwin backup file, if present.

It also printed an `rm` error when I ran it on my test VM:

```console
…
removing service org.nixos.activate-system
removing service org.nixos.nix-daemon
configuring networking...
rm: cannot remove '/Library/LaunchDaemons/org.nixos.nix-daemon.plist': No such file or directory
grep: /etc/bashrc: No such file or directory
Found no nix-daemon.sh reference in /etc/bashrc
add this snippet back to /etc/bashrc:
…
```

Maybe that's expected though.

emilazy (changes requested):
Thank you for working on this! I've left review comments for various nits; the other things I'd like to see before merging are:

* The hash for the `nix.conf` produced by the latest version of the Determinate Systems installer with default settings to be added, as the error for an unknown `nix.conf` will be confusing for new years and leaves the system in a half-activated state. The SHA-256 is `f3e03d851c240c1aa7daccd144ee929f0f5971982424c868c434eb6030e961d4` and the contents are:

  ```
  # Generated by https://github.com/DeterminateSystems/nix-installer, version 0.10.0.
  experimental-features = nix-command flakes auto-allocate-uids
  build-users-group = nixbld
  auto-optimise-store = true
  ```

  (plus trailing newline)

* Instructions for uninstalling a flakes-based system in `README.md`, whether in the uninstallation section (preferable as I think that's where people are most likely to look first) or the flakes section.

emilazy:
Thinking about https://github.com/LnL7/nix-darwin/pull/687/commits/77936c4e541d12b99bbddf088ff2ddeeb95be2c1, it might be good to move the hash checks into the `checks` activation script so that mismatches don't cause partially-activated configurations, which is the thing it's trying to avoid in the first place and which I ran into when testing this.

README.md
LnL7: FYI there's is #676 which would change the behaviour slightly.

LnL7:
That's a good idea, if we want to change a bunch of etc behaviour might be easier to do that in a separate PR however.

README.md
emilazy: Yeah, I'm agnostic to which we use; I might slightly prefer the current solution as I've had Bonjour clashes before. Mostly just want to document the correct one whichever it is.

README.md
Enzime: Updated :+1:

README.md
Enzime: Updated :+1:

README.md
Enzime: Updated :+1:

README.md
Enzime: Updated :+1:

flake.nix
Enzime: I think the `darwin-uninstaller` needs extra work and my preference would be to focus on that in another PR

modules/nix/default.nix
Enzime: Added `nix.conf` from the latest DeterminateSystems nix-installer :+1:

Enzime:
I've incorporated some of @emilazy's suggestions as well as adding a new commit just to use the suffix `before-nix-darwin` instead of `orig` as it makes the source of the file clearer

modules/services/tailscale.nix
```diff
@@ -74,10 +74,10 @@ in
       "2c28f4fe3b4a958cd86b120e7eb799eee6976daa35b228c885f0630c55ef626c"
     ];
 
-    # Cleaning up the .orig file is necessary as any files in /etc/resolver will be used.
+    # Cleaning up the .pre-nix-darwin file is necessary as any files in /etc/resolver will be used.
     system.activationScripts.etc.text = mkAfter ''
-      if [ -e /etc/resolver/ts.net.orig ]; then
-        rm /etc/resolver/ts.net.orig
+      if [ -e /etc/resolver/ts.net.pre-nix-darwin ]; then
+        rm /etc/resolver/ts.net.pre-nix-darwin
```
emilazy, line 80: Should be `.before-nix-darwin`, this will presumably break Tailscale.

emilazy (changes requested):
Alright, I think this is good to go. The installer and `nix.conf` issues are a pain but we can solve them separately and I don't want to hold this up any further. Just one more thing though...

modules/services/tailscale.nix
Enzime: Good catch, fixed :+1:

emilazy (approved):
Let's ship it! I'll open an issue for the uninstaller stuff but this looks like a great improvement to me. Thank you for all your work on this.

PR #692, opened 2023-06-16
fonts: fontDir: only warn once about hard links

sielicki:
/nix being on another filesystem is more common than not, so avoid excessively warning for something that is generally expected, and print only once. This warning should not be removed entirely as it provides a useful hint to the user that a file is being duplicated and copied outside of /nix.


LnL7 (approved):
Hmm won't that warning just always trigger, the nix installer has been using a separate apfs volume for quite some time now.

PR #706, opened 2023-06-22
activity monitor: change settings to null

zowoq:
treewide these seem to be the only settings that aren't unset by default which seems inconsistent

there wasn't any discussion about this on the original PR https://github.com/LnL7/nix-darwin/pull/447


modules/system/defaults/ActivityMonitor.nix
```diff
@@ -18,7 +18,7 @@ with lib;
             105: Active Processes
             106: Inactive Processes
             107: Windowed Processes
-          Default is 100.
+          Default is null.
```
LnL7, line 21: The documentation option usually describe what the macOS default is, not configured is't very helpful 😄.

LnL7 (approved):
Indeed, this isn't intended

LnL7 (approved):
Thanks!

PR #704, opened 2023-06-21
eval-config.nix: readd `lib.mdDoc` temporarily

emilazy:
Fixes #701. Note that this is *not* a viable long-term solution; currently, all `mdDoc`-tagged documentation is omitted from the manual build entirely(!). We need to move over to the new NixOS Markdown-based documentation generator ASAP; the DocBook output won't live for much longer. I'm working on this and it looks like it won't be too difficult (thanks @pennae for pointing me to [nix-doc-munge](https://github.com/pennae/nix-doc-munge)!), but the changes will be more extensive and it's not going to be possible to maintain simultaneous compatibility with older nixpkgs (at least for the manual build). I'm sending this just to unbreak people's systems while I work on a longer-term solution.

PR #707, opened 2023-06-23
Migrate option documentation to Markdown

emilazy:
NixOS has completed the the transition to using Markdown for option documentation, and `lib.mdDoc` has become a no-op. This poses a problem for us, as our creaky old copy of the NixOS option documentation generator can't even process Markdown. https://github.com/LnL7/nix-darwin/pull/704 was a stop-gap solution, but still doesn't give us a correct manual (we currently have in-tree modules using `lib.mdDoc` that simply don't have their docs in the manual), and not moving to Markdown will make it increasingly more difficult to share code between NixOS and nix-darwin modules, as well as keeping it painful to write well-formatted documentation.

This PR migrates nix-darwin to use the new NixOS documentation generator and converts all option documentation to Markdown. The process is:

* Pin an old version of nixpkgs that still allows DocBook.
* Switch to using NixOS's new `nixosOptionsDoc` to process the option documentation.
* Do manual adjustments and conversions for a few options.
* Batch convert the rest with [my fork of `nix-doc-munge`](https://github.com/emilazy/nix-doc-munge).
* Remove the pin.
* Switch the manual generation over to `nixos-render-docs`.

This removes a lot of complexity and duplication and produces a near-identical manual. It necessarily requires dropping support for building the manual with 22.11; its version of the documentation generator cannot handle the Markdown constructs the conversion produces. (23.05 works fine, of course.)

The number of commits might be a little daunting but I tried to factor them out for easy review :) The big automated conversion is https://github.com/LnL7/nix-darwin/pull/707/commits/e65131e69cab1b18fb49b572b983a18720502e1b and should be fully reproducible with the command in the commit message; the other commits are manually written or are simple find-and-replace jobs that I reviewed manually. I did not add a `.git-blame-ignore-revs` file for the mass change as NixOS doesn't seem to have for its equivalent migration, but I can add one if desired.

I will try to send a PR to Home Manager to migrate their module documentation to Markdown after this is merged; if anyone knows of any other third-party nix-darwin modules using DocBook documentation I will look into fixing them too.

cc @LnL7 @domenkozar


emilazy:
(Rebased to reorganize some commits in a more logical manner. No change to the resulting tree.)

.github/workflows/test.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
 
 env:
-  CURRENT_STABLE_CHANNEL: nixpkgs-22.11-darwin
+  CURRENT_STABLE_CHANNEL: nixpkgs-23.05-darwin
```
LnL7, line 7: With regard to support for releases, having only 1 option for a stable release while it was only recently created feels a little early to me.
That said, given that the conversion is rather impactful and only for documentation, would it be possible to make a 22.11 build possible if we set the default for `documentation.enable = false` for systems with an old nixpkgs?

LnL7 (approved):
Awesome, thank you for tackling this! While I added the documentation module at some point, it borrows heavily from the home-manager implementation so I'm not particularly familiar with the details.

.github/workflows/test.yml
LnL7: Actually evaluation works as is on 22.11 so this should be all that's needed. Would be nice if something similar is still possible with the next step.

```patch
diff --git a/modules/documentation/default.nix b/modules/documentation/default.nix
index 15ccfb2..3f8179d 100644
--- a/modules/documentation/default.nix
+++ b/modules/documentation/default.nix
@@ -99,12 +99,14 @@ in
     programs.man.enable = cfg.man.enable;
     programs.info.enable = cfg.info.enable;

-    environment.systemPackages = mkMerge [
+    environment.systemPackages = mkIf (lib.versionAtLeast lib.version "23.05") (mkMerge [
       (mkIf cfg.man.enable [ manual.manpages ])
       (mkIf cfg.doc.enable [ manual.manualHTML helpScript ])
-    ];
+    ]);

     system.build.manual = manual;

+    warnings = optional (lib.versionOlder lib.version "22.11") "documentation.`enable` now requires 23.05 to build documentation for `configuration.nix`, only manpages for system packages will be enabled.";
+
   };
 }
```

doc/manual/default.nix
```diff
@@ -1,99 +1,51 @@
...
+            let
+              subpath = lib.removePrefix "/"
+                (lib.removePrefix (toString prefix) (toString decl));
+            in {
+              url = "https://github.com/LnL7/nix-darwin/blob/${revision}/${subpath}";
+              name = "<nix-darwin/${subpath}>";
+            }
+          # TODO: handle this in a better way (may require upstream
+          # changes to nixpkgs)
+          else if decl == "lib/modules.nix" then
```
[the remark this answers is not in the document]
emilazy: Yeah, it's the same for the [NixOS option documentation](https://nixos.org/manual/nixos/unstable/options.html). I considered modifying `modules-list.nix` to solve it but it'd be best to just address it upstream. A big advantage of this is that once I send another PR to move the last DocBook infrastructure to `nixos-render-docs` we'll benefit from any improvements the NixOS manual gets :)

.github/workflows/test.yml
emilazy: Do we want to support stable releases that upstream nixpkgs don't consider supported, in general? It's possible to pin an older version of nix-darwin and search.nixos.org lists 22.11 as deprecated already (although [the release notes](https://nixos.org/manual/nixos/stable/release-notes.html#sec-release-22.11) imply support will stop at the end of June, which I guess would give it another week from now). My impression is that people following the stable channels of nixpkgs/NixOS are expected to ride the upgrade train pretty proactively (which admittedly is one of the reasons I've never seen the point).

It'd be easy to disable documentation on 22.11, though, so I'm happy to make this change if you'd prefer it; just wondering about what the general approach is here.

LnL7: Yeah, I personally find a couple of months rather short. Given that's the policy upstream there's no point to bend over backwards, but if we can make it work without docs using a few lines let's do that.

doc/manual/default.nix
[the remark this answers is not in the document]
LnL7: Alright, let's not add workarounds that might break once it's fixed in NixOS then.

.github/workflows/test.yml
emilazy: Okay, I've handled this in the newest push and confirmed that things still seem to build fine on 22.11. I added stub packages that replace the manual with notices explaining they require 23.05; since `documentation.enable` is on by default it seemed excessive to print a warning on every activation even if people aren't using the manual at all, but if you'd prefer to do the simpler thing I can rip that out.

emilazy:
Rebased to fix conflicts and address reviews. I also did a few minor tweaks and, more importantly, completed the migration to `nixos-render-docs`; there is now no DocBook toolchain involved at all and no warnings about the deprecated `optionsDocBook` output. Net change: -865 lines and +better manual output :)

I noticed that `darwin-option` is pretty broken after this (the `eval` hack causes it to try running anything in backticks, which is kind of scary), but it didn't feel worth delaying this to address. It'd probably be best to rebase it on the upstream C++ `nixos-option`, which I believe would also enable flakes support; maybe I'll look into that after I finish doing all this again for Home Manager.

Would be nice to get this merged and done with, since it's a little bit of a pain to rebase, but I understand there's a good amount to review even in the non-automated commits.

rycee:
@emilazy Hello, I don't know anything about the new documentation generator. Looks quite interesting, though 🙂

I just have a few considerations that maybe you know the answer to. Mainly I'm curious whether the build closure size changes a lot? Any support for custom CSS styling of the generated HTML? Would it be difficult to add support for Asciidoc?

emilazy:
Didn't check the build closure size but I assume it's comparable or went down as the new generator was specifically written for NixOS and doesn't bring in an XML toolchain; it's custom-written Python. The output is almost exactly identical to the previous one down to markup structure (you can compare the manuals on the NixOS site) and supports full CSS and JS customization. It's all-in on Markdown: current nixpkgs has completely switched over to its own flavour of Markdown and won't accept anything else. I like AsciiDoc more than Markdown in many ways myself, but I don't think there's much use in swimming against the tide here, now that the RFCs are done and the migration has been in progress for many months; all NixOS documentation has been converted, nixpkgs has Markdown documentation in options itself, all the tools take only one input format with DocBook support having been removed, so I don't think there would be any appetite to add additional ones now. It'd impede sharing code with NixOS too.

(I was going to work on doing the same conversion for Home Manager based on https://github.com/nix-community/home-manager/issues/4142#issuecomment-1604831136, but if you have reservations then I'll hold off.)

rycee:
@emilazy Thanks for the response. Will have to do some experiments to see what would be the best course of action. For now, at least, I've updated nmd to handle the recent Nixpkgs change.

emilazy:
Sure; either way the Home Manager NixOS/nix-darwin module needs to move over to Markdown documentation, as it's broken right now under latest nixpkgs.

.github/workflows/test.yml
LnL7: I think explicit warnings are fine in this case, but the stub is also good.

ps. After thinking about it a bit more. With a nixpkgs release nothing really breaks, it will just stop receiving updates until they explicitly update. So handling a bit more compatibility here is kind of the tradeoff for not having releases (which seems a bit overkill given the low rate of change).

LnL7 (approved):
I only roughly went through the changes, but it's all either the documentation build or option descriptions so worst case there are some quirks that can easily be fixed afterwards.

LnL7:
Thanks again for tackling this!

Issue #709, opened 2023-06-25
Github pages doc site broken


vedkothavade:
[The doc site](https://daiderd.com/nix-darwin/manual/index.html#sec-options) now returns the error `The nix-darwin documentation now requires nixpkgs 23.05 to build.`

I'd guess the Github Action which is used for the site was not correctly updated in #707, but not familiar enough with Nix or Github Actions to be sure.

For anyone else looking for an alternative for the time being, as the README says, you can use `man 5 configuration.nix`

emilazy:
Yeah I missed https://github.com/LnL7/nix-darwin/blob/ace0ef2a26bc428e768a5926eae95a63cfcf8b52/.github/workflows/update-manual.yml, whoops. I'll send a PR later if nobody else gets to it first.

PR #710, opened 2023-06-25
ci: update manual updater to 23.05

emilazy:
Fix https://github.com/LnL7/nix-darwin/issues/709.


LnL7 (approved):
That's not the intent of the stubs 😁

emilazy:
Yeah I guess an error would have been more useful here ^^;

Looks like it's working now!

PR #711, opened 2023-06-26
tailscale: improve MagicDNS setup

Enzime:
- Removed `services.tailscale.domain` as Tailscale automatically sets the search domains by creating `/etc/resolver/search.tailscale`
- Fixed MagicDNS when `Override local DNS` is enabled and `100.100.100.100` is not set as a DNS server by creating `/etc/resolver/ts.net` (https://github.com/tailscale/tailscale/issues/8436)
- Removed `services.tailscale.magicDNS` as MagicDNS now works with or without adding `100.100.100.100` as a DNS server
- Added option for implementing `Override local DNS` by setting `100.100.100.100` as the sole DNS server locally until it is implemented in tailscaled (https://github.com/tailscale/tailscale/issues/1338)

cc @sellout

### DIFF
--- a/modules/services/tailscale.nix
+++ b/modules/services/tailscale.nix
@@ -7,13 +7,12 @@ let
 
 in
 {
-  options.services.tailscale = {
-    domain = mkOption {
-      type = types.str;
-      default = "";
-      description = lib.mdDoc "The Tailscale domain. This is displayed at the top left of https://login.tailscale.com/admin, next to the Tailscale logo.";
-    };
+  imports = [
+    (mkRemovedOptionModule [ "services" "tailscale" "domain" ] "Tailscale no longer requires setting the search domain manually.")
+    (mkRemovedOptionModule [ "services" "tailscale" "magicDNS" ] "MagicDNS no longer requires overriding the DNS servers, if this is necessary you can use `services.tailscale.overrideLocalDns`.")
+  ];
 
+  options.services.tailscale = {
     enable = mkEnableOption (lib.mdDoc "Tailscale client daemon");
 
     package = mkOption {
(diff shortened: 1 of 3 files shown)


sellout:
I’ll try to test this out today.

I’ve had a note to implement the third bullet point for months, but your change looks more complete, and likely fixes the issue I’ve been having with mine 🤞🏼

modules/services/tailscale.nix
```diff
@@ -23,26 +22,32 @@ in
...
     };
 
-    magicDNS = {
-      enable = mkOption {
-        type = types.bool;
-        default = false;
-        example = true;
-        description = lib.mdDoc "Whether to configure networking to work with Tailscale's MagicDNS.";
-      };
+    overrideLocalDns = mkOption {
```
sellout, line 25: I think the general style across Nix is to uppercase acronyms like DNS (there are a few other examples just in this diff), but I’m not a stickler.

sellout (approved):
Sorry this took longer to get to than I planned.

I just tested it, and it works for me. And it appears to fix the intermittent issue I had been having with MagicDNS.

Enzime:
@domenkozar can you take a look at this PR?

domenkozar:
How does this compare to NixOS module, usually we try to keep them in sync.

Enzime:
@domenkozar The services are quite different due to platform specific differences.

`services.tailscale.magicDNS` and `services.tailscale.domain` never existed on NixOS as Tailscale is able to set those values correctly without our assistance. Similarly `services.tailscale.overrideLocalDns` is only necessary on macOS as Tailscale needs us to set the DNS servers for it.

domenkozar (approved):

PR #713, opened 2023-06-27
Dock icon magnification settings

pik694:
- add dock icon magnification settings

### DIFF
--- a/modules/system/defaults/dock.nix
+++ b/modules/system/defaults/dock.nix
@@ -165,18 +165,17 @@ in {
 
     system.defaults.dock.magnification = mkOption {
       type = types.nullOr types.bool;
-      default = false;
+      default = null;
       description = lib.mdDoc ''
         Magnify icon on hover. The default is false.
       '';
     };
 
     system.defaults.dock.largesize = mkOption {
-      type = types.nullOr types.int;
+      type = types.nullOr (types.ints.between 16 128);
       default = null;
       description = lib.mdDoc ''
-        Hover maginified icon size. The default is null.
-        Valid values can range from 16 up to 128.
+        Magnified icon size on hover. The default is 16.
       '';
     };
    
--- a/modules/system/defaults/dock.nix
+++ b/modules/system/defaults/dock.nix
@@ -163,6 +163,23 @@ in {
       '';
     };
 
+    system.defaults.dock.magnification = mkOption {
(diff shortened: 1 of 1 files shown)


modules/system/defaults/dock.nix
```diff
@@ -163,6 +163,23 @@ in {
...
+    system.defaults.dock.magnification = mkOption {
+      type = types.nullOr types.bool;
+      default = false;
+      description = lib.mdDoc ''
+        Magnify icon on hover.  The default is false.
+      '';
+    };
+
+    system.defaults.dock.largesize = mkOption {
+      type = types.nullOr types.int;
```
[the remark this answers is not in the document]
pik694: for me the values can range from 16 up to 128 - I can add it in the documentation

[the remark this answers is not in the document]
Enzime:
```suggestion
      type = types.nullOr (types.ints.between 16 128);
```

modules/system/defaults/dock.nix
```diff
@@ -163,6 +163,24 @@ in {
...
+      description = lib.mdDoc ''
+        Magnify icon on hover. The default is false.
+      '';
+    };
+
+    system.defaults.dock.largesize = mkOption {
+      type = types.nullOr types.int;
+      default = null;
+      description = lib.mdDoc ''
+        Hover maginified icon size. The default is null.
```
Enzime, line 178:
```suggestion
        Magnified icon size on hover.
```

modules/system/defaults/dock.nix
```diff
@@ -163,6 +163,24 @@ in {
       '';
     };
 
+    system.defaults.dock.magnification = mkOption {
+      type = types.nullOr types.bool;
+      default = false;
```
Enzime, line 168: Should we make `null` the default so we don't override this setting for users who don't set a value in `nix-darwin`?

Enzime (changes requested):
Thanks for the PR :+1:

modules/system/defaults/dock.nix
pik694: sure, you're right! I have prepared these changes sometime ago, and should have pruned the code before creating PR 🙂

Enzime (approved):
LGTM

PR #716, opened 2023-06-29
fix(launchd): improve `StartCalendarInterval`

tmillr:
Stricter launchd -> StartCalendarInterval type:

- Verify that the integers passed to `Minute`, `Hour`, etc. are within range.

- When provided, the value for StartCalendarInterval must be a non-empty list of calendar intervals and must not contain duplicates entries (throw an error otherwise).

- For increased flexibility and backwards-compatibility, allow an attrset to be passed as well (which will be type-checked and is functionally equivalent to passing a singleton list). Allowing an attrset or list is precisely in-line with what `launchd.plist(5)` accepts for StartCalendarInterval.

Migrate `nix.gc.interval` and `nix.optimise.interval` over to use this new type, and update their defaults to run weekly instead of daily.

Create `modules/launchd/types.nix` file for easier/modular use of launchd types needed in multiple files.

Documentation:

- Update and improve wording/documentation of launchd's `StartCalendarInterval`.

- Improve wording/documentation of `nix.gc.interval` and `nix.optimise.interval` ("time interval" can be misleading as it's actually a "calendar interval"; e.g. `{ Hour = 3; Minute = 15;}` runs daily, not every 3.25 hours).

### DIFF
--- a/modules/services/nix-gc/default.nix
+++ b/modules/services/nix-gc/default.nix
@@ -6,6 +6,7 @@ with lib;
 
 let
   cfg = config.nix.gc;
+  launchdTypes = import ../../launchd/types.nix { inherit config lib; };
 in
 
 {
@@ -35,9 +36,13 @@ in
       };
 
       interval = mkOption {
-        type = types.attrs;
-        default = { Hour = 3; Minute = 15; };
-        description = "The time interval at which the garbage collector will run.";
+        type = launchdTypes.StartCalendarInterval;
+        default = [{ Weekday = 7; Hour = 3; Minute = 15; }];
+        description = ''
+          The calendar interval at which the garbage collector will run.
+          See the {option}`serviceConfig.StartCalendarInterval` option of
+          the {option}`launchd` module for more info.
+        '';
       };
 
       options = mkOption {
@@ -63,7 +68,7 @@ in
       command = "${config.nix.package}/bin/nix-collect-garbage ${cfg.options}";
       environment.NIX_REMOTE = optionalString config.nix.useDaemon "daemon";
       serviceConfig.RunAtLoad = false;
-      serviceConfig.StartCalendarInterval = [ cfg.interval ];
+      serviceConfig.StartCalendarInterval = cfg.interval;
       serviceConfig.UserName = cfg.user;
     };
 
--- a/modules/launchd/launchd.nix
+++ b/modules/launchd/launchd.nix
@@ -2,6 +2,10 @@
 
 with lib;
 
+let
+  launchdTypes = import ./types.nix { inherit config lib; };
+in
+
 {
   options = {
     Label = mkOption {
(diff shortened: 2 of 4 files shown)


modules/services/nix-gc/default.nix
```diff
@@ -35,9 +36,12 @@ in
       };
 
       interval = mkOption {
-        type = types.attrs;
+        type = let
+          t = launchdConfig.options.StartCalendarInterval.type;
+        in
+          types.addCheck (t // {description = lib.removePrefix "null or" t.description;}) (x: x != null);
```
emilazy, line 42: Thank you, this is great! Would you mind making a `modules/launchd/types.nix` and exporting the type without the `nullOr` from it, so that it can be used directly in both `launchd.nix` and `nix-gc/default.nix` instead of having to do this hack?

emilazy (changes requested):

steveej:
hey, since #611 was dropped in favor of this i'm interested in getting this PR merged.

@tmillr thanks for the work on this. do you have bandwidth to see this through or would you prefer someone else continues here?

tmillr:
I'm not sure when I'll be able to finish this. I'm fine with someone else working on this pr or finishing it. Thanks

tmillr:
I've had more time lately and so I've decided I'd try to get this finished once and for all (hopefully within the next day or 2). Sorry for the delay.

tmillr:
## Updates

### Changes

- Calendar interval must be a list (when provided). I did this because allowing an attrset as well ruined the documentation.
- Create `launchd/types.nix` file
- Update option documentation
- Add type-check for duplicate interval specs (i.e. attrsets) within the provided list
- Saner default for gc interval (i.e. the default is now to run weekly instead of daily)

### Questions/Concerns/Etc.

- An error message produced by the new type-checking might not be very detailed/helpful? (e.g. an error due to finding a duplicate interval spec within the list, or, an error due to the list being empty)
- A warning should probably be emitted if an empty attrset is found in the list? (because I am pretty sure that this will result in a job which runs every minute). I don't think Nix recommends emitting warnings within the type-check itself though, so I imagine that this would have to be done anywhere/everywhere `StartCalendarInterval` is used?
- Similarly, there's no check for if there's a list containing an empty attrset in it along with others which aren't empty (wacky edge-case, probably not a big deal? considering the above is implemented at some point)

emilazy:
Thanks for all your work on this! I will try to take a look in the next couple days.

steveej:
@emilazy if i'm not mistaken this is waiting for your review.

tmillr:
Hi @emilazy. I think it's good to go now if you'd like to review it again. I've addressed the merge conflicts (which mainly had to do with the recent removal of `lib.mdDoc` in def1e23). All checks are passing and I've also done some manual testing locally. Let me know if anything else needs to be changed. Thanks

emilazy:
Thank you for all the work you’ve put into this, and I’m sorry for my prolonged absence! Catching up now and will try to review this in the next few days. Please do feel free to ping me if it slips through the cracks.

emilazy (approved):
Looks great, thank you for sticking with this for so long! I’m wondering what the failure case would be on the launchd end if a duplicate was inadvertently included? The attention to detail to getting a good error message on the nix-darwin end for that case is much appreciated, but I wonder how much we’re getting for the implementation complexity. I’m going to land this now, though, and we can worry about that tradeoff later. Thank you again for all your work on this.

Issue #649, opened 2023-05-03
flakes: error when using `nix-darwin` input from registry


edrex:
Hi, I'm just getting my first nix-darwin host set up using flakes and flake-parts, and I hit a little snag.

The flake registry entry is [called]([modules/system/flake-overrides.nix](https://github.com/LnL7/nix-darwin/blob/master/modules/system/flake-overrides.nix)) `nix-darwin`, while [modules/system/flake-overrides.nix](https://github.com/LnL7/nix-darwin/blob/master/modules/system/flake-overrides.nix) assumes the nix-darwin input is named `darwin` (along with all the example code I've seen). This results in the following error when omitting the explicit input from the flake:

```
error: attribute 'darwin' missing

       at /nix/store/pkwvghi6y6abzdfk4yfyxayixs5wzq77-source/modules/system/flake-overrides.nix:5:4:

            4|
            5| let
             |    ^
            6|   inherit (inputs) darwin nixpkgs;
```

Workaround:

```
    inputs.nix-darwin.inputs.nixpkgs.follows = "nixpkgs"; # this line included just for context
    inputs.darwin.follows = "nix-darwin";
```

I think the registry entry just needs to be renamed. Without context, `nix-darwin` does seem preferable since `darwin` sounds like the actual darwin codebase, but having the registry entry follow the established precedent seems like the right choice.

Reporting here rather than there first because the registry is a self-service commons, and it will affect this repo.

emilazy:
This should be fixed by https://github.com/LnL7/nix-darwin/pull/723. Changing the flake registry entry is still an option, but it might be better for us to standardize on the `nix-darwin` name instead, as it matches the repository name and what people call the project in practice.

Issue #718, opened 2023-07-01
Updating darwin channel stops building 


johnykifle:
After trying to to update my darwin channel with

``` nix-channel --update darwin  ```

I no longer can build my configuration using  `darwin-rebuild switch` .

###Error
``` 
error: attribute 'nixos-render-docs' missing

       at /nix/store/p5r83kdljsz507jyk2l6cb00lz7v2gsi-darwin/darwin/doc/manual/default.nix:108:29:

          107|   manpages = runCommand "darwin-manpages"
          108|     { nativeBuildInputs = [ buildPackages.nixos-render-docs ];
             |                             ^
          109|       allowedReferences = ["out"];
```

OS: Mac Ventura 13.4.1

NB.  Am using home-manager (v 22.11)  together with nix-darwin.

**Comment by StephenWithPH** [comment]
It seems that recent commit 69648c6cbbecf34c327d73e36b5aed32edfb9ed9 introduced this as part of #707.

Looking around, I think this might be similar (somehow?) to #701. Of note, disabling documentation via home-manager didn't help.

Of note, I'm also using home-manager (with flakes).

(cc @emilazy)

**Comment by StephenWithPH** [comment]
Ok, setting both of the the *nix-darwin* options https://daiderd.com/nix-darwin/manual/index.html#opt-documentation.doc.enable and https://daiderd.com/nix-darwin/manual/index.html#opt-documentation.man.enable to `false` allows me to avoid the error. Of course, this is a stopgap workaround.

Those settings prevent both https://github.com/LnL7/nix-darwin/blob/b06bab83bdf285ea0ae3c8e145a081eb95959047/doc/manual/default.nix#L60-L61 and https://github.com/LnL7/nix-darwin/blob/b06bab83bdf285ea0ae3c8e145a081eb95959047/doc/manual/default.nix#L107-L108 from breaking.

**Comment by emilazy** [comment]
Thanks for the report; sorry for breaking your systems!

Can I ask what nixpkgs version you are both using? Is it 22.11 (guessing from @johnykifle's Home Manager version)?

**Comment by StephenWithPH** [comment]
> Can I ask what nixpkgs version you are both using? Is it 22.11

Indeed it is:

```
  inputs = {
    nixpkgs-darwin.url = "github:nixos/nixpkgs/nixpkgs-22.11-darwin";
...
```

emilazy:
[...]

Issue #669, opened 2023-05-26
Flake: When `inputs` are in `specialArgs` the wrong nixpkgs is used


dacioromero:
https://github.com/LnL7/nix-darwin/issues/365#issuecomment-1445300112
https://github.com/LnL7/nix-darwin/issues/365#issuecomment-1490761162

---

I also ran into the same oddity as @n8henrie and I thought it deserved a new issue.

I pass all of my Flake `inputs` to `specialArgs` as `{ inherit inputs }`. This ended up meaning even when I attempted to have nix-darwin's `nixpkgs` input follow my own `nixpkgs-unstable` input (nixos-22.11) it would still use my `nixpkgs` input (nixos-unstable).

In NixOS `_module.args.pkgs` is actually set by the nixpkgs module (https://github.com/NixOS/nixpkgs/blob/c2f6e54e7ffd06d8976846d1dfd94df68912cb17/nixos/modules/misc/nixpkgs.nix#L339-L341) and the line linked before (https://github.com/NixOS/nixpkgs/blob/463643afa8b70b153ef1728493afce53817514ff/nixos/lib/eval-config.nix#L57) is for if you provide `pkgs` to `nixosSystem`.

I'm not sure what the best solution is, but my gut feeling is that the inputs of nix-darwin's flake could be `_inputs` as to not collide with user-provided inputs.

dacioromero:
Related problem I've had is that https://github.com/LnL7/nix-darwin/blob/master/modules/system/flake-overrides.nix will error if the name of my input for this repo is just `darwin` instead of `nix-darwin`.

n8henrie:
Hmmm, I'm AFK at the moment but the merged PR mentions it is a breaking change.

Is there a summary somewhere of what changes may be required, and under what conditions? Based on my quick-and-dirty read of the PR, it looks like this brings nix-darwin in line with NixOS with regards to handling specialArgs. Does that sound right?

emilazy:
Nothing should break unless you were importing `eval-config.nix` directly (and you'd know if you were). Any working configuration used via `darwin-rebuild`, flakes or not, should act as before, modulo some potential very minor changes (different/missing Nixpkgs revisions in system configuration labels and the manual). The main change is an overhaul of how `pkgs` and `lib` are constructed and passed to fix some nasty bugs, but as a side effect nix-darwin now doesn't care what your `inputs` are at all.

n8henrie:
Thanks for clarification.

PR #723, opened 2023-07-08
Rationalize handling of Nixpkgs

emilazy:
This is a big one to fix some very broken behaviour (and broken configurations) as described in https://github.com/LnL7/nix-darwin/issues/718 and https://github.com/LnL7/nix-darwin/issues/669. I'm submitting this as a draft for now as it's a big change and I don't trust myself to give it a final check until I've slept, but I wanted to get this out for people to look at. See the commit messages for copious verbiage justifying and explaining these changes :)


modules/nix/nixpkgs.nix
```diff
@@ -44,72 +90,258 @@ let
     check = lib.isFunction;
     merge = lib.mergeOneOption;
   };
+
+  pkgsType = types.pkgs // {
```
reckenrode, line 94: `types.pkgs` was added in 23.05. What’s the minimum nixpkgs supported by nix-darwin?

emilazy: We don't really have any, but dropping 22.11 support immediately in the docs PR was considered too soon so I've pushed another backporting workaround. (We really need release branches.)

emilazy:
Okay, I slept on this and I think it's pretty good. Marking as ready for review.

cc @LnL7 @domenkozar; the https://github.com/LnL7/nix-darwin/issues/718#issuecomment-1626123258 bug is pretty bad even when not breaking configuration evaluation and it looks like multiple people have been running into it. Sadly clearing it up to prevent future reoccurrences required some rework.

emilazy:
Okay never mind all the tests are failing, sorry. I'll fix this today.

emilazy:
I forgot that `or` has weird precedence and that you need to be able to run the installer without a configuration :) All tests pass now, and the one `default.nix` backwards-compatibility break is gone.

emilazy:
As suggested on Matrix, here's the diff between the current NixOS `nixpkgs` module and the modified version in this PR, to ease review. The differences amount to

* Backporting a few required `lib` things for compatibility down to Nixpkgs 22.11
* Removing legacy options that are deprecated in NixOS (so that we don't add born-deprecated options)
* Adding the `nixpkgs.source` option that we need due to being out of tree
* Exporting the "was `pkgs` constructed by this module?" logic so other modules can use it
* A few simple NixOS → nix-darwin adaptations (e.g. documentation, examples, imports)

```diff
diff --git a/nixpkgs.nix b/modules/nix/nixpkgs.nix
index f9d8bcc..52aec9a 100644
--- a/nixpkgs.nix
+++ b/modules/nix/nixpkgs.nix
@@ -3,6 +3,48 @@
 with lib;
 
 let
+
+  # Backport from Nixpkgs 23.05
+  defaultOverridePriority =
+    lib.modules.defaultOverridePriority or lib.modules.defaultPriority;
+
+  # Backport from Nixpkgs 23.11
+  mergeAttrDefinitionsWithPrio = lib.mergeAttrDefinitionsWithPrio or (opt:
+    let
+        # Inlined to avoid warning about using internal APIs 🥴
+        pushDownProperties = cfg:
+          if cfg._type or "" == "merge" then
+            concatMap pushDownProperties cfg.contents
+          else if cfg._type or "" == "if" then
+            map (mapAttrs (n: v: mkIf cfg.condition v)) (pushDownProperties cfg.content)
+          else if cfg._type or "" == "override" then
+            map (mapAttrs (n: v: mkOverride cfg.priority v)) (pushDownProperties cfg.content)
+          else # FIXME: handle mkOrder?
+            [ cfg ];
+
+        defsByAttr =
+          lib.zipAttrs (
+            lib.concatLists (
+              lib.concatMap
+                ({ value, ... }@def:
+                  map
+                    (lib.mapAttrsToList (k: value: { ${k} = def // { inherit value; }; }))
+                    (pushDownProperties value)
+                )
+                opt.definitionsWithLocations
+            )
+          );
+    in
+      assert opt.type.name == "attrsOf" || opt.type.name == "lazyAttrsOf";
+      lib.mapAttrs
+            (k: v:
+              let merging = lib.mergeDefinitions (opt.loc ++ [k]) opt.type.nestedTypes.elemType v;
+              in {
+                value = merging.mergedValue;
+                inherit (merging.defsFinal') highestPrio;
+              })
+            defsByAttr);
+
   cfg = config.nixpkgs;
   opt = options.nixpkgs;
 
@@ -49,7 +91,9 @@ let
     merge = lib.mergeOneOption;
   };
 
-  pkgsType = types.pkgs // {
+  # TODO: Remove backwards compatibility hack when dropping
+  # 22.11 support.
+  pkgsType = types.pkgs or (types.uniq types.attrs) // {
     # This type is only used by itself, so let's elaborate the description a bit
     # for the purpose of documentation.
     description = "An evaluation of Nixpkgs; the top level attribute set of packages";
@@ -64,9 +108,7 @@ let
   buildPlatformLine = optionalString hasBuildPlatform "${showOptionWithDefLocs opt.buildPlatform}";
 
   legacyOptionsDefined =
-    optional (opt.localSystem.highestPrio < (mkDefault {}).priority) opt.system
-    ++ optional (opt.localSystem.highestPrio < (mkOptionDefault {}).priority) opt.localSystem
-    ++ optional (opt.crossSystem.highestPrio < (mkOptionDefault {}).priority) opt.crossSystem
+    optional (opt.system.highestPrio < (mkDefault {}).priority) opt.system
     ;
 
   defaultPkgs =
@@ -83,12 +125,13 @@ let
               localSystem = cfg.hostPlatform;
             };
       in
-      import ../../.. ({
+      import cfg.source ({
         inherit (cfg) config overlays;
       } // systemArgs)
     else
-      import ../../.. {
-        inherit (cfg) config overlays localSystem crossSystem;
+      import cfg.source {
+        inherit (cfg) config overlays;
+        localSystem = { inherit (cfg) system; };
       };
 
   finalPkgs = if opt.pkgs.isDefined then cfg.pkgs.appendOverlays cfg.overlays else defaultPkgs;
@@ -96,52 +139,28 @@ let
 in
 
 {
-  imports = [
-    ./assertions.nix
-    ./meta.nix
-    (mkRemovedOptionModule [ "nixpkgs" "initialSystem" ] "The NixOS options `nesting.clone` and `nesting.children` have been deleted, and replaced with named specialisation. Therefore `nixpgks.initialSystem` has no effect anymore.")
-  ];
-
   options.nixpkgs = {
-
     pkgs = mkOption {
-      defaultText = literalExpression ''
-        import "''${nixos}/.." {
-          inherit (cfg) config overlays localSystem crossSystem;
-        }
-      '';
       type = pkgsType;
       example = literalExpression "import <nixpkgs> {}";
       description = lib.mdDoc ''
-        If set, the pkgs argument to all NixOS modules is the value of
+        If set, the pkgs argument to all nix-darwin modules is the value of
         this option, extended with `nixpkgs.overlays`, if
-        that is also set. Either `nixpkgs.crossSystem` or
-        `nixpkgs.localSystem` will be used in an assertion
-        to check that the NixOS and Nixpkgs architectures match. Any
-        other options in `nixpkgs.*`, notably `config`,
+        that is also set. The nix-darwin and Nixpkgs architectures must
+        match. Any other options in `nixpkgs.*`, notably `config`,
         will be ignored.
 
-        If unset, the pkgs argument to all NixOS modules is determined
-        as shown in the default value for this option.
+        The default value imports the Nixpkgs from
+        [](#opt-nixpkgs.source). The `config`, `overlays`, `localSystem`,
+        and `crossSystem` are based on this option's siblings.
 
-        The default value imports the Nixpkgs source files
-        relative to the location of this NixOS module, because
-        NixOS and Nixpkgs are distributed together for consistency,
-        so the `nixos` in the default value is in fact a
-        relative path. The `config`, `overlays`,
-        `localSystem`, and `crossSystem` come
-        from this option's siblings.
-
-        This option can be used by applications like NixOps to increase
+        This option can be used to increase
         the performance of evaluation, or to create packages that depend
         on a container that should be built with the exact same evaluation
         of Nixpkgs, for example. Applications like this should set
         their default value using `lib.mkDefault`, so
         user-provided configuration can override it without using
         `lib`.
-
-        Note that using a distinct version of Nixpkgs with NixOS may
-        be an unexpected source of problems. Use this option with care.
       '';
     };
 
@@ -191,14 +210,12 @@ in
 
     hostPlatform = mkOption {
       type = types.either types.str types.attrs; # TODO utilize lib.systems.parsedPlatform
-      example = { system = "aarch64-linux"; config = "aarch64-unknown-linux-gnu"; };
+      example = { system = "aarch64-darwin"; config = "aarch64-apple-darwin"; };
       # Make sure that the final value has all fields for sake of other modules
       # referring to this. TODO make `lib.systems` itself use the module system.
       apply = lib.systems.elaborate;
-      defaultText = literalExpression
-        ''(import "''${nixos}/../lib").lib.systems.examples.aarch64-multiplatform'';
       description = lib.mdDoc ''
-        Specifies the platform where the NixOS configuration will run.
+        Specifies the platform where the nix-darwin configuration will run.
 
         To cross-compile, set also `nixpkgs.buildPlatform`.
 
@@ -209,16 +226,16 @@ in
     buildPlatform = mkOption {
       type = types.either types.str types.attrs; # TODO utilize lib.systems.parsedPlatform
       default = cfg.hostPlatform;
-      example = { system = "x86_64-linux"; config = "x86_64-unknown-linux-gnu"; };
+      example = { system = "x86_64-darwin"; config = "x86_64-apple-darwin"; };
       # Make sure that the final value has all fields for sake of other modules
       # referring to this.
       apply = lib.systems.elaborate;
       defaultText = literalExpression
         ''config.nixpkgs.hostPlatform'';
       description = lib.mdDoc ''
-        Specifies the platform on which NixOS should be built.
-        By default, NixOS is built on the system where it runs, but you can
-        change where it's built. Setting this option will cause NixOS to be
+        Specifies the platform on which nix-darwin should be built.
+        By default, nix-darwin is built on the system where it runs, but you can
+        change where it's built. Setting this option will cause nix-darwin to be
         cross-compiled.
 
         For instance, if you're doing distributed multi-platform deployment,
@@ -229,64 +246,9 @@ in
       '';
     };
 
-    localSystem = mkOption {
-      type = types.attrs; # TODO utilize lib.systems.parsedPlatform
-      default = { inherit (cfg) system; };
-      example = { system = "aarch64-linux"; config = "aarch64-unknown-linux-gnu"; };
-      # Make sure that the final value has all fields for sake of other modules
-      # referring to this. TODO make `lib.systems` itself use the module system.
-      apply = lib.systems.elaborate;
-      defaultText = literalExpression
-        ''(import "''${nixos}/../lib").lib.systems.examples.aarch64-multiplatform'';
-      description = lib.mdDoc ''
-        Systems with a recently generated `hardware-configuration.nix`
-        do not need to specify this option, unless cross-compiling, in which case
-        you should set *only* {option}`nixpkgs.buildPlatform`.
-
-        If this is somehow not feasible, you may fall back to removing the
-        {option}`nixpkgs.hostPlatform` line from the generated config and
-        use the old options.
-
-        Specifies the platform on which NixOS should be built. When
-        `nixpkgs.crossSystem` is unset, it also specifies
-        the platform *for* which NixOS should be
-        built.  If this option is unset, it defaults to the platform
-        type of the machine where evaluation happens. Specifying this
-        option is useful when doing distributed multi-platform
-        deployment, or when building virtual machines. See its
-        description in the Nixpkgs manual for more details.
-
-        Ignored when `nixpkgs.pkgs` or `hostPlatform` is set.
-      '';
-    };
-
-    # TODO deprecate. "crossSystem" is a nonsense identifier, because "cross"
-    #      is a relation between at least 2 systems in the context of a
-    #      specific build step, not a single system.
-    crossSystem = mkOption {
-      type = types.nullOr types.attrs; # TODO utilize lib.systems.parsedPlatform
-      default = null;
-      example = { system = "aarch64-linux"; config = "aarch64-unknown-linux-gnu"; };
-      description = lib.mdDoc ''
-        Systems with a recently generated `hardware-configuration.nix`
-        may instead specify *only* {option}`nixpkgs.buildPlatform`,
-        or fall back to removing the {option}`nixpkgs.hostPlatform` line from the generated config.
-
-        Specifies the platform for which NixOS should be
-        built. Specify this only if it is different from
-        `nixpkgs.localSystem`, the platform
-        *on* which NixOS should be built. In other
-        words, specify this to cross-compile NixOS. Otherwise it
-        should be set as null, the default. See its description in the
-        Nixpkgs manual for more details.
-
-        Ignored when `nixpkgs.pkgs` or `hostPlatform` is set.
-      '';
-    };
-
     system = mkOption {
       type = types.str;
-      example = "i686-linux";
+      example = "x86_64-darwin";
       default =
         if opt.hostPlatform.isDefined
         then
@@ -298,34 +260,43 @@ in
         else
           throw ''
             Neither ${opt.hostPlatform} nor the legacy option ${opt.system} has been set.
-            You can set ${opt.hostPlatform} in hardware-configuration.nix by re-running
-            a recent version of nixos-generate-config.
-            The option ${opt.system} is still fully supported for NixOS 22.05 interoperability,
+            The option ${opt.system} is still fully supported for interoperability,
             but will be deprecated in the future, so we recommend to set ${opt.hostPlatform}.
           '';
       defaultText = lib.literalMD ''
-        Traditionally `builtins.currentSystem`, but unset when invoking NixOS through `lib.nixosSystem`.
+        Traditionally `builtins.currentSystem`, but unset when invoking nix-darwin through `lib.darwinSystem`.
       '';
       description = lib.mdDoc ''
-        This option does not need to be specified for NixOS configurations
-        with a recently generated `hardware-configuration.nix`.
+        Specifies the Nix platform type on which nix-darwin should be built.
+        It is better to specify `nixpkgs.hostPlatform` instead.
 
-        Specifies the Nix platform type on which NixOS should be built.
-        It is better to specify `nixpkgs.localSystem` instead.
-        ```
-        {
-          nixpkgs.system = ..;
-        }
-        ```
-        is the same as
-        ```
-        {
-          nixpkgs.localSystem.system = ..;
-        }
-        ```
-        See `nixpkgs.localSystem` for more information.
+        Ignored when `nixpkgs.pkgs` or `nixpkgs.hostPlatform` is set.
+      '';
+    };
 
-        Ignored when `nixpkgs.pkgs`, `nixpkgs.localSystem` or `nixpkgs.hostPlatform` is set.
+    # nix-darwin only
+
+    source = mkOption {
+      type = types.path;
+      defaultText = literalMD ''
+        `<nixpkgs>` or nix-darwin's `nixpkgs` flake input
+      '';
+      description = lib.mdDoc ''
+        The path to import Nixpkgs from. If you're setting a custom
+        [](#opt-nixpkgs.pkgs) or `_module.args.pkgs`, setting this
+        to something with `rev` and `shortRev` attributes (such as a
+        flake input or `builtins.fetchGit` result) will also set
+        `system.nixpkgsRevision` and related options.
+        (nix-darwin only)
+      '';
+    };
+
+    constructedByUs = mkOption {
+      type = types.bool;
+      internal = true;
+      description = ''
+        Whether `pkgs` was constructed by this module. This is false when any of
+        `nixpkgs.pkgs` or `_module.args.pkgs` is set. (nix-darwin only)
       '';
     };
   };
@@ -339,39 +310,29 @@ in
         # which is somewhat costly for Nixpkgs. With an explicit priority, we only
         # evaluate the wrapper to find out that the priority is lower, and then we
         # don't need to evaluate `finalPkgs`.
-        lib.mkOverride lib.modules.defaultOverridePriority
+        lib.mkOverride defaultOverridePriority
           finalPkgs.__splicedPackages;
     };
 
-    assertions = let
-      # Whether `pkgs` was constructed by this module. This is false when any of
-      # nixpkgs.pkgs or _module.args.pkgs is set.
-      constructedByMe =
-        # We set it with default priority and it can not be merged, so if the
-        # pkgs module argument has that priority, it's from us.
-        (lib.modules.mergeAttrDefinitionsWithPrio options._module.args).pkgs.highestPrio
-          == lib.modules.defaultOverridePriority
-        # Although, if nixpkgs.pkgs is set, we did forward it, but we did not construct it.
-          && !opt.pkgs.isDefined;
-    in [
+    nixpkgs.constructedByUs =
+      # We set it with default priority and it can not be merged, so if the
+      # pkgs module argument has that priority, it's from us.
+      (mergeAttrDefinitionsWithPrio options._module.args).pkgs.highestPrio
+        == defaultOverridePriority
+      # Although, if nixpkgs.pkgs is set, we did forward it, but we did not construct it.
+        && !opt.pkgs.isDefined;
+
+    assertions = [
       (
         let
-          nixosExpectedSystem =
-            if config.nixpkgs.crossSystem != null
-            then config.nixpkgs.crossSystem.system or (lib.systems.parse.doubleFromSystem (lib.systems.parse.mkSystemFromString config.nixpkgs.crossSystem.config))
-            else config.nixpkgs.localSystem.system or (lib.systems.parse.doubleFromSystem (lib.systems.parse.mkSystemFromString config.nixpkgs.localSystem.config));
-          nixosOption =
-            if config.nixpkgs.crossSystem != null
-            then "nixpkgs.crossSystem"
-            else "nixpkgs.localSystem";
           pkgsSystem = finalPkgs.stdenv.targetPlatform.system;
         in {
-          assertion = constructedByMe -> !hasPlatform -> nixosExpectedSystem == pkgsSystem;
-          message = "The NixOS nixpkgs.pkgs option was set to a Nixpkgs invocation that compiles to target system ${pkgsSystem} but NixOS was configured for system ${nixosExpectedSystem} via NixOS option ${nixosOption}. The NixOS system settings must match the Nixpkgs target system.";
+          assertion = cfg.constructedByUs -> !hasPlatform -> cfg.system == pkgsSystem;
+          message = "The nix-darwin nixpkgs.pkgs option was set to a Nixpkgs invocation that compiles to target system ${pkgsSystem} but nix-darwin was configured for system ${darwinExpectedSystem} via nix-darwin option nixpkgs.system. The nix-darwin system settings must match the Nixpkgs target system.";
         }
       )
       {
-        assertion = constructedByMe -> hasPlatform -> legacyOptionsDefined == [];
+        assertion = cfg.constructedByUs -> hasPlatform -> legacyOptionsDefined == [];
         message = ''
           Your system configures nixpkgs with the platform parameter${optionalString hasBuildPlatform "s"}:
           ${hostPlatformLine
@@ -385,7 +346,4 @@ in
       }
     ];
   };
-
-  # needs a full nixpkgs path to import nixpkgs
-  meta.buildDocsInSandbox = false;
 }
```

README.md
```diff
@@ -87,21 +87,21 @@ A minimal example of using an existing configuration.nix:
...
 accessible as an argument `inputs`, similar to `pkgs` and `lib`, inside the configuration.
 
 ```nix
 darwin.lib.darwinSystem {
-  system = "x86_64-darwin";
-  modules = [ ./configuration.nix ];
-  inputs = { inherit darwin dotfiles nixpkgs; };
+  modules = [
+    ./configuration.nix
+    { _module.args.inputs = { inherit darwin dotfiles nixpkgs; }; }
```
Enzime, line 103:
```suggestion
  modules = [ ./configuration.nix ];
  specialArgs = { inputs = { inherit darwin dotfiles nixpkgs; }; };
```

I think we should direct them to use `specialArgs`. It could be worthwhile to rewrite this section as `inputs` is no longer an input to `darwinSystem`, possibly showing the entire flake and showing how to get all the flake inputs as `inputs`.

flake.nix
```diff
@@ -4,19 +4,36 @@
...
+              _module.args.pkgs = lib.mkForce args.pkgs;
+            })
+            # Backwards compatibility shim; TODO: warn?
+            ++ nixpkgs.lib.optional (args ? system) ({ lib, ... }: {
+              nixpkgs.system = lib.mkDefault args.system;
+            })
+            # Backwards compatibility shim; TODO: warn?
+            ++ nixpkgs.lib.optional (args ? inputs) {
+              _module.args.inputs = args.inputs;
+            }
```
Enzime, line 25:
```suggestion
```

I wonder if it would be a good idea to remove this in the future (definitely in a different PR)

flake.nix
```diff
@@ -4,19 +4,36 @@
...
+            ++ [
+              self.darwinModules.flakeOverrides
+
+              ({ lib, ... }: {
+                nixpkgs.source = lib.mkDefault nixpkgs;
+
+                system.darwinVersionSuffix = ".${self.shortRev or "dirty"}";
+                system.darwinRevision = lib.mkIf (self ? rev) self.rev;
+              })
```
Enzime, line 34:
```suggestion
              ({ lib, ... }: {
                nixpkgs.source = lib.mkDefault nixpkgs;
                
                system.checks.verifyNixPath = lib.mkDefault false;
                system.checks.verifyNixChannels = lib.mkDefault false;

                system.darwinVersionSuffix = ".${self.shortRev or "dirty"}";
                system.darwinRevision = lib.mkIf (self ? rev) self.rev;
              })
```

Might be worth just dropping the `flake-overrides` module altogether

modules/system/version.nix
```diff
@@ -56,62 +57,67 @@ in
...
       type = types.str;
+      default = cfg.nixpkgsRelease + cfg.nixpkgsVersionSuffix;
       description = lib.mdDoc "The full nixpkgs version (e.g. `16.03.1160.f2d4ee1`).";
     };
 
     system.nixpkgsVersionSuffix = mkOption {
       internal = true;
       type = types.str;
+      default = if useSourceVersion
+        then ".${lib.substring 0 8 (nixpkgsSrc.lastModifiedDate or nixpkgsSrc.lastModified or "19700101")}.${nixpkgsSrc.shortRev or "dirty"}"
```
Enzime, line 98:
```suggestion
        then ".${lib.substring 0 8 (nixpkgsSrc.lastModifiedDate or nixpkgsSrc.lastModified or "19700101")}.${nixpkgsSrc.shortRev or nixpkgsSrc.dirtyShortRev or "dirty"}"
```

emilazy:
Rebased to factor the last commit into a few more to make review easier. No change in the resulting tree other than one variable rename and one whitespace change, and I checked that tests pass with all commits.

README.md
emilazy: Agreed. I will change accordingly.

flake.nix
emilazy: Yeah, I would like to, but I'd want to put up a deprecation warning first, and I'd rather not introduce warnings in a commit trying to fix people's configuration ^^;

If we had proper release branches I might just rip this out ASAP, but I've erred on the side of just keeping everything working the same as before.

flake.nix
emilazy: Sure.

modules/system/version.nix
emilazy: I'm reluctant to do this because it doesn't match what NixOS does. This is taken directly from the Nixpkgs `flake.nix`.

emilazy:
Rebased to address some of @Enzime's suggestions.

modules/system/version.nix
Enzime: Fair enough, I'll look into submitting a PR upstream in future

README.md
```diff
@@ -87,39 +87,25 @@ A minimal example of using an existing configuration.nix:
...
 accessible as an argument `inputs`, similar to `pkgs` and `lib`, inside the configuration.
 
 ```nix
 darwin.lib.darwinSystem {
-  system = "x86_64-darwin";
   modules = [ ./configuration.nix ];
-  inputs = { inherit darwin dotfiles nixpkgs; };
+  specialArgs = { inherit inputs; };
 }
 # in configuration.nix:
```
Enzime, line 104:
```suggestion
}

# in configuration.nix
```

Alternatively, moving the `configuration.nix` into a separate code block could look better

README.md
```diff
@@ -87,39 +87,25 @@ A minimal example of using an existing configuration.nix:
...
+Inputs from the flake can also be passed into `darwinSystem`. These inputs are then
 accessible as an argument `inputs`, similar to `pkgs` and `lib`, inside the configuration.
 
 ```nix
 darwin.lib.darwinSystem {
-  system = "x86_64-darwin";
   modules = [ ./configuration.nix ];
-  inputs = { inherit darwin dotfiles nixpkgs; };
+  specialArgs = { inherit inputs; };
 }
```
Enzime, line 103:
```suggestion
{
  outputs = inputs@{ self, darwin, nixpkgs }: {
    darwinConfigurations."Johns-MacBook" = darwin.lib.darwinSystem {
      modules = [ ./configuration.nix ];
      specialArgs = { inherit inputs; };
    };
  };
}
```

I think this will help show new users how to define `inputs`

README.md
emilazy: Done.

README.md
emilazy: I added `inputs@` to the flake in the previous code block and edited the `configuration.nix` one to reflect the inputs that are in that one. I think that's probably good enough, since it's immediately after the example flake that sets the context for the suggestion. (Otherwise the entire example would have to be duplicated: `darwin` is (frustratingly) not the name in the global Nix registry, and I'm a bit leery about suggesting people use `nixpkgs` with no branch reference.)

emilazy:
Also fixes https://github.com/LnL7/nix-darwin/issues/649 (apparently people run into this a lot!).

Enzime (approved):
LGTM

modules/nix/nixpkgs.nix
```diff
@@ -44,72 +90,260 @@ let
...
+
+  finalPkgs = if opt.pkgs.isDefined then cfg.pkgs.appendOverlays cfg.overlays else defaultPkgs;
+
 in
 
 {
-  options = {
-    nixpkgs.config = mkOption {
+  options.nixpkgs = {
+    pkgs = mkOption {
```
LnL7, line 143: I find that using the full option namespace makes things easier to find, but `options.nixpkgs` is the irrelevant part joined together 😄 

```suggestion
    nixpkgs.pkgs = mkOption {
```

modules/nix/nixpkgs.nix
```diff
@@ -44,72 +90,260 @@ let
...
+        # After all, to determine a definition priority, we need to evaluate `._type`,
+        # which is somewhat costly for Nixpkgs. With an explicit priority, we only
+        # evaluate the wrapper to find out that the priority is lower, and then we
+        # don't need to evaluate `finalPkgs`.
+        lib.mkOverride defaultOverridePriority
+          finalPkgs.__splicedPackages;
+    };
 
-    # _module.args.pkgs is defined in ../../eval-config.nix
+    nixpkgs.constructedByUs =
```
LnL7, line 317: This is rather complicated to the point I can't really follow. Why does it matter where pkgs was set?

default.nix
```diff
@@ -2,23 +2,20 @@
 , configuration ? <darwin-config>
 , lib ? pkgs.lib
 , pkgs ? import nixpkgs { inherit system; }
-, system ? builtins.currentSystem
+, system ? builtins.currentSystem or null
```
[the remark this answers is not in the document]
emilazy: Good question. The flake uses `eval-config.nix` directly so I no longer remember why I did this. There are uses for pure evaluation mode outside of flakes, though. I can remove this change if you'd prefer.

flake.nix
```diff
@@ -4,22 +4,37 @@
...
+        // nixpkgs.lib.optionalAttrs (args ? pkgs) { inherit (args.pkgs) lib; }
+        // builtins.removeAttrs args [ "system" "pkgs" "inputs" ]
+        // {
+          modules = modules
+            ++ nixpkgs.lib.optional (args ? pkgs) ({ lib, ... }: {
+              _module.args.pkgs = lib.mkForce args.pkgs;
+            })
+            # Backwards compatibility shim; TODO: warn?
+            ++ nixpkgs.lib.optional (args ? system) ({ lib, ... }: {
+              nixpkgs.system = lib.mkDefault args.system;
```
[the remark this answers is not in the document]
emilazy: In NixOS it is intentionally done like this to allow implicit default systems to be overridden by the configuration. However I guess in the case of the flake entry point there's never any default. I would lean towards keeping the behaviour as-is for consistency with NixOS, but I could also add logic for an error here if you would prefer; I didn't bother doing anything fancy as I'd prefer to deprecate passing `system` altogether and make it warn unconditionally at some point.

flake.nix
```diff
@@ -4,22 +4,37 @@
...
+            })
+            # Backwards compatibility shim; TODO: warn?
+            ++ nixpkgs.lib.optional (args ? inputs) {
+              _module.args.inputs = args.inputs;
+            }
+            ++ [ ({ lib, ... }: {
+              nixpkgs.source = lib.mkDefault nixpkgs;
+
+              system.checks.verifyNixPath = lib.mkDefault false;
+              system.checks.verifyNixChannels = lib.mkDefault false;
```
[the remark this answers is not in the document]
emilazy: Sure, I can remove that line.

modules/nix/nixpkgs.nix
emilazy: This is directly from the upstream NixOS module. I would prefer to avoid any unnecessary divergence, especially cosmetic, to make future synchronization go smoother.

modules/nix/nixpkgs.nix
emilazy: This logic is directly from the upstream NixOS module; it is used to guard the assertions below. I exported it so that the version module can correctly determine whether it knows the origin of Nixpkgs and avoid setting incorrect version information as currently happens; if you override the package set then the `nixpkgs.source` defaults of nix-darwin's `nixpkgs` flake input or your `<nixpkgs>` channel have nothing to do with where `pkgs` came from or its version. Setting `nixpkgs.source` explicitly is available in those cases.

default.nix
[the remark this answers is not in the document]
LnL7: Yes, but with pure evaluation the system argument just becomes mandatory.

flake.nix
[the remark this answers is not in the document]
LnL7: Yeah, guess we can also do that.

modules/nix/nixpkgs.nix
emilazy: cf.

https://github.com/LnL7/nix-darwin/blob/93f92130aec7d7f20d24bfa04b9a6c9be5fb4c9a/modules/system/version.nix#L20-L25

The consequences can be worse than just incorrect links in docs: I almost used it to gate the 22.11 docs backwards compatibility, which would have broken even more configurations than https://github.com/LnL7/nix-darwin/issues/718.

modules/nix/nixpkgs.nix
LnL7: Alright makes sense for this module then, but as a general style guide I lean towards the absolute attributes to make searching easier.

modules/nix/nixpkgs.nix
LnL7: Ah, I thought this was specific to the darwin module.

flake.nix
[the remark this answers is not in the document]
LnL7: Maybe the check isn't useful, but I feel like being able to run eg. `nix-build '<nixpkgs>' -A hello` or a file with `import <nixpkgs> {}` is somewhat part of a working installation.

default.nix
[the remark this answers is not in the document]
emilazy: Fair enough; people can always set it to `null`. Have removed, will push with the other changes.

flake.nix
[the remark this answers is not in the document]
emilazy: FWIW I can put in warnings now if you'd prefer, I was just trying to avoid the PR getting any bigger or the same change that fixes people's systems also warning about them :)

flake.nix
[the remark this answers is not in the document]
emilazy: Did you mean both these checks or only the latter? `verifyNixChannels` seems harmless to me, since it looks like it only fails if you have *broken* channels rather than no channels at all. `verifyNixPath` I think would be a regression for flake setups, since it requires `darwin-config` which doesn't really make sense in that context. (I personally keep a completely blank Nix path and use `NIX_PATH=nixpkgs=flake:nixpkgs` when I have to.)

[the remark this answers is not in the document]
LnL7: Yeah, there's more to say for not verifying NIX_PATH. I meant the broken channels specifically.

modules/nix/nixpkgs.nix
LnL7: WRT the version checking, yeah here you clearly notice (especially with flakes) that nixpkgs-lib and nixpkgs-pkgs should be separate packages.

emilazy:
Pushed a reversion of the `system` fallback in `default.nix` and re-adding `system.checks.verifyNixChannels` for flake configurations. Didn't adjust the `system`/`inputs` compatibility handling but if you want me to add warnings for them I can.

LnL7 (approved):

PR #724, opened 2023-07-09
linux-builder: fix evaluation errors

Enzime:
- `modules` cannot be a nested list
- `nix.settings.extra-trusted-users` might not be set


Enzime:
@domenkozar can you take a look at this hotfix?

emilazy (approved):
Fixes the evaluation for me on `x86_64-darwin`. Can't easily test the build because it seems like Hydra hasn't built the machinery for `linux-builder` yet. Discussed an issue with an assertion on Matrix, but it's orthogonal to this bug.

LnL7 (approved):

PR #725, opened 2023-07-10
trezord: Add launchd user agent service module for configuring trezord

mitchmindtree:
This adds a small module for configuring the trezor-bridge service, trezord. This service enables users to interact with their Trezor hardware wallet through the trezor suite web interface, or to use the device for U2F auth, SSH login, GPG or password mgmt. https://trezor.io/learn/a/what-is-trezor-bridge

The options were copied directly from the nixos service module here: https://github.com/NixOS/nixpkgs/blob/9d6e454b857fb472fa35fc8b098fa5ac307a0d7d/nixos/modules/services/hardware/trezord.nix#L16

The implementation was adapted from the nixos module's systemd service to a launchd user agent.

Tested successfully locally on an Air M2.

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -75,6 +75,7 @@
   ./services/synapse-bt.nix
   ./services/synergy
   ./services/tailscale.nix
+  ./services/trezord.nix
   ./services/wg-quick.nix
   ./services/yabai
   ./services/nextdns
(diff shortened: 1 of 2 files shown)


mmahut (approved):

Enzime (approved):
LGTM

PR #731, opened 2023-07-12
etc: check for existing files during checks stage

emilazy:
This ensures that activation fails early if there are any `/etc` files with unexpected state, rather than leaving the system half-activated.

Here be dragons, obviously; it would be quite bad to get anything wrong. I tried to write this as clearly as possible and I think it's a lot easier to follow than the previous version, but it should still be carefully reviewed. It worked for all the cases I tested. I endeavoured to follow the logic of the previous version as closely as possible, except for two cases, which should both be more conservative than the previous version:

* We don't treat an `/etc` file containing the string `/etc/static` as carte blanche to overwrite it. (Why would we...? This one really confuses me. @LnL7 do you remember what's going on there?)
* We delete links into `/etc/static` only when the file in `/etc/static` directly doesn't exist, not when recursively dereferencing symbolic links leads to a nonexistent file.

### DIFF
--- a/modules/system/etc.nix
+++ b/modules/system/etc.nix
@@ -52,40 +52,40 @@ in
       ln -sfn "$(readlink -f $systemConfig/etc)" /etc/static
 
       errorOccurred=false
-      for f in $(find /etc/static/* -type l); do
-        l=/etc/''${f#/etc/static/}
-        d=''${l%/*}
-        if [ ! -e "$d" ]; then
-          mkdir -p "$d"
+      for etcStaticFile in $(find /etc/static/* -type l); do
+        etcFile=/etc/''${etcStaticFile#/etc/static/}
+        etcDir=''${etcFile%/*}
+        if [ ! -e "$etcDir" ]; then
+          mkdir -p "$etcDir"
         fi
-        if [ -e "$f".copy ]; then
-          cp "$f" "$l"
+        if [ -e "$etcStaticFile".copy ]; then
+          cp "$etcStaticFile" "$etcFile"
           continue
         fi
-        if [ -e "$l" ]; then
-          if [ "$(readlink "$l")" != "$f" ]; then
-            if ! grep -q /etc/static "$l"; then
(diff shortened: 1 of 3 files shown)


modules/system/etc.nix
```diff
@@ -42,64 +39,103 @@ in
...
-
-              if [ -z "$h" ]; then
-                echo "[1;31merror: not linking environment.etc.\"''${l#/etc/}\" because $l already exists, skipping...[0m" >&2
-                echo "[1;31mexisting file has unknown content $o, move and activate again to apply[0m" >&2
-                errorOccurred=true
-              fi
-            fi
+
+        if [[ -e $etcFile ]]; then
+          if [[ $(readlink "$etcFile") == $etcStaticFile ]]; then
```
Enzime, line 118:
```suggestion
          if [[ $(readlink "$etcFile") == "$etcStaticFile" ]]; then
```

```
    if [[ $(readlink "$etcFile") == $etcStaticFile ]]; then
                                    ^------------^ SC2053 (warning): Quote the right-hand side of == in [[ ]] to prevent glob matching.
```

Enzime (changes requested):
I ran `shellcheck` on `activate` :smile:

emilazy:
Pushed with ShellCheck suggestions. I am once again recalibrating my metric for how defensive I should be when programming in shell.

Enzime (approved):
LGTM, I suspect this would've been easier to review if the variable renaming was separated from the rearranging of code, but that's totally fair if that would've been way more work to split the commits that way

emilazy:
Separated out variable renames to hopefully make the diff easier to read for future spelunkers. No change to the final tree. (Have to ask for re-approval, unfortunately.)

Enzime:
@emilazy it's still approved FYI as you didn't make any overall changes

emilazy:
Didn't realize it worked that way! Okay, merging this; fingers crossed. (We need VM tests...)

PR #732, opened 2023-07-12
flake: use `nix-darwin` instead of `darwin`

Enzime:
After #723, `nix-darwin` no longer uses the name of the flake input (previously using `darwin`), which means users can use any name they like. One benefit for users using `nix-darwin` instead of `darwin` as the name of their flake input is if they generate `nix.registry` entries from `inputs` (like [`nix.generateRegistryFromInputs`](https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/master/lib/options.nix) from `flake-utils-plus.darwinModules.autoGenFromInputs`), then their registry will have `nix-darwin` pointing to the version used in their configs, rather having both `darwin` and `nix-darwin` in the registry.

Closes https://github.com/NixOS/flake-registry/pull/47

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,8 +106,8 @@ Add the following to `flake.nix` in the same folder as `configuration.nix`:
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-23.05-darwin";
-    darwin.url = "github:LnL7/nix-darwin/master";
-    darwin.inputs.nixpkgs.follows = "nixpkgs";
+    nix-darwin.url = "github:LnL7/nix-darwin/master";
+    nix-darwin.inputs.nixpkgs.follows = "nixpkgs";
   };
 
   outputs = inputs@{ self, darwin, nixpkgs }: {
@@ -147,7 +147,7 @@ accessible as an argument `inputs`, similar to `pkgs` and `lib`, inside the conf
 
 ```nix
 # in flake.nix
-darwin.lib.darwinSystem {
+nix-darwin.lib.darwinSystem {
   modules = [ ./configuration.nix ];
   specialArgs = { inherit inputs; };
 }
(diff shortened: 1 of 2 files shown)


emilazy:
I'm in favour of this, or at least I'm in favour of either this or https://github.com/NixOS/flake-registry/pull/47. There's a lot of ambiguity around between "darwin" and "nix-darwin" and I personally lean towards standardizing on the latter; "darwin" is pretty ambiguous so we obviously can't use it everywhere, so "nix-darwin" seems the obvious option. (One wonders how far to go though: `system.nixDarwinVersion`? `lib.nixDarwinSystem`? `nix-darwin-rebuild`? `nixDarwinConfigurations`? I think I'd theoretically support the first two of those and maybe all of them. Though of course actually trying to do that requires thinking about backwards compatibility and migration in a way this change doesn't.)

(You can't close someone else's PR in another repository with a PR! Although it'd be funny if you could.)

Enzime:
Personally, I think just the flake ref being `nix-darwin` and then everything else being `darwin` is totally fine in my eyes. I think everything else is clearly referring to `nix-darwin` in Nix code, so the `nix` is extra redundant in that situation and it would require a significant amount of migrating. In my experience, the flake ref doesn't get used much so not much migration would be necessary for users who want to change the flake ref.

emilazy:
I think `darwin-{rebuild,option,{,un}installer} are pretty confusing names - it doesn't rebuild Darwin, they aren't native Darwin options, and they don't install or uninstall Darwin – but I agree that the case for the rest is weaker. I feel like `nix-` is equally implied in the flake registry context, though. But I try to say "nix-darwin" whenever I can.

emilazy (approved):
Seems like we have consensus for "nix-darwin" when it's being used as the name of the project. I'll merge this, but we can always revisit at a later point if necessary.

PR #733, opened 2023-07-12
Add `darwin-version` command

### LINKED ISSUE #327: `darwin-rebuild --help` runs `man darwin-rebuild` but no such manpage exists
When I run `darwin-rebuild --help` it prints an option summary followed by running `man darwin-rebuild`. However, no such manpage exists. This doesn't appear to be a misconfiguration, I don't see any evidence of a `darwin-rebuild` manpage in the nix-darwin source.

Enzime:
Adapted from https://github.com/NixOS/nixpkgs/blob/3b5456608a680a0cf7e488c7ddd250b6fef3eac1/nixos/modules/installer/tools/nixos-version.sh

Fixes #327

### DIFF
--- a/modules/examples/flake/flake.nix
+++ b/modules/examples/flake/flake.nix
@@ -27,6 +27,9 @@
       programs.zsh.enable = true;  # default shell on catalina
       # programs.fish.enable = true;
 
+      # Set Git commit hash for darwin-version.
+      system.configurationRevision = self.rev or self.dirtyRev or null;
+
       # Used for backwards compatibility, please read the changelog before changing.
       # $ darwin-rebuild changelog
       system.stateVersion = 4;
--- a/modules/examples/lnl.nix
+++ b/modules/examples/lnl.nix
@@ -125,7 +125,7 @@
     set -g pane-border-style fg=black
     set -g status-bg black
     set -g status-fg white
-    set -g status-right '#[fg=white]#(id -un)@#(hostname)   #(cat /run/current-system/darwin-version)'
+    set -g status-right '#[fg=white]#(id -un)@#(hostname)   #(jq --raw-output '.darwinLabel' /run/current-system/darwin-version.json)'
   '';
 
   environment.etc."nix/user-sandbox.sb".text = ''
--- a/flake.nix
+++ b/flake.nix
@@ -30,14 +30,17 @@
 
               system.checks.verifyNixPath = lib.mkDefault false;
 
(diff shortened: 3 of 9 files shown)


emilazy (commented):
Could you base this on https://github.com/NixOS/nixpkgs/commit/fd4e080885305b7c6093b70d929f38249470490d so we can avoid repeating NixOS's mistakes as far as unnecessary rebuilds and programs consuming script output rather than structured data goes? (From https://github.com/NixOS/nixpkgs/pull/238557, but the current state of the PR has become more involved due to the need to handle backwards compatibility.)

Enzime:
@emilazy I've updated this PR with the changes in the commit linked 👍

modules/examples/lnl.nix
```diff
@@ -125,7 +125,7 @@
     set -g pane-border-style fg=black
     set -g status-bg black
     set -g status-fg white
-    set -g status-right '#[fg=white]#(id -un)@#(hostname)   #(cat /run/current-system/darwin-version)'
+    set -g status-right '#[fg=white]#(id -un)@#(hostname)   #(jq --raw-output '.darwinLabel' /run/current-system/darwin-version.json)'
```
emilazy, line 128: This `jq` invocation seems to be ~15 times more expensive than a `cat`, which seems unfortunate for a shell prompt. I guess it's probably still cheaper than something like `git`, but maybe we could continue exposing `/run/current-system/darwin-version` (or rename it to `darwin-label`)? It's not my prompt though so I don't care that much ultimately I guess :)

modules/examples/flake/flake.nix
```diff
@@ -27,6 +27,9 @@
       programs.zsh.enable = true;  # default shell on catalina
       # programs.fish.enable = true;
 
+      # Set Git commit hash for darwin-version.
+      system.configurationRevision = self.rev or self.dirtyRev or null;
```
emilazy, line 31: Needing this boilerplate in every flake is unfortunate... but I don't presently see any way around it, so it's probably fine.

emilazy (approved):
A couple comments, but this looks fine to me as-is.

modules/examples/lnl.nix
Enzime: From my limited searching on GitHub, it doesn't seem like many people depend on the existing `/run/current-system/darwin-version` file, so I've gone ahead and merged it as is, however if more people report wanting `darwin-label` as a separate file, I'm happy to add it back

PR #735, opened 2023-07-12
Allow `AppleKeyboardUIMode` to be `2`

OliverJAsh:
This is the value of the following setting, when enabled:

image


Samasaur1:
Superceded by #1501

PR #736, opened 2023-07-12
cachix-agent: clarify what should be in the token file

domenkozar:


modules/services/cachix-agent.nix
```diff
@@ -37,7 +37,9 @@ in {
       type = types.path;
       default = "/etc/cachix-agent.token";
       description = lib.mdDoc ''
-        Required file that needs to contain CACHIX_AGENT_TOKEN=...
+        Required file that needs to contains:
```
Enzime, line 40:
```suggestion
        Required file that needs to contain:
```

Enzime (changes requested):

Enzime (approved):
LGTM

PR #739, opened 2023-07-13
use `types.lines` for yabai.extraConfig option

### LINKED ISSUE #729: yabai extraConfig can only be set once
Hey, first, nix-darwin is the business, and I really appreciate all the work that went into it!

I'm trying to set up my yabai config using two "feature" `.nix` files, one for my base config, and one for the scripting additions, which require SIP to be partially disabled.

I'd like to set the `extraConfig` option in both configs, but I get the error:

```
error: The option `services.yabai.extraConfig' has conflicting definition values:
       - In `/nix/store/c6a3c4vm09083a5vf4qr4zjf8gblnvdf-source/darwin/features/yabai-scripting-additions.nix':
           ''
             yabai -m signal --add event=dock_did_restart action="sudo yabai --load-sa"
             sudo yabai --load-sa
           ''
       - In `/nix/store/c6a3c4vm09083a5vf4qr4zjf8gblnvdf-source/darwin/features/yabai.nix':
           ''
             yabai -m rule --add app='System Settings' manage=off
             yabai -m rule --add app='Zoom' manage=off
             yabai -m rule --add app='zoom.us' manage=off
             yabai -m config layout bsp
           ''
       Use `lib.mkForce value` or `lib.mkDefault value` to change the priority on any of these definitions.
```

This could be solved by setting the option type to `types.lines` instead of `types.str`, which would merge the definitions together instead of flagging them as conflicting. 

I'd be happy to open a quick PR if you'd be open to the change. Thanks :)

yusefnapora:
This closes #729 by changing the type of the `services.yabai.extraConfig` option from `str` to `lines`. This allows defining `extraConfig` in multiple places in your config, with the final value being all definitions concatenated with `\n`.

### DIFF
--- a/modules/services/yabai/default.nix
+++ b/modules/services/yabai/default.nix
@@ -62,7 +62,7 @@ in
     };
 
     services.yabai.extraConfig = mkOption {
-      type = str;
+      type = lines;
       default = "";
       example = literalExpression ''
         yabai -m rule --add app='System Preferences' manage=off


emilazy (approved):
Thank you for the contribution!

Issue #740, opened 2023-07-13
New flake install instructions


bestlem:
They don"t work.

The documentation needs to also say when editing flake.nix

Edit the two "simple" in the end of the file to the first part of the machine name (not including any .local or other domain) m

9OK mine now also fails due to lack of all the build users - but I think that is a bug #681 The default for new users should not need the extra line that is recommennded. Begginer documentation MUST be exact so the user can copy and paste the example and comments like edit this file and not good but can be dealt with but asking a begginner to set a variable does not work as where in the file does that edit get done.
And the response https://github.com/LnL7/nix-darwin/pull/681#issuecomment-1584586205 looks like it is wrong to me as the steps listed here don"t work without the added line to suppress the check.

emilazy:
> The documentation needs to also say when editing flake.nix
> 
> Edit the two "simple" in the end of the file to the first part of the machine name (not including any .local or other domain) m

The flake installation instructions say the following:

> Make sure to replace all occurrences of `simple` with your short hostname which you can find by running `hostname -s`.

Could you suggest how we could make this more clear for beginners?

> 9OK mine now also fails due to lack of all the build users - but I think that is a bug #681 The default for new users should not need the extra line that is recommennded. Begginer documentation MUST be exact so the user can copy and paste the example and comments like edit this file and not good but can be dealt with but asking a begginner to set a variable does not work as where in the file does that edit get done. And the response [#681 (comment)](https://github.com/LnL7/nix-darwin/pull/681#issuecomment-1584586205) looks like it is wrong to me as the steps listed here don"t work without the added line to suppress the check.

Are you using the Determinate Systems experimental [nix-installer](https://github.com/DeterminateSystems/nix-installer)? That's not an officially-supported way to install Nix, and has some issues currently, although it's good software and we do try to support it. They tried avoiding user creation on macOS as an experiment, but it [doesn't work properly](https://github.com/DeterminateSystems/nix-installer/issues/521) and they [disabled it](https://github.com/DeterminateSystems/nix-installer/pull/524). That was released as [v0.10.0](https://github.com/DeterminateSystems/nix-installer/releases/tag/v0.10.0), so if you do a clean install with the same installer now you shouldn't have the same problem.

We could disable the build users check unconditionally, but it may make things more brittle for people using the official installer or the latest version of Determinate Systems' unofficial one. `auto-allocate-uids` doesn't really work on macOS yet, unfortunately, so it's not clear if we should be supporting it.

[...]

PR #741, opened 2023-07-15
Set default flake directory as `/etc/nix-darwin`

### LINKED ISSUE #721: Use flake.nix by default if it exists
Both `nixos-rebuild switch` and `home-manager switch` seem to allow using a flake without passing any arguments if a `flake.nix` file exists in their respective directories - `/etc/nixos` and `~/.config/home-manager` respectively. They both seem to follow symlinks, using the actual directory containing `flake.nix` rather than simply those default directories.

My proposal is that darwin-rebuild should use a similar behavior - if `flake.nix` exists in a specific default location (though I'm not sure what that location would be), it would use that without any additional arguments being passed.

Enzime:
Backported from https://github.com/NixOS/nixpkgs/blob/9995a5d59c59fde96f4ad7d811a8ed832357440f/pkgs/os-specific/linux/nixos-rebuild/nixos-rebuild.sh#L349-L353

Fixes #721

### DIFF
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -11,8 +11,8 @@ showSyntax() {
   echo "               [-Q] [{--max-jobs | -j} number] [--cores number] [--dry-run]" >&2
   echo "               [--keep-going] [-k] [--keep-failed] [-K] [--fallback] [--show-trace]" >&2
   echo "               [-I path] [--option name value] [--arg name value] [--argstr name value]" >&2
-  echo "               [--flake flake] [--update-input input flake] [--impure] [--recreate-lock-file]" >&2
-  echo "               [--no-update-lock-file] [--refresh]" >&2
+  echo "               [--flake flake] [--no-flake] [--update-input input flake] [--impure]" >&2
+  echo "               [--recreate-lock-file] [--no-update-lock-file] [--refresh]" >&2
   echo "               [--offline] [--substituters substituters-list] ..." >&2
   exit 1
 }
@@ -34,6 +34,7 @@ extraProfileFlags=()
 profile=@profile@
 action=
 flake=
+noFlake=
 
 while [ $# -gt 0 ]; do
   i=$1; shift 1
@@ -76,6 +77,9 @@ while [ $# -gt 0 ]; do
     --flake)
       flake=$1
       shift 1
+      ;;
+    --no-flake)
+      noFlake=1
       ;;
     -L|-vL|--print-build-logs|--impure|--recreate-lock-file|--no-update-lock-file|--no-write-lock-file|--no-registries|--commit-lock-file|--refresh)
       extraLockFlags+=("$i")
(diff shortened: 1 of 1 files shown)


emilazy:
If we do this, should we also switch the non-flakes default (I guess that might be backwards-incompatible?) and update the README?

Enzime:
@emilazy my preference would not be to do any breaking changes while we don't have release channels, and this PR only affects behaviour for users who have a flake present in `/etc/nix-darwin` which should be no one currently. With regard to the documentation, I'm thinking of updating it separately as the instructions there still work but it'll take some rework to incorporate explaining putting your config in `/etc/nix-darwin` vs `$HOME` (and how to symlink `/etc/nix-darwin`).

emilazy:
Okay, sure; I'm in favour. Could maybe make the instructions clearer around https://github.com/LnL7/nix-darwin/issues/740 when adjusting the docs too.

LnL7:
I'm not sure that introducing inconsistent locations is a good idea. For changing this in general with compatibility in mind, we could add detection for the global location and fallback to a per-user config so both cases work.

With NIX_PATH it is relatively easy for a user to specify their own default location, does something equivalent exist for nixos with the flake location?

emilazy:
I think there's already inconsistency since the installer etc. use `~/.nixpkgs/darwin-configuration.nix` but the flake docs talk about `~/.config/darwin`, and it's been like that for a long time now. Maybe we could introduce this and change the docs for flakes, and then handle moving non-flakes configurations over and the attendant backwards compatibility concerns afterwards, since this wouldn't be introducing any new inconsistency?

> does something equivalent exist for nixos with the flake location

Unfortunately I don't think so. Though maybe one has been added since I last checked. I can think of multiple ways to do it but I don't know which are any good. I always just manually specified the location. Some people `sudo ln -s ~/flake /etc/nixos`, which seems reasonable.

Enzime:
I've updated this PR to search `/etc/nix-darwin` first then fallback to `~/.config/nix-darwin` (current directory mentioned in the README) and `~/.nixpkgs` which is hardcoded in a lot of locations in the codebase.

I think we should migrate to `/etc/nix-darwin` in a different PR and properly deprecate the old paths and update the docs.

This PR is now ready to merge as it is no longer opinionated about the default flake directory.

Samasaur1 (approved):

pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -137,6 +141,19 @@ if [ -z "$action" ]; then showSyntax; fi
...
+# Search the default locations for the system-wide nix-darwin configuration
+if [[ -z $flake && -z $noFlake ]]; then
+  if [[ -e /etc/nix-darwin/flake.nix ]]; then
+    flake="$(dirname "$(readlink -f /etc/nix-darwin/flake.nix)")"
+  elif [[ -e ~/.config/nix-darwin/flake.nix ]]; then
+    flake="$(dirname "$(readlink -f ~/.config/nix-darwin/flake.nix)")"
+  elif [[ -e ~/.nixpkgs/flake.nix ]]; then
+    flake="$(dirname "$(readlink -f ~/.nixpkgs/flake.nix)")"
+  fi
+fi
```
emilazy, line 153: None of these were defaulted to previously, right? I’m a bit wary about giving preferential behaviour to directories we want to get rid of – `~/.nixpkgs` especially, that’s a totally inappropriate place for us to be using, especially for a flake setup – and making it more painful for us to do a deprecation process later. I think we could just use `/etc/nix-darwin` here and worry about documentation etc. later.

`~/.config/nix-darwin` makes some amount of sense while the scourge of user activation is still with us, but since nobody’s existing setup is using this feature, I feel like it’s better to take an incremental step forward than one that sanctions directories we’d like to move away from.

emilazy (approved):
Made a symlink and my configuration activated successfully; LGTM.

PR #743, opened 2023-07-21
treewide: fix `mkEnableOption` docs

emilazy:
Adventures in the world's least obvious calling convention.


modules/homebrew.nix
```diff
@@ -536,8 +536,7 @@ in
       [](#opt-homebrew.onActivation).
 
       This module also provides a few options for modifying how Homebrew commands behave when
-      you manually invoke them, under [](#opt-homebrew.global)
-    '');
+      you manually invoke them, under [](#opt-homebrew.global)'');
```
Enzime, line 539: Is it important to remove the trailing newlines?

modules/security/pam.nix
```diff
@@ -39,17 +39,24 @@ in
 
 {
   options = {
-    security.pam.enableSudoTouchIdAuth = mkEnableOption (lib.mdDoc ''
-      Enable sudo authentication with Touch ID
+    security.pam.enableSudoTouchIdAuth = mkEnableOption "" // {
+      description = lib.mdDoc ''
```
Enzime, line 43: What's the context behind these changes? Is `nixpkgs-unstable` no longer evaluating, or is it cosmetic, and are they backwards compatible?

modules/homebrew.nix
emilazy: Yes, otherwise the additional newline renders as an incorrect space before the end of the sentence in the documentation; you can see this [in the current online manual](https://daiderd.com/nix-darwin/manual/index.html#opt-homebrew.enable) (until the docs change). Sorry, I've been in the docs weeds for so long that it's obvious to me in a way that it definitely wasn't before :) I've pushed a more detailed commit message to explain.

modules/security/pam.nix
emilazy: It's cosmetic insofar as the previous call generates [an incoherent option description](https://daiderd.com/nix-darwin/manual/index.html#opt-security.pam.enableSudoTouchIdAuth): "Whether to enable Enable sudo authentication with Touch ID" and an additional period after the parenthetical note. Unfortunately `mkEnableOption` is intuitive when you're using it like `mkEnableOption "this thingy"` and very confusing when you want complete paragraphs; `mkEnableOption "this thingy. It's very cool and useful"` feels kind of like performing SQL injection on English text.

Enzime (approved):
LGTM

Maybe it would be nice if `mkEnableOption` did different things based on the type passed in (i.e. strings get prefixed with "Whether to enable ", while multiline strings don't get prefixed), not sure if that's an upstream thing or we could have our own version

emilazy:
I think two variants would be better because your suggestion would behave counterintuitively when you're using multiple lines just to wrap, but it's for upstream Nixpkgs; we don't touch `lib` at all and even if we plumbed in our own library machinery, getting people to use the right one would probably be as hard as just fixing the errors.

PR #744, opened 2023-07-21
eval-config: set `class`

emilazy:
See [the Nixpkgs manual]. This allows modules to declare themselves as being only for NixOS or nix-darwin, reducing compatibility risks. Unmarked modules will continue to behave as before. Technically a breaking change, but any configuration importing a module explicitly marked as NixOS-specific was probably broken already.

[the Nixpkgs manual]: https://nixos.org/manual/nixpkgs/unstable/#module-system-lib-evalModules-param-class

### DIFF
--- a/eval-config.nix
+++ b/eval-config.nix
@@ -16,6 +16,7 @@ let
   };
 
   eval = lib.evalModules (builtins.removeAttrs args [ "lib" ] // {
+    class = "darwin";
     modules = modules ++ [ argsModule ] ++ baseModules;
     specialArgs = { modulesPath = builtins.toString ./modules; } // specialArgs;
   });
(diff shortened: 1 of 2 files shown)


eval-config.nix
```diff
@@ -16,6 +16,7 @@ let
   };
 
   eval = lib.evalModules (builtins.removeAttrs args [ "lib" ] // {
+    class = "darwin";
```
Enzime, line 19: > `This would be remove in the future, Prefer _module.args option instead.`

From https://github.com/NixOS/nixpkgs/blob/6cee3b5893090b0f5f0a06b4cf42ca4e60e5d222/lib/modules.nix#L84

Would it be better to add it into the `argsModule` above then?

emilazy: I think that comment is misplaced and referring to the `args` argument. I don't see how you could set `class` through `_module.args`, from the rest of the code.

Enzime (approved):
LGTM, my dotfiles still build using this PR 👍

PR #745, opened 2023-07-23
eternal-terminal: add module

ryane:
Adds an eternal-terminal module. Much of the implementation is borrowed from the corresponding nixos module:
https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/networking/eternal-terminal.nix

### DIFF
--- a/modules/services/eternal-terminal.nix
+++ b/modules/services/eternal-terminal.nix
@@ -1,0 +1,92 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let cfg = config.services.eternal-terminal;
+in {
+  options = {
+    services.eternal-terminal = {
+
+      enable = mkEnableOption (lib.mdDoc "Eternal Terminal server");
+
+      package = mkOption {
+        type = types.path;
+        default = pkgs.eternal-terminal;
+        defaultText = "pkgs.eternal-terminal";
+        description = lib.mdDoc
+          "This option specifies the eternal-terminal package to use.";
+      };
+
+      port = mkOption {
+        default = 2022;
+        type = types.port;
+        description = lib.mdDoc ''
+          The port the server should listen on. Will use the server's default (2022) if not specified.
+
+          Make sure to open this port in the firewall if necessary.
+        '';
+      };
+
+      verbosity = mkOption {
+        default = 0;
+        type = types.enum (lib.range 0 9);
+        description = lib.mdDoc ''
+          The verbosity level (0-9).
+        '';
+      };
+
+      silent = mkOption {
+        default = false;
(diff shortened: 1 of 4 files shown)


modules/services/eternal-terminal.nix
```diff
@@ -0,0 +1,92 @@
...
+    # the (remote) et client runs the `etterminal` binary when it
+    # connects.
+    environment.systemPackages = [ cfg.package ];
+
+    launchd.daemons.eternal-terminal = {
+      path = [ cfg.package ];
+      serviceConfig = {
+        ProgramArguments = [
+          "${cfg.package}/bin/etserver"
+          "--daemon"
```
emilazy, line 68: It looks like `--daemon` causes the program to fork off, going by the `Type = "forking"` in the systemd services. I believe that launchd doesn't handle such programs properly (though my understanding may be out of date); is there any way to get the server to run in the foreground?

ryane: I think you are right. I was not aware of that issue and I included `--daemon` in the arguments originally because that is what the [homebrew formula](https://github.com/MisterTea/homebrew-et/blob/master/et.rb#L25) does and I've been running eternal-terminal that way for a long time without noticing any problems. But, this prompted me to look around and I found that the project repository includes a [launchd config](https://github.com/MisterTea/EternalTerminal/blob/master/init/launchd/homebrew.mxcl.et.plist) that does not include `--daemon`. It was removed in this [commit](https://github.com/MisterTea/EternalTerminal/commit/1d9cd2be9dc1ec7694e9472004b7910bbb5c34cf) for reasons that sound like what you are referring to. Anyway, I updated the agent config here to align more with what's in the source repository.

emilazy: Thanks! The lack of `KeepAlive` is strange to me, but if it's in the upstream launchd daemon file then I assume there's probably some reason for it.

emilazy (approved):
Thank you for your contribution!

modules/services/eternal-terminal.nix
Enzime: I saw that you changed `KeepAlive` to `false` to match the upstream plist, however it looks like the reason why it is there in the first place is because they used to use `--daemonize` and launchd would lose track of the process and presumably keep launching it, so I think it makes sense to set it to `true`

emilazy: Ah, whoops, that makes perfect sense; merged before getting this comment. @ryane do you mind opening a PR to toggle `KeepAlive` back on?

ryane: ah, that does make sense. here is the follow-up pr: #748

PR #748, opened 2023-07-25
eternal-terminal: enable keep alive

ryane:
See https://github.com/LnL7/nix-darwin/pull/745#discussion_r1272806159


Enzime (approved):
Thanks for making this PR, might be worth raising this upstream as well :+1:

PR #753, opened 2023-08-02
linux-builder: add maxJobs option

### LINKED ISSUE #751: `nix.linux-builder` needs a `maxJobs` configuration
The new `nix.linux-builder` module is great, but it has no way to customize the `nix.buildMachines` entry. In particular, I can't set the `maxJobs` key, which means it defaults to 1 job at a time. This is rather unfortunate as I'm configuring my builder to use multiple cores and so I want to be able to dispatch multiple jobs at a time to it. I don't know if there are any other knobs on the builder that should be exposed for configuration, but `maxJobs` definitely should be.

Incidentally, speaking of the builder, it doesn't set `protocol`, and so it defaults to `ssh`; perhaps it should be setting `protocol = "ssh-ng"` as well?

lilyball:
Fixes #751

### DIFF
--- a/modules/nix/linux-builder.nix
+++ b/modules/nix/linux-builder.nix
@@ -42,6 +42,17 @@ in
         without changing this option otherwise you may not be able to build the Linux builder.
       '';
     };
+
+    maxJobs = mkOption {
+      type = types.ints.positive;
+      default = 1;
+      example = 4;
+      description = lib.mdDoc ''
+        This option specifies the maximum number of jobs to run on the Linux builder at once.
+
+        This sets the corresponding `nix.buildMachines.*.maxJobs` option.
+      '';
+    };
   };
 
   config = mkIf cfg.enable {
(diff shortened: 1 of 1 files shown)


Enzime (approved):
LGTM

emilazy:
It looks like the VM unconditionally uses 1 core, maybe we could set that based on this option? Not sure if that would cause problematic rebuilds.

lilyball:
@emilazy I'm not opposed to the idea of defaulting `virtualisation.cores` to this value, but I don't have enough experience here to know if it's a reasonable assumption that the maximum number of jobs should match the number of cores. I also don't know if that will cause a rebuild (I hope not, I hope it just changes the configuration passed to qemu on the host, but I haven't dug into this).

lilyball:
Another idea is to just change the example for `nix.linux-builder.modules` to set `virtualisation.cores` instead of `virtualisation.darwin-builder.hostPort` (especially since setting `hostPort` right now will not work with the generated ssh config).

PR #755, opened 2023-08-03
tests: fix uses of `!`

emilazy:
`set -e` does not trigger on negations, so none of these were actually being tested (and one appears to have been incorrectly negated all along).

### DIFF
--- a/tests/checks-nix-gc.nix
+++ b/tests/checks-nix-gc.nix
@@ -14,6 +14,6 @@ in
 
     echo checking nix-gc service in /Library/LaunchDaemons >&2
     grep "<string>org.nixos.nix-gc</string>" ${config.out}/Library/LaunchDaemons/org.nixos.nix-gc.plist
-    ! grep "<key>UserName</key>" ${config.out}/Library/LaunchDaemons/org.nixos.nix-gc.plist
+    (! grep "<key>UserName</key>" ${config.out}/Library/LaunchDaemons/org.nixos.nix-gc.plist)
   '';
 }
--- a/release.nix
+++ b/release.nix
@@ -19,6 +19,10 @@ let
 
   makeTest = test:
     let
+      testName =
+        builtins.replaceStrings [ ".nix" ] [ "" ]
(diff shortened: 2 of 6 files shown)


tests/checks-nix-gc.nix
```diff
@@ -14,6 +14,6 @@ in
 
     echo checking nix-gc service in /Library/LaunchDaemons >&2
     grep "<string>org.nixos.nix-gc</string>" ${config.out}/Library/LaunchDaemons/org.nixos.nix-gc.plist
-    ! grep "<key>UserName</key>" ${config.out}/Library/LaunchDaemons/org.nixos.nix-gc.plist
+    ! grep "<key>UserName</key>" ${config.out}/Library/LaunchDaemons/org.nixos.nix-gc.plist || false
```
Enzime, line 17:
```suggestion
    grep "<key>UserName</key>" ${config.out}/Library/LaunchDaemons/org.nixos.nix-gc.plist || false
```

I'm not sure what the order of operations is but it seems clearer without using `!`

Enzime (commented):
Good catch, I knew that `!` doesn't trigger `set -e` as it's useful in the `if ! cmd` construction but I never thought about it with regards to our tests.

tests/checks-nix-gc.nix
emilazy: Not the same; it would have to be `&& exit 1` which is longer (albeit possibly clearer). Another option would be to use subshells.

Enzime: I think using subshells would be clearer than `&& exit 1`

emilazy: Sure; force-pushed. My feeling was that they're all confusing so I kept the one that's more efficient and kept the `!` at the start, but I'm happy to defer here :)

There already a couple `&& return`s in the tests but I didn't touch them since they work as-is. (Although the error is that they failed to create `out` rather than an explicit error code, which is unfortunate.)

Enzime (approved):
LGTM

PR #757, opened 2023-08-05
Fix example configuration for flake migration

aequitas:


Enzime (approved):
Thanks for the PR, LGTM

Issue #632, opened 2023-04-12
Enable useBabelfish error is used but not defined


mudrii:
By enabling Fish with useBabelfish
```nix
programs.fish.enable = true;
programs.fish.babelfishPackage = true;
```

```sh
error: The option `programs.fish.babelfishPackage' is used but not defined.
(use '--show-trace' to show detailed location information)
```

System
```sh
nix-shell -p nix-info --run "nix-info -m"                                                                                                                            2308ms
 - system: `"aarch64-darwin"`
 - host os: `Darwin 22.4.0, macOS 13.3.1`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.13.3`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
```

Looking at the code
https://github.com/LnL7/nix-darwin/blob/master/modules/programs/fish.nix
is bool but nix is not interpreting it

Full trace below

```sh
error:
       … while evaluating the attribute 'buildCommand' of the derivation 'darwin-system-23.05.20230412.6b70761+darwin4.0259125'

       at /nix/store/mcqspn5p5dy263y3nfb0dfzi7inms7j0-source/pkgs/stdenv/generic/make-derivation.nix:302:7:

          301|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
          302|       name =
             |       ^
          303|         let

       … while evaluating the attribute 'buildCommand' of the derivation 'etc'

       at /nix/store/mcqspn5p5dy263y3nfb0dfzi7inms7j0-source/pkgs/stdenv/generic/make-derivation.nix:302:7:

          301|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
          302|       name =
             |       ^
          303|         let

       … while evaluating call site

       at /nix/store/40b7ddimv45vi7bzbp0d8163ml92b7lk-source/modules/system/etc.nix:40:11:

           39|         ${concatMapStringsSep "\n" (attr: "mkdir -p $(dirname '${attr.target}')") etc}
           40|         ${concatMapStringsSep "\n" (attr: "ln -s '${attr.source}' '${attr.target}'") etc}
             |           ^
           41|       '';

       … while calling 'concatMapStringsSep'

[...]

PR #758, opened 2023-08-05
fish: add default for `babelfishPackage`

emilazy:
This option should probably be dropped but I'd prefer to batch that with using Babelfish by default (or unconditionally, as Home Manager does) so as to avoid multiple independent breaking changes.

Fixes: #632

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -68,6 +68,7 @@ in
 
       babelfishPackage = mkOption {
         type = types.package;
+        default = pkgs.babelfish;
         description = lib.mdDoc ''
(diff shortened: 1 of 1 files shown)


Enzime (dismissed):
LGTM

Is this line unnecessary? https://github.com/LnL7/nix-darwin/pull/758/files#diff-5d54768a584d2d34e6d79dba9bc82f7b3cd8f558480899059fafc5d91e14f99dR41

emilazy:
Seems so. Dropped it.

Enzime (approved):
LGTM

Issue #2, opened 2016-12-22
Documentation, what's missing?


LnL7:
The readme is pretty basic at the moment. What should be added?

yurrriq:
I'd like to add some documentation of using this with the fish shell. The syntax is only slightly different, but examples in the README.md would've saved me a couple minutes.

Aside: This is fantastic! Thanks so much for making and releasing it. Since I occasionally have iOS work, I'm basically stuck on Darwin (for now) and this might be the final push I need to fully ditch Homebrew in favor of Nix. I'm happy to help develop and test this in any way I can. Don't hesitate to reach out.

LnL7:
If you could contribute, would be great! I personally don't use fish so I'm not very familiar with it.
But I can probably add a `programs.fish` module myself based on [the one from nixos](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/fish.nix) that sets up the nix-darwin environment properly.

yurrriq:
Sounds good. I might have a go tonight, depending on whether we have company or not.

dejanr:
What can we do to simplify installation process? I am not a big fun of current approach.

I was thinking about publishing nix-darwin to nixpkgs repo as a package, and just having a single cli nix-darwin

For me perfect scenario would be:

- install nix normally : curl https://nixos.org/nix/install | sh
- then install nix- darwin : nix-env -i nix-darwin
- optionally setup dotfiles : git clone git@github.com:user/dotfiles.git ~/.dotfiles && ~/.dotfiles/install.sh
- and just rebuild by : nix-darwin rebuild

nix-darwin should like now read default config, and optionally we could define NIX_PATH to use different darwin.

What do you think?

LnL7:
Installing the nix-darwin scripts with nix-env doesn't really make a lot of sense to me, this entire project might move to nixpkgs at some point but that's a different story.

I should probably create a bootstrap script like https://nixos.org/nix/install that people can use to install.
Since nix-darwin supports launchd services, it's also possible to provide a multi user installer for nix that uses the nix-daemon service, but that will be a bit harder to get right.

yurrriq:
Could we add an example of cross building with/for `nixops`?

LnL7:
[...]

PR #761, opened 2023-08-13
chore(installer): capitalize the default option when reading user input

sevillaarvin:
Just my 2 cents w.r.t. https://github.com/LnL7/nix-darwin/issues/2.

Feel free to accept / reject.

### DIFF
--- a/pkgs/darwin-installer/default.nix
+++ b/pkgs/darwin-installer/default.nix
@@ -63,7 +63,7 @@ stdenv.mkDerivation {
     # Skip when stdin is not a tty, eg.
     # $ yes | darwin-installer
     if test -t 0; then
-        read -p "Would you like to edit the default configuration.nix before starting? [y/n] " i
+        read -p "Would you like to edit the default configuration.nix before starting? [y/N] " i
         case "$i" in
             y|Y)
                 PATH=$_PATH ''${EDITOR:-nano} "$config"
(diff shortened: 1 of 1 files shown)


pkgs/darwin-installer/default.nix
```diff
@@ -75,7 +75,7 @@ stdenv.mkDerivation {
     darwinPath=$(NIX_PATH=$HOME/.nix-defexpr/channels nix-instantiate --eval -E '<darwin>' 2> /dev/null) || true
     if ! test -e "$darwinPath"; then
         if test -t 0; then
-            read -p "Would you like to manage <darwin> with nix-channel? [y/n] " i
+            read -p "Would you like to manage <darwin> with nix-channel? [Y/n] " i
```
Enzime, line 78:
```suggestion
            read -p "Would you like to manage <darwin> with nix-channel? [y/N] " i
```

The default value here is actually `n`, the `y` only gets used if stdin is not a TTY e.g.

```
echo "apple" | darwin-installer
```

Enzime (changes requested):
Thanks for the PR :+1:

pkgs/darwin-installer/default.nix
sevillaarvin: I see, thanks for the info!

Enzime (approved):

Issue #554, opened 2022-10-20
Multiple bugs when setting up multiple macos users


busti:
I am using nix-darwin on my macbook, I have it configured using flakes for which I mostly followed this gist:  
https://gist.github.com/jmatsushita/5c50ef14b4b96cb24ae5268dab613050  

I have recently added another user using the MacOS system preferences to better split my system up between work and private usage.  

There are multiple issues with the services provided by nix-darwin on that account which do not exist on the user I originally set it up from. 

Specifically I have encountered the following problems so far:  
- homebrew and any commands installed by it are not present on that user  
   (casks are present, but only the applications, not the commands, i.e. vscode is installed but the `code` command cannot be used)
- zsh prompts the following every time I open a shell:  
   ```
   zsh compinit: insecure directories and files, run compaudit for list.
   Ignore insecure directories and files and continue [y] or abort compinit [n]?
   ```
- `system.defaults` options are not set up for that user

Also since the configuration itself resides in the homedir of the other user, I cannot access it by default.  
I have since moved it to `/opt` but I am not sure if that is the right place to keep it.  

Is there anything special I need to do to setup nix-darwin for multiple users? The manual seems to be not have any information on it.

busti:
I was able to solve most of my issues, but it took a lot of digging.  

1. `programs.homebrew.enable = true;` does not install the `brew` command. I was under the impression that it did, but I was mislead by a script from work installing the command instead without me noticing.  
   It can simply be installed for the other user using the usual console command.  
2. The zhs compinit errors can be ignored by appending a `-u` to the `compinit` command.  
   To be more complete, I added the following to my nix darwin config: `programs.zsh.enableCompletion = false;` and this line to my `home.nix` where zsh is actually configured: `programs.zsh.completionInit = "autoload -U compinit && compinit -u";`
   
   Still no idea why the defaults are not being set up tho.

hraban:
This is being discussed in #96

dlubawy:
This issue has been a major pain point for me too. I've found a few things out while messing with nix-darwin.
* nix-darwin has options (ex: `system.defaults`) which are red herrings and better left to the scope of a user environment manager like home-manager
  - These options require per user activation scripts run by each user which can be problematic for nix-darwin as it only runs under one admin user
[...]

PR #763, opened 2023-08-16
feat: repeat per-user config for every user

hraban:
Setting up the dock, menubar, etc. Should be run for every user, not just for the admin running nix-darwin.

I think there are three types of activation code:

- system-wide which should be run non-root (e.g. brew)
- system-wide which should be run root (e.g. networking)
- user-local

The last option being invoked by the second one, using sudo -u username.

This is very POC, mostly to get others' opinions on it. The idea is to expand that userProfileScript with more options as appropriate.

It seems to work on my machine but I wouldn't be surprised to be missing some edge cases.

What do you think?

This addresses #554.

I have a follow-up PR that expands on this to use chsh to set the user's shell from `config.users.users.<name>.shell`: https://github.com/hraban/nix-darwin/compare/per-user...hraban:nix-darwin:chsh

### DIFF
--- a/modules/system/activation-scripts.nix
+++ b/modules/system/activation-scripts.nix
@@ -36,6 +36,21 @@ in
 
   config = {
 
+    # Configuration of user-local settings only, intended to be run as that
+    # user themselves (throug sudo).
+    system.activationScripts.userProfileScript.source = pkgs.writeShellScript "activate-user-profile" ''
+      ${cfg.activationScripts.userDefaults.text}
+      ${cfg.activationScripts.userLaunchd.text}
+    '';
+
+    # Per-user config
+    system.activationScripts.configureUsers.text = builtins.concatStringsSep "\n" (builtins.map (user: ''
+      sudo -u "${user}" ${cfg.activationScripts.userProfileScript.source}
+    '') config.users.knownUsers);
+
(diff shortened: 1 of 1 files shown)


emilazy:
Thanks for the PR! I agree this is definitely an area that could use improvement. While this is certainly better than the status quo (although I think the defaults/launchd settings should be scoped under `users.users` in this case), the `activate`/`activate-user` split in general is problematic (see https://github.com/LnL7/nix-darwin/issues/96) and I think in many cases it's stuff we shouldn't be handling at all or that is best delegated to Home Manager. I'm not sure it's a great idea to change semantics incompatibly here without addressing the problem at the root, at least until we have an idea of the destination we're aiming at.

I have a local branch for the issue I linked that kills off `activate-user` entirely and mostly just drops user-specific stuff, which gets rid of a bunch of weird quirks and pulls us more in line with NixOS's approach of managing the system rather than being sensitive to the user running activation, but it needs a lot of polish and thinking about backwards compatibility before it could go upstream here. (Retaining some of the functionality under `users.users` in that scheme would certainly be an option too, though.)

hraban:
I misunderstood how this all worked and found a better way to set the shell in #818 . In combination with the other comments, that makes this PR obsolete.

PR #765, opened 2023-08-17
add progress to homebrew package installations

tnxz:


Enzime:
Can you post a comparison (ideally using asciinema) to show the difference in how much `brew` outputs when you use the `--verbose` flag? I just wanted to make sure that this change won't result in a significant amount of terminal spam for Homebrew users.

tnxz:
Screenshot 2023-08-17 at 10 18 54 AM

something similar to this

Enzime:
What does it look like without the `--verbose` flag?

tnxz:
Screenshot 2023-08-17 at 10 31 15 AM

**the cursor keeps blinking untill the package is installed**

Screenshot 2023-08-17 at 10 31 48 AM

**final output**

Enzime (approved):
LGTM, thanks for the contribution

emilazy:
This greatly increases the noise on every activation:

```
Homebrew bundle...
Skipping install of 1password cask. It is already installed.
Using 1password
Skipping install of 1password-cli cask. It is already installed.
…
```

Homebrew is already the loudest part of system activation; I would prefer to revert this unless these new "Skipping …" messages can be suppressed. (Ideally I'd prefer it print nothing at all for things that are already installed.)

Enzime:
Alternatively we could add an option like `hombrew.onActivation.verbose` and  possibly `homebrew.onActivation.flags` or `bundleFlags`

PR #771, opened 2023-09-08
Add support for dfont to the fonts module

Benjamin-L:
There aren't actually any dfont-formatted fonts in nixpkgs, but I'm using one in my personal config. This used to work, prior to bc776e4940106a2578998dd3117a76d62ec0a8c, but the font file is now ignored.

Supporting a format that isn't supported by nixpkgs feels a bit weird, but I think it's okay because nixpkgs includes a lot of formats that are not supported on darwin.

### DIFF
--- a/modules/fonts/default.nix
+++ b/modules/fonts/default.nix
@@ -41,7 +41,7 @@ in
       { preferLocalBuild = true; }
       ''
         mkdir -p $out/Library/Fonts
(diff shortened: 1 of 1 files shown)


Enzime (approved):
Thanks for the contribution, happy to merge this, would be great if you could package a `dfont` font on Nixpkgs if you have the time

PR #772, opened 2023-09-10
Use scutil to obtain the default local hostname for flake configs

purcell:
When a VPN is in use (Wireguard in my case), `hostname` can spuriously report the short name of the current host as a partial host name on the VPN network, e.g. `client-100-34-5-1`. Elsewhere in `nix-darwin`, the options for setting the network host name use `scutil`, so this change should ensure that the applied name is reliably used to find the config later.


purcell:
Closing as duplicate of #676

PR #777, opened 2023-09-13
Expose _module attr in realised sysconfig

### LINKED ISSUE #609: [proposal] expose more attributes in darwin system config similar to nixos system config
After restructuring my system flake, which includes linux and darwin hosts, I was bitten by an ["incompatibility" between nixos and nix-darwin config handling](https://discourse.nixos.org/t/module-args-not-accessible-in-realised-darwinconfiguration/25652).

I noticed that it is actually [trivial to alter the behaviour](https://github.com/LnL7/nix-darwin/compare/master...ppenguin:nix-darwin:expose-more-config-attr) to make this issue "go away".

This begs the question: for system attributes that both appear in a `nixosConfiguration` and a `darwinConfiguration`, is there a strong reason to not expose the same ones in `darwin` as in `nixos`? Otherwise one could say that more "compatibility" between the two is better and it would be better to expose these attributes?

(This seems to be only a few, at least `_module`, I didn't check for others like `extendModules`, `type`, etc.)

ppenguin:
In short: expose `_module` attribute to improve compatibility between `darwinSystem` and `nixosSystem` in support of multi-system flakes.

See https://github.com/LnL7/nix-darwin/issues/609 and https://discourse.nixos.org/t/module-args-not-accessible-in-realised-darwinconfiguration/25652/4 for rationale.

Closes #609


eval-config.nix
```diff
@@ -24,6 +24,6 @@ in
 {
   inherit (eval._module.args) pkgs;
   inherit (eval) options config;
-
+  inherit _module;
```
Enzime, line 27: Did you mean?

```suggestion
  inherit (eval) _module;
```

Enzime (changes requested):

eval-config.nix
ppenguin: You're right. I could've sworn that it worked before, but since `nixos-unstable` got updated it blew up in my face.
Thanks!

Enzime (approved):
Thanks for the contribution, LGTM

PR #781, opened 2023-09-15
time: bury confusing `systemsetup -settimezone` output

### LINKED ISSUE #359: `systemsetup` error during 'configuring time' step on macOS 11.6
Since updating to macOS Big Sur 11.6 (which provides an important security update, as I understand it), I've been receiving the following non-blocking error during rebuilds:

```
configuring time...
2021-09-16 13:08:57.284 systemsetup[46864:2995101] ### Error:2 File:/System/Volumes/Data/SWE/macOS/BuildRoots/38cf1d983f/Library/Caches/com.apple.xbs/Sources/Admin/Admin-876.1/InternetServices.m Line:379
```

I haven't seen any notable side effects from this, and the rebuild completes successfully.

Considering that `Caches` is in the path, maybe it's some leftover state from the previous macOS version?

montchr:
Fixes #359

As noted in #359, the timezone is still set correctly despite the error message. So, the message appears to be irrelevant. No one in that thread has reported observing other side effects that may relate.

### DIFF
--- a/modules/time/default.nix
+++ b/modules/time/default.nix
@@ -11,7 +11,7 @@ let
       echo "${cfg.timeZone} is not a valid timezone. The command 'listtimezones' will show a list of valid time zones." >&2
       false
     fi
-    systemsetup -settimezone "${cfg.timeZone}" > /dev/null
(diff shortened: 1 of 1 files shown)


Enzime (approved):
LGTM

PR #782, opened 2023-09-19
Create $out/darwin-version

domenkozar:
This fixes backwards compatibility with code that expects it.

cc @Enzime


Enzime (approved):
LGTM

PR #785, opened 2023-09-28
linux-builder: change from `modules` to `config`

Enzime:
Using the type `deferredModule` fixes `pkgs` not being accessible as a module argument (apart from `config._module.args.pkgs`).

Changing the interface from a list to a module means that options can be easily set with fully quantified path syntax like

```
nix.linux-builder.config.virtualisation.darwin-builder.hostPort = 22;
```

### DIFF
--- a/modules/nix/linux-builder.nix
+++ b/modules/nix/linux-builder.nix
@@ -8,11 +8,15 @@ let
   cfg = config.nix.linux-builder;
 
   builderWithOverrides = cfg.package.override {
-    inherit (cfg) modules;
+    modules = [ cfg.config ];
   };
 in
 
 {
+  imports = [
+    (mkRemovedOptionModule [ "nix" "linux-builder" "modules" ] "This option has been replaced with `nix.linux-builder.config` which allows setting options directly like `nix.linux-builder.config.networking.hostName = \"banana\";.")
+  ];
+
   options.nix.linux-builder = {
     enable = mkEnableOption (lib.mdDoc "Linux builder");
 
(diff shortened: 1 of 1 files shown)


malob:
I like this change, but it won't be backwards compatible, so there should probably be a check of some sort that triggers an error message explaining the change.

malob:
Something like this maybe with a helpful message: https://github.com/LnL7/nix-darwin/blob/e236a1e598a9a59265897948ac9874c364b9555f/modules/nix/default.nix#L143

Enzime:
I've replaced `mkRenamedOptionModule` with `mkRemovedOptionModule` which should make usage clearer 👍

Enzime:
It looks like options for the Linux builder can only be defined in one place at the moment, I'll mark this as a draft until I have time to fix it

@roberth do you have any ideas how to fix this?

Enzime:
My previous comment was wrong, I made a mistake when testing this previously, setting options for the Linux builder can be done in multiple places 👍

roberth (approved):
LGTM.
If you want to be really nice, you could support both `modules` and `config` simultaneously and deprecate `.modules` at some point.

Enzime:
From a quick search online, it doesn't look like too many people are using this interface yet, so I'm happy to replace it with `config`

https://github.com/search?utf8=%E2%9C%93&q=path%3A*.nix+linux-builder+modules&type=Code

domenkozar (approved):

lilyball:
> From a quick search online, it doesn't look like too many people are using this interface yet, so I'm happy to replace it with `config`

The problem with this claim is people like me use private repos as we have stuff in our config that we can't make publicly visible. That said, I always read through the nix-darwin update history every time I update, specifically to find breaking changes like this.

Issue #784, opened 2023-09-28
macOS Sonoma changes `/etc/pam.d/sudo`


lilyball:
In macOS Sonoma, the `/etc/pam.d/sudo` file now starts with

```
auth       include        sudo_local
```

and there's a separate `sudo_local.template` file with the contents

```
# sudo_local: local config file which survives system update and is included for sudo
# uncomment following line to enable Touch ID for sudo
#auth       sufficient     pam_tid.so
```

Given this, the `security.pam.enableSudoTouchIdAuth` module should probably be changed to use `environment.etc."pam.d/sudo_local"` and then have the `sed` script check for the `sudo_local` line and insert that if it's not present. This way on macOS Sonoma the change will persist across OS updates.

lilyball:
I just tested applying this change to my macOS Ventura machine and it works there (as expected), although of course software updates will reset `/etc/pam.d/sudo` and so it won't persist until macOS Sonoma.

stgarf:
You can add yubikeys as well if you are so inclined. Auth is parsed in order so if yubi fails/times out it falls back to tid next.

```
# sudo_local: local config file which survives system update and is included for sudo

# yubikey auth
auth       sufficient     /opt/homebrew/opt/pam_yubico/lib/security/pam_yubico.so  mode=challenge-response

# uncomment following line to enable Touch ID for sudo
auth       sufficient     pam_tid.so
```

dwt:
See #1020

PR #787, opened 2023-09-29
pam: add touchIdAuth to sudo_local

hgl:
Apple sanctioned sudo_local in Sonoma, which makes it possible for sudo related pam settings to survive system upgrade. We replicate this design for earlier macOS versions to ensure backward compatibility.

Changes:
- Rename option to security.pam.touchIdAuth
- Add pam_reattach.so support
- Use sudo_local as the entrypoint
- Include different files for different features

Related PRs:

#784 directly addressed.
#591 no longer necessary. The current design is influenced by @malob and @emilazy's suggestion that `/etc/pam.d/*` should be regular files instead of symlinks.
#662 no longer necessary, and it demonstrates using patches is probably not a good idea, since they are not very composable.

Design choices:
- `reattach` is nested in `touchIdAuth` since it doesn't really make sense to use the former without the latter, if I'm not wrong.
- Make `reattach.enable` and `reattach.ignoreSSH` default to `true` if `touchIdAuth` is enabled. Since users should generally want them enabled, and there is no harm in enabling them. Lets me know if that's not the case.

This PR currently serves as an overture to design discussion, thus old option migration is currently omitted.

Care has been taken to ensure script error doesn't result in a broken sudo. Let me know if you find otherwise.

### DIFF
--- a/modules/security/pam.nix
+++ b/modules/security/pam.nix
@@ -1,69 +1,135 @@
 { config, lib, pkgs, ... }:
 
 with lib;
-
 let
   cfg = config.security.pam;
+  touchIdAuth = ''
+    sudo_file=/etc/pam.d/sudo
+    sudo_local_file=/etc/pam.d/sudo_local
+    tid_file=/etc/pam.d/nix-darwin-touchIdAuth
+    del_tid() {
+      local first=1
+      local f
+      for f; do
+        if [[ ! -e $f ]]; then
+          continue
+        fi
+        if [[ -n $first ]]; then
+          first=""
+          set --
+        fi
+        set -- "$@" "$f"
+      done
+      if [[ $# != 0 ]]; then
+        ${pkgs.gawk}/bin/awk -i inplace '
+          /^[[:space:]]*#/ || !NF {
+            print
+            next
+          }
+          !($1 == "auth" && $3 ~ /(\/|^)pam_(tid|reattach)\.so$/) {
+            print
+          }
+        ' "$@"
+      fi
+    }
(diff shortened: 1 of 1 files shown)


lilyball:
> The current design is influenced by @malob and @emilazy's suggestion that `/etc/pam.d/*` should be regular files instead of symlinks.

`/etc/pam.d/sudo` certainly should be, but `/etc/pam.d/sudo_local` can be a symlink. PAM clearly works when the target of an `include` directive doesn't exist at all, so I just tested by having `sudo_local` be a symlink to a path that doesn't exist and that worked. I also tried having it be a symlink to a path that exists but isn't readable by anyone, that also worked. So it should be safe enough to have `/etc/pam.d/sudo_local` be a symlink into the store as long as `/etc/pam.d/sudo` itself is still a regular file.

hgl:
It's possible that sudo_local already exists before it's deployed. Current awk approach to remove tid & reattach entries should be good enough. But I think it's a good idea to turn `/etc/pam.d/nix-darwin-touchIdAuth` into a symlink.

mvillafuertem:
@hgl any update? thanks

emilazy:
Do we really need to support an already‐existing `sudo_local` here? If we could just generate a file from scratch it seems like we’d save a lot of logic and get a simpler and more robust result. It seems like if you’re managing sudo PAM configuration with nix-darwin it’s okay to expect you to go all‐in, unless there’s existing software that creates the file that we’d really want to be compatible with.

Enzime:
Superseded by #1344

PR #791, opened 2023-10-11
linux-builder: avoid /tmp for certs

### LINKED ISSUE #768: `nix.linux-builder` loses certs after a few days
I use `nix.linux-builder` and twice now I've had the linux builder lose its SSL certificates. The symptom is a remote build on the builder fails with an SSL error, and if I SSH into the builder I find that the `/etc/ssl/certs` mount has vanished.

I'm pretty sure what's going on is this mount is mapped to a directory located inside of `/tmp`, and macOS will, by default, clean up anything in `/tmp` that hasn't been accessed in 3 days. So I think what's going on is if I don't use the builder for a few days (or possibly this happens if I leave my laptop asleep for several days in a row) the temporary directory being used by the linux builder is deleted, and this causes the mount to vanish.

There's also a sibling directory `xchg` in the temporary directory, I'm not sure what that's for but it's also being set up by `qemu-vm.nix` as a shared directory, and presumably that's vanishing as well. Also a `store.img` file.

I'm not entirely sure what the right solution here is. Perhaps the launchd daemon could simply point `$TMPDIR` at some location that isn't periodically cleaned up, although it would presumably have to then manually clean it up.

lilyball:
macOS has a daily task that cleans files in `/tmp` that haven't been accessed for 3+ days. This task can clean up the folder used for sharing certs with the linux-builder VM, which would cause the VM to lose the ability to connect to https URLs (such as the nixpkgs binary cache). Using a subfolder of `/run` avoids this issue.

Fixes #768

### DIFF
--- a/modules/nix/linux-builder.nix
+++ b/modules/nix/linux-builder.nix
@@ -10,6 +10,18 @@ let
   builderWithOverrides = cfg.package.override {
     inherit (cfg) modules;
   };
+
+  # create-builder uses TMPDIR to share files with the builder, notably certs.
+  # macOS will clean up files in /tmp automatically that haven't been accessed in 3+ days.
+  # If we let it use /tmp, leaving the computer asleep for 3 days makes the certs vanish.
+  # So we'll use /run/org.nixos.linux-builder instead and clean it up ourselves.
+  script = pkgs.writeShellScript "linux-builder-start" ''
+    export TMPDIR=/run/org.nixos.linux-builder USE_TMPDIR=1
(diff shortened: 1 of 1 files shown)


lilyball:
Note: I haven't actually run this code in nix-darwin yet. I've been running this as a set overrides in my local config. I just copied the script from my local config to nix-darwin for this PR.

Enzime (approved):
LGTM, working locally for me 👍

Enzime:
Thanks for the contribution

Issue #798, opened 2023-11-06
Fix darwin-rebuild for Cyberark EPM enabled machines


rawmannovak:
Cyberark EPM is a popular Enterprise EndpointPrivilegeManager, and it tries to secure end devices by setting certain system settings that prevent end users from messing with the system. https://www.cyberark.com/products/endpoint-privilege-manager/

One of the settings is found in `/private/etc/sudoers.cyberark` with the content:
```
Defaults timestamp_timeout=0
Defaults secure_path="/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Applications/CyberArk EPM.app/Contents/Helpers"
```
As darwin-rebuild expects the path to be preserved during sudo runs, this no longer works, and errors with:
```
sudo: nix-env: command not found
```

PR #799, opened 2023-11-06
fix(#798): darwin-rebuild support for Cyberark EPM

rawmannovak:
restore sudo environment PATH when Cyberark EPM is setting secure_path in sudo config. Fixes LnL7/nix-darwin/issues/798

### DIFF
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -187,7 +187,7 @@ fi
 
 if [ "$action" = list ] || [ "$action" = rollback ]; then
   if [ "$USER" != root ] && [ ! -w $(dirname "$profile") ]; then
-    sudo -H nix-env -p "$profile" "${extraProfileFlags[@]}"
+    sudo -H --preserve-env=PATH env nix-env -p "$profile" "${extraProfileFlags[@]}"
   else
     nix-env -p "$profile" "${extraProfileFlags[@]}"
   fi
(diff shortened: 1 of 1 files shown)


pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -187,7 +187,7 @@ fi
 
 if [ "$action" = list ] || [ "$action" = rollback ]; then
   if [ "$USER" != root ] && [ ! -w $(dirname "$profile") ]; then
-    sudo -H nix-env -p "$profile" "${extraProfileFlags[@]}"
+    sudo -H --preserve-env=PATH env nix-env -p "$profile" "${extraProfileFlags[@]}"
```
Enzime, line 190: Is the `env` command necessary if you’re already using the `--preserve-env=PATH` flag?

rawmannovak: Hey @Enzime , thanks for taking a look. Exactly, `--preserve-env=PATH` will preserve the ENV variable in the scope of executing the `sudo` command, but it's not available when `sudo` is actually looking for the binary to execute. That's the reason we use `env` so it _evaluates_ this ENV variable first and searches for the binary in the PATH. 
Does this make sense?

Enzime: Yeah that makes sense :+1:

pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -215,7 +215,7 @@ if [ "$action" = switch ] || [ "$action" = activate ] || [ "$action" = rollback
   "$systemConfig/activate-user"
 
   if [ "$USER" != root ]; then
-    sudo -H "$systemConfig/activate"
+    sudo -H --preserve-env=PATH env "$systemConfig/activate"
```
Enzime, line 218:
```suggestion
    sudo -H --preserve-env=PATH "$systemConfig/activate"
```

Shouldn't be necessary here as `$systemConfig/activate` should be an absolute path

Enzime (changes requested):
Apart from the one change, LGTM :+1:

Enzime (approved):

PR #801, opened 2023-11-07
ssh: add extraConfig

kloenk:


RealHarshThakur:
Thank you @Kloenk for this PR.
I'm trying to setup a Linux builder on a non-default ssh port and it doesn't seem possible without this PR. Curious if there's an offline discussion about getting this merged or if something is blocking you

vifino:
Hey, this is something I stumbled over as well, I'd really like this!

PR #803, opened 2023-11-09
Fix unbound variable `etcProblems`

felixscheinost:
I have another activationScript which traps errors. Currently `etcProblems` can be unbound.

### DIFF
--- a/modules/system/etc.nix
+++ b/modules/system/etc.nix
@@ -48,7 +48,7 @@ in
           etc}
       )
 
-      declare -a etcProblems
+      declare -a etcProblems=()
 
       while IFS= read -r -d "" configFile; do
         subPath=''${configFile#"$systemConfig"/etc/}


modules/system/etc.nix
```diff
@@ -48,7 +48,7 @@ in
           etc}
       )
 
-      declare -a etcProblems
+      declare -a etcProblems=()
```
Enzime, line 51: From my testing this doesn’t fix the problem:

```
$ set -u
$ declare -a etcProblems=()
bash: etcProblems: unbound variable
```

Enzime (changes requested):

felixscheinost:
I think the test you did was incorrect. Can you test the following?

```
$ bash -u -c 'declare -a etcProblems; echo "number of problems: ${#etcProblems[@]}"'
bash: line 1: etcProblems: unbound variable

$ bash -u -c 'declare -a etcProblems=(); echo "number of problems: ${#etcProblems[@]}"'
number of problems: 0
```

It seems that as `etcProblems` is declared as an array it may only be used as such.

Also I am successfully running this branch, while master doesn't work for me. So it seems to be at least a a step in the right direction 😃

Enzime (approved):

felixscheinost:
Cool, thanks!

PR #805, opened 2023-11-10
Add security.pki.installCACerts config

yacinehmito:
Made is possible to disable the management of /etc/ssl/certs/ca-certificates.crt by nix-darwin.

Use case: we distribute corporate CA certificates through our MDM. Keeping them in sync with our shared nix config is a hassle. Instead, we plan on adding a nix-darwin module that sets up a daemon that will periodically export the bundle of certificates from macOS keychain into /etc/ssl/certs/ca-certificates.crt. To make this possible, we need to put /etc/ssl/certs/ca-certificates.crt out of the control of nix-darwin.

### DIFF
--- a/modules/security/pki/default.nix
+++ b/modules/security/pki/default.nix
@@ -21,6 +21,14 @@ in
 
 {
   options = {
+    security.pki.installCACerts = mkOption {
+      type = types.bool;
+      default = true;
+      description = lib.mdDoc ''
+        Whether to enable certificate management with nix-darwin.
+      '';
+    };
+
     security.pki.certificateFiles = mkOption {
       type = types.listOf types.path;
       default = [];
@@ -71,7 +79,7 @@ in
     };
   };
 
-  config = {
+  config = mkIf cfg.installCACerts {
 
     security.pki.certificateFiles = [ "${cacertPackage}/etc/ssl/certs/ca-bundle.crt" ];
 
--- a/modules/services/nix-daemon.nix
+++ b/modules/services/nix-daemon.nix
@@ -63,7 +63,10 @@ in
 
       serviceConfig.EnvironmentVariables = mkMerge [
         config.nix.envVars
-        { NIX_SSL_CERT_FILE = mkDefault config.environment.variables.NIX_SSL_CERT_FILE;
+        {
+          NIX_SSL_CERT_FILE = mkIf
+            (config.environment.variables ? NIX_SSL_CERT_FILE)
+            (mkDefault config.environment.variables.NIX_SSL_CERT_FILE);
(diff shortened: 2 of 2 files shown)


modules/security/pki/default.nix
```diff
@@ -10,20 +10,27 @@ let
...
-     }
+    }
     ''
       cat $files > $out
     '';
 in
 
 {
   options = {
+    security.pki.enable = mkOption {
```
Enzime, line 25: Can you change the name to match the NixOS option?

```suggestion
    security.pki.installCACerts = mkOption {
```

Enzime (changes requested):

modules/security/pki/default.nix
yacinehmito: Sure!

yacinehmito:
@Enzime I renamed as suggested, fixed the issue that popped up in the CI and [made it so that setting NIX_SSL_CERT_FILE becomes optional](https://github.com/LnL7/nix-darwin/pull/805/files#diff-743a678fbcc2f1ee20002a4c5b9ac32e26ca676bea877dd3e356312fa1476ad2R67-R69) (otherwise, setting `security.pki.installCACerts = false;` without setting `environment.variables.NIX_SSL_CERT_FILE` breaks the evaluation of `nix-daemon`).

modules/services/nix-daemon.nix
```diff
@@ -63,7 +63,10 @@ in
 
       serviceConfig.EnvironmentVariables = mkMerge [
         config.nix.envVars
-        { NIX_SSL_CERT_FILE = mkDefault config.environment.variables.NIX_SSL_CERT_FILE;
+        {
+          NIX_SSL_CERT_FILE = mkIf
+            (builtins.hasAttr "NIX_SSL_CERT_FILE" config.environment.variables)
```
Enzime, line 68:
```suggestion
            config.environment.variables ? NIX_SSL_CERT_FILE
```

Enzime: I think you might still need the brackets around the condition but possibly not the expression after it

yacinehmito: I want to pass the whole of `mkDefault config.environment.variables.NIX_SSL_CERT_FILE` as the second argument of `mkIf` so I think it is needed.

Enzime (approved):
Thanks for the contribution, LGTM

yacinehmito:
Thanks for the review!
Would it be interesting for this project that I submit our module that performs the syncing of certificates from the keychains?

Enzime:
Yeah that sounds interesting, contributions are definitely welcome

PR #806, opened 2023-11-10
Install darwin-uninstall by default, which includes the 'empty' config it switches to

### LINKED ISSUE #804: Include the uninstaller in the default closure?
As part of the Determinate Nix Installer work, we've run into a lot of users having frustrating experiences uninstalling nix-darwin. We've moved to detecting nix-darwin in the uninstall step to prevent breaking stuff, but actually doing the uninstall is still a bit frustrating for users.

My thinking is it might be nicer to install `darwin-uninstaller` by default in the system, so the uninstall experience is just running that command.

Another thing I think people would appreciate is if the uninstall command didn't need to run nix-build, and came with the closure and nix-darwin system built in to the script. Something like this:

```nix
{ stdenv, nix, pkgs, makeDarwinSystem }:

let
  uninstallSystem = makeDarwinSystem [ ./configuration.nix ];
in

stdenv.mkDerivation {
  name = "darwin-uninstaller";


  # ...

  shellHook = ''
    #!${stdenv.shell}
    set -e

    # ...

    ${uninstallSystem.system}/sw/bin/darwin-rebuild switch

    # ...
```

Would a PR implementing these two pieces be interesting / wanted?

grahamc:
I ran `nix flake check` and that worked, and I ran `nix-build . -A uninstaller`, and that worked, but I haven't tested it beyond that yet.

Closes #804

### DIFF
--- a/pkgs/darwin-uninstaller/default.nix
+++ b/pkgs/darwin-uninstaller/default.nix
@@ -1,18 +1,17 @@
-{ stdenv, nix, pkgs, nix-darwin }:
+{ stdenv, lib, pkgs }:
 
 let
-  configuration = builtins.path {
-    name = "nix-darwin-uninstaller-configuration";
-    path = ./.;
-    filter = name: _type: name != toString ./default.nix;
+  uninstallSystem = import ../../eval-config.nix {
+    inherit lib;
+    modules = [
+      ./configuration.nix
+      {
+        nixpkgs.source = pkgs.path;
+        nixpkgs.hostPlatform = pkgs.system;
+        system.includeUninstaller = false;
+      }
+    ];
   };
-
-  nixPath = pkgs.lib.concatStringsSep ":" [
-    "darwin-config=${configuration}/configuration.nix"
-    "darwin=${nix-darwin}"
-    "nixpkgs=${pkgs.path}"
-    "$NIX_PATH"
-  ];
 in
 
 stdenv.mkDerivation {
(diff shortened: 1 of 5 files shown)


grahamc:
cc @Enzime

pkgs/darwin-uninstaller/default.nix
```diff
@@ -62,10 +51,7 @@ stdenv.mkDerivation {
         esac
     fi
 
-    export nix=${nix}
-    export NIX_PATH=${nixPath}
-    system=$($nix/bin/nix-build '<darwin>' -A system)
-    $system/sw/bin/darwin-rebuild switch
+    ${uninstallSystem.system}/sw/bin/darwin-rebuild activate
```
grahamc, line 54: Is it sufficient to activate this profile, instead of using `switch`? `activate` really wants to build.

Enzime: Yeah, last time I checked `activate` is `switch` without the `build`

grahamc:
Ok, I think this is good to go!

Enzime:
I've tested this and both the flake and non-flake versions build and uninstall correctly.

I don't have permissions to push to this branch so I made a new branch with a refactored version, which I thought was a bit cleaner as it doesn't introduce a new argument to `lib.darwinSystem` and stays more in line with all of the `nix-tools` being added to `environment.systemPackages` as part of the `nix-darwin` module.

https://github.com/LnL7/nix-darwin/compare/master...Enzime:nix-darwin:easier-uninstall

@grahamc feel free to take a look and tell me what you think of these changes

grahamc:
Looks good to me! Would you like me to push those to my branch?

Enzime:
Yeah that would be great if you could 👍

Enzime:
@grahamc could you push the changes to your branch when you have a chance?

grahamc:
Updated, @Enzime!

Enzime (approved):

PR #807, opened 2023-11-11
Adds support for declarative /etc/hosts config

willemml:
Copies the `/etc/hosts` configuration ability from NixOS. This allows users to configure their hosts declaratively.

On my system `/etc/hosts` was locked, however this appears to not always be the case. I was able to unlock it by running `sudo chflags nouchg,noschg /etc/hosts`.

### DIFF
--- a/modules/networking/default.nix
+++ b/modules/networking/default.nix
@@ -24,6 +24,8 @@ let
 in
 
 {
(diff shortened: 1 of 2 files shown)


Enzime:
Superseded by #939

PR #809, opened 2023-11-16
etc: add more known hashes

Enzime:
This should be less of an issue as of DetSys' nix-installer 0.15.0 which no longer includes the version number in `nix.conf`.


Enzime:
@domenkozar can you take a look at this trivial PR? it's a blocker for installing `nix-darwin` with the latest versions of DetSys' Nix installer

domenkozar (approved):

PR #813, opened 2023-11-19
Power related options (sleep, wake-on-lan, restart after power failure, ...)

PhilipBorgesen:
Hey there, I've recently picked up nix-darwin to configure a Mac Mini M2 as a home server.
I need to configure some power related options that nix-darwin currently does not offer, so I would like to contribute them.

This pull request adds:
- `networking.wakeOnLan.enable`
- `power.restartAfterPowerFailure`
- `power.restartAfterFreeze`
- `power.sleep.computer`
- `power.sleep.display`
- `power.sleep.harddisk`
- `power.sleep.allowSleepByPowerButton`

The options are implemented by means of the `systemsetup` command that also is used for the `time.timezone` option.
I believe this is the command that System Settings uses behind the screen.

I've tried to guess where it made most sense to add the new options.
I've added a new `power` module; maybe it would make more sense to place the restartXXX and sleep options under `system` instead. I'm prepared to move them elsewhere as per your feedback.

Thanks for creating nix-darwin!

### DIFF
--- a/modules/networking/default.nix
+++ b/modules/networking/default.nix
@@ -9,6 +9,8 @@ let
 
   emptyList = lst: if lst != [] then lst else ["empty"];
 
+  onOff = cond: if cond then "on" else "off";
+
   setNetworkServices = optionalString (cfg.knownNetworkServices != []) ''
     networkservices=$(networksetup -listallnetworkservices)
     ${concatMapStringsSep "\n" (srv: ''
@@ -93,6 +95,16 @@ in
       default = [];
       description = "The list of search paths used when resolving domain names.";
     };
+
+    networking.wakeOnLan.enable = mkOption {
+      type = types.nullOr types.bool;
+      default = null;
+      description = ''
+        Enable Wake-on-LAN for the device.
+
+        Battery powered devices may require being connected to power.
+      '';
+    };
   };
 
   config = {
@@ -116,6 +128,10 @@ in
       ''}
 
       ${setNetworkServices}
+
+      ${optionalString (cfg.wakeOnLan.enable != null) ''
+        systemsetup -setWakeOnNetworkAccess '${onOff cfg.wakeOnLan.enable}' &> /dev/null
+      ''}
     '';
 
   };
--- a/modules/power/default.nix
+++ b/modules/power/default.nix
@@ -1,0 +1,47 @@
+{ config, lib, ... }:
+
+let
+  cfg = config.power;
+
+  types = lib.types;
+
+  onOff = cond: if cond then "on" else "off";
+in
+
+{
+  options = {
+    power.restartAfterPowerFailure = lib.mkOption {
+      type = types.nullOr types.bool;
+      default = null;
+      description = ''
+        Whether to restart the computer after a power failure.
+      '';
+    };
+
+    power.restartAfterFreeze = lib.mkOption {
+      type = types.nullOr types.bool;
+      default = null;
+      description = ''
+        Whether to restart the computer after a system freeze.
+      '';
+    };
+  };
+
+  config = {
+
+    system.activationScripts.power.text = ''
+      echo "configuring power..." >&2
+
+      ${lib.optionalString (cfg.restartAfterPowerFailure != null) ''
+        systemsetup -setRestartPowerFailure \
+          '${onOff cfg.restartAfterPowerFailure}' &> /dev/null
+      ''}
+
+      ${lib.optionalString (cfg.restartAfterFreeze != null) ''
+        systemsetup -setRestartFreeze \
+          '${onOff cfg.restartAfterFreeze}' &> /dev/null
+      ''}
+    '';
+
+  };
+}
--- a/modules/power/sleep.nix
+++ b/modules/power/sleep.nix
@@ -1,0 +1,80 @@
+{ config, lib, ... }:
+
+let
+  cfg = config.power.sleep;
+
+  types = lib.types;
+
+  onOff = cond: if cond then "on" else "off";
+in
+
+{
+  options = {
+    power.sleep.computer = lib.mkOption {
+      type = types.nullOr (types.either types.ints.positive (types.enum ["never"]));
+      default = null;
+      example = "never";
+      description = ''
+        Amount of idle time (in minutes) until the computer sleeps.
+
+        `"never"` disables computer sleeping.
+
+        The system might not be considered idle before connected displays sleep, as
+        per the `power.sleep.display` option.
+      '';
+    };
+
+    power.sleep.display = lib.mkOption {
+      type = types.nullOr (types.either types.ints.positive (types.enum ["never"]));
+      default = null;
+      example = "never";
+      description = ''
+        Amount of idle time (in minutes) until displays sleep.
+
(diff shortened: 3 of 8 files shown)


modules/networking/default.nix
```diff
@@ -94,6 +95,16 @@ in
       default = [];
       description = lib.mdDoc "The list of search paths used when resolving domain names.";
     };
+
+    networking.wakeOnLan = mkOption {
```
PhilipBorgesen, line 99: Name of option and placement inspired by the NixOS option: [`networking.interfaces.<name>.wakeOnLan.enable`](https://search.nixos.org/options?channel=23.05&show=networking.interfaces.%3Cname%3E.wakeOnLan.enable)

PhilipBorgesen: Would it make sense to append `.enable`?

modules/networking/default.nix
```diff
@@ -117,6 +128,10 @@ in
       ''}
 
       ${setNetworkServices}
+
+      ${optionalString (cfg.wakeOnLan != null) ''
+        systemsetup -setWakeOnNetworkAccess '${onOff cfg.wakeOnLan}' &> /dev/null
```
PhilipBorgesen, line 133: The `systemsetup` stdout and stderr are silenced:
* The changed value is echoed on stdout, e.g. `setwakeonnetworkaccess: On`.
* An error is reported on stderr, but the command succeeds as reflected by the corresponding get command and System Settings. Example error: `2023-11-18 15:47:54.640 systemsetup[4589:139580] ### Error:-99 File:/AppleInternal/Library/BuildRoots/11aa8fb2-5f4b-11ee-bc7f-926038f30c31/Library/Caches/com.apple.xbs/Sources/Admin/InternetServices.m Line:379`

The above observations apply to the new `power` and `power.sleep` options as well.

PhilipBorgesen: I see this matches the work done here: https://github.com/LnL7/nix-darwin/pull/781

mattpolzin:
@Enzime I hope you don't mind if I ping you on this to get your thoughts. Not my PR but configuring power management would be nice for me now that I am considering retiring my old Mac to be headless.

anderslundstedt:
Any progress on this?

modules/power/default.nix
```diff
@@ -0,0 +1,47 @@
+{ config, lib, ... }:
+
+with lib;
+
+let
+  cfg = config.power;
+
+  onOff = cond: if cond then "On" else "Off";
```
Samasaur1, line 8: This produces `On` or `Off`, while the other two `onOff` implementations produce `on` or `off`. If this is deliberately different, please add a comment detailing why. Otherwise, let's keep them all the same

modules/power/sleep.nix
```diff
@@ -0,0 +1,80 @@
...
+let
+  cfg = config.power.sleep;
+
+  onOff = cond: if cond then "on" else "off";
+in
+
+{
+  options = {
+    power.sleep.computer = mkOption {
+      type = types.nullOr (types.either types.ints.unsigned (types.enum ["never"]));
```
Samasaur1, line 14: If `0` and `never` behave identically, I think the standard way that is handled in nix-darwin is not to accept a "never" option and to just note in the description that `0` disables the feature.

Samasaur1 (changes requested):

modules/power/default.nix
PhilipBorgesen: Difference was not intentional. Fixed.

modules/power/sleep.nix
PhilipBorgesen: I can't remember my original reasoning for writing the module the way I did, but I would much rather disable sleep by specifying `never` than having to remember what `0` means or having to consult the documentation; `never` is more readable.

The `systemsetup` man page describes to use `never` or `off` to disable the sleeping, but does not specify what happens if `0` is set. That they behave identically have been derived empirically.

If it's an issue that `0` and `never` behave identically, then I would rather change the type to `types.ints.positive` and require `never` for disabling sleep.

Enzime:
```suggestion
      type = types.nullOr (types.either types.ints.positive (types.enum ["never"]));
```

Sounds good to me 👍

PhilipBorgesen: Done

modules/power/default.nix
```diff
@@ -0,0 +1,47 @@
+{ config, lib, ... }:
+
+with lib;
```
Enzime, line 3:
```suggestion
```

modules/power/default.nix
```diff
@@ -0,0 +1,47 @@
...
+
+  onOff = cond: if cond then "on" else "off";
+in
+
+{
+  options = {
+    power.restartAfterPowerFailure = mkOption {
+      type = types.nullOr types.bool;
+      default = null;
+      description = lib.mdDoc ''
```
Enzime, line 16:
```suggestion
      description = ''
```

All the `lib.mdDoc` calls are deprecated and need to be removed

modules/power/default.nix
```diff
@@ -0,0 +1,47 @@
...
+
+let
+  cfg = config.power;
+
+  onOff = cond: if cond then "on" else "off";
+in
+
+{
+  options = {
+    power.restartAfterPowerFailure = mkOption {
```
Enzime, line 13:
```suggestion
    power.restartAfterPowerFailure = lib.mkOption {
```

modules/power/sleep.nix
```diff
@@ -0,0 +1,80 @@
+{ config, lib, ... }:
+
+with lib;
```
Enzime, line 3:
```suggestion
```

modules/power/sleep.nix
```diff
@@ -0,0 +1,80 @@
...
+      default = null;
+      description = lib.mdDoc ''
+        Whether the power button can sleep the computer.
+      '';
+    };
+  };
+
+  config = {
+
+    system.activationScripts.sleep.text = ''
```
Enzime, line 60:
```suggestion
    system.activationScripts.power.text = lib.mkAfter ''
```

I think adding a `sleep` activation script is unnecessary, I'd rather see it use the `power` one

Enzime (changes requested):
Could you also add some tests?

modules/power/default.nix
PhilipBorgesen: Why? I see this pattern is being used in the other modules.

Enzime: We're in the process of removing `with lib;` like `nixpkgs`

PhilipBorgesen: I assume that the work to do so already is in progress, in a different branch/fork.

Otherwise it would be cleaner to make the change in one commit.

I'll remove `with lib;`.

Enzime: I don't think anyone's started writing it yet, however ideally preferably preventing new code from using `with lib;` will reduce the amount of work in the future refactor

PhilipBorgesen:
Tests added

Enzime (approved):
LGTM

Thanks for the contribution

PR #814, opened 2023-11-19
[yabai] Configure scripting addition

### LINKED ISSUE #750: Is yabai being properly setup?
Yabai recently changed how scripting additions and it's service work, I wanted to know if those changes are reflected somewhere. I see some configuration in the service but the yabai wiki says a sudoers file needs to be created and that some boilerplate commands need to be added to config, which I don't see in the launch agent or yabairc loaded from the launch agent. 

From  [`modules/services/yabai/default.nix` ](https://github.com/LnL7/nix-darwin/blob/master/modules/services/yabai/default.nix) 
```nix
    # TODO: [@cmacrae] Handle removal of yabai scripting additions
    (mkIf (cfg.enableScriptingAddition) {
      launchd.daemons.yabai-sa = {
        script = ''
          if [ ! $(${cfg.package}/bin/yabai --check-sa) ]; then
            ${cfg.package}/bin/yabai --install-sa
          fi

          ${cfg.package}/bin/yabai --load-sa
        '';

      };
    })
```

From the [yabai wiki](https://github.com/koekeishiya/yabai/wiki/Installing-yabai-(latest-release)#configure-scripting-addition)

# Configure scripting addition
yabai uses the macOS Mach APIs to inject code into Dock.app; this requires elevated (root) privileges. You can configure your user to execute yabai --load-sa as the root user without having to enter a password. To do this, we add a new configuration entry that is loaded by /etc/sudoers.

```shell
# create a new file for writing - visudo uses the vim editor by default.
# go read about this if you have no idea what is going on.

sudo visudo -f /private/etc/sudoers.d/yabai

# input the line below into the file you are editing.
#  replace <yabai> with the path to the yabai binary (output of: which yabai).
#  replace <user> with your username (output of: whoami). 
#  replace <hash> with the sha256 hash of the yabai binary (output of: shasum -a 256 $(which yabai)).
#   this hash must be updated manually after running brew upgrade.

<user> ALL=(root) NOPASSWD: sha256:<hash> <yabai> --load-sa
```
After the above edit has been made, add the command to load the scripting addition at the top of your yabairc config file

```shell
# for this to work you must configure sudo such that
# it will be able to run the command without password

yabai -m signal --add event=dock_did_restart action="sudo yabai --load-sa"
sudo yabai --load-sa

# .. more yabai startup stuff
```

I also don't think you need `yabai --install-sa` and yabai also has a built in service now, though maybe nix needs to manage that because configs are stored in non default locations. 

The actual package seems like it just downloads the latest release instead of compiling. 

I'm pretty new to the nix ecosystem, so It's possible this may be happening somewhere else, but I thought I'd ask. 

Additional:

There is a `yabai-sa-start` file in my nix store, however this would need to be run with sudo permissions to load the scripting additions and `--check-sa` and `--install-sa` are no longer valid options. 

```shell
bash /nix/store/1w813awldn35yzixvsfzwbz1s3r39nv6-yabai-sa-start                                                                                                                                                                                                                                                            
#yabai: '--check-sa' is not a valid option!                                                                                                                                                                                                                                                                                   
#yabai: '--install-sa' is not a valid option!                                                                                                                                                                                                                                                                                 
#yabai: scripting-addition must be loaded as root!  
```

aeolyus:
Hey there, I've recently picked up nix to configure a MacBook and was setting up yabai with nix-darwin when I noticed that the yabai scripting addition was not configured properly. See [[here in the yabai install wiki]](https://github.com/koekeishiya/yabai/wiki/Installing-yabai-(latest-release)#configure-scripting-addition).

> yabai uses the macOS Mach APIs to inject code into Dock.app; this requires elevated (root) privileges. You can configure your user to execute yabai --load-sa as the root user without having to enter a password. To do this, we add a new configuration entry that is loaded by /etc/sudoers.

This PR will allow admins to execute `yabai --load-sa` as the root user without having to enter a password.

This should partly fix #750. However, I have found that I still need to include
```
yabai -m signal --add event=dock_did_restart action="sudo yabai --load-sa"
sudo yabai --load-sa
```
in my `extraConfig` in order for this to work. Not sure if the current `org.nixos.yabai-sa` service is working as expected.

Disclaimer: this is my very first nix pr and my third day of learning about the nix ecosystem so would love any feedback! :)

Thanks for creating nix-darwin!

### DIFF
--- a/modules/services/yabai/default.nix
+++ b/modules/services/yabai/default.nix
@@ -7,9 +7,10 @@ let
 
   toYabaiConfig = opts:
     concatStringsSep "\n" (mapAttrsToList
-      (p: v: "yabai -m config ${p} ${toString v}") opts);
+      (p: v: "yabai -m config ${p} ${toString v}")
+      opts);
 
-  configFile = mkIf (cfg.config != {} || cfg.extraConfig != "")
+  configFile = mkIf (cfg.config != { } || cfg.extraConfig != "")
     "${pkgs.writeScript "yabairc" (
       (if (cfg.config != {})
        then "${toYabaiConfig cfg.config}"
@@ -42,7 +43,7 @@ in
 
     services.yabai.config = mkOption {
       type = attrs;
-      default = {};
+      default = { };
       example = literalExpression ''
         {
           focus_follows_mouse = "autoraise";
@@ -77,7 +78,7 @@ in
 
       launchd.user.agents.yabai = {
         serviceConfig.ProgramArguments = [ "${cfg.package}/bin/yabai" ]
-                                         ++ optionals (cfg.config != {} || cfg.extraConfig != "") [ "-c" configFile ];
+          ++ optionals (cfg.config != { } || cfg.extraConfig != "") [ "-c" configFile ];
 
         serviceConfig.KeepAlive = true;
         serviceConfig.RunAtLoad = true;
@@ -101,6 +102,12 @@ in
         serviceConfig.RunAtLoad = true;
         serviceConfig.KeepAlive.SuccessfulExit = false;
       };
+
+      environment.etc."sudoers.d/yabai".text =
+        let
+          sha = builtins.hashFile "sha256" "${cfg.package}/bin/yabai";
+        in
+        "%admin ALL=(root) NOPASSWD: sha256:${sha} ${cfg.package}/bin/yabai --load-sa";
     })
   ];
 }


Enzime (approved):
Thanks for the contribution

LGTM :+1:

Enzime:
If you're interested, feel free to create another PR that adds those lines you mentioned above to `configFile` if `services.yabai.enableScriptingAddition` is enabled

PR #816, opened 2023-11-22
Add `nix.linux-builder.systems` option to set corresponding `nix.buildMachines.*.systems` option

ethnt:
I've set up my local `linux-builder` to emulate `x86_64-linux` as well as `aarch64-linux` (jury's out as to how well that will work), but I still need to be able to set that option in `/etc/nix/machines`. I've added an option for this in `nix.linux-builder`. Potentially in the future it would be a better idea to simply allow `extraBuilderConfig` or something similar, but I stuck with convention in this PR.

### DIFF
--- a/modules/nix/linux-builder.nix
+++ b/modules/nix/linux-builder.nix
@@ -136,9 +136,9 @@ in
       hostName = "linux-builder";
       sshUser = "builder";
       sshKey = "/etc/nix/builder_ed25519";
-      system = "${stdenv.hostPlatform.uname.processor}-linux";
+      systems = [ "${stdenv.hostPlatform.uname.processor}-linux" ] ++ cfg.systems;
       publicHostKey = "c3NoLWVkMjU1MTkgQUFBQUMzTnphQzFsWkRJMU5URTVBQUFBSUpCV2N4Yi9CbGFxdDFhdU90RStGOFFVV3JVb3RpQzVxQkorVXVFV2RWQ2Igcm9vdEBuaXhvcwo=";
-      inherit (cfg) maxJobs supportedFeatures systems;
+      inherit (cfg) maxJobs supportedFeatures;
     }];
 
     nix.settings.builders-use-substitutes = true;
--- a/modules/nix/linux-builder.nix
+++ b/modules/nix/linux-builder.nix
@@ -57,6 +57,22 @@ in
       '';
     };
 
+    systems = mkOption {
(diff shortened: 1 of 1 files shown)


modules/nix/linux-builder.nix
```diff
@@ -120,7 +136,7 @@ in
       hostName = "linux-builder";
       sshUser = "builder";
       sshKey = "/etc/nix/builder_ed25519";
-      system = "${stdenv.hostPlatform.uname.processor}-linux";
+      systems = [ "${stdenv.hostPlatform.uname.processor}-linux" ] ++ cfg.systems;
```
Enzime, line 139:
```suggestion
      inherit (cfg) systems;
```

Enzime (changes requested):
I think it makes more sense to specify the default `system` as part of the new `systems` option

modules/nix/linux-builder.nix
```diff
@@ -57,6 +57,22 @@ in
       '';
     };
 
+    systems = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
```
Enzime, line 62:
```suggestion
      default = [ "${stdenv.hostPlatform.uname.processor}-linux" ];
```

modules/nix/linux-builder.nix
```diff
@@ -57,6 +57,22 @@ in
...
+      example = literalExpression ''
+        [
+          "x86_64-linux"
+          "aarch64-linux"
+        ]
+      '';
+      description = ''
```
Enzime, line 69:
```suggestion
      description = lib.mdDoc ''
```

Enzime (changes requested):

ethnt:
@Enzime Thanks for the review, I've made those changes!

modules/nix/linux-builder.nix
```diff
@@ -57,6 +57,22 @@ in
       '';
     };
 
+    systems = mkOption {
+      type = types.listOf types.str;
+      default = [ "${stdenv.hostPlatform.uname.processor}-linux" ];
```
Enzime, line 62:
```suggestion
      default = [ "${stdenv.hostPlatform.uname.processor}-linux" ];
      defaultText = literalExpression ''[ "''${stdenv.hostPlatform.uname.processor}-linux" ]'';
```

Enzime (changes requested):

PR #818, opened 2023-11-23
fix: user shell path handling

hraban:
Properly detect the binary name (not just /nix/store/...-bash, but include the .../bin/bash), and use the symlinked name which also appears in /etc/shells.

### DIFF
--- a/modules/users/default.nix
+++ b/modules/users/default.nix
@@ -24,6 +24,14 @@ let
   deletedUsers = filter (n: isDeleted cfg.users n) cfg.knownUsers;
 
   packageUsers = filterAttrs (_: u: u.packages != []) cfg.users;
+
+  # convert a valid argument to user.shell into a string that points to a shell
+  # executable. Logic copied from modules/system/shells.nix.
+  shellPath = v:
+    if types.shellPackage.check v
+    then "/run/current-system/sw${v.shellPath}"
+    else v;
+
 in
 
 {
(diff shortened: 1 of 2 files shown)


modules/users/default.nix
```diff
@@ -158,12 +166,14 @@ in
...
           dscl . -create '/Users/${v.name}' NFSHomeDirectory '${v.home}'
-          dscl . -create '/Users/${v.name}' UserShell '${v.shell}'
           ${optionalString v.createHome "createhomedir -cu '${v.name}'"}
         else
           if [ "$u" -ne ${toString v.uid} ]; then
             echo "[1;31mwarning: existing user '${v.name}' has unexpected uid $u, skipping...[0m" >&2
           fi
+
+          # Always set the shell path, in case it was updated
+          dscl . -create '/Users/${v.name}' UserShell ${lib.escapeShellArg (shellPath v.shell)}
```
Enzime, line 176: From the way I understand the logic in this file, your changes won't always set the shell, they'll only run when the user already exists so they won't get set when your user gets created for the first time.

Ideally it should always run regardless of whether we're creating the user for the first time, and it shouldn't run when the existing user has an unexpected `uid`.

Enzime (changes requested):

modules/users/default.nix
hraban: Are you sure that's not already the case? Note that in the if guard above, this code is in the else branch. See also the error message just above this one ("existing user..."). See also the definition of createdUsers, which eventually boils down to isCreated, which is just a check if the user exists in your config. That also makes sense if you think about it because the existence of users is state, so nix code can't detect it, it would have to be in the activation script.

I also currently use this PR locally and it has the expected effect on existing users.

hraban: Never mind I missed the "always" in your comment, good point!

rounakdatta:
Can we get this merged?

hraban:
@LnL7 @Enzime anything I can help with to get this over the line?

Enzime (approved):
LGTM

PR #819, opened 2023-11-23
fix: tests failing to build on unstable Nix

Enzime:
Use Nix 2.13.6 which is the default version for NixOS 23.05.


Enzime:
@domenkozar could you take a look at this PR? tests are currently failing on master and some PR branches meaning they can't be merged because the unstable version of Nix we're currently using is segfaulting

domenkozar (approved):

PR #820, opened 2023-11-30
Add `universalaccess.reduceMotion` option

mhumeSF:
Adding config to manage `reduce motion` option:
https://support.apple.com/guide/mac-help/stop-or-reduce-onscreen-motion-mchlc03f57a1/mac


Enzime (approved):

PR #822, opened 2023-12-04
checks: disable `nixGarbageCollector` for multi-user

zowoq:
Currently if you run activation as root you need to set `nix.gc.user = "root";` or `nix.gc.user = "";`.

https://github.com/LnL7/nix-darwin/blob/4b9b83d5a92e8c1fbfd8eb27eda375908c11ec4d/modules/system/checks.nix#L184-L192

### DIFF
--- a/modules/system/checks.nix
+++ b/modules/system/checks.nix
@@ -240,7 +240,7 @@ in
       (mkIf (config.nix.useDaemon && cfg.verifyBuildUsers) buildUsers)
       (mkIf (!config.nix.useDaemon) singleUser)
       nixStore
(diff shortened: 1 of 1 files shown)


zowoq:
Rebased to fix merge conflict.

emilazy:
I believe checks currently run as part of user activation; I’m not sure we support running that as root. I would like to move all activation to be run as root, but as it stands I’m worried that this might just mask other issues with that configuration.

zowoq:
> I’m not sure we support running that as root.

No, not officially.

> move all activation to be run as root

I'll close the PR as this is the better option.

PR #825, opened 2023-12-07
Add support for system preferences that are stored as dictionaries

davish:
Some plist attributes are represented by dictionaries. This PR adds support for dictionary system settings backed by attribute sets. An example custom preference that failed with `invalid value type` before but works now:

```nix
system.defaults.CustomUserPreferences = {
    "com.google.Chrome" = {
        "NSUserKeyEquivalents" = {
            "Open Location..." = "~d";
        };
    };
};
```


Enzime:
I believe this should be fixed by #897

PR #831, opened 2023-12-15
darwin-rebuild: fix sudo invocation on High Sierra

rkitover:
darwin-rebuild: fix sudo invocation on High Sierra

The version of sudo on macOS 10.13 High Sierra does not support the
`--preserve-env=LIST` option syntax, which is used here to preserve
PATH.

Override sudo with a shell function that checks for the availability of
this option syntax, and do not pass PATH otherwise.

Also move the prefix env command in the sudo invocations into the new
sudo function.

### DIFF
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -16,6 +16,14 @@ showSyntax() {
   exit 1
 }
 
+sudo() {
+  if command sudo --help | grep -- --preserve-env= >/dev/null; then
+    command sudo -H --preserve-env=PATH env "$@"
+  else
+    command sudo -H "$@"
+  fi
+}
+
 # Parse the command line.
 origArgs=("$@")
 extraMetadataFlags=()
@@ -187,7 +195,7 @@ fi
 
 if [ "$action" = list ] || [ "$action" = rollback ]; then
   if [ "$USER" != root ] && [ ! -w $(dirname "$profile") ]; then
-    sudo -H --preserve-env=PATH env nix-env -p "$profile" "${extraProfileFlags[@]}"
+    sudo nix-env -p "$profile" "${extraProfileFlags[@]}"
   else
     nix-env -p "$profile" "${extraProfileFlags[@]}"
   fi
@@ -205,7 +213,7 @@ if [ -z "$systemConfig" ]; then exit 0; fi
 
 if [ "$action" = switch ]; then
   if [ "$USER" != root ] && [ ! -w $(dirname "$profile") ]; then
-    sudo -H --preserve-env=PATH env nix-env -p "$profile" --set "$systemConfig"
+    sudo nix-env -p "$profile" --set "$systemConfig"
   else
     nix-env -p "$profile" --set "$systemConfig"
   fi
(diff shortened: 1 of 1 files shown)


pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -16,6 +16,14 @@ showSyntax() {
   exit 1
 }
 
+sudo() {
+  if command sudo --help | grep -- --preserve-env= >/dev/null; then
+    command sudo -H --preserve-env=PATH "$@"
+  else
+    command sudo -H --preserve-env "$@"
```
Enzime, line 23:
```suggestion
    command sudo -H "$@"
```

If we can't preserve just `$PATH`, I reckon it should be fine to just preserve `$HOME`

Enzime (changes requested):

rkitover:
How's this?

pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -16,6 +16,14 @@ showSyntax() {
   exit 1
 }
 
+sudo() {
+  if command sudo --help | grep -- --preserve-env= >/dev/null; then
+    command sudo -H --preserve-env=PATH "$@"
+  else
+    command sudo -H -- sh -c 'PATH='"'$PATH'"' "$@"' -- "$@"
```
Enzime, line 23: Not a big fan of how brittle that command looks, I would personally much rather not have `$PATH` passed, as it was only implemented for #798 and I'd consider it if more people ran into the same issue

```suggestion
    command sudo -H "$@"
```

Enzime (changes requested):

pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -16,6 +16,14 @@ showSyntax() {
   exit 1
 }
 
+sudo() {
+  if command sudo --help | grep -- --preserve-env= >/dev/null; then
+    command sudo -H --preserve-env=PATH "$@"
```
Enzime, line 21:
```suggestion
    command sudo -H --preserve-env=PATH env "$@"
```

As the `env` command was added to ensure that the preserved `PATH` gets searched for the binary, I think it'd be better to move it here

pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -187,7 +195,7 @@ fi
 
 if [ "$action" = list ] || [ "$action" = rollback ]; then
   if [ "$USER" != root ] && [ ! -w $(dirname "$profile") ]; then
-    sudo -H --preserve-env=PATH env nix-env -p "$profile" "${extraProfileFlags[@]}"
+    sudo env nix-env -p "$profile" "${extraProfileFlags[@]}"
```
Enzime, line 198:
```suggestion
    sudo nix-env -p "$profile" "${extraProfileFlags[@]}"
```

pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -205,7 +213,7 @@ if [ -z "$systemConfig" ]; then exit 0; fi
 
 if [ "$action" = switch ]; then
   if [ "$USER" != root ] && [ ! -w $(dirname "$profile") ]; then
-    sudo -H --preserve-env=PATH env nix-env -p "$profile" --set "$systemConfig"
+    sudo env nix-env -p "$profile" --set "$systemConfig"
```
Enzime, line 216:
```suggestion
    sudo nix-env -p "$profile" --set "$systemConfig"
```

rkitover:
Updated.

Enzime (approved):
Thanks for the contribution

LGTM

PR #832, opened 2023-12-17
vim: don't use deprecated vam attribute

getchoo:
As of https://github.com/NixOS/nixpkgs/pull/179039 (~22.11), the `vam` attribute of `vimrcConfig` has been deprecated. This moves the module to the recommend `packages` attribute, removing the trace most systems will come across currently and making adding plugins a little bit easier. The default "sensible" plugins have also been replaced with the less opinionated `vim-sensible` - similar to home-manager.

***This is a breaking change***

### DIFF
--- a/modules/programs/vim.nix
+++ b/modules/programs/vim.nix
@@ -28,31 +28,10 @@ in
       description = lib.mdDoc "Enable sensible configuration options for vim.";
     };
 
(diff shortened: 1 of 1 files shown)


chrisportela:
This is still an issue today on the latest versions of nix-darwin.

PR #836, opened 2023-12-22
etc: allow overwriting files with unknown hashes

Enzime:
For users encountering #149

### DIFF
--- a/modules/lib/write-text.nix
+++ b/modules/lib/write-text.nix
@@ -58,6 +58,15 @@ in
       type = types.listOf types.str;
       default = [];
     };
+
+    force = mkOption {
+      type = types.bool;
+      default = false;
+      internal = true;
+      description = lib.mdDoc ''
+        Whether this file should be replaced even if an unknown version of the file already exists.
+      '';
+    };
   };
 
   config = {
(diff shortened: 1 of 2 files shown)


emilazy:
Is this still desirable now that we have known hashes for `/etc/shells` and `/etc/nix/nix.conf`, by the far the most common occurrences of this? I’m especially wary of a configuration‐based approach, as opposed to something more one‐time like an environment variable, as I suspect people will opt in and then forget about it. Overall I lean towards handling this cleanly upstream whenever we can and letting users manually `mv` things out of the way for the cases we can’t deal with ourselves.

Enzime:
I agree that making this an option that users can set and forget in their configuration is not ideal, I'm happy to close this PR for now

The main reason I created this commit in the first place was for my darwin test VM where `/etc/nix/nix.conf` would not have a fixed content and I just wanted to ignore that, I'll have to think more potentially about how I'll implement it

emilazy:
A potential simple solution would be an environment variable containing filenames or hashes that are okay to overwrite. You could also just add `rm /etc/nix/nix.conf` to the activation scripts if you wanted to do something really hacky :)

PR #837, opened 2023-12-23
etc: add known hash for `/etc/shells`

Enzime:
#149


Enzime:
@domenkozar can you review this PR? it should unblock users that are setting `environment.shells` and have to manually deal with `/etc/shells` getting reset every update

domenkozar (approved):

PR #839, opened 2023-12-29
GlobalPreferences: fix mouse scaling example

khaneliman:
Follow up to https://github.com/LnL7/nix-darwin/pull/838 . 

The change requires a float number but example is integer and will not build.


Enzime (approved):
Good catch, thanks for the contribution :+1:

PR #840, opened 2023-12-29
sketchybar: add logFile and log output

khaneliman:
Sketchybar configs contain a lot of scripts running and it's useful to have the log output directed somewhere to see what's failing and why. Adding a stdout and stderr output file to monitor sketchybar service.

### DIFF
--- a/modules/services/sketchybar/default.nix
+++ b/modules/services/sketchybar/default.nix
@@ -43,6 +43,12 @@ in
         and [example](https://github.com/FelixKratz/SketchyBar/blob/master/sketchybarrc).
       '';
     };
+
+    logFile = mkOption {
+      type = types.nullOr types.path;
+      default = "/var/tmp/sketchybar.log";
+      description = lib.mdDoc "Absolute path to log all stderr and stdout";
+    };
   };
 
   config = mkIf cfg.enable {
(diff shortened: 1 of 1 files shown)


Samasaur1 (changes requested):
same comment as #1231, I assume we want separate standard output and standard error log files. otherwise good.

modules/services/sketchybar/default.nix
```diff
@@ -43,17 +54,39 @@ in
...
+      example = "/Users/khaneliman/Library/Logs/sketchybar.log";
+      description = "Absolute path to log all stderr output.";
+    };
+
+    outLogFile = mkOption {
+      type = types.nullOr types.path;
+      default = "/var/tmp/sketchybar.out.log";
+      example = "/Users/khaneliman/Library/Logs/sketchybar.log";
+      description = "Absolute path to log all stdout output.";
+    };
```
Samasaur1, line 70: Ideally this would match upstream as well, which I believe is under `/var/log`:

https://github.com/FelixKratz/homebrew-formulae/blob/8cd8fae90b2e154e3fa3092cc5f8e3a6e963acd0/sketchybar.rb#L61-L68

but this is a user agent and that's not a user-owned directory. I grepped and we don't seem to have a consistent location. Personally I would use `/tmp`, but this is fine.

Samasaur1 (approved):

PR #841, opened 2023-12-30
Fix type error

carlthome:
Not sure if this is intended but looks like `serviceConfig.StartCalendarInterval` has to wrapped in a list so would be good if the example follows that assumption as well.

```
error: A definition for option `launchd.user.agents.[...].serviceConfig.StartCalendarInterval' is not of type `null or (list of (submodule))'. Definition values:
       - In `<unknown-file>':
           {
             Hour = 0;
             Minute = 0;
           }
```

### DIFF
--- a/modules/launchd/launchd.nix
+++ b/modules/launchd/launchd.nix
@@ -342,7 +342,7 @@ with lib;
 
     StartCalendarInterval = mkOption {
       default = null;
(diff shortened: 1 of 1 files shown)


Enzime (approved):
LGTM, thanks for the contribution

PR #842, opened 2023-12-31
linux-builder: remove trusted user requirement

Enzime:
If you set up a signing key for the `linux-builder` and add that as trusted public key on your machine, you won't need to be a trusted user at all.


Enzime:
cc @domenkozar

emilazy (approved):

PR #850, opened 2024-01-14
linux-builder: Delete VM's disk image on service restart 

tfc:
After enabling the linux builder on a new macOS system, the first thing I do is adding more configuration (bigger disk, more cores and RAM, etc.) and then rebuild the system again.

The problem is that such settings (especially disk space) do not propagate automatically into the disk image or can be incompatible with the state in the disk image.

This PR does:

- dedupe the `/var/lib/darwin-builder` string literal
- Delete the `nixos.qcow2` image of the builder VM upon service restart

This should only harm build time as it drops nix store contents, too.
If this is considered too harsh, we could also add a config attribute `nix.linux-builder.restart-prunes-disk = true/false;` or so that is `false` by default.


modules/nix/linux-builder.nix
```diff
@@ -11,6 +11,8 @@ let
     modules = [ cfg.config ];
   };
 
+  WorkingDirectory = "/var/lib/darwin-builder";
```
Enzime, line 14: I think it would be good if you could make this into an option

modules/nix/linux-builder.nix
```diff
@@ -20,6 +22,7 @@ let
     rm -rf $TMPDIR
     mkdir -p $TMPDIR
     trap "rm -rf $TMPDIR" EXIT
+    rm -f ${WorkingDirectory}/nixos.qcow2
```
Enzime, line 25: If you could also make an option that controls this, possibly named something like `wipeOnRestart` or `ephemeral`

Enzime (changes requested):

modules/nix/linux-builder.nix
tfc: done.

modules/nix/linux-builder.nix
tfc: Done.

tfc:
@Enzime thank you for your actionable feedback. It's all incorporated now.

modules/nix/linux-builder.nix
```diff
@@ -20,6 +20,7 @@ let
     rm -rf $TMPDIR
     mkdir -p $TMPDIR
     trap "rm -rf $TMPDIR" EXIT
+    ${lib.optionalString cfg.ephemeral "rm -f ${cfg.workingDirectory}/nixos.qcow2"}
```
Enzime, line 23:
```suggestion
    ${lib.optionalString cfg.ephemeral "rm ${cfg.workingDirectory}/{cfg.config.networking.hostname}.qcow2"}
```

modules/nix/linux-builder.nix
```diff
@@ -77,6 +78,22 @@ in
...
+        The working directory of the Linux builder daemon process.
+      '';
+    };
+
+    ephemeral = mkEnableOption (lib.mdDoc ''
+      delete the builder's disk image on ever service restart.
+
+      This is useful whenever you change the builder's system configuration or
+      simply don't want it to accumulate state. Leave it disabled if you need to
+      rely on secrets or other information in the builder's file system.
```
Enzime, line 95:
```suggestion
      Wipe the builder's filesystem on every restart.

      This is disabled by default as maintaining the builder's Nix Store reduces
      rebuilds. You can enable this if you don't want your builder to accumulate
      state.
```

Enzime (changes requested):

modules/nix/linux-builder.nix
tfc: btw the small first letter is due to the fact that `mkEnableOption` prefixes the doc string with `"Whether to "`, or is that disabled in nix-darwin?

tfc: done

modules/nix/linux-builder.nix
tfc: had to slightly change this but it's done.

tfc:
@Enzime all done

Enzime (approved):
LGTM

PR #854, opened 2024-01-19
fix shell escaping in networking config

kallisti-dev:
Fix a bug where using a single quote in any of the networking names caused the following error:

```
/nix/store/8zbx7q8xxrsd46m0n0bspnjvm27ak3sr-darwin-system-24.05.20240108.317484b+darwin4.0dd382b/activate: line 264: unexpected EOF while looking for matching `''
```

Example config:

```nix
  networking.hostName = "Adams-MacBook-Pro";
  networking.computerName = "Adam's MacBook Pro";
```

### DIFF
--- a/modules/networking/default.nix
+++ b/modules/networking/default.nix
@@ -107,13 +107,13 @@ in
       echo "configuring networking..." >&2
 
       ${optionalString (cfg.computerName != null) ''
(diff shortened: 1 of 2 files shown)


Enzime (approved):
LGTM

Thanks for the contribution :+1:

PR #857, opened 2024-01-19
Allow launchd serivceConfig.LimitLoadToSessionType to be a list

thanegill:
Per: https://developer.apple.com/library/archive/technotes/tn2083/_index.html#:~:text=If%20you%20want%20to%20run%20in%20more%20than%20one%20session%20type%2C%20you%20can%20set%20LimitLoadToSessionType%20to%20an%20array%2C%20where%20each%20element%20is%20a%20session%20type%20string.

LimitLoadToSessionType can also be an array if more than one session type is desired.

### DIFF
--- a/modules/launchd/launchd.nix
+++ b/modules/launchd/launchd.nix
@@ -91,7 +91,7 @@ with lib;
     };
 
     LimitLoadToSessionType = mkOption {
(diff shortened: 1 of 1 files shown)


Enzime (approved):
LGTM

PR #860, opened 2024-01-28
Use nixpkgs generators.toPlist for launchd service generation

### LINKED ISSUE #93: Use lib.toPlist for launchd
Just so that we have more users, it would be nice to get launchd stuff here using lib.toPlist from Nixpkgs. I can contribute fixes there if implementation needs to change. Otherwise it should just be this line:

https://github.com/LnL7/nix-darwin/blob/master/modules/launchd/default.nix#L13

to something like

```
 value.text = lib.generators.toPlist {} value.serviceConfig;
```

thanegill:
Fixes #93


Enzime (approved):
LGTM

Thanks for the contribution

PR #861, opened 2024-01-28
Add zsh completions to darwin-rebuld by default

thanegill:
Added a few more options that weren't in the example. Not checked to see if all arguments/flags have completions.

### DIFF
--- a/pkgs/nix-tools/darwin-rebuild.zsh-completions
+++ b/pkgs/nix-tools/darwin-rebuild.zsh-completions
@@ -7,7 +7,7 @@ local -a _1st_arguments
 _1st_arguments=(
   'edit:Open nix-darwin config in $EDITOR'\
   'switch:Build, activate, and update the current generation'\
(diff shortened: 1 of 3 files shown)


pkgs/nix-tools/darwin-rebuild.zsh-completions
```diff
@@ -0,0 +1,38 @@
+#compdef darwin-rebuild
+#autoload
+
+_nix-common-options
+
+local -a _1st_arguments
+_1st_arguments=(
+  'edit:Open nix-darwin config in $EDITOR'\
+  'switch:Build, activate, and update the current generation'\
+  'activate:Activate, and update the current generation'\
```
Enzime, line 10:
```suggestion
  'activate:Activate an already built store path and update the current generation'\
```

Enzime (changes requested):

Enzime (approved):
LGTM

PR #863, opened 2024-02-01
services/yabai: Remove IFD

nrabulinski:
`enableScriptingAddition` no longer triggers IFD by using runCommand to generate `sudoers.d/yabai`, instead of `builtins.hashFile` and interpolating the string in nix.


Enzime (approved):
Thanks for the contribution

LGTM

PR #866, opened 2024-02-11
security.sudo.extraConfig: init

### LINKED ISSUE #793: Add `security.sudo.extraConfig`
On NixOS, there's an option `security.sudo.extraConfig`. It allows me to do things like:

```nix
security.sudo.extraConfig = ''
  Defaults pwfeedback
  Defaults lecture = always
  Defaults lecture_file = ${./sudoers.txt}
'';
```

Although the same option does not exist in nix-darwin, I was able to replicate it by doing:

```nix
environment.etc."sudoers.d/prompt".text = ''
  Defaults pwfeedback
  Defaults lecture = always
  Defaults lecture_file = ${./sudoers.txt}
'';
```

which works because the default macOS `/etc/sudoers` file includes these three lines at the end by default:

```
## Read drop-in files from /private/etc/sudoers.d
## (the '#' here does not indicate a comment)
#includedir /private/etc/sudoers.d
```

nix-darwin should add this option, whether it be via the same method I used or some other method. If I have time, I will attempt to add this, but if somebody else wants to do so, they can.

Samasaur1:
This adds the option `security.sudo.extraConfig`, which puts a file in `/etc/sudoers.d`. Because macOS' `/etc/sudoers` file includes these lines:
```
## Read drop-in files from /private/etc/sudoers.d
## (the '#' here does not indicate a comment)
#includedir /private/etc/sudoers.d
```
this allows extra configuration, while also matching the NixOS option and thus allowing common config across operating systems.

Closes #793

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -7,6 +7,7 @@
   ./security/pam.nix
   ./security/pki
   ./security/sandbox
+  ./security/sudo.nix
   ./system
   ./system/base.nix
   ./system/checks.nix
--- a/modules/security/sudo.nix
+++ b/modules/security/sudo.nix
@@ -1,0 +1,26 @@
+{ config, lib, ... }:
+
+with lib;
+
+let
+  cfg = config.security.sudo;
+in
+{
+  meta.maintainers = [
+    lib.maintainers.samasaur or "samasaur"
+  ];
+
+  options = {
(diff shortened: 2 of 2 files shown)


Enzime (approved):
Thanks for the contribution

LGTM

PR #872, opened 2024-02-17
Add NSStatusItemSpacing and NSStatusItemSelectionPadding 

robharrop:
Add `NSStatusItemSpacing` and `NSStatusItemSelectionPadding` to `NSGlobalDomain` options.

These options control the spacing between and padding inside status icons in the menu bar. With these options it's possible to squeeze more items on to menu bar, something that's especially useful on machines with a notch!

### DIFF
--- a/modules/system/defaults/NSGlobalDomain.nix
+++ b/modules/system/defaults/NSGlobalDomain.nix
@@ -232,12 +232,12 @@ in {
         Whether to enable moving window by holding anywhere on it like on Linux. The default is false.
       '';
     };
-    
+
     system.defaults.NSGlobalDomain.NSStatusItemSpacing = mkOption {
       type = types.nullOr types.int;
       default = null;
       example = 12;
-      description = lib.mdDoc ''
+      description = ''
         Sets the spacing between status icons in the menu bar. The default is null.
       '';
     };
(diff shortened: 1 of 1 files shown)


Enzime (changes requested):
Can you squash the two commits into one? It looks like they're exactly the same

modules/system/defaults/NSGlobalDomain.nix
```diff
@@ -225,6 +225,24 @@ in {
       '';
     };
 
+    system.defaults.NSGlobalDomain.NSStatusItemSpacing = mkOption {
+      type = types.nullOr types.int;
+      default = null;
+      example = 12;
+      description = lib.mdDoc ''
+        Sets the spacing between status icons in the menu bar. The default is null.
```
Enzime, line 233: The description should specify the macOS default like the rest of the options

https://github.com/LnL7/nix-darwin/blob/master/modules/system/defaults/NSGlobalDomain.nix#L11-L17

robharrop: Those are the MacOS defaults. Both of these options are unset by default.

robharrop:
> Can you squash the two commits into one? It looks like they're exactly the same

Done. Apologies for the delay, I was travelling all last week.

Samasaur1:
You should squash the commit removing `lib.mdDoc` into the previous commit

PR #888, opened 2024-02-27
security.sudo.extraConfig: fix default behavior

Samasaur1:
Due to where I put the `lib.mkIf`, the file `/etc/sudoers.d/10-nix-darwin-extra-config` was generated with no content by default on every nix-darwin user's system. This PR fixes the default behavior, while not affecting any users who have set an actual value for the option.


Enzime (approved):
LGTM

Issue #618, opened 2023-03-17
feature request: `security.wrappers`


ezracelli:
NixOS has [`security.wrappers`](https://search.nixos.org/options?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=security.wrappers), a mechanism to enable certain security features (SUID/GUID, capabilities, ownership/permissions) without directly modifying the Nix store.

It’d be great to have this functionality in `nix-darwin`, too &mdash; I’d like to set the SUID bit on `htop` (it’s the only way to see info from all users’ processes without `sudo`).

NixOS module implementation: https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/security/wrappers/default.nix

LnL7:
Once upon a time I spent a little time on this, next to setuid on darwin there's also a usecase for codesinging certain binaries. Perhaps https://github.com/LnL7/nix-darwin/commit/a008b2e8fdbda6220a73ee6255aa9008a5604b9d could still be useful as a reference.

PR #890, opened 2024-02-27
`security.wrappers`: init

Samasaur1:
This PR adds `security.wrappers` to nix-darwin, closing #618. It's heavily modeled after how NixOS implements the same option. On Darwin, it allows setting ownership, permissions, setuid/setgid, and (not yet implemented) codesigning. I've tested this PR, and it does meet the minimal use case of providing a setuid wrapper to a given program.

However, I have some concerns that should be addressed before it is merged:
- Is the generated wrapper "secure"? The NixOS implementation took a lot of steps to ensure security, some of which I could not replicate on Darwin. For example, multiple lists of unsecure environment variables — loaded from other derivations, not defined as constants — were combined into the wrapper. I've copied what the env vars were at the time of writing, but this isn't great for maintenance
- Supporting codesigning. I wasn't exactly clear on what the goal was here, only that it was mentioned in #618. I left a stub in this implementation which should hopefully allow implementing codesigning without a huge amount of reworking.
- Am I correct that setting capabilities is not a thing on Darwin?
- NixOS does not let you use setcap and setuid on the same wrapper. Although I didn't implement setcap at all, I do allow codesigning and setuid/setgid on the same wrapper. Is there a reason to disallow this?

These are just the concerns that I have — if others have concerns (about the concept as a whole or about my implementation), please leave a comment.


Samasaur1:
Is it possible that codesigning is desired so that entitlements can be applied to binaries?

Samasaur1:
Okay, I have removed codesigning from this PR, so it now only supports setuid/setgid. I've left the commented-out codesigning code in place for now, but I can remove it if desired.

I'm marking this PR as ready for review, and hopefully reviewers can look at `wrapper.c`/`wrapper.nix` in this PR versus [wrapper.c](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/security/wrappers/wrapper.c) and [wrapper.nix](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/security/wrappers/wrapper.nix) in nixpkgs, with a focus on the security of the wrappers.

Samasaur1:
Pinging @Enzime for the change to Karabiner

Enzime (commented):
The Karabiner change looks fine to me 👍

modules/security/wrappers/wrapper.c
```diff
@@ -0,0 +1,128 @@
...
+// of this wrapper would leave them intact.
+#define UNSECURE_ENVVARS_TUNABLES \
+    "MALLOC_CHECK_\0" \
+    "MALLOC_TOP_PAD_\0" \
+    "MALLOC_PERTURB_\0" \
+    "MALLOC_MMAP_THRESHOLD_\0" \
+    "MALLOC_TRIM_THRESHOLD_\0" \
+    "MALLOC_MMAP_MAX_\0" \
+    "MALLOC_ARENA_MAX\0" \
+    "MALLOC_ARENA_TEST\0"
```
emilazy, line 62: Not sure if any of these apply to macOS.

modules/security/wrappers/wrapper.c
```diff
@@ -0,0 +1,128 @@
...
+  "LD_DEBUG\0"								      \
+  "LD_DEBUG_OUTPUT\0"							      \
+  "LD_DYNAMIC_WEAK\0"							      \
+  "LD_HWCAP_MASK\0"							      \
+  "LD_LIBRARY_PATH\0"							      \
+  "LD_ORIGIN_PATH\0"							      \
+  "LD_PRELOAD\0"							      \
+  "LD_PROFILE\0"							      \
+  "LD_SHOW_AUXV\0"							      \
+  "LD_USE_LOAD_BIAS\0"							      \
```
emilazy, line 78: These are specific to the Linux dynamic loader interface – **meaning that their Darwin `DYLD_*` equivalents are currently not sanitized at all**. I fear that this means the current version of the wrapper is likely vulnerable.

I suspect that there are other Linux‐specific variables still in this list that I’m less confident about, and almost surely Darwin‐specific ones that we’re missing. That makes me really uneasy.

modules/security/wrappers/wrapper.c
```diff
@@ -0,0 +1,128 @@
...
+    // Since we expect the wrappers to be invoked from either $PATH or /run/wrappers/bin,
+    // there should be no reason to pass any particularly large values here, so we can
+    // be strict for strictness' sake.
+    ASSERT(strlen(argv[0]) < 512);
+
+    int debug = getenv(wrapper_debug) != NULL;
+
+    // Drop insecure environment variables explicitly
+    //
+    // glibc does this automatically in SUID binaries, but we'd like to cover this:
```
emilazy, line 103: It’s possible that Darwin’s dynamic loader also does this, in which case we could sleep a lot easier about all this. It would be good if someone could investigate that.

modules/security/wrappers/wrapper.c
```diff
@@ -0,0 +1,128 @@
...
+    //
+    //  a) before it gets to glibc
+    //  b) in binaries that are only granted ambient capabilities by the wrapper,
+    //     but don't run with an altered effective UID/GID, nor directly gain
+    //     capabilities themselves, and thus don't run in secure mode.
```
emilazy, line 108: Not relevant to us, except maybe for codesigning entitlements.

modules/security/wrappers/wrapper.c
```diff
@@ -0,0 +1,128 @@
...
+    //
+    // We're using musl, which doesn't drop environment variables in secure mode,
+    // and we'd also like glibc-specific variables to be covered.
```
emilazy, line 111: This (and other glibc‐related comments) would need to be updated.

emilazy (changes requested):
So, to be upfront about my bias here: I really dislike set{u,g}id. They’re mechanisms for running processes in highly‐privileged contexts but – critically – with an environment that is almost entirely under the control of an *untrusted* context. Privileged daemons interacting with untrusted clients are already a security minefield because of the confused deputy problem, but while those get started by a service manager in a predictable and controlled environment, set{u,g}id multiply the problem by allowing the invoking process to arrange pretty much any context for the hapless privileged code they desire. Naturally, it’s extremely difficult to code to this standard, so this leads to all kinds of awful vulnerabilities.

Most of the time, you are better off setting up a privileged daemon or a restricted `sudo` rule. `sudo` itself is a set{u,g}id program, of course, but it does a lot of the job of setting up a more predictable environment for its child processes itself, and gets more security attention than the majority of set{u,g}id code out there or especially our wrappers. (Despite this, it still has a lot of vulnerabilities! systemd’s `run0`, Ian Jackson’s `userv`, or even `ssh` to localhost are fundamentally better approaches here.)

In this case, I understand the examples people have brought up are:

* `ping`: This actually works fine without `setuid`. You only need privileges for raw pings, which aren’t that common to use. A simple restricted `sudo` rule is likely a better option here if ordinary users being able to send raw pings without privilege elevation is critical.

* `mtr`: Also a good candidate for `sudo`.

* `top`/`htop`: These ones are trickier. I’m not sure about `top`, but you probably don’t want to give people unrestricted access to run `htop` as root, since it can kill processes and the like. *However,* since it’s already coded to be able to be run as setuid, I believe you can instruct `sudo` to set the real/effective UIDs and GIDs such that you don’t get any additional privileges.

I would feel much more comfortable with reviewing and merging options that try to set up locked down `sudo` rules and wrappers that call out to `sudo` for cases like these.

For what it’s worth, I think that codesigning would actually be a great use case for this kind of interface, though; being able to set entitlements on executables would be tighter‐scoped than full set{u,g}id wrappers in the same way that Linux capabilities are, and unlock functionality we really have no other way of providing at the moment.

With all that said, support for set{u,g}id wrappers is already an unfortunate reality on NixOS, so if people are really clamouring for it then I’m not going to stop this going through. I just want to be clear that *here be dragons* and we should expect it to lead to vulnerabilities; even NixOS’s wrapper has had security issues. I wish I could say I feel fully qualified to sign off on an implementation, but I’d feel more comfortable getting a review from someone who is more intimately familiar with the fine details of Darwin’s dynamic loader. If we can’t get that then we might just want to put a big warning on the options before merging.

modules/security/wrappers/wrapper.c
Samasaur1: The dynamic loader does drop all the `DYLD_*` env vars that were listed in `man dyld`

modules/security/wrappers/wrapper.c
Samasaur1: I have updated the wrapper to unset the Darwin-specific env vars, although I have continued to unset these env vars anyway

modules/security/wrappers/wrapper.c
Samasaur1: Me neither, but I'm leaving them in to err on the side of safety

modules/security/wrappers/wrapper.c
Samasaur1: I've updated this comment to refer to binaries with entitlements

dwt:
How could an alternative look that used sudo rules to achieve similar goals?

For `mtr` could that be replacing the `mtr-packet` application with a shell wrapper that `exec`'s the original binary through sudo to elevate it's permissions?

That could be relatively straightforward to implement.

I am trying to gage what a simple and secure implementation could look like.

auscyber:
i am trying to add ccache to nix-darwin and nix-ccache is a great usage of wrappers

PR #892, opened 2024-02-28
`system.startup.chime`: init

### LINKED ISSUE #885: How to disable macos startup sound?
I've searched the documentation but haven't found a way to disable MacOS startup sound: `System Settings -> Sound -> Play sound on startup`

Samasaur1:
Adds an option `system.startup.chime`, which can either be `null` (don't change the current setting), `true` (enable the startup chime), or `false` (disable the startup chime).

Because the startup chime settings are stored in NVRAM, I also added an internal `system.nvram.variables` option, rather than implement NVRAM settings specifically for the startup chime.

Will close #885

### DIFF
--- a/modules/system/nvram.nix
+++ b/modules/system/nvram.nix
@@ -1,0 +1,40 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.system;
+
+  mkNvramVariables =
+    lib.attrsets.mapAttrsToList
+      (name: value: "nvram ${lib.escapeShellArg name}=${lib.escapeShellArg value}")
+      cfg.nvram.variables;
+in
+
+{
+  meta.maintainers = [
+    lib.maintainers.samasaur or "samasaur"
+  ];
+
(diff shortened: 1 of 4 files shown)


rmgpinto (approved):

modules/system/nvram.nix
```diff
@@ -0,0 +1,36 @@
...
+
+  options = {
+    system.nvram.variables = lib.mkOption {
+      type = with lib.types; attrsOf str;
+      default = {};
+      internal = true;
+      example = ''
+        # produces `nvram StartupMute=%01` (or `%00`)
+        system.nvram."StartupMute" = if chime then "%00" else "%01";
+      '';
```
Enzime, line 25: This should be in `description`, whereas `example` is for values like `example = { StartupMute = "%00"; }`

Samasaur1: whoops, my bad

Enzime (approved):
LGTM

PR #893, opened 2024-02-28
github-runners: adapt to NixOS module

veehaitch:
While #859 added basic support for configuring GitHub runners through nix-darwin, it did not yet support all of the options the NixOS module offers.

I am aware that this is a rather big overhaul. I think, however, that it's worth it:

- Copies the `options.nix` from the [NixOS module] with only minor adaptations. This should help to keep track of any changes to it.
- Respect the `workDir` config option. So far, the implementation didn't even read the value of the option.
- Allow configuring a custom user and group. If both are `null`, nix-darwin manages the `_github-runner` user shared among all instances. Take care of creating your own users if that's not what you want.
- Also creates the necessary directories for state, logs and the working directory (unless `workDir != null`). It uses the following locations: * state: `/var/lib/github-runners/${name}` * logs: `/var/log/github-runners/${name}` * work: The value of `workDir` or `/var/run/github-runners/${name}` if (`workDir == null`). We have to create the logs directory before starting the service since launchd expects that the `Standard{Error,Out}Path` exist. We do this by prepending to [`system.activationScripts.launchd.text`]. All directories belong to the configured `user` and `group`.
- Warn if a `tokenFile` points to the Nix store.

[NixOS module]: https://github.com/NixOS/nixpkgs/blob/3c30c56/nixos/modules/services/continuous-integration/github-runner/options.nix
[`system.activationScripts.launchd.text`]: https://github.com/LnL7/nix-darwin/blob/bbde06b/modules/system/launchd.nix#L99-L123


veehaitch:
Cc @domenkozar

domenkozar (approved):

PR #896, opened 2024-03-03
Support new nix 2.20 command syntax

marksisson:
This commit updates the nix.conf validation logic to accommodate different versions of Nix. It introduces a conditional assignment of the `showCommand` variable, which determines the appropriate command to use based on the Nix version. For versions at least "2.20pre", it uses "config show"; otherwise, it falls back to "show-config". This change ensures compatibility across various Nix releases.

Same as fix for NixOS - https://github.com/NixOS/nixpkgs/pull/271423


Enzime (approved):
LGTM

Thanks for the contribution

PR #897, opened 2024-03-03
fix writing values with containers

dfrankland:
Currently, the `-array` value for `writeValue` produces incorrect arguments for `defaults` like so

```ts
-array -string 'blah' -string 'blah'
```

Complex container values like `-array` and `-dict` have their own DSL which does not allow specifying all data types. Instead of using the DSL use plist fragments instead. 

The man page for `defaults` writes the following, below. This doesn't give much guidance on passing values with containers, but there's a feature to pass plist fragments for much more complex use-cases, mentioned as `<value_rep>` from `defaults --help`:
https://shadowfile.inode.link/blog/2018/06/advanced-defaults1-usage/#container-types

```
...
     write domain key 'value'
                  Writes value as the value for key in domain.  value must be a property list, and must be enclosed in single quotes.  For example:

                        defaults write com.companyname.appname "Default Color" '(255, 0, 0)'

                  sets the value for Default Color to an array containing the strings 255, 0, 0 (the red, green, and blue components). Note that the key is enclosed in quotation marks because it contains a space.

     write domain 'plist'
                  Overwrites the defaults information in domain with that given as plist.  plist must be a property list representation of a dictionary, and must be enclosed in single quotes.  For example:

                        defaults write com.companyname.appname '{ "Default Color" = (255, 0, 0);
                                                        "Default Font" = Helvetica; }';

                  erases any previous defaults for com.companyname.appname and writes the values for the two names into the defaults system.
...
     -array      Allows the user to specify an array as the value for the given preference key:

                       defaults write somedomain preferenceKey -array element1 element2 element3

                 The specified array overwrites the value of the key if the key was present at the time of the write. If the key was not present, it is created with the new value.

     -array-add  Allows the user to add new elements to the end of an array for a key which has an array as its value. Usage is the same as -array above. If the key was not present, it is created with the specified
                 array as its value.
```

Supersedes old PR: #790

### DIFF
--- a/tests/fixtures/system-defaults-write/activate.txt
+++ b/tests/fixtures/system-defaults-write/activate.txt
@@ -1,0 +1,10 @@
+defaults write /Library/Preferences/SystemConfiguration/com.apple.smb.server 'NetBIOSName' $'<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<string>IMAC-000000</string>
+</plist>'
+defaults write /Library/Preferences/SystemConfiguration/com.apple.smb.server 'ServerDescription' $'<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<string>Darwin\\\\U2019\'s iMac</string>
+</plist>'
--- a/tests/system-defaults-write.nix
+++ b/tests/system-defaults-write.nix
@@ -1,4 +1,4 @@
-{ config, pkgs, ... }:
+{ config, pkgs, lib, ... }:
 
 {
   system.defaults.NSGlobalDomain.AppleShowAllFiles = true;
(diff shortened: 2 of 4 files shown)


tests/system-defaults-write.nix
```diff
@@ -50,7 +50,7 @@
   system.defaults.screensaver.askForPassword = true;
   system.defaults.screensaver.askForPasswordDelay = 5;
   system.defaults.smb.NetBIOSName = "IMAC-000000";
-  system.defaults.smb.ServerDescription = ''Darwin\\\\U2019s iMac'';
+  system.defaults.smb.ServerDescription = ''Darwin\\\\U2019's iMac'';
```
dfrankland, line 53: testing that single quotes get escaped

tests/fixtures/system-defaults-write/activate.txt
```diff
@@ -0,0 +1,10 @@
+defaults write /Library/Preferences/SystemConfiguration/com.apple.smb.server 'NetBIOSName' $'<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<string>IMAC-000000</string>
+</plist>'
+defaults write /Library/Preferences/SystemConfiguration/com.apple.smb.server 'ServerDescription' $'<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<string>Darwin\\\\U2019\'s iMac</string>
```
dfrankland, line 9: this single quote got escaped

Enzime (approved):
LGTM

PR #619, opened 2023-03-20
Add support for persistent-apps in dock

dfrankland:
Applications added to the dock point to an old derivation when a new derivation of an application is added to the Nix store. This allows configuring the persistent applications to display system apps or apps in the Nix store.

Updates require re-login or restarting the Dock process e.g. `killall Dock`.

Dependent on:
#790

Related / supersedes:
https://github.com/LnL7/nix-darwin/issues/590
https://github.com/LnL7/nix-darwin/pull/616

PR #898, opened 2024-03-03
Add support for persistent-apps in dock

dfrankland:
Applications added to the dock point to an old derivation when a new derivation of an application is added to the Nix store. This allows configuring the persistent applications to display system apps or apps in the Nix store.

Updates require re-login or restarting the Dock process e.g. `killall Dock`.

Dependent on: #897
Supersedes the old PR: #619

### DIFF
--- a/modules/system/defaults/dock.nix
+++ b/modules/system/defaults/dock.nix
@@ -123,6 +123,19 @@ in {
       '';
     };
 
+    system.defaults.dock.persistent-apps = mkOption {
+      type = types.nullOr (types.listOf (types.either types.path types.str));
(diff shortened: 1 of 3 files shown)


Enzime:
Can you rebase this?

Also, in future, when you create a PR you can target your own branch, so it'll only show the changes that are unique to this branch and then when that branch gets merged, GitHub will automatically update the target branch to `master`

dfrankland:
Rebased. Will do in the future—I wasn't sure which way was preferred 👍

Enzime (approved):
LGTM

PR #909, opened 2024-03-21
Add support for zsh-fast-syntax-highlighting

motheki:
Hello,

I would like to add support for zsh-fast-syntax-highlighting which has marginally better performance than zsh-syntax-highlighting, to the zsh configuration module in nix-darwin. 

The repository can be found [here](https://github.com/zdharma-continuum/fast-syntax-highlighting)

There is an existing nix package for zsh-fast-syntax-highlighting [here](https://github.com/zdharma-continuum/fast-syntax-highlighting)

### DIFF
--- a/modules/programs/zsh/default.nix
+++ b/modules/programs/zsh/default.nix
@@ -107,6 +107,12 @@ in
       default = false;
       description = "Enable zsh-syntax-highlighting.";
     };
+
+    programs.zsh.enableFastSyntaxHighlighting = mkOption {
+      type = types.bool;
+      default = false;
+      description = lib.mdDoc "Enable zsh-fast-syntax-highlighting.";
+    };
   };
 
   config = mkIf cfg.enable {
@@ -115,7 +121,8 @@ in
       [ # Include zsh package
         pkgs.zsh
       ] ++ optional cfg.enableCompletion pkgs.nix-zsh-completions
-        ++ optional cfg.enableSyntaxHighlighting pkgs.zsh-syntax-highlighting;
+        ++ optional cfg.enableSyntaxHighlighting pkgs.zsh-syntax-highlighting
+        ++ optional cfg.enableFastSyntaxHighlighting pkgs.zsh-fast-syntax-highlighting;
 
     environment.pathsToLink = [ "/share/zsh" ];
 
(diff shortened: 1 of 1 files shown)


Samasaur1:
Would there be any problems if both zsh-syntax-highlighting and zsh-fast-syntax-highlighting are enabled at the same time?

Trevor-Opiyo:
@Samasaur1 Not from my testing. I have had both enabled at multiple points in time and zsh-fast-syntax-highlighting seems to take precedence.

Samasaur1:
I think there should still probably be a warning when both options are enabled, since zsh-syntax-highlighting won't take effect if both are enabled.

Trevor-Opiyo:
@Samasaur1 Thank you for your patience. I took some time to add a warning that Zsh-Syntax-highlighting and zsh-fast-synatx-highlighting are mutually exclusive. Let me know if there is anything else that needs to be addressed.

modules/programs/zsh/default.nix
```diff
@@ -107,15 +107,25 @@ in
...
+      default = false;
+      description = lib.mdDoc "Enable zsh-fast-syntax-highlighting.";
+    };
   };
 
   config = mkIf cfg.enable {
 
+    warnings = mkIf (cfg.enableFastSyntaxHighlighting && cfg.enableSyntaxHighlighting) [
+      "zsh-fast-syntax-highlighting and zsh-syntax-highlighting are mutually exclusive. Disable one of them."
+    ];
```
Enzime, line 122:
```suggestion
    assertions = [
      {
        assertion = !(cfg.enableSyntaxHighlighting && cfg.enableFastSyntaxHighlighting);
        message = "zsh-syntax-highlighting and zsh-fast-syntax-highlighting are mutually exclusive, please disable one of them.";
      }
    ];
```

modules/programs/zsh/default.nix
```diff
@@ -107,15 +107,25 @@ in
       default = false;
       description = "Enable zsh-syntax-highlighting.";
     };
+
+    programs.zsh.enableFastSyntaxHighlighting = mkOption {
+      type = types.bool;
+      default = false;
+      description = lib.mdDoc "Enable zsh-fast-syntax-highlighting.";
+    };
```
Enzime, line 115:
```suggestion
    programs.zsh.enableFastSyntaxHighlighting = mkEnableOption "zsh-fast-syntax-highlighting";
```

Enzime (changes requested):

modules/programs/zsh/default.nix
motheki: Thanks for the changes @Enzime I had forgotten about this commit.

motheki: @Samasaur1 @Enzime has shown that Fast Syntax Highlighting is mutually exclusive with Zsh-syntax-highlighting and has provided alterations.

Enzime (approved):
LGTM

Thanks for the contribution

Issue #693, opened 2023-06-18
After building with flakes, git is always failing during darwin-rebuild switch


ryanbooker:
One of my machines has recently started displaying the following error when I try to switch after `nix build` using flakes. It was previously working fine, and two other machines using the same flake set up are working fine.
```
./result/sw/bin/darwin-rebuild switch --show-trace --flake .
error:
       … while fetching the input 'git+file:///path/to/the/flake/directory'

       error: program 'git' failed due to signal 9 (Killed: 9)
```
I ran `nix doctor` on all machines, and the failing machine's output is slightly different. The working machines all output the following line about being trusted, the failing machine doesn't output this line:
```
[INFO] You are trusted by store uri: daemon
```
Is this potentially the issue? How do I get the user to be trusted? All machines have exactly the same `/etc/nix/nix.conf`.

I'm not sure what I'm looking for, so if anyone has any tips on what to look for or how to get more useful information, that would be greatly appreciated. Thanks.

bestlem:
Which git is being found on the machine and what architecture is it and what architecture is the machine?

ryanbooker:
The broken machine is `arm64`, with an `arm64` git binary at `/etc/profiles/per-user/myuser/bin/git. One of the working machines is `arm64` and one is `x86_64`.

ryanbooker:
For what it's worth this appears to have been the lack of `[INFO] You are trusted by store uri: daemon`, because it's now working again and the user is trusted… though I have no idea how. Perhaps it fixed itself after a reboot.

ryanbooker:
Sorry, it has started happening again, `nix doctor` still reports the user as trusted, so I guess it wasn't that. Not sure what I'm looking for now.

ryanbooker:
Unfortunate update, this issue has now spread to a second machine…

emilazy:
This is very strange, but I can't imagine it's not an upstream Nix bug. Perhaps the Nix version changes on activation (Nixpkgs is currently using 2.15 rather than the latest stable Nix release of 2.16)? Do you have the sandbox on (though I don't think that should be relevant for fetching flake inputs)?

ryanbooker:
I deleted `~/.cache/nix` and the local `result` link in the flake's folder, then did a `nix-collect-garbage` and tried again… 

Everything worked…

Quite odd. When I'm back home I'll try that on the other computer and see if it fixes things.

bestlem:
[...]

PR #915, opened 2024-03-25
Add `nix.optimise` module

malob:
Add `nix.optimise` based of NixOS module. Implementation in `nix-darwin` based of `nix.gc`.

Would appreciate a sanity check on changes to `module/system/checks.nix` in particular since I'm least certain about that.

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -71,6 +71,7 @@
   ./services/netbird.nix
   ./services/nix-daemon.nix
   ./services/nix-gc
+  ./services/nix-optimise
   ./services/ofborg
   ./services/postgresql
   ./services/privoxy
--- a/modules/services/nix-optimise/default.nix
+++ b/modules/services/nix-optimise/default.nix
@@ -1,0 +1,73 @@
+# Based off:
+# https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/misc/nix-optimise.nix
+# When making changes please try to keep it in sync.
+{ config, lib, ... }:
+
+
+let
+  inherit (lib)
+    mdDoc
+    mkIf
+    mkOption
+    mkRemovedOptionModule
+    optionalString
+    types
+    ;
+
+  cfg = config.nix.optimise;
+in
+
+{
+  imports = [
+    (mkRemovedOptionModule [ "nix" "optimise" "dates" ] "Use `nix.optimise.interval` instead.")
+  ];
+
+  ###### interface
+
+  options = {
+
+    nix.optimise = {
+
+      automatic = mkOption {
+        type = types.bool;
+        default = false;
+        description = mdDoc "Automatically run the nix store optimiser at a specific time.";
+      };
+
+      # Not in NixOS module
+      user = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = mdDoc "User that runs the store optimisation.";
+      };
+
+      interval = mkOption {
+        type = types.attrs;
+        default = { Hour = 3; Minute = 15; };
+        description = mdDoc "The time interval at which the optimiser will run.";
+      };
+
+    };
+
+  };
+
+
+  ###### implementation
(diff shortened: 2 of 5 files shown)


malob:
Supersedes #588

Samasaur1:
The service should probably use `/bin/wait4path`

malob:
@Samasaur1, thanks! I can do that.

As an aside, should [`nix.gc`](https://github.com/LnL7/nix-darwin/blob/bcc8afd06e237df060c85bad6af7128e05fd61a3/modules/services/nix-gc/default.nix#L63) use it as well?

malob:
Alrighty, I think this is ready to go.

Enzime (approved):
LGTM

emilazy:
As much as I appreciate the work to improve NixOS compatibility here, I think this module is a huge footgun and that we should remove it, or at least add very loud scary warnings when it’s turned on. Store optimization is basically completely broken on macOS and causes many bizarre and hard‐to‐debug issues; cf. https://github.com/NixOS/nix/issues/7273 https://github.com/DeterminateSystems/nix-installer/pull/542 https://github.com/LnL7/nix-darwin/issues/693. I have run into explicit breakage every time I have tried to use it, but what’s worse is that it can fail in ways that could cause silent corruption (as far as I can tell it’s only luck that made that last issue result in crashes rather than buggy behaviour). It's at the top of my list when someone has strange behaviour that seems like it violates the Nix model, so I don’t think that anyone on macOS should have this option turned on in its current upstream state.

emilazy:
(Just to add: I know some of the issues are more prominent with `auto-optimise-store` vs. a batch process, presumably due to a lower chance of race conditions, but I’ve run into serious problems along these lines even with manual `nix-store --optimise` runs and have consequently given up on the feature entirely on macOS.)

malob:
Interesting. I implemented this because of the issues with `auto-optimise-store`, and haven't myself experienced, or heard of, issues with manually running store optimization. How prominent is the latter? Are these issues documented anywhere online?

emilazy:
I’ll give an immediate mea culpa here: I have much clearer memories and confidence in automatic optimization being broken on macOS than manual, and it’s entirely possible that I’m misremembering running into unambiguous issues with manual optimisation. So, sorry for jumping the gun on this one – I have an instinctive twitch reaction to the topic of store optimization at this point, and it got ahead of me in this case.

To be more explicit about my levels of confidence: I know for certain that automatic store optimization on macOS is horribly broken and can lead to scary corruption, and that even manual store optimization can break things on some platforms (e.g., Linux with SELinux). I suspect that manual optimization on macOS can probably cause similar issues to automatic store optimization if it happens to race with a build. I vaguely recall that I did manual store optimization on macOS and then gave up due to issues, but it’s possible I’m confusing it with the SELinux case. Unfortunately I don’t have great documentation about all these issues (even “automatic store optimization is broken on macOS” is basically folklore, sadly – there should probably be upstream documentation about it, or even a warning).

Seems like the best way forward here is to see if this causes any issues when people use it; I might consider turning it on myself to be a guinea pig. The disk savings are nice; I had just resigned myself to missing out on them because of all the issues I’ve had in the past.

malob:
No worries, I’ve been driven half crazy in the past trying to debug issues on macOS arising from enabling `auto-optimise-store`, so I definitely get your concern 🙂.

FYI, as far as I’m aware, the best “documentation” of issues with that on macOS is this issue: https://github.com/NixOS/nix/issues/7273

Your plan here sounds reasonable to me. I’ve got `nix.optimise` enabled, so I’ll make sure to keep an eye out for any issues myself.

emilazy:
Yeah, I’m mostly just twitchy because it’s such a nondeterministic thing, and I worry that making it a scheduled batch process only works more reliably because it reduces the likelihood of it racing with a build. If only use we could use something like Antithesis to try and get a deterministic reproducer…

PR #917, opened 2024-03-26
Use valid maintainer in meta example

HyunggyuJang:
Unless will fail on nix eval, e.g.,
```sh
nix eval .#darwinConfigurations.<your-host-name>.options
```
will fail.


Enzime (approved):
LGTM

Thanks for your contribution

PR #920, opened 2024-03-30
Expose `extendModules` attr

hiroqn:
expose extendModules

implementation reference is https://github.com/nix-community/home-manager/blob/3142bdcc470e1e291e1fbe942fd69e06bd00c5df/modules/default.nix#L43

ref: https://github.com/LnL7/nix-darwin/pull/777
ref: https://github.com/LnL7/nix-darwin/issues/609

### DIFF
--- a/eval-config.nix
+++ b/eval-config.nix
@@ -82,12 +82,10 @@ let
     modules = modules ++ [ argsModule ] ++ baseModules;
     specialArgs = { modulesPath = builtins.toString ./modules; } // specialArgs;
   });
-in
-
-{
-  inherit (eval._module.args) pkgs;
-  inherit (eval) options config;
-  inherit (eval) _module;
-
-  system = eval.config.system.build.toplevel;
-}
+  
+  withExtraAttrs = configuration: configuration // {
+    inherit (configuration._module.args) pkgs;
+    system = configuration.config.system.build.toplevel;
+    extendModules = args: withExtraAttrs (configuration.extendModules args);
+  };
+in withExtraAttrs eval


remi-gelinas:
@LnL7 Any chance this could get a pair of eyes? I need `extendModules` to be passed through for a composite use case I have, and I figured I'd see if anyone tried before raising a PR myself.

eval-config.nix
```diff
@@ -82,12 +82,12 @@ let
...
-
-{
-  inherit (eval._module.args) pkgs;
-  inherit (eval) options config;
-  inherit (eval) _module;
-
-  system = eval.config.system.build.toplevel;
-}
+  
+  withExtraAttrs = module:
```
Enzime, line 86: Can you rename `module` to `configuration` and merge this set with `configuration` to match NixOS's implementation:

https://github.com/NixOS/nixpkgs/blob/b1f2198021490b51fb92b8b09db97b9ba2a7b4ce/nixos/lib/eval-config.nix

Enzime (changes requested):

eval-config.nix
hiroqn: Thanks for your feedback @Enzime !
I fixed it to match NixOS's implementation. But another attributes(`_type`, `class`, `type`) is exposed, is this ok? or should I remove them?
NixOS configuration (`nixpkgs.lib.nixosSystem`) already expose these attributes.

Enzime (approved):
LGTM

Thanks for the contribution

PR #926, opened 2024-04-06
Add offline and substituters flags

### LINKED ISSUE #608: Allow `--offline` flag
`nix build --offline` exists and since it is not an `--option` there's no way to have `darwin-rebuild` pass it

Rubikoid:
Add missing `--offline` and `--substituters` options to `darwin-rebuild` which is impossible to use now.

Closes #608

### DIFF
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -117,7 +117,17 @@ while [ $# -gt 0 ]; do
       shift 1
       ;;
     --offline)
+      extraMetadataFlags+=("$i")
       extraBuildFlags+=("$i")
+      ;;
+    --substituters)
+      if [ -z "$1" ]; then
+        echo "$0: '$i' requires an argument"
+        exit 1
+      fi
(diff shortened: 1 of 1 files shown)


Rubikoid:
Hi, @Enzime, can you have a quick look at this PR?

pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -116,6 +117,19 @@ while [ $# -gt 0 ]; do
       fi
       shift 1
       ;;
+    --offline)
+      extraMetadataFlags+=("$i")
+      extraBuildFlags+=("$i")
+      ;;
```
Enzime, line 123:
```suggestion
```

Can you merge this with the option on line 47?

Enzime (changes requested):

pkgs/nix-tools/darwin-rebuild.sh
Rubikoid: Done!

Enzime (approved):
LGTM

Thanks for the contribution

PR #939, opened 2024-04-22
Add networking.hosts and .hostFiles from nixos 

ibizaman:
This code was taken nearly verbatim from
https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/config/networking.nix

The few changes were related to making the default `/etc/hosts` match the Apple's stock one. This implied no `127.0.0.2` and forcing the IPV6 `::1` entry.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,12 +47,14 @@ jobs:
         sudo cp modules/examples/simple.nix /etc/nix-darwin/configuration.nix
 
         nixConfHash=$(shasum -a 256 /etc/nix/nix.conf | cut -d ' ' -f 1)
+        etcHostsHash=$(shasum -a 256 /etc/hosts | cut -d ' ' -f 1)
         sudo /usr/bin/sed -i.bak \
           "s/# programs.fish.enable = true;/ \
             imports = [ \
               ({ options, ... }: { \
                 nix.settings.access-tokens = [ \"github.com=\${{ secrets.GITHUB_TOKEN }}\" ]; \
                 environment.etc.\"nix\/nix.conf\".knownSha256Hashes = [ \"$nixConfHash\" ]; \
+                environment.etc.hosts.knownSha256Hashes = [ \"$etcHostsHash\" ]; \
                 nix.nixPath = \
                   [ { darwin = \"${PWD////\/}\"; } ] \
                   ++ options.nix.nixPath.default; \
@@ -102,8 +104,9 @@ jobs:
         pushd /etc/nix-darwin
           sudo nix flake init -t $darwin
           nixConfHash=$(shasum -a 256 /etc/nix/nix.conf | cut -d ' ' -f 1)
+          etcHostsHash=$(shasum -a 256 /etc/hosts | cut -d ' ' -f 1)
           sudo /usr/bin/sed -i.bak \
-            "s/# programs.fish.enable = true;/nix.settings.access-tokens = [ \"github.com=\${{ secrets.GITHUB_TOKEN }}\" ]; environment.etc.\"nix\/nix.conf\".knownSha256Hashes = [ \"$nixConfHash\" ];/" \
+            "s/# programs.fish.enable = true;/nix.settings.access-tokens = [ \"github.com=\${{ secrets.GITHUB_TOKEN }}\" ]; environment.etc.\"nix\/nix.conf\".knownSha256Hashes = [ \"$nixConfHash\" ]; environment.etc.hosts.knownSha256Hashes = [ \"$etcHostsHash\" ];/" \
             flake.nix
           sudo /usr/bin/sed -i.bak \
             's/darwinConfigurations."simple"/darwinConfigurations."'$(scutil --get LocalHostName)'"/g' \
--- a/modules/networking/default.nix
+++ b/modules/networking/default.nix
@@ -1,4 +1,4 @@
-{ config, lib, ... }:
+{ config, lib, pkgs, ... }:
 
 with lib;
 
@@ -22,6 +22,8 @@ let
       esac
     '') cfg.knownNetworkServices}
   '';
+
+  localhostMultiple = any (elem "localhost") (attrValues (removeAttrs cfg.hosts [ "127.0.0.1" "::1" ]));
 in
 
 {
@@ -105,9 +107,50 @@ in
         Battery powered devices may require being connected to power.
       '';
     };
+
+    networking.hosts = lib.mkOption {
+      type = types.attrsOf (types.listOf types.str);
+      example = literalExpression ''
+        {
+          "127.0.0.1" = [ "foo.bar.baz" ];
+          "192.168.0.2" = [ "fileserver.local" "nameserver.local" ];
+        };
+      '';
+      description = ''
+        Locally defined maps of hostnames to IP addresses.
+      '';
+      default = {};
+    };
+
+    networking.hostFiles = lib.mkOption {
+      type = types.listOf types.path;
+      defaultText = literalMD "Hosts from {option}`networking.hosts` and {option}`networking.extraHosts`";
+      example = literalExpression ''[ "''${pkgs.my-blocklist-package}/share/my-blocklist/hosts" ]'';
+      description = ''
+        Files that should be concatenated together to form {file}`/etc/hosts`.
+      '';
+    };
+
+    networking.extraHosts = lib.mkOption {
+      type = types.lines;
+      default = "";
+      example = "192.168.0.1 lanlocalhost";
+      description = ''
+        Additional verbatim entries to be appended to {file}`/etc/hosts`.
+        For adding hosts from derivation results, use {option}`networking.hostFiles` instead.
+      '';
+    };
   };
 
   config = {
+    assertions = [{
+      assertion = !localhostMultiple;
+      message = ''
+        `networking.hosts` maps "localhost" to something other than "127.0.0.1"
+        or "::1". This will break some applications. Please use
+        `networking.extraHosts` if you really want to add such a mapping.
+      '';
+    }];
 
     warnings = [
       (mkIf (cfg.knownNetworkServices == [] && cfg.dns != []) "networking.knownNetworkServices is empty, dns servers will not be configured.")
@@ -134,5 +177,36 @@ in
       ''}
     '';
 
+    networking.hostFiles = let
+      # Note: localhostHosts has to appear first in /etc/hosts so that 127.0.0.1
+      # resolves back to "localhost" (as some applications assume) instead of
+      # the FQDN!
+      localhostHosts = pkgs.writeText "localhost-hosts" ''
+        ##
+        # Host Database
+        #
+        # localhost is used to configure the loopback interface
+        # when the system is booting.  Do not change this entry.
+        ##
+        127.0.0.1	localhost
+        255.255.255.255	broadcasthost
+        ::1             localhost
+      '';
+      stringHosts =
+        let
+          oneToString = set: ip: ip + " " + concatStringsSep " " set.${ip} + "\n";
+          allToString = set: concatMapStrings (oneToString set) (attrNames set);
+        in pkgs.writeText "string-hosts" (allToString (filterAttrs (_: v: v != []) cfg.hosts));
+      extraHosts = pkgs.writeText "extra-hosts" cfg.extraHosts;
+    in mkBefore [ localhostHosts stringHosts extraHosts ];
+
+    environment.etc.hosts = {
+      knownSha256Hashes = [
+        # Comes from MacOS: Darwin 24.1.0 Darwin Kernel Version 24.1.0.
+        "c7dd0e2ed261ce76d76f852596c5b54026b9a894fa481381ffd399b556c0e2da"
+      ];
+
+      source = pkgs.concatText "hosts" cfg.hostFiles;
+    };
   };
 }
--- a/tests/networking-hosts.nix
+++ b/tests/networking-hosts.nix
@@ -1,0 +1,20 @@
+{ config, pkgs, ... }:
+
+{
+  networking.hosts = {
+    "127.0.0.1" = [ "my.super.host" ];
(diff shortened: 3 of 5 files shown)


willemml:
Possible duplicate of #807?

ibizaman:
Sorry @willemml I missed your PR. I didn’t intend to duplicate work.

secana:
This PR seems fine but is blocked as a review is required. How can we help to get it merged?

niklasravnsborg:
Oh this would be awesome :)

tests/networking-hosts.nix
```diff
@@ -0,0 +1,16 @@
...
+  networking.hosts = {
+    "127.0.0.1" = [ "my.super.host" ];
+    "10.0.0.1" = [ "my.super.host" "my.other.host" ];
+  };
+
+  test = ''
+    echo checking /etc/hosts file >&2
+
+    file=${config.out}/etc/hosts
+    grep '127.0.0.1' $file
```
Enzime, line 13: Probably also worth testing that `localhost` is mapped to `127.0.0.1` and `::1` as they're the defaults

Enzime (changes requested):
Can you rebase this PR?

ibizaman:
@Enzime rebasing is done. I tried to do the hash thing but I think I got the hash wrong.

Btw I did `cp /etc/hosts hosts`, edited the file to remove the modifications I made, then `sha256sum hosts` and that's the hash I used. I suppose the hash is wrong though? Could someone provide me with the hash of a stock `/etc/hosts` file?

emilazy:
It looks right to me:

```
emily@yuyuko ~> cat /etc/hosts
##
# Host Database
#
# localhost is used to configure the loopback interface
# when the system is booting.  Do not change this entry.
##
127.0.0.1	localhost
255.255.255.255	broadcasthost
::1             localhost
emily@yuyuko ~> shasum -a 256 /etc/hosts
c7dd0e2ed261ce76d76f852596c5b54026b9a894fa481381ffd399b556c0e2da  /etc/hosts
```

emilazy:
However the file in `doc/` seems to be missing the trailing newline.

ibizaman:
@emilazy good catch! Indeed:

Without the newline:
```bash
$ sha256sum doc/known-files/c7dd0e2ed261ce76d76f852596c5b54026b9a894fa481381ffd399b556c0e2da
1b6fb3e08d41ae31b6bfe1f66fecf3ef53e302047993f6235570bc9391f291bb  doc/known-files/c7dd0e2ed261ce76d76f852596c5b54026b9a894fa481381ffd399b556c0e2da
```

With:
```bash
$ sha256sum doc/known-files/c7dd0e2ed261ce76d76f852596c5b54026b9a894fa481381ffd399b556c0e2da
c7dd0e2ed261ce76d76f852596c5b54026b9a894fa481381ffd399b556c0e2da  doc/known-files/c7dd0e2ed261ce76d76f852596c5b54026b9a894fa481381ffd399b556c0e2da
```

Enzime:
Looks like the tests are broken now

ibizaman:
@Enzime indeed 😞 I have two questions related to the two failures I see.

How can I fix the `/etc/hosts` file exists issue? Does the error know comes from me including a wrong /etc/hosts file? How can I print the /etc/hosts file from CI?

I looked and looked at the test output and can’t see the error printed out. How can I run the tests locally if I use flakes?

Enzime:
The ability to run the tests through a flake interface will be added in #1140, until that's merged you can run your test locally with `nix-build release.nix -A tests.networking-hosts`

I've accessed the GitHub runner through tmate (I suspect only maintainers have access but I'm not sure, it's the last step in the `install-against-stable` and `install-against-unstable` jobs that runs if the job failed) and extracted the `/etc/hosts` file:

```
$ cat /etc/hosts
##
# Host Database
#
# localhost is used to configure the loopback interface
# when the system is booting.  Do not change this entry.
##
127.0.0.1	localhost
255.255.255.255	broadcasthost
::1             localhost
192.168.64.25 Mac-1730942908330.local Mac-1730942908330
192.168.64.25 Mac-1730942908330.local Mac-1730942908330
$ shasum -a 256 /etc/hosts
31ce667128143b67ac1e7c3b64fda0891d4ff011259c48e6c9bdbd704bc29118  /etc/hosts
```

I suspect that every GitHub Mac will have a different `/etc/hosts` so you'll most likely need to update these lines to generate a hash of `/etc/hosts` and set `environment.etc.hosts.knownSha256Sums = [ ... ]` like is already done with `nix.conf`: https://github.com/LnL7/nix-darwin/blob/98e7dba87238e4fa4eac609dc44f07dab40661c4/.github/workflows/test.yml#L143-L146

Enzime:
Also if you could rebase this PR that would be great, let us know if you're still interested in working on this PR

emilazy:
Probably simplest to just move the config files out of the way rather than hashing them.

Enzime:
In the case of `nix.conf` I chose to hash it because of `access-tokens` and the changes get reflected immediately when I last tested but that should be fine for `/etc/hosts`

Teebor-Choka:
I'd love this functionality, any ETA on finalizing this PR and merging it, please?

Enzime:
There hasn't been @ibizaman in 3 weeks so if you're interested in picking up the work, feel free to take the changes in this PR and rebase them on top of `master` as well as adding the changes mentioned in https://github.com/LnL7/nix-darwin/pull/939#issuecomment-2461162546

Teebor-Choka:
> There hasn't been @ibizaman in 3 weeks so if you're interested in picking up the work, feel free to take the changes in this PR and rebase them on top of `master` as well as adding the changes mentioned in [#939 (comment)](https://github.com/LnL7/nix-darwin/pull/939#issuecomment-2461162546)

I'm afraid I'm not at the skill level of debugging and developing nix components at this point, just a happy user.

ibizaman:
Sorry about the hiatus. I’ll pick this up this week. But of course I don’t mind if someone beats me to it 😛

ibizaman:
Well it seems I did a bad job rebasing 😅  more rebasing incoming.

.github/workflows/test.yml
```diff
@@ -51,8 +51,9 @@ jobs:
         cp modules/examples/simple.nix ~/.config/nix-darwin/configuration.nix
 
         nixConfHash=$(shasum -a 256 /etc/nix/nix.conf | cut -d ' ' -f 1)
+        etcHostsHash=$(shasum -a 256 /etc/hosts | cut -d ' ' -f 1)
         /usr/bin/sed -i.bak \
-          "s/# programs.fish.enable = true;/nix.settings.access-tokens = [ \"github.com=\${{ secrets.GITHUB_TOKEN }}\" ]; environment.etc.\"nix\/nix.conf\".knownSha256Hashes = [ \"$nixConfHash\" ];/" \
+          "s/# programs.fish.enable = true;/nix.settings.access-tokens = [ \"github.com=\${{ secrets.GITHUB_TOKEN }}\" ]; environment.etc.\"nix\/nix.conf\".knownSha256Hashes = [ \"$nixConfHash\" ]; environment.etc.\"etc\/hosts\".knownSha256Hashes = [ \"etcHostsHash\" ];/" \
```
Enzime, line 56:
```suggestion
          "s/# programs.fish.enable = true;/nix.settings.access-tokens = [ \"github.com=\${{ secrets.GITHUB_TOKEN }}\" ]; environment.etc.\"nix\/nix.conf\".knownSha256Hashes = [ \"$nixConfHash\" ]; environment.etc.hosts.knownSha256Hashes = [ \"etcHostsHash\" ];/" \
```

Enzime (changes requested):

.github/workflows/test.yml
ibizaman: Oh of course it's already under /etc 👍

ibizaman:
I'm not sure where the issue comes from. I tried to reproduce locally but I don't have more log output. Any idea?

tests/networking-hosts.nix
```diff
@@ -0,0 +1,27 @@
+{ config, pkgs, ... }:
+
+{
+  networking.hosts = {
+    "127.0.0.1" = [ "my.super.host" ];
+    "10.0.0.1" = [ "my.super.host" "my.other.host" ];
+  };
+
+  test = ''
+    echo checking /etc/hosts file >&2
```
Enzime, line 10:
```suggestion
    set -v

    echo checking /etc/hosts file >&2
```

For debugging

ibizaman: TIL -v

tests/networking-hosts.nix
```diff
@@ -0,0 +1,35 @@
...
+
+    file=${config.out}/etc/hosts
+
+    if [[ ! $(grep '127.0.0.1' $file | head -n1) =~ localhost$ ]]; then
+      cat $file
+      echo "'$(grep '127.0.0.1' $file | head -n1)'"
+      exit 1
+    fi
+    if [[ ! $(grep '127.0.0.1' $file | tail -n1) =~ my.super.host$ ]]; then
```
Enzime, line 19:
```suggestion
    grep '127.0.0.1' $file | tail -n1 | grep 'my.super.host$'
```

`set -e` should be enabled so if you can construct these as a single chain of commands then if statements shouldn't necessary

Enzime (changes requested):

tests/networking-hosts.nix
ibizaman: I verified and indeed if I meddle with the second grep, like replacing `host` with `hst`, the test fails.

tests/networking-hosts.nix
```diff
@@ -0,0 +1,27 @@
...
+  };
+
+  test = ''
+    echo checking /etc/hosts file >&2
+
+    file=${config.out}/etc/hosts
+
+    if ! grep '127.0.0.1' $file | head -n1 | grep localhost$; then
+      exit 1
+    fi
```
Enzime, line 16:
```suggestion
    grep '127.0.0.1' $file | head -n1 | grep localhost$
```

The `if !; then exit` isn't necessary

modules/networking/default.nix
```diff
@@ -135,5 +178,35 @@ in
...
+        let
+          oneToString = set: ip: ip + " " + concatStringsSep " " set.${ip} + "\n";
+          allToString = set: concatMapStrings (oneToString set) (attrNames set);
+        in pkgs.writeText "string-hosts" (allToString (filterAttrs (_: v: v != []) cfg.hosts));
+      extraHosts = pkgs.writeText "extra-hosts" cfg.extraHosts;
+    in mkBefore [ localhostHosts stringHosts extraHosts ];
+
+    environment.etc.hosts = {
+      knownSha256Hashes = [
+        "c7dd0e2ed261ce76d76f852596c5b54026b9a894fa481381ffd399b556c0e2da"
```
Enzime, line 206: Can you add a comment explaining where you got this file from, like the macOS version you're running

.github/workflows/test.yml
```diff
@@ -147,8 +149,9 @@ jobs:
         pushd ~/.config/nix-darwin
           nix flake init -t $darwin
           nixConfHash=$(shasum -a 256 /etc/nix/nix.conf | cut -d ' ' -f 1)
-          /usr/bin/sed -i.bak \
-            "s/# programs.fish.enable = true;/nix.settings.access-tokens = [ \"github.com=\${{ secrets.GITHUB_TOKEN }}\" ]; environment.etc.\"nix\/nix.conf\".knownSha256Hashes = [ \"$nixConfHash\" ];/" \
+        etcHostsHash=$(shasum -a 256 /etc/hosts | cut -d ' ' -f 1)
+        /usr/bin/sed -i.bak \
+          "s/# programs.fish.enable = true;/nix.settings.access-tokens = [ \"github.com=\${{ secrets.GITHUB_TOKEN }}\" ]; environment.etc.\"nix\/nix.conf\".knownSha256Hashes = [ \"$nixConfHash\" ]; environment.etc.hosts.knownSha256Hashes = [ \"$etcHostsHash\" ];/" \
```
Enzime, line 154: Looks like these commands should be indented more

Enzime (changes requested):
Can you squash all the commits into one commit with a commit message like:

    networking: add `hosts` option for managing `/etc/hosts`

Samasaur1:
The comment in the default `/etc/hosts` file
```
# localhost is used to configure the loopback interface
# when the system is booting.  Do not change this entry.
```
does make me a little apprehensive about symlinking this into place instead of copying it.

Teebor-Choka:
Almost there!

tests/networking-hosts.nix
```diff
@@ -0,0 +1,20 @@
...
+    "10.0.0.1" = [ "my.super.host" "my.other.host" ];
+  };
+
+  test = ''
+    set -v
+    echo checking /etc/hosts file >&2
+
+    file=${config.out}/etc/hosts
+
+    grep '127.0.0.1' $file | head -n1 | grep localhost$ || exit 1
```
Enzime, line 15:
```suggestion
    grep '127.0.0.1' $file | head -n1 | grep localhost$
```

Can you remove the `|| exit` clauses?

Enzime (commented):
Can you squash all your changes into one commit and rebase on top of master?

Other than that it looks all good now 👍

tests/networking-hosts.nix
ibizaman: Are you sure? I was under the impression the goal was to know which statement was failing because of the numbered exit codes.

Enzime: My preference would be to not have different exit codes on failure

Teebor-Choka:
Ready?

Enzime (approved):
LGTM

Thanks for the contribution

oderwat:
This caught me cold and I think that there may be problems with that because Docker-Desktop changes the `/etc/hosts` and some other tools for Mac too. How to opt out of having home-manager to manage the `/etc/hosts` ?

PR #942, opened 2024-04-29
programs.nh: init module

alinnow:
Based on the [NixOS module](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/nh.nix) of the same name.

I used the same checks that `nix.gc` uses, as it performs exactly that function.

### DIFF
--- a/modules/programs/nh.nix
+++ b/modules/programs/nh.nix
@@ -1,0 +1,110 @@
+# Based off: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/nh.nix
+# When making changes please try to keep it in sync.
+{ config
+, lib
+, pkgs
+, ...
+}:
+
+with lib;
+
+let
+  cfg = config.programs.nh;
+in
+{
+  meta.maintainers = [
+    maintainers.alanpearce or "alanpearce"
+  ];
+
+  imports = [
+    (mkRemovedOptionModule [ "programs" "nh" "clean" "dates" ] "Use `programs.nh.clean.interval` instead.")
+  ];
+
+  options.programs.nh = {
+    enable = mkEnableOption "nh, yet another Nix CLI helper";
+
+    package = mkPackageOption pkgs "nh" { };
+
+    flake = mkOption {
+      type = with types; nullOr (either singleLineStr path);
+      default = null;
+      description = ''
+        The path that will be used for the `NH_FLAKE` environment variable.
+
+        `NH_FLAKE` is used by nh as the default flake for performing actions, like `nh os switch`.
+      '';
+    };
+
+    clean = {
+      enable = mkEnableOption "periodic garbage collection with nh clean all";
+
+      # Not in NixOS module
+      user = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = "User that runs the garbage collector.";
+      };
+
+      interval = mkOption {
+        type = types.attrs;
+        default = { Hour = 3; Minute = 15; };
+        description = ''
+          How often cleanup is performed.
+
+          The format is described in
+          {manpage}`launchd.plist(5)`.
+        '';
+      };
+
+      extraArgs = mkOption {
+        type = types.singleLineStr;
+        default = "";
+        example = "--keep 5 --keep-since 3d";
+        description = ''
+          Options given to {file}`nh` clean when the service is run automatically.
+
+          See {file}`nh clean all --help` for more information.
+        '';
+      };
+    };
+  };
+
+  config = {
+    warnings =
+      if (!(cfg.clean.enable -> !config.nix.gc.automatic)) then [
+        "programs.nh.clean.enable and nix.gc.automatic are both enabled. Please use one or the other to avoid conflict."
+      ] else [ ];
+
+    assertions = [
+      # Not strictly required but probably a good assertion to have
+      {
+        assertion = cfg.clean.enable -> cfg.enable;
+        message = "programs.nh.clean.enable requires programs.nh.enable";
+      }
+
+      {
+        assertion = (cfg.flake != null) -> !(hasSuffix ".nix" cfg.flake);
+        message = "nh.flake must be a directory, not a nix file";
+      }
+    ];
+
+    environment = mkIf cfg.enable {
+      systemPackages = [ cfg.package ];
+      variables = mkIf (cfg.flake != null) {
+        NH_FLAKE = cfg.flake;
+      };
+    };
+
+    launchd = mkIf cfg.clean.enable {
+      daemons.nh-clean = {
+        command = "${getExe cfg.package} clean all ${cfg.clean.extraArgs}";
+        serviceConfig = {
+          RunAtLoad = false;
+          StartCalendarInterval = [ cfg.clean.interval ];
+          UserName = cfg.clean.user;
+        };
+        path = [ config.nix.package ];
+      };
+    };
+  };
+}
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -105,6 +105,7 @@
   ./programs/gnupg.nix
   ./programs/man.nix
   ./programs/info
+  ./programs/nh.nix
   ./programs/nix-index
   ./programs/ssh
   ./programs/tmux.nix
--- a/modules/system/checks.nix
+++ b/modules/system/checks.nix
@@ -195,6 +195,17 @@ let
     fi
   '';
 
+  nhCleanAll = ''
+    if test -O /nix/store; then
+        echo "[1;31merror: A single-user install can't run nh clean all as root, aborting activation[0m" >&2
+        echo "Configure nh clean all to run as the current user:" >&2
+        echo >&2
+        echo "    programs.nh.clean.user = \"$USER\";" >&2
+        echo >&2
+        exit 2
+    fi
+  '';
+
   # TODO: Remove this a couple years down the line when we can assume
   # that anyone who cares about security has upgraded.
   oldSshAuthorizedKeysDirectory = ''
@@ -279,6 +290,7 @@ in
       (mkIf cfg.verifyBuildUsers preSequoiaBuildUsers)
       (mkIf cfg.verifyBuildUsers buildGroupID)
       (mkIf config.nix.enable nixDaemon)
+      (mkIf (config.programs.nh.clean.enable && config.programs.nh.clean.user == null) nhCleanAll)
       nixInstaller
       (mkIf cfg.verifyNixPath nixPath)
       oldSshAuthorizedKeysDirectory
--- a/release.nix
+++ b/release.nix
@@ -80,6 +80,7 @@ in {
 
   tests.activation-scripts = makeTest ./tests/activation-scripts.nix;
   tests.autossh = makeTest ./tests/autossh.nix;
+  tests.checks-nh-clean = makeTest ./tests/checks-nh-clean.nix;
   tests.environment-path = makeTest ./tests/environment-path.nix;
   tests.environment-terminfo = makeTest ./tests/environment-terminfo.nix;
   tests.homebrew = makeTest ./tests/homebrew.nix;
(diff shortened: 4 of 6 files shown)


skykanin:
Does the `nh` CLI even support `nix-darwin`?

alanpearce:
> Does the `nh` CLI even support `nix-darwin`?

Yes? I wouldn't have worked on this otherwise.  Only the `os` subcommand doesn't work

viperML:
nh doesn't wrap nix-darwin. And I don't use macOs, so thought has not been put into making sure the rest works there.

alanpearce:
> nh doesn't wrap nix-darwin. And I don't use macOs, so thought has not been put into making sure the rest works there.

Could you explain why you said that? I'm not sure what point you're trying to make.

viperML:
I don't know what is to explain, nh doesn't support `nix-darwin` in the sense that it doesn't have anything to do with `nix-darwin`, not nix on darwin.

ToyVo:
The `nh os switch` does not support Darwin due to differences between nixos-rebuild and darwin-rebuild. I have a fork that adds support for darwin. https://github.com/ToyVo/nh

The `nh clean` command does work on Darwin on the original and my fork.

alanpearce:
> I don't know what is to explain, nh doesn't support `nix-darwin` in the sense that it doesn't have anything to do with `nix-darwin`, not nix on darwin.

I think perhaps the meaning of "support" here is overloaded. I'll try to clarify, please correct me if I'm wrong:
- `nh` _works_ on darwin as in I can install the package and run it and it works, except:
- the `os` subcommand does not work on darwin as you don't support darwin and [don't want to](https://github.com/viperML/nh/pull/97#issuecomment-2078654074), which is totally fine.

As `nh` has more functionality than the `os` subcommand, it is reasonable to include in nix-darwin. All the functionality of `nh` exposed as options for nix-darwin in this PR work—that is, there is no functionality here that relies on the `os` subcommand.

bestlem:
@alanpearce 

My understanding is that nh was written by @viperML and this version does not support darwin as @viperML does not have a Mac to be able to test.
@ToyVo has produced a fork of the original nh and this does support darwin ie os switch works. (I have used this and it does work).

The nh module here is using @ToyVo s fork and so is fully functional on macOS.

viperML:
To add the comment from @bestlem, there are 2 side of the "not supported":

A) Not supporting darwin as the run-time platform for `nh home` or `nh clean`. I guess they work fine but I also don't know any of the corner cases for darwin.
B) Not having a nix-darwin wrapper, which is done in the fork.

alanpearce:
@viperML I'm still not sure what the point is you're trying to make. What is the reason for you saying that darwin isn't supported? What do you hope to achieve? Is there something that you think I should do with regard to this PR?

bestlem:
@alanpearce What @ViperML is saying that his code might or might not run on darwin and if you find an issue with his original on darwin he can't investigate so can't fix it.

You run his executable at your own risk and the author can't answer anything about issues.

@viperML is not being awkward here they are just saying that they can't deal with darwin issues. This is fair as they can't test,

To test things you have to have all the different hardware and operating systems  that your customers use. You can do this if you are reasonably sized organisation but no-one can expect single person FOSS developers to do this. This is no doubt one of the reason nix-darwin exists it can deal with issues that are different to nixes.

SolidRhino:
@LnL7 when will you merge this? This is a great edition.

ToyVo:
@SolidRhino the nh-darwin fork does have a Darwin module that defines the same options if you are impatient for this pr.

I have also recently closed an issue that pointed out that there was indeed a bug with nh os clean on macos that I have since fixed. That bug was macos uids start at 501 so the original nh would clean xdg directories for users 1000-1100, now on nh-dawrin the appropriate users 501-601 are cleaned.

kamushadenes:
Hey, can we get this merged? Thanks!

zhongjis:
Wondering what is blocking us from merging this PR? If possible can any of the maintainers providing some specifications or concerns so we can address them?

kimjongbing:
> Wondering what is blocking us from merging this PR? If possible can any of the maintainers providing some specifications or concerns so we can address them?

someone needs to get the maintainer a macbook lol

bestlem:
@kimjongbing No - this refers to a fork for the macOS version by ToyVo

modules/programs/nh.nix
```diff
@@ -0,0 +1,110 @@
...
+  ];
+
+  imports = [
+    (mkRemovedOptionModule [ "programs" "nh" "clean" "dates" ] "Use `programs.nh.clean.interval` instead.")
+  ];
+
+  options.programs.nh = {
+    enable = mkEnableOption "nh, yet another Nix CLI helper";
+
+    package = mkPackageOption pkgs "nh" { };
```
Eveeifyeve, line 26: Instead could we use https://github.com/ToyVo/nh_darwin instead of the default nh package as it doesn't support `darwin-rebuild` when nh_darwin does!

Perhaps making it a input?

alanpearce: That would make sense, but requires nh_darwin to be in nixpkgs. As it is, users will be able to choose to use that package if they have imported it.

bestlem: nh makes errors under darwin - darwin has to use the nh_darwin fork.

For example there are checks on the user id - nh uses the Linux rules and so only checks for ids >= 1000 nh_darwin understands that macOS users start at 501. This affect nh clean as far as I remember

I don't think there is any point in a darwin module that does not use the nh_darwin fork.

Eveeifyeve: > That would make sense, but requires nh_darwin to be in nixpkgs. As it is, users will be able to choose to use that package if they have imported it.

I agree with this, @ToyVo Are you willing to make the package nh_darwin in nixpkgs?

Eveeifyeve: I mean I could but the problem is there is no ref tags yet until that is a thing I really can't make the package.

CC @ToyVo perhaps make a tag for each update and go based on what nh does already.

bestlem: It does not need to go in nixpkgs.

Just add it to the inputs in your flake.nix (As far as I see nh only deals with flakes so you will have a flake.nix to control your system) See the nix-darwin setup in https://github.com/ToyVo/nh_darwin

Eveeifyeve: > It does not need to go in nixpkgs.
> 
> Just add it to the inputs in your flake.nix (As far as I see nh only deals with flakes so you will have a flake.nix to control your system) See the nix-darwin setup in https://github.com/ToyVo/nh_darwin

The only problem for package types all packages are required to be on nixpkgs.

ToyVo: I have migrated to nh_plus, I've made the first tag of 3.6.0-0. This puts it in a better spot to bring into nixpkgs

Eveeifyeve: I will make the package later today, as I am at school.

ToyVo: I didn't see a pr for it so https://github.com/NixOS/nixpkgs/pull/357106

ToyVo: Change of plans, with base nh 4.0 nix-darwin support has been added. nh 4.0 is less flake-centric and updating flake inputs has been removed so I might keep my fork alive for that if it isn't re-added.

ToyVo:
Just reiterating outside of the last thread, with nh 4.0 nix-darwin is now supported. So when that update hits nixpkgs this PR will be good to merge

modules/programs/nh.nix
Eveeifyeve: Nh 4.0 has officially released on [unstable](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=nh) so this no longer an issue.

modules/programs/nh.nix
```diff
@@ -0,0 +1,110 @@
...
+      {
+        assertion = (cfg.flake != null) -> !(hasSuffix ".nix" cfg.flake);
+        message = "nh.flake must be a directory, not a nix file";
+      }
+    ];
+
+    environment = mkIf cfg.enable {
+      systemPackages = [ cfg.package ];
+      variables = mkIf (cfg.flake != null) {
+        FLAKE = cfg.flake;
```
ToyVo, line 94:
```suggestion
        NH_FLAKE = cfg.flake;
```
With 4.0 this is the var name otherwise LGTM

alanpearce: changed this and the docs that mentioned `FLAKE`

modules/programs/nh.nix
```diff
@@ -0,0 +1,110 @@
...
+    (mkRemovedOptionModule [ "programs" "nh" "clean" "dates" ] "Use `programs.nh.clean.interval` instead.")
+  ];
+
+  options.programs.nh = {
+    enable = mkEnableOption "nh, yet another Nix CLI helper";
+
+    package = mkPackageOption pkgs "nh" { };
+
+    flake = mkOption {
+      type = types.nullOr types.path;
```
gigamonster256, line 29: See https://github.com/nix-community/home-manager/pull/6468 and https://github.com/nix-community/home-manager/pull/6898 (allows flake specs like "github:owner/repo")
```suggestion
      type = with types; nullOr (either singleLineStr path);
```

modules/programs/nh.nix
```diff
@@ -0,0 +1,110 @@
...
+      extraArgs = mkOption {
+        type = types.singleLineStr;
+        default = "";
+        example = "--keep 5 --keep-since 3d";
+        description = ''
+          Options given to {file}`nh` clean when the service is run automatically.
+
+          See {file}`nh clean all --help` for more information.
+        '';
+      };
```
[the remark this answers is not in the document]
alanpearce: That would be nice, but the launchd calendar format is completely different from the systemd one, unfortunately.

skykanin:
`nh` 4.x.x with darwin support has landed in nixpkgs. What's keeping this PR from being merged if anything?

dwt:
This is very similar to #1547 can we perhaps join forces?

dwt:
I have just created a custom config for my machine that runs nh with these settings:

```nix
  # This works with a feature of nix-direnv that automatically touches used direnv-gc-roots on environment activation
  launchd.daemons.nh-nix-garbage-collector = {
    command = "${lib.getExe pkgs.nh} clean all --keep 5 --keep-since 2w --verbose --optimise";
    path = [ config.nix.package ];
    serviceConfig = {
      LowPriorityIO = true;
      LowPriorityBackgroundIO = true;
      ProcessType = "Background";
      StandardOutPath = "/var/log/nh-gc.log";
      StandardErrorPath = "/var/log/nh-gc-error.log";
      StartInterval =
        let
          weekly = 7 * 24 * 60 * 60;
        in
        weekly;
      WorkingDirectory = "/var/root";
    };
  };
```

That I also very similar, but perhaps the `LowPriorityIO` and `LowPriorityBackgroundIO` and `ProcessType` would be good additions to ensure that this interrupts the user as little as possible. I am not sure if setting `LowPriorityBackgroundIO` in addition to the others actually does something, perhaps someone more knowledgeable can chime in on that.

I have also added log paths for debugging - not sure if it's better to just let the unified macOS logging do this for us? My investigations so far have only brought up something like this: `nh clean all ... 2>&1 | /usr/bin/logger -t nh-gc` to actually get the logs into the unified logging of darwin, to later be able to query them like this:

```sh
log show --predicate 'eventMessage CONTAINS "nh-gc"' --last 2d --style syslog
log stream --predicate 'eventMessage CONTAINS "nh-gc"' --style syslog
```

dwt:
friendly ping @alanpearce hey, are you still around and willing to continue this? If not, I and (probably) @quinneden would like to continue this work to get it merged.

alinnow:
Still around, but lost the motivation to continue.
If there's more to do than a rebase, feel free to build on it (or not, it's just code :) ).

dwt:
@quinneden you already started a pull request, would you be willing to merge in the changes and suggestions from here?

Issue #667, opened 2023-05-23
Default prompt walters interferes with zsh printing output that doesn't contain new line.


takeda:
So this looks like this has been forever (according to git blame it's been at least for 7 years).

The zsh has a mechanism where if output doesn't end with a new line, it will add it automatically and mark that with inverted percent sign:

https://unix.stackexchange.com/questions/167582/why-zsh-ends-a-line-with-a-highlighted-percent-symbol

The Walters default prompt which is defined if the user doesn't change `promgrams.zsh.promptInit` has a fancy way of showing current directory on the right side of the screen. This unfortunately interferes with the former and causes the last line that has no new line at the end to be covered by the prompt.

This happens even if one installs custom prompts, either via oh-my-zsh or via starship.

Since I use a custom prompt, so for me the solution to this problem was to set `promgrams.zsh.promptInit` to empty string. I think setting `prompt_cr` and `prompt_sp` options in zsh could also be solution if somebody wants to use prompt like Walters, but then setting prompt like Walters might have other unintended consequences for somebody who uses different prompt anyway.

Enzime:
Closed by #954

PR #954, opened 2024-05-20
Change zsh default prompt

thegleich:
Zsh prompt Walters causes issues as the default prompt and is a strange default, e.g.
it sets the RPS1.

We should use the same prompt as NixOS uses:
https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/zsh/zsh.nix#L93

Resolves:
https://github.com/LnL7/nix-darwin/issues/667

Related discussion:
https://github.com/NixOS/nixpkgs/pull/38535

### DIFF
--- a/modules/programs/zsh/default.nix
+++ b/modules/programs/zsh/default.nix
@@ -55,7 +55,7 @@ in
 
     programs.zsh.promptInit = mkOption {
       type = types.lines;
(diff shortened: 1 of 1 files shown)


Enzime (approved):
LGTM

Thanks for the contribution

Issue #924, opened 2024-04-06
`cachix-agent` service dies at launch


dhess:
This started happening sometime in the last month or two with `nixpkgs-unstable`:

```
cachix: security: createProcess: posix_spawnp: does not exist (No such file or directory)
```

It has something to do with the `PATH` set by the `launchd` service. I looked at the Cachix agent source and didn't see anything obvious, and have run out of time to debug it properly, but FYI adding `/usr/bin` to the `launchd` service's `PATH` fixes the problem.

(I suspected `coreutils` and added `coreutils-full` to the path, but that did not fix the issue on my system.)

domenkozar:
I've only seen it now, looks similar to https://github.com/NixOS/nixpkgs/issues/297775

sandydoo:
So one of the Haskell TLS libraries is calling `security`, which isn't in the path.

We can either add `/usr/bin` to the launchctl `paths`, or do something like this https://github.com/tweag/rules_haskell/commit/31171a520f49f263895112678ac93c7ed958ead1

sandydoo:
`config.environment.systemPaths` would probably also work?

PR #955, opened 2024-05-21
hercules-ci-agent: fix crash calling `security`

zowoq:
hercules ci agent has the same problem as was reported in https://github.com/LnL7/nix-darwin/issues/924.

I had originally fixed this downstream in nix-community by wrapping `/usr/bin/security` (https://github.com/nix-community/infra/commit/5da85a9b7276f15a64052dcc820ab210d742b914) but here I've reused the same fix that was used for cachix agent.

cc @roberth

### DIFF
--- a/modules/services/hercules-ci-agent/default.nix
+++ b/modules/services/hercules-ci-agent/default.nix
@@ -25,7 +25,7 @@ in
     launchd.daemons.hercules-ci-agent = {
       script = "exec ${cfg.package}/bin/hercules-ci-agent --config ${cfg.tomlFile}";
 
-      path = [ config.nix.package ];
+      path = [ config.nix.package config.environment.systemPath ];
       environment = {
         NIX_SSL_CERT_FILE = "${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt";
       };


roberth:
Calling security sounds alarming. More seriously though, do you know why it's called? I don't see an explicit call to it, neither from the Nix code base, nor hercules-ci-agent. I suspected it might be called by `sandbox-exec`, which would explain why I haven't seen this before, assuming most installations have been multi-user so far. I suppose then `cachix-agent` would only fail if it had to fall back to building something.

Regardless, this fix looks alright to me, and I'd be happy to upstream it into the agent repo as well.

roberth (approved):

zowoq:
> More seriously though, do you know why it's called? I don't see an explicit call to it, neither from the Nix code base, nor hercules-ci-agent.

I looked as well at first and also couldn't find it, AFAIK it is in one of the haskell libs. https://github.com/LnL7/nix-darwin/issues/924#issuecomment-2115301197

First noticed this problem after this haskell updates merge with the ghc 9.4 -> 9.6 bump. https://github.com/NixOS/nixpkgs/pull/279413

Also seems to be causing problems for other haskell packages. https://github.com/NixOS/nixpkgs/issues/297775

Enzime (approved):
LGTM

Thanks for the contribution

sandydoo:
> I looked as well at first and also couldn't find it, AFAIK it is in one of the haskell libs. [#924 (comment)](https://github.com/LnL7/nix-darwin/issues/924#issuecomment-2115301197)

At least for cachix, I think it's coming from here: https://github.com/kazu-yamamoto/crypton-certificate/blob/a91c3f4bd8ccbbb0568f41e185a0329855a05b90/x509-system/System/X509/MacOS.hs#L22

PR #957, opened 2024-05-22
homebrew: add extraMasApps option for composability of mas apps

abayomi185:
I thought it might be good to have masApps as a list such that it is automatically merged and can be composed in different files/modules like with casks and brews.

Attempted to fit it into `masApps` but struggled to get the tests to pass so opted for `extraMasApps` option instead. Happy to make changes if this approach is not ideal.

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -520,6 +520,32 @@ let
           + optionalString (sCfgSubset != { }) ", ${mkBrewfileLineOptionsListString sCfgSubset}";
       };
   };
+
+  masOptions = { config, ... }: {
+    options = {
+      name = mkOption {
+        type = types.nullOr (types.str);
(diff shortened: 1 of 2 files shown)


malob:
I think I'm confused about what problem this is trying to solve, e.g., in my config I have the following:
https://github.com/malob/nixpkgs/blob/8b609120370553e12aeca01c5b65afdf541af2f7/darwin/homebrew.nix#L43-L67

I added the following to a [separate file](https://github.com/malob/nixpkgs/blob/master/darwin/general.nix) that's part of my config,
```nix
homebrew.masApps = {
  _11111 = 123;
};
```

and `nix repl` shows that it merges them:
```
nix-repl> darwinConfigurations.MaloBookPro.config.homebrew.masApps 
{ "1Password for Safari" = 1569813296; [...] _11111 = 123; }
```

PR #961, opened 2024-06-01
feat: check if homebrew exists

### LINKED ISSUE #778: homebrew.enable fails silently when homebrew is not installed
Description:

Running `darwin-rebuild switch --flake .#my-machine` with a flake that has `homebrew.enable = true;` completes without a warning or an error.

Expectation:

I would expect to at least be warned that `homebrew` is not installed, instead of failing silently.

Version:

```shell
> darwin-version 
23.11pre-git+darwin4.4496ab2
```

codingCoffee:
fixes #778

- [ ] unsure if `checkHomebrewInstall` should be a conditional check

cc: @Enzime please recommend changes

### DIFF
--- a/modules/system/checks.nix
+++ b/modules/system/checks.nix
@@ -28,6 +28,7 @@ let
     fi
   '';
 
+
   oldBuildUsers = ''
     if dscl . -list /Users | grep -q '^nixbld'; then
         echo "[1;31merror: Detected old style nixbld users, aborting activation[0m" >&2
@@ -283,6 +284,19 @@ let
         exit 2
     fi
   '';
+
+  homebrewInstalled = ''
+    if [[ ! -f ${escapeShellArg config.homebrew.brewPrefix}/brew ]]; then
+        echo "[1;31merror: Using the homebrew module requires homebrew installed, aborting activation[0m" >&2
+        echo "Homebrew doesn't seem to be installed. Please install homebrew separately." >&2
+        echo "You can install homebrew using the following command:" >&2
+        echo >&2
+        # shellcheck disable=SC2016
+        echo '    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"' >&2
+        echo >&2
+        exit 2
+    fi
+  '';
 in
 
 {
@@ -331,6 +345,7 @@ in
       nixInstaller
       (mkIf cfg.verifyNixPath nixPath)
       oldSshAuthorizedKeysDirectory
+      (mkIf config.homebrew.enable homebrewInstalled)
     ];
 
     system.activationScripts.checks.text = ''


modules/system/checks.nix
```diff
@@ -44,6 +44,18 @@ let
     fi
   '';
 
+  checkHomebrewInstall = ''
+    if [ ! -f /usr/local/bin/brew ] && [ ! -f /opt/homebrew/bin/brew ]; then
+        echo "[1;31merror: Using the homebrew module requires homebrew installed, aborting activation[0m" >&2
+        echo "Homebrew doesn't seem to be installed. Please install homebrew separately." >&2
+        echo "You can install homebrew using the following command" >&2
+        echo >&2
+        echo '    /bin/bash -c \"$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)\"' >&2
```
joaothallis, line 53: `\` is being printed but the command works with `\` but in the [website](https://brew.sh) it is without the `\`
```suggestion
        echo '    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"' >&2
```

modules/system/checks.nix
```diff
@@ -44,6 +44,18 @@ let
     fi
   '';
 
+  checkHomebrewInstall = ''
+    if [ ! -f /usr/local/bin/brew ] && [ ! -f /opt/homebrew/bin/brew ]; then
```
joaothallis, line 48: For the case when someone is using a custom directory for brew

https://stackoverflow.com/a/677212
```suggestion
    if [ command -v brew ]; then
```

codingCoffee: The thing is, this does not work. The `brew` binary is not in path by default. At the end of the homebrew installation, we're asked to put the following line `eval "$(/usr/local/bin/brew shellenv)"` in our `.zprofile` or `.bashrc`. This is what loads `brew` into the path. Ergo, using something like `command -v brew` will always fail when nix is running since it doesn't evaluate the user's zsh / bash profile.

codingCoffee: But I do recognize my solution would fail for someone using a custom directory for brew. Not sure what an optimal solution for that case should be. We cannot even `eval "$(/usr/local/bin/brew shellenv)"` in the nix code, since that too would require knowledge about `brew`'s installation path.

codingCoffee:
Also @joaothallis is there a way to evaluate the `checkHomebrewInstall` check only if the user has `homebrew.enable = true` in this nix configuration?

modules/system/checks.nix
emilazy: This should use `${config.homebrew.brewPrefix}/brew`, like the Homebrew module itself does.

emilazy:
You can use `mkIf` in the `system.checks.text` entry.

modules/system/checks.nix
```diff
@@ -237,6 +249,7 @@ in
       darwinChanges
       runLink
       oldBuildUsers
+      checkHomebrewInstall
```
Enzime, line 252:
```suggestion
      (mkIf config.homebrew.enable checkHomebrewInstall)
```

modules/system/checks.nix
Enzime:
```suggestion
    if [ ! -f ${config.homebrew.brewPrefix}/brew ]; then
```

Enzime (changes requested):
Please let us know if you're still interested on working on this PR

codingCoffee:
Hey @Enzime, definitely interested in submitting the PR. I've commited the suggessions and resolved the conflicts. Please do tell if I need to make any more changes.

modules/system/checks.nix
```diff
@@ -28,6 +28,18 @@ let
     fi
   '';
 
+  checkHomebrewInstall = ''
+    if [ ! -f ${config.homebrew.brewPrefix}/brew ]; then
```
Enzime, line 32:
```suggestion
    if [[ ! -f ${escapeShellArg config.homebrew.brewPrefix}/brew ]]; then
```

Enzime (changes requested):
Can you squash all the commits into one?

codingCoffee:
Hey @Enzime, I just ended up creating a separate PR: https://github.com/LnL7/nix-darwin/pull/1160. Squashing was proving to be difficult since this is such an old PR. If it's okay can you close this and merge the other one?

modules/system/checks.nix
```diff
@@ -28,6 +28,18 @@ let
     fi
   '';
 
+  checkHomebrewInstall = ''
+    if [[ ! -f ${escapeShellArg config.homebrew.brewPrefix}/brew ]]; then
+        echo "[1;31merror: Using the homebrew module requires homebrew installed, aborting activation[0m" >&2
+        echo "Homebrew doesn't seem to be installed. Please install homebrew separately." >&2
+        echo "You can install homebrew using the following command" >&2
+        echo >&2
+        echo '    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"' >&2
```
Enzime, line 37:
```suggestion
        # shellcheck disable=SC2016
        echo '    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"' >&2
```

Enzime (changes requested):
To squash the commits on this PR, the easiest solution would be while rebasing them on top of the latest master

First you need to make sure you have this repo as a remote:

```
$ git remote -v
upstream  https://github.com/LnL7/nix-darwin.git (fetch)
upstream  https://github.com/LnL7/nix-darwin.git (push)
```

The name can be anything, and if it’s not there you can use `git remote add upstream https://github.com/LnL7/nix-darwin.git`

```
$ git fetch
$ git checkout homebrewcheck
$ git rebase -i upstream/master
```

And in the rebase editor you can change all the `pick` commands except the first to `squash`

Enzime (approved):
LGTM

Thanks for the contribution

PR #964, opened 2024-06-05
Fix NextDNS arguments

kamushadenes:
It seems that `escapeShellArgs` doesn't play nicely with launchd, as it surrounds strings with single quotes that are interpreted literally by launchd and passed as arguments (think passing `"'-profile'"` instead of `"-profile"`).

```
Unrecognized parameter: '-profile'
```

```
Unrecognized parameter: '-profile 192.168.4.0/24=XXXXXX'
```

This PR maps the single version of `escapeShellArg` to each argument, and removes the resulting surrounding quotes.

### DIFF
--- a/modules/services/nextdns/default.nix
+++ b/modules/services/nextdns/default.nix
@@ -31,7 +31,7 @@ in {
     launchd.daemons.nextdns = {
       path = [ nextdns ];
       serviceConfig.ProgramArguments =
-        [ "${pkgs.nextdns}/bin/nextdns" "run" (escapeShellArgs cfg.arguments) ];
+        [ "${pkgs.nextdns}/bin/nextdns" "run" ] ++ cfg.arguments;
       serviceConfig.KeepAlive = true;
       serviceConfig.RunAtLoad = true;
     };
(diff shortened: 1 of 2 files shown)


emilazy:
I’m not sure `escapeShellArg` is correct here? I don’t think the launchd `ProgramArguments` are passed through a shell at all (hence it being a list of arguments rather than a string), so I don’t think they need any escaping at all.

kamushadenes:
I asked myself the same thing, but couldn't find a documentation stating how launch parses stuff. I'll remove the escaping.

kamushadenes:
Done.

kamushadenes:
@emilazy can you take a look?

emilazy:
This looks good to me; sorry for not getting to this sooner. Could you squash this into one commit, e.g. `nextdns: fix argument handling`?

kamushadenes:
@emilazy not a problem, done!

emilazy (approved):
Thanks!

emilazy:
Sorry for the delay, they were being slow today :)

PR #973, opened 2024-06-11
module : `jankyborders` for window borders Configuration

amsynist:
[](url)Hi @LnL7,

I hope this message finds you well. I am submitting this PR to add a new NixOS module for [jankyborders](https://search.nixos.org/packages?channel=24.05&show=jankyborders&from=0&size=50&sort=relevance&type=packages&query=jankyborders). This module provides configuration options to enable the service and customize various parameters such as width, active and inactive colors, HiDPI mode, and focus settings.

Official Repo : https://github.com/FelixKratz/JankyBorders

### Summary of Changes
- Added a new module `services.jankyborders` to the NixOS configuration.
- Included configuration options for:
  - `width`
  - `active_color`
  - `inactive_color`
  - `hidpi`
  - `ax_focus`
  - `background_color`
  - `blur_radius`
  - `blacklist`
  - `whitelist`
- Implemented a test case to verify the correct setup of the `jankyborders` service.

### Module Usage Example
```nix
{ pkgs, ... }:
{
  services.jankyborders = {
    enable = true;
    package = pkgs.jankyborders;
    width = 5.0;
    active_color = "gradient(top_left=0xffDDB6F2,bottom_right=0xff96CDFB)";
    inactive_color = "0x9992B3F5";
    hidpi = true;
    ax_focus = true;
    whitelist = ["Arc" "USB Drive" "WezTerm"];
  };
}
```

### DIFF
--- a/modules/services/jankyborders/default.nix
+++ b/modules/services/jankyborders/default.nix
@@ -1,0 +1,157 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}: let
+  inherit (lib) maintainers mkEnableOption mkIf mkPackageOptionMD mkOption types;
+
+  cfg = config.services.jankyborders;
+  joinStrings = strings: builtins.concatStringsSep "," strings;
+
+  optionalArg = arg: value:
+    if value != null && value != ""
+    then
+      if lib.isList value
+      then lib.map (val: "${arg}=${val}") value
+      else ["${arg}=${value}"]
+    else [];
+in {
+  meta.maintainers = [
+    maintainers.amsynist or "amsynist"
+  ];
+
+  options.services.jankyborders = {
+    enable = mkEnableOption "Enable the jankyborders service.";
+
+    package = mkPackageOptionMD pkgs "jankyborders" {};
+
+    width = mkOption {
+      type = types.float;
+      default = 5.0;
+      description = ''
+        Determines the width of the border. For example, width=5.0 creates a border 5.0 points wide.
+      '';
+    };
+
+    hidpi = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        If set to on, the border will be drawn with retina resolution.
+      '';
+    };
+
+    active_color = mkOption {
+      type = types.str;
+      default = "0xFFFFFFFF";
+      example = "0xFFFFFFFF";
+      description = ''
+        Sets the border color for the focused window (format: 0xAARRGGBB). For instance, active_color="0xff00ff00" creates a green border.
+        For Gradient Border : active_color="gradient(top_right=0x9992B3F5,bottom_left=0x9992B3F5)"
+      '';
+    };
+
+    inactive_color = mkOption {
+      type = types.str;
+      default = "0xFFFFFFFF";
+      example = "0xFFFFFFFF";
+      description = ''
+        Sets the border color for all windows not in focus (format: 0xAARRGGBB).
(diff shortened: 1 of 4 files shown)


modules/services/jankyborders/default.nix
```diff
@@ -0,0 +1,148 @@
...
+        Once this list is populated, only applications listed here are considered for receiving a border. If the whitelist is empty (default) it is inactive.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    environment.systemPackages = [cfg.package];
+
+    launchd.user.agents.jankyborders = {
+      path = [config.environment.systemPath];
```
Samasaur1, line 121: Is it necessary for the service to have the entire system path rather than a more limited path to the specific subset of packages that ti needs?

modules/services/jankyborders/default.nix
```diff
@@ -0,0 +1,148 @@
...
+
+    whitelist = mkOption {
+      type = types.str;
+      default = "";
+      example = "Safari,kitty";
+      description = ''
+        Once this list is populated, only applications listed here are considered for receiving a border. If the whitelist is empty (default) it is inactive.
+      '';
+    };
+  };
```
Samasaur1, line 115: I think these should both be exposed as a list of strings (e.g. `[ "Safari" "kitty" ]`) and should be transformed into the required format inside the module. When doing so, you should also make sure to handle applications with spaces in their names (such as "Google Chrome.app").

Also, if it doesn't make sense for both a whitelist and a blacklist to be defined, then you should put an assertion in the module to check that that is not the case.

Samasaur1 (changes requested):

modules/services/jankyborders/default.nix
amsynist: @Samasaur1  thanks for the review , yeah did commit to correct this and points to the pkg path instead whole system path

modules/services/jankyborders/default.nix
amsynist: Updated to list and handling the white spaces along with assertion to to the whitelist/blacklist options

modules/services/jankyborders/default.nix
```diff
@@ -0,0 +1,158 @@
...
+    assertions = [
+      {
+        assertion = !(cfg.blacklist != [] && cfg.whitelist != []);
+        message = "Cannot define both a blacklist and a whitelist for jankyborders.";
+      }
+    ];
+    environment.systemPackages = [cfg.package];
+
+    launchd.user.agents.jankyborders = {
+      path = [cfg.package];
```
Samasaur1, line 131: I actually think you don't need to set `path` at all, since you're calling the package via its absolute path in `ProgramArguments`.

amsynist: @Samasaur1 Corrected this and just reverified !

Samasaur1 (approved):
LGTM but I do not have merge permissions

Demianeen:
Are there any updates on janky borders support with nix?

amsynist:
> Are there any updates on janky borders support with nix?

Waiting this to be merged :(

emilazy:
Sorry for the belated review; this seems fine to me. Could you add the missing trailing newline to `modules/services/jankyborders/default.nix` and squash the fix commits into the first one?

amsynist:
> Sorry for the belated review; this seems fine to me. Could you add the missing trailing newline to `modules/services/jankyborders/default.nix` and squash the fix commits into the first one?

@emilazy  added the missing trailing newline and squashed

emilazy (approved):
Thank you!

PR #974, opened 2024-06-13
linux-builder: make compatible with cross-arch builder package

nicknovitski:
Before this commit, aarch64 users building the following configuration would end up with an aarch64-linux builder, while after it, they get the x86_64-linux builder they expect:
```nix
 nix.linux-builder = {
  enable = true;
  package = pkgs.darwin.linux-builder-x86_64;
};
```

Before, in order to get an x86_64-linux builder, they would have needed to use this configuration instead:
```nix
 nix.linux-builder = {
  enable = true;
  config.nixpkgs.hostPlatform = "x86_64-linux";
  systems = ["x86_64-linux"];
};
```

The reason for this is that the linux-builder module calls `override` on the package option, and the `linux-builder-x86_64` package is also defined using override:
```nix
linux-builder-x86_64 = linux-builder.override {
  modules = [ { nixpkgs.hostPlatform = "x86_64-linux"; } ];
};
```

The module was effectively discarding the `nixpkgs.hostPlatform` option.

Example issue: https://github.com/NixOS/nixpkgs/issues/313784

### DIFF
--- a/modules/nix/linux-builder.nix
+++ b/modules/nix/linux-builder.nix
@@ -7,9 +7,11 @@ let
 
   cfg = config.nix.linux-builder;
 
-  builderWithOverrides = cfg.package.override {
-    modules = [ cfg.config ];
-  };
+  builderWithOverrides = cfg.package.override (previousArguments: {
+    # the linux-builder packages require a list `modules` argument, so it's
+    # always non-null.
+    modules = previousArguments.modules ++ [ cfg.config ];
+  });
 
   # create-builder uses TMPDIR to share files with the builder, notably certs.
   # macOS will clean up files in /tmp automatically that haven't been accessed in 3+ days.
@@ -133,8 +135,10 @@ in
 
     systems = mkOption {
       type = types.listOf types.str;
-      default = [ "${stdenv.hostPlatform.uname.processor}-linux" ];
-      defaultText = literalExpression ''[ "''${stdenv.hostPlatform.uname.processor}-linux" ]'';
+      default = [ builderWithOverrides.nixosConfig.nixpkgs.hostPlatform.system ];
+      defaultText = ''
+        The `nixpkgs.hostPlatform.system` of the build machine's final NixOS configuration.
+      '';
       example = literalExpression ''
         [
           "x86_64-linux"


modules/nix/linux-builder.nix
```diff
@@ -7,9 +7,9 @@ let
 
   cfg = config.nix.linux-builder;
 
-  builderWithOverrides = cfg.package.override {
-    modules = [ cfg.config ];
-  };
+  builderWithOverrides = cfg.package.override (previousArguments: {
+    modules = previousArguments.modules ++ [ cfg.config ];
```
Samasaur1, line 11: What happens if `previousArguments.modules` is null? Can that ever happen?

When I'm doing overrides like this, I generally use this pattern:
```suggestion
    modules = (previousArguments.modules or []) ++ [ cfg.config ];
```

Samasaur1: If we know for a fact that `previousArguments.modules` will never be null, then feel free to ignore this suggestion (though a comment indicating that we know it can never be null might still be helpful)

nicknovitski: Good question! I had to check to be sure. Yes, modules is a currently a required argument of the function used to create these packages, there is no default.  I'll add a comment.

modules/nix/linux-builder.nix
```diff
@@ -133,8 +135,8 @@ in
 
     systems = mkOption {
       type = types.listOf types.str;
-      default = [ "${stdenv.hostPlatform.uname.processor}-linux" ];
-      defaultText = literalExpression ''[ "''${stdenv.hostPlatform.uname.processor}-linux" ]'';
+      default = [ builderWithOverrides.nixosConfig.nixpkgs.hostPlatform.system ];
+      defaultText = literalExpression ''[ config.nix.linux-builder.package.nixosConfig.nixpkgs.hostPlatform.system " ]'';
```
Enzime, line 139:
```suggestion
      defaultText = literalExpression ''[ config.nix.linux-builder.package.nixosConfig.nixpkgs.hostPlatform.system ]'';
```

Enzime (changes requested):

modules/nix/linux-builder.nix
emilazy: Could we have descriptive text in `defaultText` instead here, if it’s not possible to provide a correct expression? In this case the user configuration has an effect on the resulting `hostPlatform` so the claimed default isn’t really accurate. Another alternative would be to export `builderWithOverrides` as `config.nix.linux-builder.finalPackage` or something (I vaguely recall there being a naming convention for such options in NixOS, but not what it is).

nicknovitski: I'd be happy to put a description instead.

emilazy (approved):
LGTM, thank you!

emilazy (approved):

Enzime (approved):
LGTM

PR #932, opened 2024-04-14
treewide: remove lib.mdDoc

wegank:
`lib.mdDoc` has been removed in NixOS/nixpkgs#303841.

PR #975, opened 2024-06-13
Drop compatibility for unsupported Nixpkgs versions

emilazy:
https://github.com/LnL7/nix-darwin/pull/932 already broke these versions, so we can do some clean‐up.


Enzime (approved):
LGTM

My configuration builds on this branch :+1:

PR #976, opened 2024-06-15
ssh: use symlinks for `authorizedKeys` options

emilazy:
This is a security fix over a year in the making; I wish I had been able to get it out sooner. Anyone using `AuthorizedKeysCommand` for something else will need to ensure they set it in a later file than `101-authorized-keys.conf`, but hopefully anyone with that customized an SSH setup knows what they’re doing, and I’m not sure how we could better signal that; I doubt anyone reads the changelog but not merged PRs.

This does mean that the Nix store failing to mount could lead to an SSH lock‐out; I’m not sure how we could handle that elegantly and it seems difficult to recover from without direct access anyway (what if your shell is in the Nix store anyway?). It’s possible we could do an ad‐hoc recreation of `copy` here that operates unconditionally on the entire directory, I guess.

### DIFF
--- a/modules/programs/ssh/default.nix
+++ b/modules/programs/ssh/default.nix
@@ -3,7 +3,7 @@
 with lib;
 
 let
-  cfg  = config.programs.ssh;
+  cfg = config.programs.ssh;
 
   knownHosts = map (h: getAttr h cfg.knownHosts) (attrNames cfg.knownHosts);
 
@@ -81,8 +81,7 @@ let
   };
 
   authKeysFiles = let
-    mkAuthKeyFile = u: nameValuePair "ssh/authorized_keys.d/${u.name}" {
-      copy = true;
+    mkAuthKeyFile = u: nameValuePair "ssh/nix_authorized_keys.d/${u.name}" {
       text = ''
         ${concatStringsSep "\n" u.openssh.authorizedKeys.keys}
         ${concatMapStrings (f: readFile f + "\n") u.openssh.authorizedKeys.keyFiles}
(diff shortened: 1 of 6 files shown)


emilazy:
Incidentally it seems like our `/etc` code doesn’t clean up empty directories that consisted entirely of Nix store symlinks. That seems unfortunate, but I’m not sure how we could fix it. In this case it seems like it would be best for the entire `/etc/ssh/nix_authorized_keys.d` directory to be a symlink; it would be easy to handle deleting them in that case. Not sure why it doesn’t work like that presently.

modules/programs/ssh/default.nix
```diff
@@ -97,28 +96,16 @@ let
 in
 
 {
+  imports = [
+    (mkRemovedOptionModule [ "services" "openssh" "authorizedKeysFiles" ] "No `nix-darwin` equivalent to this NixOS option.")
+  ];
```
Enzime, line 101: I added this because SSH keys specified in `~/.ssh/authorized_keys` stopped working (see #677), ideally it would be nice if this PR could maintain that, potentially by adding it to the `AuthorizedKeysCommand`

Enzime:
Everything else in the PR LGTM

Thanks for fixing this issue, I've known about it for a while but I never got around to fixing it as I was planning to add functionality that tracked which files were created by `nix-darwin` and then wipe them if they're no longer necessary (similar to the NixOS boot partition manager script)

modules/programs/ssh/default.nix
emilazy: I think that should work out of the box now, since we don’t override `AuthorizedKeysFile`:

              The  program should produce on standard output zero or more lines
              of authorized_keys output (see AUTHORIZED_KEYS in sshd(8)) .  Au‐
              thorizedKeysCommand is tried after the  usual  AuthorizedKeysFile
              files  and will not be executed if a matching key is found there.
              By default, no AuthorizedKeysCommand is run.

emilazy:
Yeah, I was talking with @Samasaur1 on Matrix and it seems like we might be able to implement a more robust `copy` mode by diffing against the currently‐active generation, but I don’t feel 100% confident in our ability to make it reliably clean up without ever trashing user data at present.

Enzime:
My preference would be to write proper state management logic and write a program that writes to a manifest file (JSON probably) that tracks all the generated files by `nix-darwin` with their original hashes so we know what we can safely override. Alternatively, we could go down the `home-manager` path which is any time we want to override a file, we can error out and tell the user to deal with it.

Enzime (approved):
LGTM

Happy to merge this PR first and if any future consumers want to use `.copy` they can readd it in their PR with better state management logic

PR #978, opened 2024-06-19
checks.nix: disable verifyBuildUsers for auto-allocate-uids

usertam:
Follow https://github.com/LnL7/nix-darwin/pull/681, but disable the check automatically if `config.nix.settings.auto-allocate-uids` is explicitly set to `true`.

Note that by default `config.nix.settings = {}`, and this should not affact people who do not configure `nix.settings.auto-allocate-uids`.

### DIFF
--- a/modules/system/checks.nix
+++ b/modules/system/checks.nix
@@ -220,7 +220,7 @@ in
 
     system.checks.verifyBuildUsers = mkOption {
       type = types.bool;
-      default = true;
+      default = !(config.nix.settings.auto-allocate-uids or false);
       description = "Whether to run the Nix build users validation checks.";
     };
 


modules/system/checks.nix
```diff
@@ -220,7 +220,7 @@ in
 
     system.checks.verifyBuildUsers = mkOption {
       type = types.bool;
-      default = true;
+      default = !(hasAttr "auto-allocate-uids" config.nix.settings && config.nix.settings.auto-allocate-uids);
```
emilazy, line 223: Nit: `!(config.nix.settings.auto-allocate-uids or false)`.

emilazy (changes requested):
Sure, seems like a good idea to me. I must disclaim that this setting is known to break a bunch of builds so you might not enjoy having it on, but we might as well not gratuitously break it further.

modules/system/checks.nix
usertam: Thanks!

emilazy (approved):
Thanks! I hope this can be the default one day, if we ever figure out how to deal with builds that want to know about their current user.

PR #988, opened 2024-06-30
programs.jdks: init

Samasaur1:
This PR allows you to install Java versions system-wide.

I've left stubs in place to allow setting the JDK to use, but I have not yet implemented that, so this only adds support for `programs.jdk.installed`

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -91,6 +91,7 @@
   ./programs/direnv.nix
   ./programs/fish.nix
   ./programs/gnupg.nix
+  ./programs/jdks.nix
   ./programs/man.nix
   ./programs/info
   ./programs/nix-index
--- a/modules/programs/jdks.nix
+++ b/modules/programs/jdks.nix
@@ -1,0 +1,98 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.programs.jdks;
+in
+
+{
+  meta.maintainers = [
+    lib.maintainers.samasaur or "samasaur"
+  ];
+
+  options.programs.jdks = {
+    installed = lib.mkOption {
+      type = with lib.types; listOf package;
+      default = [];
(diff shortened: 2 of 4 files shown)


Samasaur1:
I'm marking this PR as a draft because the consensus we came to in the nix-darwin Matrix room was that we'd like to wait for an overhaul of managing files outside of the Nix store. If anyone has been waiting on this PR, though, we can reconsider.

Samasaur1:
Depends on #1205, after which it should be as simple as something like
```nix
system.file = builtins.listToAttrs
  (builtins.map (jdk: 
    let
      versionElems = lib.strings.splitString "." jdk.version;
      majorVersion = builtins.elemAt versionElems 0;
    in
      lib.nameValuePair
      "Library/Java/JavaVirtualMachines/zulu-${majorVersion}.jdk"
      { source = "${jdk}/zulu-${majorVersion}.jdk"; }
  ) cfg.jdks);
```

Samasaur1:
As written, this PR only supports the Zulu JDKs in nixpkgs (which have macOS JDK bundles under the derivations). It should also be able to support other JDKs in nixpkgs (such as `temurin-bin`, which itself is a JDK bundle), so that's also on the TODO list

Samasaur1:
https://github.com/NixOS/nixpkgs/pull/375212 should make it easier to find JDK bundles from any JDK in nixpkgs

Samasaur1:
With that nixpkgs PR, it becomes as simple as
```nix
home.file = builtins.listToAttrs
  (builtins.map (jdk:
    lib.nameValuePair
      "Library/Java/JavaVirtualMachines/${builtins.baseNameOf jdk.bundle}"
      { source = jdk.bundle; }
  ) cfg);
```
for home-manager, and the equivalent with `system.file` for nix-darwin

PR #993, opened 2024-07-03
`services.nix-daemon.environmentVariables`

lrworth:
https://github.com/LnL7/nix-darwin/commit/a9e0f71c50fc9a72e22e991e323a6a7e50bfc0d7 made it possible to modify the environment passed to nix-daemon. This turns out to also be useful for [other things](https://github.com/NixOS/nix/issues/2161#issuecomment-2204763331), so let's make it a little more idiomatic and document it.

### DIFF
--- a/modules/services/nix-daemon.nix
+++ b/modules/services/nix-daemon.nix
@@ -22,7 +22,9 @@ in
 
     services.nix-daemon.environmentVariables = mkOption {
       type = types.attrsOf types.str;
-      default = { };
+      default =
+        # See https://github.com/LnL7/nix-darwin/commit/a9e0f71c50fc9a72e22e991e323a6a7e50bfc0d7.
+        config.nix.envVars;
       example = {
         AWS_CONFIG_FILE = "/etc/nix/aws/config";
         AWS_SHARED_CREDENTIALS_FILE  = "/etc/nix/aws/credentials";
@@ -74,7 +76,6 @@ in
 
       serviceConfig.EnvironmentVariables = mkMerge [
         cfg.environmentVariables
-        config.nix.envVars
         {
           NIX_SSL_CERT_FILE = mkIf
             (config.environment.variables ? NIX_SSL_CERT_FILE)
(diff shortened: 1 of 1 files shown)


emilazy:
Thanks for the contribution! The current behaviour [matches NixOS](https://github.com/NixOS/nixpkgs/blob/67ab30a1aac28c793dc119a74ceab95643a330f8/nixos/modules/services/system/nix-daemon.nix#L189C27-L189C34), so I’m a little worried about introducing divergence with their module. Maybe it’d be a good idea to get this upstream in NixOS too so we can minimize the differences and users can share their configuration? On the other hand I don’t want to double your work. Does setting `launchd.daemons.nix-daemon.serviceConfig.EnvironmentVariables` work currently? That at least signals that the same option won’t work on NixOS too.

lrworth:
As far as I can tell, `config.nix.envVars` is not documented in NixOS either so perhaps upstreaming this change is a good idea; but I don't have a NixOS machine to test with, nor time to set one up.

> Does setting launchd.daemons.nix-daemon.serviceConfig.EnvironmentVariables work currently?

I'm not sure what you are asking. That field is already used to configure the environment variables in `/Library/LaunchDaemons/org.nixos.nix-daemon.plist`.

emilazy:
Totally understood re: NixOS. I’ll try to find the time to raise it there, but I suspect the answer would be “just use the `systemd.*` settings”, which is why I brought up `launchd.*`.

> I'm not sure what you are asking. That field is already used to configure the environment variables in `/Library/LaunchDaemons/org.nixos.nix-daemon.plist`.

Ah, the difference is that `nix.nixVars` also gets added to your shell environment. I assumed that the idea of this option was for things that you wanted the daemon and your user to have separate values for, which is why I suggested setting them directly in the `launchd.*` options.

lrworth:
Just revisiting this, and I'm not sure what I was thinking before. Putting these vars in `launchd.daemons.nix-daemon.serviceConfig.EnvironmentVariables` works fine. Thanks @emilazy :)

PR #996, opened 2024-07-06
Add lix-installer to known files

### LINKED ISSUE #995: Unable to install on top of a fresh Lix install
I was looking to get started with `nix-darwin`, and tried to do a clean-slate install after uninstalling everything Nix from my Mac and then reinstalling with [Lix](https://lix.systems/install/), but ran into this (I think something similar to #880):
```console
$ nix run nix-darwin -- switch --flake .
error: Unexpected files in /etc, aborting activation
The following files have unrecognized content and would be overwritten:

  /etc/nix/nix.conf

Please check there is nothing critical in these files, rename them by adding .before-nix-darwin to the end, and then try again.
```

I don't want to delete that file as it seems like it would be important:
```ini
# Generated by https://install.lix.systems/.
# See `/nix/nix-installer --version` for the version details.

build-users-group = nixbld
experimental-features = nix-command flakes
max-jobs = auto
substituters = https://cache.nixos.org https://cache.lix.systems
trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= cache.lix.systems:aBnZUw8zA7H35Cz2RyKFVs3H4PlGTLawyY5KRbvJR8o=
```
The installer:
```console
$ /nix/nix-installer --version
lix-installer 0.17.1
$ sha256sum /etc/nix/nix.conf
24797ac05542ff8b52910efc77870faa5f9e3275097227ea4e50c430a5f72916  /etc/nix/nix.conf
```

Is this just a matter of adding another "known hash" like #881 ? I think there might also exist a variant of the `nix.conf` that would not use `experimental-features = nix-command flakes`, since Lix asks whether it should enable flakes at installation time. I'm not sure if the output is totally deterministic, it seems like it could vary a bit: https://git.lix.systems/lix-project/lix-installer#installation-differences

Meanwhile, is there any workaround to get past this and install anyway (like providing my own hash or explicitly ignoring the file, or something like https://github.com/LnL7/nix-darwin/wiki/Running-a-locally-patched-version-of-nix-darwin)? Is this something that would need to be fixed by Lix upstream somehow and I should file an issue there? 

Thanks in advance! I was hoping these two could work together, as I preferred using Lix on Linux but seems like getting started might be a little trickier than I thought 😆

ian-h-chamberlain:
Closes #995

With this change, I was able to set `nix-darwin.url` to my local checkout with `path:` and perform an install successfully.

As far as I can tell, enabling/disabling flakes is the only variation that normally can happen, so hopefully these two hashes cover both cases there...

The files generated by the Lix installer don't _seem_ to be version dependent, and its installer is based on the `DeterminateSystems` installer, so hopefully it's also forward-compatible like DeterminateSystems 0.16.0+

Please let me know if there's anything else needed to merge this!


emilazy (approved):
Looks good, thanks!

Issue #967, opened 2024-06-09
Is there a better (native) way to enable AppExposé


jeeftor:
```nix
    activationScripts.postUserActivation.text = ''
      # activateSettings -u will reload the settings from the database and apply them to the current session,
      # so we do not need to logout and login again to make the changes take effect.
      /System/Library/PrivateFrameworks/SystemAdministration.framework/Resources/activateSettings -u

      # Enable App Exposé
      defaults write com.apple.dock showAppExposeGestureEnabled -bool true
      killall Dock
    '';
```

I'm just using activation scripts - but wondering if there is a native way. Couldn't find anything in the docs but I could be stupid today :)

Samasaur1:
The `defaults` command can definitely be done natively, but I don't think we have any existing options to run `activateSettings -u` or `killall Dock`. In my personal opinion, it's nicer to keep related things together, so I think what you're doing now is probably your best bet.

motschel123:
I would like to configure App Exposé with a boolean option somewhere under system.defaults (probably NSGlobalDomain), is this planned anytime soon?

noahfraiture:
> I would like to configure App Exposé with a boolean option somewhere under system.defaults (probably NSGlobalDomain), is this planned anytime soon?

any news ?

PR #999, opened 2024-07-10
defaults: restart Dock when changing settings

Enzime:


Samasaur1 (approved):
This looks good to me. The one thing that I'm wondering about is whether we want to detect when the user manually sets defaults for `com.apple.dock` rather than using the nice option, but I think if they choose to do that then it's on them to restart the dock

emilazy (commented):
A couple thoughts:

* Are there any potential downsides to this? I remember in the past killing the Dock led to some slight weirdness – e.g. dynamic app icons reverting to their closed state.

* There’s some other command people have been advocating for that applies more settings at once, right? I forget exactly what but it’s come up several times here. Do we want to run that too?

We definitely have settings that just won’t take effect unless you log back out and in again, I think, so I’m wondering whether we want to get into this game of whack‐a‐mole with user expectations for being able to set settings and have them apply to the running session. Probably it’s not so bad since some of the settings we have now do already just work in real‐time.

modules/system/defaults-write.nix
```diff
@@ -107,6 +107,11 @@ in
         ${concatStringsSep "\n" universalaccess}
         ${concatStringsSep "\n" ActivityMonitor}
         ${concatStringsSep "\n" CustomUserPreferences}
+
+        ${optionalString (length dock > 0) ''
+          echo >&2 "restarting dock..."
```
emilazy, line 112: nit:
```suggestion
          echo >&2 "restarting Dock..."
```

Enzime:
I'm not aware of any downsides to restarting the dock, most of the information/guides I've seen online just talk about restarting the dock, so I'm fairly sure it's a common thing to do.

With regard to other dynamic things, I'd be happy to merge them in in separate PRs, the less things that require a restart the better in my opinion

emilazy (approved):
The thing I was thinking of was `/System/Library/PrivateFrameworks/SystemAdministration.framework/Resources/activateSettings -u` (see https://github.com/LnL7/nix-darwin/issues/967); I don’t actually know what it does, but maybe we should investigate it. In particular, [this blog post about nix-darwin](https://medium.com/@zmre/nix-darwin-quick-tip-activate-your-preferences-f69942a93236) suggests that it can cause Dock settings to activate instantly; I’m fine merging this as‐is, but maybe we should test whether it could be replaced with that?

Enzime:
Just tested it, that command doesn't have any effect when changing the dock icons using `system.defaults.dock.persistent-apps`, I'm going to merge this PR as is, if we can figure out what settings that command actually makes a difference for then we can consider adding it later

PR #1001, opened 2024-07-10
services/yggdrasil: init

Rubikoid:
Add the yggdrasil module.

It's based on nixpkgs upstream yggdrasil module + launchd service from brew.


modules/services/yggdrasil.nix
```diff
@@ -0,0 +1,117 @@
...
+
+  settingsProvided = cfg.settings != { };
+  configFileProvided = cfg.configFile != null;
+
+  format = pkgs.formats.json { };
+in
+{
+  options = with types; {
+    services.yggdrasil = {
+      enable = mkEnableOption ("the yggdrasil system service");
```
Samasaur1, line 14: minor nit: the parentheses here are unnecessary

modules/services/yggdrasil.nix
```diff
@@ -0,0 +1,117 @@
...
+      configFile = mkOption {
+        type = nullOr path;
+        default = null;
+        example = "/run/keys/yggdrasil.conf";
+        description = lib.mdDoc ''
+          A file which contains JSON or HJSON configuration for yggdrasil. See
+          the {option}`settings` option for more information.
+
+          Note: This file must not be larger than 1 MB because it is passed to
+          the yggdrasil process via systemd‘s LoadCredential mechanism. For
```
Samasaur1, line 57: this isn't accurate on macOS, right? I would either remove this paragraph entirely or say something like

> On NixOS, this option is limited to 1M due to limitations in systemd. If you would like to share your yggdrasil configuration between nix-darwin and NixOS, you should keep this limitation in mind, even though there is no equivalent limit on macOS

Samasaur1 (changes requested):

Samasaur1 (approved):
I'm a little leery of putting the config file and example path under `/run` (as opposed to `/etc` or somewhere else), but this looks good to me

Samasaur1:
Sorry this took a couple days, the notification got buried. I don't have merge permissions anyway, so you'll be waiting for that anyway

Rubikoid:
> I'm a little leery of putting the config file and example path under /run (as opposed to /etc or somewhere else), but this looks good to me

How it seems to me, `/etc` is for something being static / symlinked to nix store, while `/run` is for something being generated.

Anyway, `/run/yggdrasil` also used for `yggdrasil.sock` controlling socket, so dropping config here quite logical.

Rubikoid:
Hi, @Enzime, can you have a quick look here?

modules/services/yggdrasil.nix
```diff
@@ -0,0 +1,117 @@
...
+with lib;
+let
+  cfg = config.services.yggdrasil;
+
+  settingsProvided = cfg.settings != { };
+  configFileProvided = cfg.configFile != null;
+
+  format = pkgs.formats.json { };
+in
+{
```
Enzime, line 11: Can you add yourself as a maintainer for this module?

modules/services/yggdrasil.nix
```diff
@@ -0,0 +1,117 @@
...
+          are supplied, they will be combined, with values from
+          {option}`configFile` taking precedence.
+
+          You can use the command `nix-shell -p yggdrasil --run "yggdrasil -genconf"`
+          to generate default configuration values with documentation.
+        '';
+      };
+
+      configFile = mkOption {
+        type = nullOr path;
```
Enzime, line 49: From what I remember there are issues with the way `types.path` works, I think it's probably better to just use `types.str`

modules/services/yggdrasil.nix
```diff
@@ -0,0 +1,117 @@
...
+              "${binYggdrasil} -genconf") + " > ${yggdrasilConf}"}
+
+            # start yggdrasil
+            ${binYggdrasil} -useconffile ${yggdrasilConf} ${lib.strings.escapeShellArgs cfg.extraArgs}
+          '';
+
+          serviceConfig = {
+            ProcessType = "Interactive";
+            StandardOutPath = "/tmp/yggdrasil.stdout.log";
+            StandardErrorPath = "/tmp/yggdrasil.stderr.log";
```
Enzime, line 110:
```suggestion
```

Generally we don't set `StandardOutPath` and `StandardErrorPath` by default, also I'm not sure if putting it in `/tmp` is a good idea as that gets wiped every 3 days and I don't think it'll create the file again after it gets wiped

modules/services/yggdrasil.nix
```diff
@@ -0,0 +1,117 @@
...
+            ${(if settingsProvided || configFileProvided then
+              "echo "
+
+              + (lib.optionalString settingsProvided
+                "'${builtins.toJSON cfg.settings}'")
+              + (lib.optionalString configFileProvided
+                "$(${binHjson} -c ${cfg.configFile})")
+              + " | ${pkgs.jq}/bin/jq -s add | ${binYggdrasil} -normaliseconf -useconf"
+            else
+              "${binYggdrasil} -genconf") + " > ${yggdrasilConf}"}
```
Enzime, line 101: What happens if the config file already exists?

Enzime (changes requested):

modules/services/yggdrasil.nix
Rubikoid: I did not contributed to nixpkgs yet, so I have no `lib.maintainers` record, so I leaved simple string with my username, is this okay?

modules/services/yggdrasil.nix
Rubikoid: According to [types.nix#L482](https://github.com/NixOS/nixpkgs/blob/468356287fe4ce68f58fa5cf16e8ed6fc40a1a7f/lib/types.nix#L482), the only differences from `types.str` to `types.path` is check for leading `/` and `isStringLike`, i.e. to `types.path` option you can pass `str` / `path` / something with `outPath`.

As how I see, here is reasonable.

I'm tried to google any problems with `types.path` and found only problems with nix's internal path type, not with `types.path` ;(

modules/services/yggdrasil.nix
Rubikoid: Moved this to separate option `logFile`.

modules/services/yggdrasil.nix
Rubikoid: Thanks for pointing this!

In the upstream ygg module this was expected behavior since it has `persistentKeys` option.

I changed this to persist config between service restarts, if it already exists.

modules/services/yggdrasil.nix
Enzime: Can you change it to `maintainers.rubikoid or "rubikoid"` so if you ever end up maintaining packages in Nixpkgs it'll still refer to you

Enzime (commented):
Could you also add a test for this module?

r4v3n6101:
Is the PR alive? I still want this one.

PR #505, opened 2022-08-26
Add Stage Manager (com.apple.WindowManager) settings as system.defaults.windowmanager

AlexOwl:

PR #1002, opened 2024-07-11
(feature) Add Stage Manager (com.apple.WindowManager) settings

malko42:
All credits go to @AlexOwl. Their [PR](https://github.com/LnL7/nix-darwin/pull/505) looked abandonned so I reported their changes and addressed the change requests.

### DIFF
--- a/modules/system/defaults-write.nix
+++ b/modules/system/defaults-write.nix
@@ -32,6 +32,7 @@ let
   trackpadBluetooth = defaultsToList "com.apple.driver.AppleBluetoothMultitouch.trackpad" cfg.trackpad;
   universalaccess = defaultsToList "com.apple.universalaccess" cfg.universalaccess;
   ActivityMonitor = defaultsToList "com.apple.ActivityMonitor" cfg.ActivityMonitor;
+  WindowManager = defaultsToList "com.apple.WindowManager" cfg.WindowManager;
   CustomUserPreferences = flatten (mapAttrsToList (name: value: defaultsToList name value) cfg.CustomUserPreferences);
   CustomSystemPreferences = flatten (mapAttrsToList (name: value: defaultsToList name value) cfg.CustomSystemPreferences);
 
@@ -85,6 +86,7 @@ in
         universalaccess
         ActivityMonitor
         CustomUserPreferences
+        WindowManager
       ]
       ''
         # Set defaults
@@ -107,6 +109,7 @@ in
         ${concatStringsSep "\n" universalaccess}
         ${concatStringsSep "\n" ActivityMonitor}
         ${concatStringsSep "\n" CustomUserPreferences}
+        ${concatStringsSep "\n" WindowManager}
 
         ${optionalString (length dock > 0) ''
           # Only restart Dock if current user is logged in
--- a/modules/system/defaults/WindowManager.nix
+++ b/modules/system/defaults/WindowManager.nix
@@ -1,0 +1,76 @@
+{ config, lib, ... }:
+
+with lib;
+{
+  options = {
+    system.defaults.WindowManager.GloballyEnabled = mkOption {
+      type = types.nullOr types.bool;
+      default = null;
+      description = ''
+        Enable Stage Manager
+        Stage Manager arranges your recent windows into a single strip for reduced clutter and quick access. Default is false.
+      '';
+    };
+
+    system.defaults.WindowManager.EnableStandardClickToShowDesktop = mkOption {
+      type = types.nullOr types.bool;
+      default = null;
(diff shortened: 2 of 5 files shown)


Samasaur1 (approved):
Assuming that you've tested it and that those are still the correct defaults keys to use for those settings, this looks good to me

malko42:
Hi @Samasaur1 and thanks for the review!
If you compare this PR with the orginal one the options are not the same. I added and tested only the options that I was interested in. I'm not sure about the default values though I took my own system as reference since I haven't touched those options AFAIK but still not 100% sure.
Is there an up-to-date list somewhere listing all `com.apple.*` settings ?

Also there are a few commonly defined options in [Nix config files](https://github.com/ryan4yin/nix-config/blob/1d66cf655daab10d1cef80897d7521046e5f8241/modules/darwin/system.nix#L126) do you want me to add them in a later PR or on this one ?

Samasaur1:
> Hi @Samasaur1 and thanks for the review!
> 
> If you compare this PR with the orginal one the options are not the same. I added and tested only the options that I was interested in. I'm not sure about the default values though I took my own system as reference since I haven't touched those options AFAIK but still not 100% sure.

I think you're right about the default values. I also don't think it _really_ matters all too much because I think if people are looking at these options, they're probably planning to set them anyway.

I did look again at your `EnableStandardClickToShowDesktop` option, and I think you should update the docs you wrote for it to say that `false` means "Only in Stage Manager" and `true` means Always

> Is there an up-to-date list somewhere listing all `com.apple.*` settings ?

Wouldn't that be nice! I wish. I know there's <https://macos-defaults.com/> and maybe other sites like it, but I don't think they're comprehensive

> Also there are a few commonly defined options in [Nix config files](https://github.com/ryan4yin/nix-config/blob/1d66cf655daab10d1cef80897d7521046e5f8241/modules/darwin/system.nix#L126) do you want me to add them in a later PR or on this one ?

I'd say you should feel free to add any options under `com.apple.WindowManager` in this PR, but other defaults options should have their own PRs

modules/system/defaults/WindowManager.nix
```diff
@@ -0,0 +1,27 @@
+{ config, lib, ... }:
+
+with lib;
+{
+  options = {
+    system.defaults.WindowManager.GloballyEnabled = mkOption {
+      type = types.nullOr types.bool;
+      default = null;
+      description = ''
+        Apple menu > System Settings > Dekstop and Dock > Desktop and Stage Manager
```
Enzime, line 10:
```suggestion
```

modules/system/defaults/WindowManager.nix
```diff
@@ -0,0 +1,27 @@
...
+        Stage Manager arranges your recent windows into a single strip for reduced clutter and quick access. Default is false.
+      '';
+    };
+
+    system.defaults.WindowManager.EnableStandardClickToShowDesktop = mkOption {
+      type = types.nullOr types.bool;
+      default = null;
+      description = ''
+        Apple menu > System Settings > Dekstop and Dock > Desktop and Stage Manager
```
Enzime, line 19:
```suggestion
```

Enzime (changes requested):
Can you rebase your branch on top of master and squash all your commits into a singular one?

malko42:
Few things to note:
- I added all the options I could effectively test
- I named the options after the ones from Apple but I'm not happy with it. I find them at best non explicit and most of the time logically opposite of their UI counterpart

modules/system/defaults-write.nix
```diff
@@ -107,7 +109,7 @@ in
         ${concatStringsSep "\n" universalaccess}
         ${concatStringsSep "\n" ActivityMonitor}
         ${concatStringsSep "\n" CustomUserPreferences}
-
+        ${concatStringsSep "\n" WindowManager}
```
Enzime, line 112:
```suggestion
        ${concatStringsSep "\n" WindowManager}

```

Enzime (changes requested):
Can you update the tests?

https://github.com/LnL7/nix-darwin/blob/master/tests/fixtures/system-defaults-write/activate-user.txt
https://github.com/LnL7/nix-darwin/blob/master/tests/system-defaults-write.nix

malko42:
> Can you update the tests?
> 
> https://github.com/LnL7/nix-darwin/blob/master/tests/fixtures/system-defaults-write/activate-user.txt https://github.com/LnL7/nix-darwin/blob/master/tests/system-defaults-write.nix

Done

Enzime (approved):
LGTM

Thanks for the contribution

Issue #982, opened 2024-06-22
Sorting in persistent-others


nklmilojevic:
Is there a way to sort and set a content as fan/stack/whatever with nix-darwin? Each time I apply settings, it resets to defaults:

![CleanShot 2024-06-22 at 12 05 32@2x](https://github.com/LnL7/nix-darwin/assets/9573356/26732231-b82d-4d89-8505-4502db2f55a5)

nklmilojevic:
For example, if I apply it manually like this it allow setting options:

```

defaults write com.apple.dock persistent-others -array '<dict>
    <key>tile-data</key>
    <dict>
        <key>arrangement</key>
        <integer>2</integer>
        <key>displayas</key>
        <integer>1</integer>
        <key>file-data</key>
        <dict>
            <key>_CFURLString</key>
            <string>file:///Users/nkl/Documents</string>
            <key>_CFURLStringType</key>
            <integer>15</integer>
        </dict>
        <key>file-type</key>
        <integer>2</integer>
        <key>showas</key>
        <integer>1</integer>
    </dict>
    <key>tile-type</key>
    <string>directory-tile</string>
</dict>'
```

nick-skriabin:
curious as well. the docs don't say anything about it, so is it safe to assume that there's no such functionality atm?

PR #1004, opened 2024-07-12
Allow more configuration for persistent-others in dock

szethh:
See #982, currently `system.defaults.dock.persistent-others` can only take paths to folders to display in the dock.
From each path, [a more complex object is constructed](https://github.com/LnL7/nix-darwin/blob/cf297a8d248db6a455b60133f6c0029c04ebe50e/modules/system/defaults/dock.nix#L139), with the actual settings for the folder. This PR is to allow users to directly define that object.

This allows users to define the sorting criterion, and multiple other display changes.

You would use it like so:

```nix
# darwin.nix
system.defaults.dock.persistent-others = [
      {
        "tile-data" = {
          arrangement = 3; # 1 = name, 2 = date added, 3 = date modified, 4 = Date Created, 5 = kind; default is 1
          displayas = 0;  # 0 = stack, 1 = folder; default is 0
          "file-data" = {
            "_CFURLString" = "file:///Users/szeth/Downloads";  # the path to the folder
            "_CFURLStringType" = 15;
          };
          #"file-label" = "Downloads";  # default is the name of the directory
          showas = 1; # 1 = fan, 2 = grid, 3 = list, 4 = automatic; default is 4
        };
        "tile-type" = "directory-tile";  # either file-tile or directory-tile
      }
];
```
(this works fine on my dock)

The changes I made simply widen the type (to accept `types.attrsOf types.anything` as well), and then simply pass the object through.

Now my initial implementation is very simple, and since I'm quite new to Nix and to this repo, I don't know what the best way to approach this is:

1. We leave the behavior as-is (ie just pass objects through), and document each option and their possible values/defaults.
2. We create a submodule, to have proper type checking. No idea how to do this properly.
Note that there are some properties that should still be generated for you, like `_CFURLStringType` and `tile-type`.


szethh:
@emilazy hi, pinging maintainers to ask whether any changes are needed to this? or any clarification?
i see from the reactions to the pr some people might be interested in this feature as well :)

emilazy:
Sorry for not getting to this earlier. Can `system.defaults.CustomUserPreferences` not be used for this advanced case? I guess maybe not because it wouldn’t let you mix normal paths with the fancier stuff? In general it feels a bit awkward to expose an undocumented/probably not entirely stable defaults structure without any type‐checking as a supported interface, but that’s somewhat of a pervasive problem already. The commit message should be e.g. `system/defaults/dock: support more configuration for persistent-others`.

szethh:
I agree that having this without typechecking is a bit hacky. I tried setting that up with the submodule type but I must be missing something because I could not get it to work :/

maybe someone with more nix knowhow can take a shot at it, for now i'm just updating the commit message as requested and leaving it without typechecks

szethh:
bump :)

sadoMasupilami:
Hi, i would also be super interested in this feature! @emilazy what would you need to be able to merge this?

sadoMasupilami:
anyone struggling as well, it can be done like this but than you can not use the persistent-others anymore:

```nix
    CustomUserPreferences = {
      # Sets Downloads folder with fan view in Dock
      "com.apple.dock" = {
        persistent-others = [
          {
            "tile-data" = {
              "file-data" = {
                "_CFURLString" = "/Users/<youruser>/Downloads";
                "_CFURLStringType" = 0;
              };
              "arrangement" = 2;  # sorting order
              "displayas" = 1;    # 1 for fan display
              "showas" = 1;       # 1 for stack view
            };
            "tile-type" = "directory-tile";
          }
          {
            "tile-data" = {
              "file-data" = {
                "_CFURLString" = "/Applications";
                "_CFURLStringType" = 0;
              };
            };
            "tile-type" = "directory-tile";
          }
        ];
      };
    };
``` 
I don't know if it is a goof solution but it works

Samasaur1:
Superseded by #1431

PR #1012, opened 2024-07-23
defaults: only restart Dock when user is logged in

Enzime:
This can occur if your primary user is not logged in and you try to do a `darwin-rebuild switch`

### DIFF
--- a/modules/system/defaults-write.nix
+++ b/modules/system/defaults-write.nix
@@ -109,8 +109,11 @@ in
         ${concatStringsSep "\n" CustomUserPreferences}
 
         ${optionalString (length dock > 0) ''
-          echo >&2 "restarting Dock..."
-          killall Dock
+          # Only restart Dock if current user is logged in
+          if pgrep -xu $UID Dock; then
+            echo >&2 "restarting Dock..."
+            killall Dock || true
+          fi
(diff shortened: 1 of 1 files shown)


modules/system/defaults-write.nix
```diff
@@ -109,8 +109,11 @@ in
         ${concatStringsSep "\n" CustomUserPreferences}
 
         ${optionalString (length dock > 0) ''
-          echo >&2 "restarting Dock..."
-          killall Dock
+          # Only restart Dock if current user is logged in
+          if pgrep -xu $UID Dock; then
```
emilazy, line 113: Shouldn’t need this with the `|| true`.

emilazy (changes requested):

modules/system/defaults-write.nix
Enzime: I would rather leave it in as it makes the intent of this code more clear (the prerequisite of restarting the Dock is when it is running as the current user)

emilazy: Alright; it’ll be a little awkward in a post‐user‐activation world but we can cross that bridge when we come to it.

emilazy (approved):

PR #1013, opened 2024-07-23
github-runnners: fix workDir missing on reboot

Enzime:
As `/var/run` gets cleared on every reboot, all jobs will fail with `Access to the path '/var/run/github-runners/<name>' is denied.`


emilazy (approved):
Odd that we don’t use `/run` here like NixOS, but if “Changing this option triggers a new runner registration” I suppose it’s not worth fixing.

PR #1018, opened 2024-07-27
github-runners: move `workDir` outside of `/run`

Enzime:
Turns out I messed up when testing #1013 and forgot to reboot when testing and the PR didn't actually work.

As `/run` gets recreated every reboot and we can't specify dependencies for launchd, creating the `workDir` every reboot will require extra complexity with a separate daemon that runs as `root` otherwise it won't have sufficient privileges.

As we clean the `workDir` when the service first starts anyway, it ends up being the same.

### DIFF
--- a/modules/services/github-runner/options.nix
+++ b/modules/services/github-runner/options.nix
@@ -22,12 +22,12 @@ with lib;
 
       * `/var/lib/github-runners/<name>`:
         State directory to store the runner registration credentials
+      * `/var/lib/github-runners/_work/<name>`:
+        Working directory for workflow files. The runner only uses this
+        directory if `workDir` is `null` (see the `workDir` option for details).
       * `/var/log/github-runners/<name>`:
         The launchd service writes the stdout and stderr streams to this
         directory.
-      * `/var/run/github-runners/<name>`:
-        Working directory for workflow files. The runner only uses this
-        directory if `workDir` is `null` (see the `workDir` option for details).
     '';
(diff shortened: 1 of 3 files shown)


Enzime:
Marking as draft until I've finished testing this properly this time :)

Enzime:
Currently, changing `workDir` won't have any effect on existing setups that don't have `ephemeral` enabled.

https://github.com/LnL7/nix-darwin/blob/0413754b3cdb879ba14f6e96915e5fdf06c6aab6/modules/services/github-runner/service.nix#L128-L137

This is because the logic in NixOS that ensures runner registration is rerun if any configuration changes hasn't been ported yet:

https://github.com/NixOS/nixpkgs/blob/6d8391a3ce154bdf1870d998f187b26de8147065/nixos/modules/services/continuous-integration/github-runner/service.nix#L123-L140

My previous PR makes setting up new GitHub Runners broken as the configure script won't be run with sufficient permissions to create the work directory. This PR will fix setting up new GitHub Runners by changing the default directory and restoring the old logic of using the `launchd` activation script to create the folder, meaning new setups won't encounter this error.

As I'm short on time at the moment, migrating the logic to rerun runner registration on configuration changes will have to wait for another PR.

As a manual workaround, anyone with an existing setup can delete the file at `/var/lib/github-runners/<name>/.runner` to force runner registration.

emilazy (approved):

Issue #985, opened 2024-06-28
Option to enable touch ID authentication for sudo doesn't work in tmux


347Online:
I really love the out-of-the-box option to enable touch ID for sudo and was surprised to find it doesn't work in tmux

I eventually stumbled across [this](https://github.com/fabianishere/pam_reattach), which is working for me, but I know it's going to get blown away when macOS does an update

I spent awhile noodling around with getting this working in my personal config to no avail, but it'd be even better to just have this as an option in nix-darwin — programs.pam-reattach.enable perhaps

emilazy:
Yeah, finally getting this in with the rework of the sudo PAM stuff is on my to‐do list. See also: https://github.com/LnL7/nix-darwin/pull/787, https://github.com/LnL7/nix-darwin/pull/662, https://github.com/LnL7/nix-darwin/pull/606, https://github.com/LnL7/nix-darwin/pull/591. (This is probably the area of nix-darwin that has evolved the slowest for how many PRs it’s had…)

My current plan is as outlined in https://github.com/LnL7/nix-darwin/pull/787#issuecomment-2165246619.

347Online:
Ahh, thank you that's fairly illuminating! My apologies if I've created extra noise here, for whatever reason I was not able to turn up any of these prior conversations when I searched

emilazy:
No worries, I just wish I’d made progress on this sooner!

gshpychka:
FWIW this works for me: https://github.com/gshpychka/dotfiles/blob/140600b6b4cec63b36f9a94d04fcd6e2f65251ee/machines/eve/touch-id.nix

You don't need `pam_watchid` - I believe that was due to a quirk of pam_tid not working with a closed lid

michaelvanstraten:
> FWIW this works for me: https://github.com/gshpychka/dotfiles/blob/140600b6b4cec63b36f9a94d04fcd6e2f65251ee/machines/eve/touch-id.nix
> 
> You don't need `pam_watchid` - I believe that was due to a quirk of pam_tid not working with a closed lid

This bricked my system, I had to reboot into recovery mode to undo this change.

Enzime:
Closed by #1344

PR #1020, opened 2024-07-27
fix: move pam configuration to sudo_local

dlubawy:
Addresses #985 and #787.

* Moves pam configuration to `/etc/pam.d/sudo_local`
* Performs configuration through `environment.etc`
* Adds a `pam_reattach` option to fix sudo TouchID in tmux

Implementation uses `environment.etc` to create the `/etc/pam.d/sudo_local` file and adds the `pkgs.pam-reattach` option provided in #662. Follows [the comment](https://github.com/LnL7/nix-darwin/pull/787#issuecomment-2165246619) by @emilazy to have nix-darwin manage the file entirely. If the file exists already, nix-darwin should handle it through the usual warning telling the user to rename the file to `sudo_local.before-nix-darwin`. As identified by @lilyball in [their comment](https://github.com/LnL7/nix-darwin/pull/787#issuecomment-1742376938), the symlink approach here shouldn't impact users since it doesn't touch the main `/etc/pam.d/sudo` file. As long as `/etc/pam.d/sudo` remains a regular file than this should work fine without disrupting sudo to apply `nix-darwin` configurations.

I recognize this may be a duplicate PR given all the other open issues/PRs, but I haven't seen movement on them in a while. Feel free to close this if it's unnecessary.

### DIFF
--- a/modules/security/pam.nix
+++ b/modules/security/pam.nix
@@ -37,11 +37,11 @@ let
   ''
     ${if isEnabled then ''
       if ! grep '${option}' ${file} > /dev/null; then
-        ${sed} -i '2iauth       sufficient     pam_tid.so # nix-darwin: ${option}' ${file}
+        ${sed} -i 's/#\(auth\s\+sufficient\s\+pam_tid.so\)/\1 # nix-darwin: ${option}/g' ${file}
       fi
     '' else ''
       if grep '${option}' ${file} > /dev/null; then
-        ${sed} -i '/${option}/d' ${file}
+        ${sed} -i 's/\(.*\)#.*${option}/#\1/g' ${file}
       fi
     ''}
   '';
--- a/modules/security/pam.nix
+++ b/modules/security/pam.nix
@@ -49,9 +49,13 @@ let
   let
     option = "security.pam.enablePamReattach";
   in
+  # NOTE: needs to check that `pkgs.pam-reattach` is in the line in case the store location updates
   ''
     ${if isEnabled then ''
-      if ! grep '${option}' ${file} > /dev/null; then
+      if ! grep '${pkgs.pam-reattach}' ${file} > /dev/null; then
+        if grep '${option}' ${file} > /dev/null; then
+          ${sed} -i '/${option}/d' ${file}
+        fi
         ${sed} -i '1iauth optional ${pkgs.pam-reattach}/lib/pam/pam_reattach.so # nix-darwin: ${option}' ${file}
       fi
     '' else ''
--- a/modules/security/pam.nix
+++ b/modules/security/pam.nix
@@ -61,7 +61,7 @@ in
 
   config =
   let
-    isPamEnabled = (cfg.enableSudoTouchIdAuth || cft.enablePamReattach);
+    isPamEnabled = (cfg.enableSudoTouchIdAuth || cfg.enablePamReattach);
   in
   {
     environment.etc."pam.d/sudo_local" = {
--- a/modules/security/pam.nix
+++ b/modules/security/pam.nix
@@ -22,6 +22,10 @@ let
     ${if isEnabled then ''
       # Check if include line is needed (macOS < 14)
       if ! grep 'sudo_local' ${file} > /dev/null; then
+        # Clear out older implementation
+        if grep '${option}' ${file} > /dev/null; then
+          ${sed} -i '/${option}/d' ${file}
+        fi
         ${sed} -i '2iauth       include        sudo_local # nix-darwin: ${option}' ${file}
       fi
     '' else ''
--- a/modules/security/pam.nix
+++ b/modules/security/pam.nix
@@ -4,6 +4,33 @@ with lib;
 
 let
   cfg = config.security.pam;
+
+  # Implementation Notes
+  #
+  # Uses `environment.etc` to create the `/etc/pam.d/sudo_local` file that will be used
+  # to manage all things pam related for nix-darwin. An activation script will run to check
+  # for the existance of the line `auth       include        sudo_local`. This is included
+  # in macOS Sonoma and later. If the line is not there already then `sed` will add it.
+  # In those cases, the line will include the name of the option root (`security.pam`),
+  # to make it easier to identify the line that should be deleted when the option is disabled.
+  mkIncludeSudoLocalScript = isEnabled:
+  let
+    file = "/etc/pam.d/sudo";
+    option = "security.pam";
+    sed = "${pkgs.gnused}/bin/sed";
+  in ''
+    ${if isEnabled then ''
+      # Check if include line is needed (macOS < 14)
+      if ! grep 'sudo_local' ${file} > /dev/null; then
+        ${sed} -i '2iauth       include        sudo_local # nix-darwin: ${option}' ${file}
+      fi
+    '' else ''
+      # Remove include line if we added it
+      if grep '${option}' ${file} > /dev/null; then
+        ${sed} -i '/${option}/d' ${file}
+      fi
+    ''}
+  '';
 in
 {
   options = {
@@ -32,7 +59,11 @@ in
     };
   };
 
-  config = {
+  config =
+  let
+    isPamEnabled = (cfg.enableSudoTouchIdAuth || cft.enablePamReattach);
+  in
+  {
     environment.etc."pam.d/sudo_local" = {
       enable = (cfg.enablePamReattach || cfg.enableSudoTouchIdAuth);
       text = lib.strings.concatStringsSep "\n" [
@@ -40,5 +71,10 @@ in
         (lib.optionalString cfg.enableSudoTouchIdAuth "auth       sufficient     pam_tid.so")
       ];
     };
+    system.activationScripts.pam.text = ''
+      # PAM settings
+      echo >&2 "setting up pam..."
+      ${mkIncludeSudoLocalScript isPamEnabled}
+    '';
   };
 }
--- a/modules/security/pam.nix
+++ b/modules/security/pam.nix
@@ -17,15 +17,17 @@ let
   let
     file = "/etc/pam.d/sudo";
     option = "security.pam";
+    deprecatedOption = "security.pam.enableSudoTouchIdAuth";
     sed = "${pkgs.gnused}/bin/sed";
   in ''
     ${if isEnabled then ''
+      # NOTE: this can be removed at some point when support for older versions are dropped
+      # Always clear out older implementation if it exists
+      if grep '${deprecatedOption}' ${file} > /dev/null; then
+        ${sed} -i '/${option}/d' ${file}
+      fi
       # Check if include line is needed (macOS < 14)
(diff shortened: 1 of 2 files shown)


emilazy:
Thanks! This is definitely the direction I want to go, but we need to keep support for macOS < 14. That means that ensuring the `auth       include        sudo_local` line is added to `/etc/pam.d/sudo` if it’s not present (and cleaning up after the old `# nix-darwin: security.pam.enableSudoTouchIdAuth` method). So we still need to do some amount of file wrangling here, but the upside is that users of Sonoma and higher won’t have to re‐apply their configuration after every update.

dlubawy:
> Thanks! This is definitely the direction I want to go, but we need to keep support for macOS < 14. That means that ensuring the `auth include sudo_local` line is added to `/etc/pam.d/sudo` if it’s not present (and cleaning up after the old `# nix-darwin: security.pam.enableSudoTouchIdAuth` method). So we still need to do some amount of file wrangling here, but the upside is that users of Sonoma and higher won’t have to re‐apply their configuration after every update.

Good call on the support for older versions and older nix-darwin support. What do you think of the recent changes? I added back an activation script like the old one which will use `grep` to check if `sudo_local` is already in the `/etc/pam.d/sudo` file. If it's not there, it then checks for the old implementation and deletes it using `sed -i '/security.pam.enableSudoTouchIdAuth/d'`. Then it will add in this line using `sed -i '2i....` if any option in `security.pam` is set:
```
auth       include        sudo_local # nix-darwin: security.pam
```
Otherwise, if all `security.pam` options are disabled, then it performs `sed -i '/security.pam/d' /etc/pam.d/sudo` which should also act to disable it for older implementations as well. Then if enabled again they will just have the single line for including `sudo_local`.

For users on macOS >=14, the entire thing will just be managed through `environment.etc` as the `/etc/pam.d/sudo` file will already include `sudo_local`. Hopefully, there will be a point where we can remove the activation script entirely as older version support is eventually dropped.

modules/security/pam.nix
```diff
@@ -7,63 +7,80 @@ let
...
 
-  config = {
+  config =
+  let
+    isPamEnabled = (cfg.enableSudoTouchIdAuth || cfg.enablePamReattach);
+  in
+  {
+    environment.etc."pam.d/sudo_local" = {
+      enable = isPamEnabled;
+      text = lib.strings.concatStringsSep "\n" [
```
Enzime, line 75:
```suggestion
      text = lib.concatLines [
```

modules/security/pam.nix
```diff
@@ -7,63 +7,80 @@ let
...
+        (lib.optionalString cfg.enablePamReattach "auth       optional       ${pkgs.pam-reattach}/lib/pam/pam_reattach.so")
+        (lib.optionalString cfg.enableSudoTouchIdAuth "auth       sufficient     pam_tid.so")
```
Enzime, line 77:
```suggestion
        (lib.mkIf cfg.enablePamReattach "auth       optional       ${pkgs.pam-reattach}/lib/pam/pam_reattach.so")
        (lib.mkIf cfg.enableSudoTouchIdAuth "auth       sufficient     pam_tid.so")
```

I think this will mean the output file won't have blank lines in it

modules/security/pam.nix
```diff
@@ -7,63 +7,80 @@ let
...
     ${if isEnabled then ''
-      # Enable sudo Touch ID authentication, if not already enabled
-      if ! grep 'pam_tid.so' ${file} > /dev/null; then
-        ${sed} -i '2i\
-      auth       sufficient     pam_tid.so # nix-darwin: ${option}
-        ' ${file}
+      # NOTE: this can be removed at some point when support for older versions are dropped
+      # Always clear out older implementation if it exists
+      if grep '${deprecatedOption}' ${file} > /dev/null; then
+        ${sed} -i '/${option}/d' ${file}
```
Enzime, line 27:
```suggestion
        ${sed} -i '/${deprecatedOption}/d' ${file}
```

modules/security/pam.nix
```diff
@@ -7,63 +7,80 @@ let
...
+  # for the existance of the line `auth       include        sudo_local`. This is included
+  # in macOS Sonoma and later. If the line is not there already then `sed` will add it.
+  # In those cases, the line will include the name of the option root (`security.pam`),
+  # to make it easier to identify the line that should be deleted when the option is disabled.
+  mkIncludeSudoLocalScript = isEnabled:
   let
-    file   = "/etc/pam.d/sudo";
-    option = "security.pam.enableSudoTouchIdAuth";
+    file = "/etc/pam.d/sudo";
+    option = "security.pam";
```
Enzime, line 19:
```suggestion
    marker = "security.pam.sudo_local";
```

modules/security/pam.nix
```diff
@@ -7,63 +7,80 @@ let
...
-      auth       sufficient     pam_tid.so # nix-darwin: ${option}
-        ' ${file}
+      # NOTE: this can be removed at some point when support for older versions are dropped
+      # Always clear out older implementation if it exists
+      if grep '${deprecatedOption}' ${file} > /dev/null; then
+        ${sed} -i '/${option}/d' ${file}
+      fi
+      # Check if include line is needed (macOS < 14)
+      if ! grep 'sudo_local' ${file} > /dev/null; then
+        ${sed} -i '2iauth       include        sudo_local # nix-darwin: ${option}' ${file}
```
Enzime, line 31:
```suggestion
        ${sed} -i '2iauth       include        sudo_local # nix-darwin: ${marker}' ${file}
```

modules/security/pam.nix
```diff
@@ -7,63 +7,80 @@ let
...
-  # As such, we resort to line addition/deletion in place using `sed`. We add a comment to the
-  # added line that includes the name of the option, to make it easier to identify the line that
-  # should be deleted when the option is disabled.
-  mkSudoTouchIdAuthScript = isEnabled:
+  # Uses `environment.etc` to create the `/etc/pam.d/sudo_local` file that will be used
+  # to manage all things pam related for nix-darwin. An activation script will run to check
+  # for the existance of the line `auth       include        sudo_local`. This is included
+  # in macOS Sonoma and later. If the line is not there already then `sed` will add it.
+  # In those cases, the line will include the name of the option root (`security.pam`),
+  # to make it easier to identify the line that should be deleted when the option is disabled.
```
Enzime, line 15: It would be good to add a comment about what happens to the `marker` line in `/etc/pam.d/sudo` when upgrading to Sonoma

modules/security/pam.nix
```diff
@@ -7,63 +7,80 @@ let
...
+  mkIncludeSudoLocalScript = isEnabled:
```
Enzime, line 16:
```suggestion
```

This being a function seems unnecessary

Enzime (changes requested):
Sorry, it took so long to getting around to reviewing this PR

Let us know if you're still interested on working on this PR

modules/security/pam.nix
```diff
@@ -7,63 +7,80 @@ let
...
+    deprecatedOption = "security.pam.enableSudoTouchIdAuth";
     sed = "${pkgs.gnused}/bin/sed";
   in ''
     ${if isEnabled then ''
-      # Enable sudo Touch ID authentication, if not already enabled
-      if ! grep 'pam_tid.so' ${file} > /dev/null; then
-        ${sed} -i '2i\
-      auth       sufficient     pam_tid.so # nix-darwin: ${option}
-        ' ${file}
+      # NOTE: this can be removed at some point when support for older versions are dropped
```
Enzime, line 24:
```suggestion
      # REMOVEME when macOS 13 no longer supported
```

emilazy:
I think we can skip the `isPamEnabled` and just unconditionally apply this stuff too, since it’s harmless to have an empty file.

Enzime:
The benefit of the marker would be that if a user uninstalled nix-darwin on a pre-Sonoma version we'd be able to restore it to closer to the original state, but either way is fine as `auth include sudo_local` still supports `sudo_local` not existing

As `environment.etc.<file>.copy` doesn't exist, that means `/etc/pam.d/sudo_local` will be a symlink to a path in the `/nix/store`, I wonder if this is going to lead to some race conditions with FDE

emilazy:
I think not being able to use touch ID very early in the user session is an acceptable failure mode if that’s all it is.

dlubawy:
@Enzime, thanks for taking the time to review. I should still be able to address the points you made and will request a re-review when I get that done.

modules/security/pam.nix
dlubawy: The intention of using the root option name as the marker was to catch both the new option and old option in the one `if` condition on line 35 for when the option is disabled. This is to prevent the edge case where someone has evaluated the config, upgraded to the new version, and then disabled the `security.pam.enableSudoTouchIdAuth` option before evaluating it again. We need to remove any line containing `security.pam` as a precaution against this.

modules/security/pam.nix
dlubawy: Need to use `lib.optional` instead because `lib.mkIf` produces a set. By using `lib.optional` we can construct a list by appending the result of the conditional list whereby a false condition appends the empty list equating to a NOP.

dlubawy:
I made the changes suggested by @Enzime excluding a few like the marker change which was deliberate to allow for a universal check on any config defined values when disabled to ensure everything is deleted in certain edge cases.

@emilazy I chose not to remove the `isPamEnabled` because a user may want to retain control over the `sudo_local` file. If we remove it then we are enforcing that file to either be blank or only the one/two lines described in the config. This may not be desirable for some.

Thank you all for the suggestions and sorry for taking a bit long to address them.

emilazy (commented):
> I made the changes suggested by @Enzime excluding a few like the marker change which was deliberate to allow for a universal check on any config defined values when disabled to ensure everything is deleted in certain edge cases.
> 
> @emilazy I chose not to remove the `isPamEnabled` because a user may want to retain control over the `sudo_local` file. If we remove it then we are enforcing that file to either be blank or only the one/two lines described in the config. This may not be desirable for some.

I personally think we can consider a user not wanting `sudo_local` to be managed out of scope, since nix-darwin takes ownership of a bunch of stuff (e.g. `/etc/shells`) already. That said, if we want it to be a toggle I would prefer it to be an on‐by‐default `security.pam.enable` rather than inferred based on the sub‐options.

We could add e.g. `security.pam.extraSudoLocalConfig` to allow an open‐ended customization point, too.

> Thank you all for the suggestions and sorry for taking a bit long to address them.

Sorry for taking a while to review :)

Other than that this LGTM!

emilazy:
This is still pending on https://github.com/LnL7/nix-darwin/pull/1020#pullrequestreview-2538317670. I regret how long it’s taken to get this merged, though, and if @dlubawy doesn’t have the time maybe someone else could take over to bring it across the finish line?

Mic92:
Follow up: https://github.com/LnL7/nix-darwin/pull/1344

Enzime:
Superseded by #1344

PR #991, opened 2024-07-02
(feature) Add swapLeftCtrlAndFn

malko42:
Use this and never find yourself again hitting fn because of muscle memory! (you can even physically swap the keycaps, at least on M series)

Keycodes have been pulled from https://hidutil-generator.netlify.app/ and the hex value has been converted to a base 10 int.

PR #1023, opened 2024-08-01
(feature) Add swapLeftCtrlAndFn

malko42:
Use this and never find yourself again hitting fn because of muscle memory! (you can even physically swap the keycaps, at least on M series)

Keycodes have been pulled from https://hidutil-generator.netlify.app/ and the hex value has been converted to a base 10 int.

(was previously opened as #991 but made a mistake)


Enzime (approved):
LGTM

Thanks for the contribution

PR #1025, opened 2024-08-03
added option `programs.bash.promptInit`

christian-burger:
Option is identical to NixOS option. Default value copied from there.

Motivation: I know `zsh` is default in macOS. But every time you enter `nix develop`, there is only a rudimentary prompt by default. Plus, on older MacBooks some people might not have transitioned to `zsh`, yet. Both `zsh` as well as `fish` provide this option as well.

Only problem is: This might break things for people using `programs.bash.interactiveShellInit` or `environment.interactiveShellInit` to configure their prompt, because it gets overwritten as I imitated the way it is implemented in the `zsh` module. I could, alternatively to the way it is done now,

1. move it up, so that it is executed during shell initialization before said option. Or 
2. provide a warning in the module, that alerts the user of the `interactiveShellInit` option to the change the next time they run `darwin-rebuild`.

Though the latter is more consistent with expected behaviour in my opinion, it will have to be removed after some grace period.

### DIFF
--- a/modules/programs/bash/default.nix
+++ b/modules/programs/bash/default.nix
@@ -31,6 +31,26 @@ in
       '';
     };
 
+    programs.bash.promptInit = mkOption {
+      type = types.lines;
+      default = ''
+        # Provide a nice prompt if the terminal supports it.
+        if [ "$TERM" != "dumb" ] || [ -n "$INSIDE_EMACS" ]; then
+          PROMPT_COLOR="1;31m"
+          ((UID)) && PROMPT_COLOR="1;32m"
+          if [ -n "$INSIDE_EMACS" ]; then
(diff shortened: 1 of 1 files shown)


Enzime (changes requested):
My preference would be if this more closely matched the logic in NixOS i.e. setting `programs.bash.interactiveShellInit` to contain `programs.bash.promptInit` and run it before `config.environment.interactiveShellInit`

https://github.com/NixOS/nixpkgs/blob/17726c73ffacc0eb7b739446d40d8a79ad7c8ef2/nixos/modules/programs/bash/bash.nix#L114-L142

christian-burger:
Okay, you mean like this? Because to me your request makes the most sense if all commands meant for interactive shells are put there. 🤷

PR #1026, opened 2024-08-03
Allow disabling channels

### LINKED ISSUE #1024: How to disable Nix channels?
I manage my system with flakes, so channels sometimes lead to surprising results.
I tried settings `nix.channel.enable = false` as it is on NixOS but found out that this is not an option.

thecaralice:
Resolves #1024.

### DIFF
--- a/modules/environment/default.nix
+++ b/modules/environment/default.nix
@@ -16,6 +16,10 @@ let
 in
 
 {
+  imports = [
+    (mkRenamedOptionModule ["environment" "postBuild"] ["environment" "extraSetup"])
+  ];
+
   options = {
     environment.systemPackages = mkOption {
       type = types.listOf types.package;
@@ -41,12 +45,6 @@ in
     environment.profiles = mkOption {
       type = types.listOf types.str;
       description = "A list of profiles used to setup the global environment.";
-    };
-
-    environment.postBuild = mkOption {
-      type = types.lines;
-      default = "";
-      description = "Commands to execute when building the global environment.";
     };
 
     environment.extraOutputsToInstall = mkOption {
@@ -147,6 +145,17 @@ in
       '';
       type = types.lines;
     };
+
+    environment.extraSetup = mkOption {
+      type = types.lines;
+      default = "";
+      description = ''
+        Shell fragments to be run after the system environment has been created.
+        This should only be used for things that need to modify the internals
+        of the environment, e.g. generating MIME caches.
+        The environment being built can be accessed at $out.
+      '';
+    };
   };
 
   config = {
@@ -188,7 +197,8 @@ in
     system.path = pkgs.buildEnv {
       name = "system-path";
       paths = cfg.systemPackages;
-      inherit (cfg) postBuild pathsToLink extraOutputsToInstall;
+      postBuild = cfg.extraSetup;
+      inherit (cfg) pathsToLink extraOutputsToInstall;
     };
 
     system.build.setEnvironment = pkgs.writeText "set-environment" ''
@@ -205,6 +215,5 @@ in
     system.build.setAliases = pkgs.writeText "set-aliases" ''
       ${concatStringsSep "\n" aliasCommands}
     '';
-
   };
 }
--- a/modules/nix/default.nix
+++ b/modules/nix/default.nix
@@ -768,17 +768,11 @@ in
     ];
 
     # Not in NixOS module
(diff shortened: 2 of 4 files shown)


modules/environment/default.nix
```diff
@@ -46,7 +46,8 @@ in
     environment.postBuild = mkOption {
       type = types.lines;
       default = "";
-      description = "Commands to execute when building the global environment.";
+      visible = false;
+      description = "Deprecated alias of {option}`environment.extraSetup`";
```
Enzime, line 50: What's the reason behind deprecating `environment.postBuild`?

modules/nix/default.nix
```diff
@@ -768,17 +768,11 @@ in
...
-      (mkIf (config.system.stateVersion > 3) (mkOrder 1200
-      [ { darwin-config = "${config.environment.darwinConfig}"; }
-        "/nix/var/nix/profiles/per-user/root/channels"
-      ]))
-    ];
+    nix.nixPath = mkIf (config.system.stateVersion < 2) (mkDefault [
+      "darwin=$HOME/.nix-defexpr/darwin"
+      "darwin-config=$HOME/.nixpkgs/darwin-configuration.nix"
+      "/nix/var/nix/profiles/per-user/root/channels"
+    ]);
```
Enzime, line 775: What does this change do?

modules/environment/default.nix
thecaralice: NixOS uses `environment.extraSetup` for that, and having two options doing the same thing is error-prone, so I thought I should introduce the NixOS option and make the old one warn on use.

modules/nix/default.nix
thecaralice: See https://github.com/LnL7/nix-darwin/issues/1024#issuecomment-2266718542

modules/system/checks.nix
```diff
@@ -236,7 +236,7 @@ in
 
     system.checks.verifyNixChannels = mkOption {
       type = types.bool;
-      default = true;
+      default = !config.nix.channel.enable;
```
Enzime, line 239:
```suggestion
      default = config.nix.channel.enable;
```

Shouldn't this be the other way around?

modules/environment/default.nix
```diff
@@ -206,5 +219,8 @@ in
       ${concatStringsSep "\n" aliasCommands}
     '';
 
+    warnings = lib.optional (
+      cfg.postBuild != ""
+    ) "The option `environment.postBuild` has been renamed to `environment.extraSetup`";
```
Enzime, line 224:
```suggestion
```

This should use `mkRenamedOption` instead

modules/environment/default.nix
```diff
@@ -147,6 +148,17 @@ in
       '';
       type = types.lines;
     };
+
+    environment.extraSetup = mkOption {
+      type = types.lines;
+      default = cfg.postBuild;
```
Enzime, line 154:
```suggestion
      default = "";
```

modules/environment/default.nix
```diff
@@ -46,7 +46,8 @@ in
     environment.postBuild = mkOption {
```
Enzime, line 46:
```suggestion
```

The option should be removed and `mkRenamedOption` should be used instead

Enzime (changes requested):

modules/system/checks.nix
thecaralice: fixed

modules/environment/default.nix
thecaralice: Didn't find that in docs, will do

modules/environment/default.nix
```diff
@@ -147,8 +141,23 @@ in
...
+        This should only be used for things that need to modify the internals
+        of the environment, e.g. generating MIME caches.
+        The environment being built can be accessed at $out.
+      '';
+    };
   };
 
+  imports = [
+    (mkRenamedOptionModule ["environment" "postBuild"] ["environment" "extraSetup"])
+  ];
```
Enzime, line 159: Can you move this to be before all the options?

Enzime (commented):
Everything else looks good to me

modules/environment/default.nix
thecaralice: Done

Enzime (approved):
LGTM

Thanks for the contribution

Issue #880, opened 2024-02-22
Installer fails with /etc/ file conflicts when running on a fresh Determinate Nix Installer install


epetousis:
I ran the nix-darwin flake installer, immediately after running the Determinate Nix installer. I got a couple of conflicts:

```
epetousis@evan-mba ~ % nix run nix-darwin -- switch --flake ~/.local/share/dotfiles
building the system configuration...
warning: Ignoring setting 'auto-allocate-uids' because experimental feature 'auto-allocate-uids' is not enabled
warning: Ignoring setting 'impure-env' because experimental feature 'configurable-impure-env' is not enabled
Password:
error: Unexpected files in /etc, aborting activation
The following files have unrecognized content and would be overwritten:

  /etc/nix/nix.conf
  /etc/zshenv

Please check there is nothing critical in these files, rename them by adding .before-nix-darwin to the end, and then try again.
```

I took a look and it's basically just a few defaults getting overwritten, plus Determinate Nix Installer's `upgrade-nix-store-path-url`, which is probably not ideal to have nix-darwin overwrite. Is there meant to be support for this setup at the moment?

epetousis:
Here's the contents of both files with their SHA256 hashes:
<details>
<summary>/etc/zshenv</summary>

```

# Set up Nix only on SSH connections
# See: https://github.com/DeterminateSystems/nix-installer/pull/714
if [ -e '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh' ] && [ -n "${SSH_CONNECTION}" ] && [ "${SHLVL}" -eq 1 ]; then
    . '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
fi
# End Nix
```

Checksum: `d07015be6875f134976fce84c6c7a77b512079c1c5f9594dfa65c70b7968b65f  /etc/zshenv.before-nix-darwin`
</details>

<details>
<summary>/etc/nix/nix.conf</summary>

```# Generated by https://github.com/DeterminateSystems/nix-installer.
# See `/nix/nix-installer --version` for the version details.

build-users-group = nixbld
experimental-features = nix-command flakes repl-flake
bash-prompt-prefix = (nix:$name)\040
max-jobs = auto
extra-nix-path = nixpkgs=flake:nixpkgs
upgrade-nix-store-path-url = https://install.determinate.systems/nix-upgrade/stable/universal
```

[...]

PR #1031, opened 2024-08-07
Add Determinate Systems 0.20.2 to known files

muradbu:
Hi, I ran into the same issue as described in #880, #995 and alike. 

Sorry if this is wrong, I don't fully understand the problem as I'm new to the Nix ecosystem, but from looking at the associated PR's with the issues above it seems like a simple fix so I've replicated it for Determinate Systems 0.20.2 which is the version I'm having issues installing nix-darwin for.


modules/nix/default.nix
```diff
@@ -678,6 +678,7 @@ in
       "c6080216f2a170611e339c3f46e4e1d61aaf0d8b417ad93ade8d647da1382c11"  # DeterminateSystems Nix installer 0.14.0
       "97f4135d262ca22d65c9554aad795c10a4491fa61b67d9c2430f4d82bbfec9a2"  # DeterminateSystems Nix installer 0.15.1
       "5d23e6d7015756c6f300f8cd558ec4d9234ca61deefd4f2478e91a49760b0747"  # DeterminateSystems Nix installer 0.16.0+
```
Enzime, line 680: Could you update this with the version that this configuration file stopped being used? I think it would be the version when they added the `upgrade-nix-store-path-url` default setting

muradbu: Yes, I'll see to it tomorrow

jporsay: Seems it was in v0.16.0 according to the commit it was added in:

* https://github.com/DeterminateSystems/nix-installer/commit/0b0172547c4666f6b1eacb6561a59d6b612505a3
* https://github.com/DeterminateSystems/nix-installer/compare/v0.15.1...v0.16.0

muradbu: > Yes, I'll see to it tomorrow

Hi everyone, sorry for going ghost on this issue. I went on vacation and was AFK, forgot this was active. Thanks for closing this with a fix

PR #1038, opened 2024-08-20
feat: system.disableInstallerTools

isabelroses:
A port of https://github.com/NixOS/nixpkgs/blob/50f054152e00aaee3e75b3041f666453605029bd/nixos/modules/installer/tools/tools.nix#L121

Why?

Other installer tools exist like [`nh_darwin`](https://github.com/ToyVo/nh_darwin)

### DIFF
--- a/modules/nix/nix-darwin.nix
+++ b/modules/nix/nix-darwin.nix
@@ -13,8 +13,19 @@ let
 in
 
 {
-  options = {
-    system.includeUninstaller = lib.mkOption {
+  options.system = {
+    disableInstallerTools = lib.mkOption {
+      type = lib.types.bool;
+      internal = true;
+      default = false;
+      description = ''
+        Disable darwin-rebuild and darwin-option. This is useful to shrink
+        systems which are not expected to rebuild or reconfigure themselves.
+        Use at your own risk!
+    '';
+    };
+
+    includeUninstaller = lib.mkOption {
       type = lib.types.bool;
       internal = true;
       default = true;
@@ -23,10 +34,10 @@ in
 
   config = {
     environment.systemPackages =
-      [ # Include nix-tools by default
+      [ darwin-version ]
+      ++ lib.optionals (!config.system.disableInstallerTools) [
         darwin-option
         darwin-rebuild
-        darwin-version
       ] ++ lib.optional config.system.includeUninstaller darwin-uninstaller;
 
     system.build = {


emilazy:
I’m a little hesitant to merge this without a warning like the one in NixOS. I’m not sure we really have the bandwidth to support third‐party activation tools, especially with changes like https://github.com/LnL7/nix-darwin/issues/96 / https://github.com/LnL7/nix-darwin/pull/1016#issuecomment-2285910068 (though the design I settled on should maintain backwards compatibility).

isabelroses:
> I’m a little hesitant to merge this without a warning like the one in NixOS

I can add the warning. (Besides you can always add the activation tools you want)

> I’m not sure we really have the bandwidth to support third‐party activation tools

You don't really have to support 3rd party activation tools since they should have to support you. And I don't think this would stop anyone from using the tooling if it existed anyways.

modules/nix/nix-darwin.nix
```diff
@@ -23,7 +34,7 @@ in
 
   config = {
     environment.systemPackages =
-      [ # Include nix-tools by default
+      lib.optionals (!config.system.disableInstallerTools) [
         darwin-option
         darwin-rebuild
         darwin-version
```
Enzime, line 40: It seems to me that you would probably still want `darwin-version`

isabelroses: thats a fair point

modules/nix/nix-darwin.nix
```diff
@@ -13,8 +13,19 @@ let
...
-    system.includeUninstaller = lib.mkOption {
+  options.system = {
+    disableInstallerTools = lib.mkOption {
+      type = lib.types.bool;
+      internal = true;
+      default = false;
+      description = ''
+        Disable darwin-rebuild, and other nix-darwin tools.
+        This is useful to shrink clouresize which are not expected to
+        be rebuild or reconfigure themselves. Use at your own risk!
```
Enzime, line 24:
```suggestion
        Disable darwin-rebuild and darwin-option. This is useful to shrink
        systems which are not expected to rebuild or reconfigure themselves.
        Use at your own risk!
```

Enzime (changes requested):

Enzime (approved):
LGTM

Thanks for the contribution

PR #1042, opened 2024-08-24
Use flake based command for uninstaller in README

### LINKED ISSUE #918: Add a uninstaller for flakes
add this comand for uninstaller if you are using flakes becuase the other uninstaller dont work if you are using flakes 

https://github.com/LnL7/nix-darwin/issues/853#issuecomment-1897541986

Enzime:
Fixes #918 

This command should work on both non-flake and flake based setups as it does not require `nixpkgs` to be set in `NIX_PATH`.

Also remove the manual install section as it is unmaintained as far as I can tell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,12 +41,16 @@ darwin-rebuild changelog
 
 ## Uninstalling
 
-There's also an uninstaller if you don't like the project and want to
-remove the configured files and services.
+To run the latest version of the uninstaller, you can run the following command:
 
-```bash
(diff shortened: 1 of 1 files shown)


emilazy (approved):
Seems like an improvement. Should we recommend the locally‐installed `darwin-uninstaller` by default instead?

Enzime:
I suspect it's better to run the latest version of the nix-darwin uninstaller, especially if we end up making further improvements to it

I can add the local `darwin-uninstaller` as the backup option if that command fails for whatever reason

emilazy (approved):

PR #1048, opened 2024-08-27
Support configuring network locations.

tjni:
This adds support for configuring DNS servers and search domains for each [location](https://support.apple.com/en-us/105129).

This is exposed through the `networking.location` option. The `networking.knownNetworkServices` option now applies globally across all locations.

The `networking.dns` and `networking.search` options are now aliases under `networking.location.Automatic`, since the "Automatic" network location is the default location on an unconfigured system.

Below is a sample from my own configuration illustrating usage. I configure DNS servers on my Home location in order to bypass my AT&T router adding AT&T's DNS server automatically:

```
networking = {
  knownNetworkServices = ["Wi-Fi"];

  location = {
    Automatic = {};

    Home = {
      # to avoid AT&T's slow DNS servers
      dns = [
        "2001:4860:4860:0:0:0:0:8888"
        "2001:4860:4860:0:0:0:0:8844"
        "8.8.8.8"
        "8.8.4.4"
      ];
    };
  };
};
```


Enzime:
Can you rebase this on master?

tjni:
Done!

dustypomerleau:
I would love to see this merged :) Using different DNS arrangements is also my use case.

tjni:
@Enzime @emilazy Would you have some time to review this PR? I've been successfully using it consistently in my setup (hence all of the pushes that you see).

tjni:
@Enzime @emilazy Another ping to request a review, in case you find some free time to spare.

modules/networking/default.nix
```diff
@@ -7,24 +7,92 @@ let
...
+    done
+
+    # switch to a location that surely does not need to be deleted
+    networksetup -switchtolocation ${strings.escapeShellArg (builtins.head (builtins.attrNames cfg.location))} > /dev/null
+
+    for location in "''${!curr_locations[@]}"; do
+      if [[ ! -v goal_locations[$location] ]]; then
+        networksetup -deletelocation "$location" > /dev/null
+      fi
+    done
```
emilazy, line 42: This is going to clobber unmanaged locations, right? That’s usually why we have `known*` stuff. That might be okay, although it’s unfortunate that going from *n* to 0 locations won’t be handled.

modules/networking/default.nix
```diff
@@ -7,24 +7,92 @@ let
...
+
+    ${concatMapStringsSep "\n" (location: ''
+      networksetup -switchtolocation ${strings.escapeShellArg location} > /dev/null
```
emilazy, line 45: Is there no way we can avoid changing the live network configuration while doing this? It seems unexpected and unfortunate.

modules/networking/default.nix
```diff
@@ -73,27 +141,54 @@ in
...
 
-    networking.search = mkOption {
-      type = types.listOf types.str;
-      default = [];
-      description = "The list of search paths used when resolving domain names.";
+        By default, a system comes with a single location called "Automatic", but you can
+        define additional locations to switch between different network configurations.
+
+        If you define any locations here, you must also explicitly define the "Automatic"
+        location if you want it to exist.
```
emilazy, line 190: Maybe it should be included by default, if feasible.

emilazy (commented):
Sorry about this, it has been a busy time for me. I do want to get around to this one; here’s a first pass, maybe Michael can help too. Please drop the formatting changes; we should do a treewide format now that NixOS has, but they need to go in `.git-blame-ignore-revs` and make review harder. (You could split them into a separate commit and add it to the ignore file if it’s too much of a pain to disable auto‐formatting in your editor, though.)

PR #1051, opened 2024-08-29
gpg: Suppress stderr from gpg-connect-agent on shell init

booxter:
In some scenarios, the command may fail, e.g. when the shell is executed with a different $HOME from where gpg agent is configured to run from.

(E.g. this happens in kitty terminal test suite.)

This patch will suppress stderr errors on tty in this situation.

Note that zsh does not allow to suppress execution of /etc/zshenv on startup, so it's impossible to skip it in the test suite environment.

An alternative would be to set IN_NIX_SHELL in the test suite, but this was rejected in upstream:

https://github.com/kovidgoyal/kitty/pull/7800

There's also a kitty package specific fix posted here but this may be unnecessary once nix-darwin is patched here:

https://github.com/NixOS/nixpkgs/pull/338070

### DIFF
--- a/modules/programs/gnupg.nix
+++ b/modules/programs/gnupg.nix
@@ -43,7 +43,7 @@ in
     '' + (optionalString cfg.agent.enableSSHSupport ''
       # SSH agent protocol doesn't support changing TTYs, so bind the agent
       # to every new TTY.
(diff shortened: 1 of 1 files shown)


booxter:
Tested this with running `./test.py` from inside the kitty `nix-shell` devenv. It now passes.

Enzime (approved):
LGTM

Thanks for the contribution

Issue #1043, opened 2024-08-24
Waiting for the nix store to mount before attempting to run launchd daemons that depend on nix derivations


will-lol:
The nix-daemon launchd daemon is configured to use the `/bin/wait4path` utility to ensure the nix store has been mounted before it runs its nix derivation.

I recently ran into an issue with the yabai scripting addition daemon where the daemon would attempt to access a script (located in the nix store) before the store had been mounted. I'd imagine that the daemon works sometimes and for some people, but I reckon we should fix this.

I added the following to my darwin configuration to fix this issue: 

```nix
launchd.daemons.yabai-sa = let
  yabaiScript = pkgs.writeShellScript "yabai-sa" ''
    ${pkgs.yabai}/bin/yabai --load-sa
  '';
in {
  script = pkgs.lib.mkForce "";
  serviceConfig.RunAtLoad = true;
  serviceConfig.KeepAlive.SuccessfulExit = false;
  serviceConfig.ProgramArguments = [ "/bin/sh" "-c" "/bin/wait4path ${yabaiScript} &amp;&amp; exec ${yabaiScript}" ];
};
```

I think that we should either:
- Fix the existing `yabai-sa` daemon to work like the nix-daemon one (and my override)
- Fix the `launchd.daemons` module so that the `script` attribute works reliably (by waiting for the script to be in the nix store in some fashion). 

I'd like to hear all of your thoughts. I'm happy to write a PR for this also, but thought some discussion would be good.

Enzime:
My preference is the latter, if you could make a PR that changes `launchd.daemons` to automatically use `wait4path` that would be appreciated.

In the same PR, you could probably drop all explicit/manual usage of `wait4path` in nix-darwin

will-lol:
Sounds good to me! I’ll tackle that next weekend.

Samasaur1:
Would also address https://github.com/LnL7/nix-darwin/issues/774 and https://github.com/LnL7/nix-darwin/issues/563

Samasaur1:
(and possibly other issues as well)

Enzime:
Fixed by #1052

PR #1052, opened 2024-08-31
Use wait4path with command and script launchd options

will-lol:
Addresses my recent issue, https://github.com/LnL7/nix-darwin/issues/1043.

I've attempted to make this compatible with both the 'command' and 'script' options of 'launchd.daemons.<name>'. This works fine on my machine, but some tests that check the launchd plist files fail because this will change the format of them slightly. 

I thought I would put this first commit here in case anyone had any objections before I went ahead and rewrote all of the tests and modules to not use wait4path anymore.

### DIFF
--- a/modules/launchd/default.nix
+++ b/modules/launchd/default.nix
@@ -20,7 +20,6 @@ let
     { config, name, ... }:
     let
 
-      cmd = config.command;
       env = config.environment // optionalAttrs (config.path != "") { PATH = config.path; };
 
     in
@@ -88,7 +87,11 @@ let
         '');
 
         serviceConfig.Label = mkDefault "${cfg.labelPrefix}.${name}";
-        serviceConfig.ProgramArguments = mkIf (cmd != "") [ "/bin/sh" "-c" "exec ${cmd}" ];
+        serviceConfig.ProgramArguments = mkIf (config.command != "") [
+          "/bin/sh"
+          "-c"
+          "/bin/wait4path /nix/store &amp;&amp; exec ${config.command}"
+        ];
         serviceConfig.EnvironmentVariables = mkIf (env != {}) env;
       };
     };
--- a/modules/nix/linux-builder.nix
+++ b/modules/nix/linux-builder.nix
@@ -4,21 +4,6 @@ with lib;
 
 let
   cfg = config.nix.linux-builder;
-
-  # create-builder uses TMPDIR to share files with the builder, notably certs.
-  # macOS will clean up files in /tmp automatically that haven't been accessed in 3+ days.
-  # If we let it use /tmp, leaving the computer asleep for 3 days makes the certs vanish.
-  # So we'll use /run/org.nixos.linux-builder instead and clean it up ourselves.
-  script = pkgs.writeShellScript "linux-builder-start" ''
-    export TMPDIR=/run/org.nixos.linux-builder USE_TMPDIR=1
-    rm -rf $TMPDIR
-    mkdir -p $TMPDIR
-    trap "rm -rf $TMPDIR" EXIT
-    ${lib.optionalString cfg.ephemeral ''
-      rm -f ${cfg.workingDirectory}/${cfg.package.nixosConfig.networking.hostName}.qcow2
-    ''}
-    ${cfg.package}/bin/create-builder
-  '';
 in
 
 {
@@ -176,11 +161,23 @@ in
       environment = {
         inherit (config.environment.variables) NIX_SSL_CERT_FILE;
       };
+
+      # create-builder uses TMPDIR to share files with the builder, notably certs.
+      # macOS will clean up files in /tmp automatically that haven't been accessed in 3+ days.
+      # If we let it use /tmp, leaving the computer asleep for 3 days makes the certs vanish.
+      # So we'll use /run/org.nixos.linux-builder instead and clean it up ourselves.
+      script = ''
+        export TMPDIR=/run/org.nixos.linux-builder USE_TMPDIR=1
+        rm -rf $TMPDIR
+        mkdir -p $TMPDIR
+        trap "rm -rf $TMPDIR" EXIT
+        ${lib.optionalString cfg.ephemeral ''
+          rm -f ${cfg.workingDirectory}/${cfg.package.nixosConfig.networking.hostName}.qcow2
+        ''}
+        ${cfg.package}/bin/create-builder
+      '';
+
       serviceConfig = {
-        ProgramArguments = [
-          "/bin/sh" "-c"
-          "/bin/wait4path /nix/store &amp;&amp; exec ${script}"
-        ];
         KeepAlive = true;
         RunAtLoad = true;
         WorkingDirectory = cfg.workingDirectory;
--- a/modules/services/karabiner-elements/default.nix
+++ b/modules/services/karabiner-elements/default.nix
@@ -38,14 +38,11 @@ in
     # the system extension is activated, so we can call activate from the manager
     # which will block until the system extension is activated.
     launchd.daemons.start_karabiner_daemons = {
-      serviceConfig.ProgramArguments = [
-        "/bin/sh" "-c"
-        "/bin/wait4path /nix/store &amp;&amp; ${pkgs.writeScript "start_karabiner_daemons" ''
+      script = ''
           ${parentAppDir}/.Karabiner-VirtualHIDDevice-Manager.app/Contents/MacOS/Karabiner-VirtualHIDDevice-Manager activate
           launchctl kickstart system/org.pqrs.karabiner.karabiner_grabber
           launchctl kickstart system/org.pqrs.karabiner.karabiner_observer
-        ''}"
-      ];
+      '';
       serviceConfig.Label = "org.nixos.start_karabiner_daemons";
       serviceConfig.RunAtLoad = true;
     };
@@ -73,11 +70,7 @@ in
     };
 
     launchd.daemons.Karabiner-DriverKit-VirtualHIDDeviceClient = {
-      serviceConfig.ProgramArguments = [
-        "/bin/sh" "-c"
-        # For unknown reasons this daemon will fail if VirtualHIDDeviceClient is not exec'd.
-        "/bin/wait4path /nix/store &amp;&amp; exec \"${pkgs.karabiner-elements.driver}/Library/Application Support/org.pqrs/Karabiner-DriverKit-VirtualHIDDevice/Applications/Karabiner-DriverKit-VirtualHIDDeviceClient.app/Contents/MacOS/Karabiner-DriverKit-VirtualHIDDeviceClient\""
-      ];
+      command = "${pkgs.karabiner-elements.driver}/Library/Application Support/org.pqrs/Karabiner-DriverKit-VirtualHIDDevice/Applications/Karabiner-DriverKit-VirtualHIDDeviceClient.app/Contents/MacOS/Karabiner-DriverKit-VirtualHIDDeviceClient";
       serviceConfig.ProcessType = "Interactive";
       serviceConfig.Label = "org.pqrs.Karabiner-DriverKit-VirtualHIDDeviceClient";
       serviceConfig.KeepAlive = true;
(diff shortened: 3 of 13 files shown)


Samasaur1:
I wonder if you still want to `exec` the actual command, which appears to be the previous behavior but looks like you removed it

modules/launchd/default.nix
```diff
@@ -23,7 +23,10 @@ let
       cmd = config.command;
       env = config.environment // optionalAttrs (config.path != "") { PATH = config.path; };
 
-    in
+      wait4Command = pkgs.writeShellScript "wait4command"
+        "/bin/wait4path $(realpath $(type -P \"$1\")) && exec \"$@\"";
```
Enzime, line 27:
```suggestion
```

This won't work as `pkgs.writeShellScript` will give you a shell script inside `/nix/store` which might not be available

modules/launchd/default.nix
```diff
@@ -88,7 +91,11 @@ let
         '');
 
         serviceConfig.Label = mkDefault "${cfg.labelPrefix}.${name}";
-        serviceConfig.ProgramArguments = mkIf (cmd != "") [ "/bin/sh" "-c" "exec ${cmd}" ];
+        serviceConfig.ProgramArguments = mkIf (cmd != "") [
+          "/bin/sh"
+          "-c"
+          "exec ${wait4Command} ${cmd}"
```
Enzime, line 97:
```suggestion
          "/bin/wait4path /nix/store && exec ${cmd}"
```

Enzime (changes requested):

niklasravnsborg:
I also had a similar problem as @will-lol described in the #1043 issue but with nextdns:

The service didn't start after reboot and the service state was stuck at `spawn scheduled` (`sudo launchctl print system/org.nixos.nextdns`). I now applied this refactor to my nix-darwin fork and now everything works like a charm. 

I had to refactor the nextdns service to use `command` instead of `serviceConfig.ProgramArguments` in https://github.com/niklasravnsborg/nix-darwin/commit/697115e51d4c6eccd6bb65738a32b1b14258f4d9. Do I see it correctly that all services using `launchd.daemons` have to be refactored in a similar fashion?

will-lol:
Hi @niklasravnsborg. I believe so. I've refactored some modules that were using the wait4path hack themselves to use 'command' or 'script' (not yet pushed). I am now working on getting tests to pass. I could include refactors of other modules that are manually using serviceConfig.ProgramArguments also. It may also be best for module maintainers to exercise their own judgement.

niklasravnsborg:
@will-lol If you want you can also pull in my nextdns fix here :) I'm looking forward to not maintaining my own fork anymore :D

emilazy:
We could implement our own `programArguments` that uses `lib.escapeShellArgs` to construct an equivalent shell command, since the list interface is nicer and it would mean fewer changes to services.

will-lol:
All of the tests broken by this change are now fixed. I wonder if we'd prefer to deal with fixing problems with other modules that aren't yet using /bin/wait4path in some fashion in a separate issue or PR? 
Also: many of the modules I migrated over from ProgramArguments and /bin/wait4path I do not use myself, and have not yet confirmed to be working on bare metal. 
If required I'll work on that next weekend and hopefully we can get this merged. Thanks all!

modules/launchd/default.nix
```diff
@@ -88,7 +88,11 @@ let
         '');
 
         serviceConfig.Label = mkDefault "${cfg.labelPrefix}.${name}";
-        serviceConfig.ProgramArguments = mkIf (cmd != "") [ "/bin/sh" "-c" "exec ${cmd}" ];
+        serviceConfig.ProgramArguments = mkIf (cmd != "") [
+          "/bin/sh"
+          "-c"
+          "/bin/wait4path /nix/store &amp;&amp; exec ${cmd}"
```
niklasravnsborg, line 94: Is there any special reason, you used `exec` here? I think we could just drop it and it would make extending the cmd string easier, because one could still use basic shell syntax.

niklasravnsborg (changes requested):

modules/launchd/default.nix
will-lol: Using 'exec' was the previous behaviour. I also received a comment (https://github.com/LnL7/nix-darwin/pull/1052#issuecomment-2323053459) suggesting that I add it back. 

I'm now of the view that chaining commands together with any special shell syntax belongs in a script. The shell that is used to execute the program arguments is also not obvious to the user of the abstraction. By using a script one would expect (by nix convention) that you are getting bash, which also has some extended syntax not supported by /bin/sh. 

Totally open to anything else tho

will-lol:
Manual testing of the refactors I made that I've done on my machine:
- tailscaled works
- nix-optimise works
- nix-daemon works
- was unable to test karabiner elements due to it currently being broken
- linux-builder, was able to reach the vm's ssh

niklasravnsborg:
@will-lol Do you know why some of the tests are broken?

will-lol:
Hah not totally sure.. I can't really see a reason why it failed that time. The error suggests hello.nix wasn't copied or checked into git? All checks are passing now.

modules/launchd/default.nix
Enzime, line 94:
```suggestion
          "/bin/wait4path /nix/store &amp;&amp; exec ${cfg.command}"
```

For clarity

modules/launchd/default.nix
```diff
@@ -88,7 +88,11 @@ let
         '');
 
         serviceConfig.Label = mkDefault "${cfg.labelPrefix}.${name}";
-        serviceConfig.ProgramArguments = mkIf (cmd != "") [ "/bin/sh" "-c" "exec ${cmd}" ];
+        serviceConfig.ProgramArguments = mkIf (cmd != "") [
```
Enzime, line 91:
```suggestion
        serviceConfig.ProgramArguments = mkIf (cfg.command != "") [
```

modules/nix/linux-builder.nix
```diff
@@ -176,11 +161,23 @@ in
       environment = {
         inherit (config.environment.variables) NIX_SSL_CERT_FILE;
       };
+
+			# create-builder uses TMPDIR to share files with the builder, notably certs.
```
Enzime, line 165: Can you fix the indentation?

modules/services/nix-daemon.nix
```diff
@@ -44,10 +44,7 @@ in
     nix.useDaemon = true;
 
     launchd.daemons.nix-daemon = {
-      serviceConfig.ProgramArguments = [
-        "/bin/sh" "-c"
-        "/bin/wait4path ${config.nix.package}/bin/nix-daemon &amp;&amp; exec ${config.nix.package}/bin/nix-daemon"
-      ];
+      command = "${config.nix.package}/bin/nix-daemon";
```
Enzime, line 47:
```suggestion
      command = lib.getExe' config.nix.package "nix-daemon";
```

modules/services/nix-optimise/default.nix
```diff
@@ -62,11 +62,8 @@ in
 
     launchd.daemons.nix-optimise = {
       environment.NIX_REMOTE = optionalString config.nix.useDaemon "daemon";
+      command = "${config.nix.package}/bin/nix-store --optimise";
```
Enzime, line 65:
```suggestion
      command = "${lib.getExe' config.nix.package "nix-store"} --optimise";
```

Enzime (changes requested):
Can you squash all the commits into one commit?

Enzime:
I wonder if the real fix is to get all users to migrate to an unencrypted Nix Store as it should still be encrypted automatically by FileVault see: https://github.com/DeterminateSystems/nix-installer/issues/753#issuecomment-1995192341

It'll probably require users to uninstall nix-darwin then Nix and then reinstall both of them

emilazy:
I personally use an encrypted store. An unencrypted store is not protected in the same way that FileVault volumes are, even though the storage is encrypted at rest; you can access such volumes in macOS Recovery without authentication. My understanding is that full encryption is also a hard requirement for many enterprise users.

I hope we can move towards a more structured interface per https://github.com/LnL7/nix-darwin/pull/1052#issuecomment-2336654030, but this seems like an improvement on the status quo.

modules/services/tailscale.nix
```diff
@@ -54,12 +54,9 @@ in
     launchd.daemons.tailscaled = {
       # derived from
       # https://github.com/tailscale/tailscale/blob/main/cmd/tailscaled/install_darwin.go#L30
+      command = "${cfg.package}/bin/tailscaled";
```
Enzime, line 57: Can you change this to `lib.getExe'`?

Enzime (changes requested):

modules/services/karabiner-elements/default.nix
```diff
@@ -95,23 +88,17 @@ in
...
       serviceConfig.RunAtLoad = true;
       serviceConfig.KeepAlive.SuccessfulExit = false;
     };
 
     launchd.user.agents.karabiner_session_monitor = {
-      serviceConfig.ProgramArguments = [
-        "/bin/sh" "-c"
-        "/bin/wait4path /run/wrappers/bin &amp;&amp; /run/wrappers/bin/karabiner_session_monitor"
-      ];
+      command = "/run/wrappers/bin/karabiner_session_monitor";
```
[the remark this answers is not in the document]
will-lol: good catch thank you

modules/services/karabiner-elements/default.nix
```diff
@@ -95,23 +95,16 @@ in
...
           install -m4555 "${pkgs.karabiner-elements}/Library/Application Support/org.pqrs/Karabiner-Elements/bin/karabiner_session_monitor" /run/wrappers/bin
-        ''}"
-      ];
+      '';
       serviceConfig.RunAtLoad = true;
       serviceConfig.KeepAlive.SuccessfulExit = false;
     };
 
     launchd.user.agents.karabiner_session_monitor = {
-      serviceConfig.ProgramArguments = [
```
Enzime, line 111: I think this needs to be readded

Enzime (changes requested):
Other than that it looks fine to me

modules/services/karabiner-elements/default.nix
will-lol: 😅 I can't believe I made that mistake... thanks for reviewing.

Enzime (approved):
LGTM

Thanks for the contribution

niklasravnsborg:
> I wonder if the real fix is to get all users to migrate to an unencrypted Nix Store as it should still be encrypted automatically by FileVault see: [DeterminateSystems/nix-installer#753 (comment)](https://github.com/DeterminateSystems/nix-installer/issues/753#issuecomment-1995192341)

I think even an unencrypted store might suffer the same problem when using in combination with launchd services, since the OS might try to start those services before the store is mounted ([while the mounting itself is done in a launchd service](https://github.com/DeterminateSystems/nix-installer/blob/main/src/action/macos/create_determinate_volume_service.rs)).

niklasravnsborg:
@will-lol Do I see correctly there some other services missing regarding this update to use the `script`, right? How should we proceed on those? Merge this first and wait for module maintainers to move their logic over as well?

I personally just noticed it regarding `nextdns` and would be available to update and test it there.

modules/services/karabiner-elements/default.nix
```diff
@@ -73,11 +70,7 @@ in
     };
 
     launchd.daemons.Karabiner-DriverKit-VirtualHIDDeviceClient = {
-      serviceConfig.ProgramArguments = [
-        "/bin/sh" "-c"
-        # For unknown reasons this daemon will fail if VirtualHIDDeviceClient is not exec'd.
-        "/bin/wait4path /nix/store &amp;&amp; exec \"${pkgs.karabiner-elements.driver}/Library/Application Support/org.pqrs/Karabiner-DriverKit-VirtualHIDDevice/Applications/Karabiner-DriverKit-VirtualHIDDeviceClient.app/Contents/MacOS/Karabiner-DriverKit-VirtualHIDDeviceClient\""
-      ];
+      command = "${pkgs.karabiner-elements.driver}/Library/Application Support/org.pqrs/Karabiner-DriverKit-VirtualHIDDevice/Applications/Karabiner-DriverKit-VirtualHIDDeviceClient.app/Contents/MacOS/Karabiner-DriverKit-VirtualHIDDeviceClient";
```
niklasravnsborg, line 73: Is it intentional that just this one uses `command` and the other ones use `script`?

Enzime:
Tested this locally and looks like it's working

ofalvai:
Came here to investigate the same issue. My `nix-darwin` follows `nixpkgs-stable` to keep using Karabiner 14.3 and updating `nix-darwin` to the latest version breaks Karabiner. Rolling back to `6374cd7e50aa057a688142eed2345083047ad884` makes it work again.

Enzime:
See #1092

PR #1055, opened 2024-09-03
nixos/github-runner: fix build failure

siriobalmelli:
Shellcheck complains:

       > args=(
       >      ^-- SC2054 (warning): Use spaces, not commas, to separate array elements.

Add a comment disabling shellcheck in this case and annotating why.

### DIFF
--- a/modules/services/github-runner/service.nix
+++ b/modules/services/github-runner/service.nix
@@ -90,7 +90,7 @@ in
           let
             configure = pkgs.writeShellApplication {
               name = "configure-github-runner-${name}";
-              text = ''
+              text = /*bash*/''
                 export RUNNER_ROOT
 
                 args=(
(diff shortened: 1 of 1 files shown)


siriobalmelli:
I see a build failure but seems to have no correlation with the minor fix being PRed.

Enzime:
Could you instead wrap the argument to `--labels` with double quotes?

siriobalmelli:
Respun and tested, works fine.

Enzime (approved):
LGTM

Thanks for the contribution

modules/services/github-runner/service.nix
```diff
@@ -90,15 +90,15 @@ in
...
+              text = /*bash*/''
                 export RUNNER_ROOT
 
                 args=(
                   --unattended
                   --disableupdate
                   --work ${escapeShellArg workDir}
                   --url ${escapeShellArg cfg.url}
-                  --labels ${escapeShellArg (concatStringsSep "," cfg.extraLabels)}
+                  --labels "${escapeShellArg (concatStringsSep "," cfg.extraLabels)}"
```
srid, line 101: > @Enzime Could you instead wrap the argument to `--labels` with double quotes?

Looks like `escapeShellArg` is adding its own quote. See #1085

siriobalmelli: My apologies for having introduced this. @Enzime looks like the original suggestion with a `#shellcheck disable` directive was the correct way to go.

PR #1061, opened 2024-09-07
activate-system: Activate user environment too

### LINKED ISSUE #420: [help] launchd.envVariables after reboot
Setting `launchd.envVariables` does not seem to survive after reboot, the script is generated to `activate`, which will not be executed in launch daemon `org.nixos.activate-system`?
Is there any way to set launchd environment variables other than manually writing another launch agent?
```nix
  launchd.envVariables = { COLORTERM = "truecolor"; };
```

booxter:
Closes #420

### DIFF
--- a/modules/services/activate-system/default.nix
+++ b/modules/services/activate-system/default.nix
@@ -42,5 +42,10 @@ in
       serviceConfig.KeepAlive.SuccessfulExit = false;
     };
 
+    launchd.user.agents.activate-user = {
+      command = "/run/current-system/activate-user";
+      serviceConfig.RunAtLoad = true;
+      serviceConfig.KeepAlive.SuccessfulExit = false;
+    };
   };
 }


modules/services/activate-system/default.nix
```diff
@@ -42,5 +42,10 @@ in
       serviceConfig.KeepAlive.SuccessfulExit = false;
     };
 
+    launchd.user.agents.activate-user = {
+      command = "/run/current-system/activate-user";
+      serviceConfig.RunAtLoad = true;
+      serviceConfig.KeepAlive.SuccessfulExit = false;
+    };
```
Enzime, line 49: This will run `activate-user` as every user on a machine which is not ideal

Enzime (changes requested):
Thanks for the PR, but currently nix-darwin doesn't have the concept of primary user or per-user activation which is necessary before something like this should be added

see: https://github.com/LnL7/nix-darwin/pull/1016#issuecomment-2285910068

modules/services/activate-system/default.nix
booxter: Thanks for this comment, I didn't realize it.

On a related note, I am now realizing that there's also `launchd.agents` module in `home-manager`. Is it a different module? Does it mean I can achieve the same effect (scoped to my user) by defining it there instead? If so, then I think I may not need this patch in `nix-darwin` at all. I can instead define the service that `launchctl setenv` in HM instead. (That's the only part of the `activate-user` script I actually care about here.)

Enzime: Yeah, putting it in home-manager will definitely work better as it’ll be specific to your user

booxter:
Abandoning as per comments, will implement this in HM instead then.

PR #1068, opened 2024-09-10
tests/{networking-hostname,users-groups}: update for `lib.escapeShellArg` change

emilazy:
Fixes the unstable tests.


Enzime (approved):
LGTM

PR #1070, opened 2024-09-11
Fix darwin-rebuild changelog not displaying

### LINKED ISSUE #329: `darwin-rebuild changelog` is completely broken
Running `darwin-rebuild changelog` just exits with no output. The problem is it isn't ever setting the `$systemConfig` var for this action, which causes the script to exit early (with no warnings or errors even).

It looks like this broke 2 years ago in 24089a93d7a34cf33c566040ae12e6be72664ca9.

antoineco:
The script returns early when `$systemConfig` is empty, which currently prevents the changelog from being displayed.

Fixes #329

---

Before, `$systemConfig=`

```console
$ darwin-rebuild changelog
(nothing)
```

After, `$systemConfig=/nix/store/22dc9r5hhm311x5bffwjjr1abr3r2jrz-darwin-system-24.05.20240830.6e99f2a+darwin5.f4f18f3`:

```console
$ darwin-rebuild changelog
CHANGELOG

2024-09-10
- The default Nix build user group ID is now set to 350 when
  `system.stateVersion` ≥ 5, to reflect the default for new Nix
  installations. This only affects installations that enable
  `nix.configureBuildUsers`, and any divergence will be detected on
  system activation. To use `nix.configureBuildUsers` with a higher
  `system.stateVersion` on installations using the old group ID, set:

      ids.gids.nixbld = 30000;

  We do not recommend trying to change the group ID with macOS user
  management tools without a complete uninstallation and reinstallation
  of Nix.

...
```

### DIFF
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -203,7 +203,7 @@ if [ "$action" = edit ]; then
   fi
 fi
 
-if [ "$action" = switch ] || [ "$action" = build ] || [ "$action" = check ]; then
+if [ "$action" = switch ] || [ "$action" = build ] || [ "$action" = check ] || [ "$action" = changelog ]; then
   echo "building the system configuration..." >&2
   if [ -z "$flake" ]; then
     systemConfig="$(nix-build '<darwin>' "${extraBuildFlags[@]}" -A system)"
@@ -252,11 +252,7 @@ if [ "$action" = switch ] || [ "$action" = activate ] || [ "$action" = rollback 
 fi
 
 if [ "$action" = changelog ]; then
-  echo >&2
-  echo "[1;1mCHANGELOG[0m" >&2
-  echo >&2
-  head -n 32 "$systemConfig/darwin-changes"
-  echo >&2
+  ${PAGER:-less} -- "$systemConfig/darwin-changes"
 fi
 
 if [ "$action" = check ]; then


emilazy:
Thanks, I was wondering why this was broken. The `head -n 32` is weird, we should probably just run `${PAGER:-less}` on the file. Do you mind doing that or do you want me to handle it in another PR?

(FWIW the convention here / in Nixpkgs is to just use `topic:` rather than conventional commits, although we’re not very strict about those standards. In Nixpkgs it’s more important since the CI tools read commit messages.)

pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -255,7 +255,7 @@ if [ "$action" = changelog ]; then
   echo >&2
   echo "[1;1mCHANGELOG[0m" >&2
   echo >&2
-  head -n 32 "$systemConfig/darwin-changes"
+  "${PAGER:-less}" -- "$systemConfig/darwin-changes"
   echo >&2
```
emilazy, line 259: The pager eats up these extra lines in practice, so this could just be the pager line I think.

Unfortunately the pedantic shell quoting (which I generally approve of!) is counterproductive here – I have `$PAGER` set to `less -R`, which breaks with this.

emilazy (changes requested):

pkgs/nix-tools/darwin-rebuild.sh
antoineco: That's a good point I hadn't really paid attention to those `echo`. And you're right, if `PAGER` contains shell words as opposed to a single word resolving to an alias, we shouldn't quote it 👍

emilazy (approved):
LGTM!

PR #1072, opened 2024-09-12
Remove nix-shell early return in /etc/{zshenv,bashrc}

antoineco:
The condition does not match the comment, and therefore not the original intention. It currently returns early in *any* type of Nix shell, not just pure ones, including 'nix develop'.

Besides being unnecessary, this check prevents Nix shells from functioning properly. For instance, it causes the initialization of the Zsh fpath to be skipped, which is critical. The fact that the user is unable to opt out of this behaviour makes this an ever bigger problem since /etc/zshenv is being loaded unconditionally by Zsh.

For reference, NixOS does not perform such check, and apparently never did based on my inspection of the [commit history](https://github.com/NixOS/nixpkgs/commits/nixos-24.05/nixos/modules/programs/zsh/zsh.nix).

For a history, see the following commits (without comment or context unfortunately):
- 71f25cba9d530e3e7f58bd4698b8595dc1373927
- 94fddd38afe7312d1409e901b927230a7d3a2199

### DIFF
--- a/modules/programs/bash/default.nix
+++ b/modules/programs/bash/default.nix
@@ -55,9 +55,6 @@ in
       if [ -n "$__ETC_BASHRC_SOURCED" -o -n "$NOSYSBASHRC" ]; then return; fi
       __ETC_BASHRC_SOURCED=1
 
-      # Don't execute this file when running in a pure nix-shell.
-      if [ "$IN_NIX_SHELL" = "pure" ]; then return; fi
-
       if [ -z "$__NIX_DARWIN_SET_ENVIRONMENT_DONE" ]; then
         . ${config.system.build.setEnvironment}
       fi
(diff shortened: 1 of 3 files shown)


emilazy:
It looks like Bash also does the same thing currently. We should probably be consistent across the shells.

A little scared to touch something that looks as load‐bearing as this, but if it works with `nix develop`, `nix develop --pure`, and the `nix-shell` equivalents, then this seems fine.

antoineco:
I originally didn't want to get out of my way since I don't use Bash as my day to day shell, but I agree that we should be consistent.

AFAIK pure shells have a reasonably restricted environment and work as intended without having to resort to these kind of early returns. At least I couldn't reproduce a situation in which this would be a problem. And as I wrote, it seems like NixOS never introduced this early return at any given point, so the current behavior just causes more inconsistencies between various Nix-managed hosts.

emilazy:
Yeah, it seems fine to do. Could you adjust the commit message to `{bash,zsh}: …`?

PR #1075, opened 2024-09-12
version: make `system.stateVersion` mandatory

emilazy:
When testing the Sequoia UID change, I discovered that @mjm didn’t have `system.stateVersion` set; I suspect this is not too uncommon. Let’s make it required now, like NixOS is trying to, to improve our backwards‐compatibility story in anticipation of starting to cut release branches.

### DIFF
--- a/modules/system/version.nix
+++ b/modules/system/version.nix
@@ -5,8 +5,6 @@ with lib;
 let
   cfg = config.system;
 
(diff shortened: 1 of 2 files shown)


Enzime (approved):
LGTM

FWIW I didn’t have it set either

PR #1078, opened 2024-09-16
feat: add dock item property overrides

aisamu:
#### feat: add dock item property overrides <sup>e8be494</sup>
This allows configuring things like "Sort by", "Display as" and "View
content as" by mirroring the exact value observed on a `defaults read`.

The existing options and values weren't formalized in the schema as I
have no expectations of stability across OS versions.

#### docs: don't use tildes on dock.persistent-others examples <sup>73a5575</sup>
It no longer works on Sonoma 14.6.1, leaving a `?` instead of the
desired folder.

#### style: whitespace <sup>31b0fb7</sup>
(nixpkgs-fmt)

Thanks for nix-darwin! ❤️

### DIFF
--- a/modules/system/defaults/dock.nix
+++ b/modules/system/defaults/dock.nix
@@ -200,17 +200,32 @@ in
     };
 
     system.defaults.dock.persistent-others = mkOption {
-      type = types.nullOr (types.listOf (types.either types.path types.str));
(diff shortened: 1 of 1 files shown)


Samasaur1:
Superseded by #1431

Issue #1082, opened 2024-09-17
`nix.registry.nixpkgs.to.path` defined multiple times within nix-darwin itself


sellout:
8714f9e28529183d65d9f42ac92cdc5d70dbb6f7 is a breaking change. When I `darwin-rebuild build` at or after this commit, I get the following error.
```
error: The option `nix.registry.nixpkgs.to.path' is defined multiple times while it's expected to be unique.

Definition values:
- In `/nix/store/bf…4i-source/modules/nix': "/nix/store/4a…rx-source"
- In `/nix/store/bf…4i-source/modules/nix/nixpkgs-flake.nix'
Use `lib.mkForce value` or `lib.mkDefault value` to change the priority on any of these definitions.
```
- [modules/nix/default.nix](https://github.com/LnL7/nix-darwin/blob/21fe31f26473c180390cfa81e3ea81aca0204c80/modules/nix/default.nix#L478-L485)
- [modules/nix/nixpkgs-flake.nix](https://github.com/LnL7/nix-darwin/blob/21fe31f26473c180390cfa81e3ea81aca0204c80/modules/nix/nixpkgs-flake.nix#L91-L94)

Both of these locations use `mkDefault`, but the latter was added just a week ago in 8714f9e28529183d65d9f42ac92cdc5d70dbb6f7.

My config sets `nix.registry.nixpkgs.flake = inputs.nixpkgs`. Removing that setting allows my configuration to build, but the issue is that my `nix.registry` and `nix.settings` are shared between my configs for nix-darwin, NixOS, and Home Manager. This change seems to make it so I can’t share them (or at least _all_ of them) any more.

But perhaps there is some other change I can make to the config to maintain cross-compatibility?

emilazy:
Yeah, NixOS does something quite bad here and we inherited it, but slightly different, in a way that breaks things…

Ideally the NixOS module should be improved but please try https://github.com/LnL7/nix-darwin/pull/1083 which should make us suboptimal in the same way. Sorry for not catching this before the merge.

mcgilly17:
Thanks for the quick update @emilazy, and thank you for your work on nix-darwin! With the latest update, are we then safe to upgrade to sequoia if we have run a build prior to upgrading?

emilazy:
As long as you see the `darwin-rebuild switch` error message, run the command it quotes, and manage to successfully `darwin-rebuild switch` after that, you should be good to upgrade.

mcgilly17:
[...]

Issue #792, opened 2023-10-13
`nixpkgs.constructedByUs` assertion message references variable that doesn't exist


lilyball:
The `nixpkgs` module has an assertion for `nixpkgs.constructedByUs` whose message references a variable `darwinExpectedSystem` that does not exist. This does not cause an issue under normal evaluation, but if the message is ever evaluated, it will throw an error.

https://github.com/LnL7/nix-darwin/blob/8b6ea26d5d2e8359d06278364f41fbc4b903b28a/modules/nix/nixpkgs.nix#L326-L333

This was introduced in PR #723. I noticed this when I decided to print the `assertions` config.

```console
> nix eval .#darwinConfigurations.$(hostname).config.assertions
error:
       … while evaluating the attribute 'message'

         at /nix/store/5gybvvfjyd29q9n4grrm7165vznl739b-source/modules/nix/nixpkgs.nix:331:11:

          330|           assertion = cfg.constructedByUs -> !hasPlatform -> cfg.system == pkgsSystem;
          331|           message = "The nix-darwin nixpkgs.pkgs option was set to a Nixpkgs invocation that compiles to target system ${pkgsSystem} but nix-darwin was configured for system ${darwinExpectedSystem} via nix-darwin option nixpkgs.system. The nix-darwin system settings must match the Nixpkgs target system.";
             |           ^
          332|         }

       error: undefined variable 'darwinExpectedSystem'

       at /nix/store/5gybvvfjyd29q9n4grrm7165vznl739b-source/modules/nix/nixpkgs.nix:331:177:

          330|           assertion = cfg.constructedByUs -> !hasPlatform -> cfg.system == pkgsSystem;
          331|           message = "The nix-darwin nixpkgs.pkgs option was set to a Nixpkgs invocation that compiles to target system ${pkgsSystem} but nix-darwin was configured for system ${darwinExpectedSystem} via nix-darwin option nixpkgs.system. The nix-darwin system settings must match the Nixpkgs target system.";
             |                                                                                                                                                                                 ^
          332|         }
```

niklasravnsborg:
I've run into this error just now. Don't exactly know why...

PR #1083, opened 2024-09-17
flake: match NixOS definition of `nixpkgs.flake.source`

emilazy:
Closes: #1082

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -27,7 +27,7 @@
             }
             ++ [ ({ lib, ... }: {
               nixpkgs.source = lib.mkDefault nixpkgs;
-              nixpkgs.flake.source = lib.mkDefault nixpkgs;
+              nixpkgs.flake.source = lib.mkDefault nixpkgs.outPath;
 
               system.checks.verifyNixPath = lib.mkDefault false;
 


emilazy:
This is still wrong because Nixpkgs isn’t necessarily from our `nixpkgs` input at all and I forgot all the lessons I learned in https://github.com/LnL7/nix-darwin/pull/723 and everything is sad and I need to do something else here. :(

sellout:
FWIW, my configuration builds successfully with this change. Good luck with finding a solution that satisfies you!

Enzime:
How does this PR relate to #1076?

antoineco:
I'm admitedly confused about this one. The error was

> `'nix.registry.nixpkgs.to.path' is defined multiple times`

but all this PR does is type the value of `nixpkgs.flake.source` from a flake attrset to a string representation of its path in the Nix store.

How does this prevent said option to be set multiple times?

antoineco (commented):
I tried this branch with an overriden `nixpkgs`:

```nix
{
  inputs = {
    nixpkgs.url = "https://flakehub.com/f/NixOS/nixpkgs/0.2405"; # nixos-24.05

    nix-darwin = {
      url = "github:emilazy/nix-darwin/push-zovpmlzlzvvm";
      inputs.nixpkgs.follows = "nixpkgs";
    };
  };
  outputs =
    {
      self,
      nixpkgs,
      nix-darwin,
    }:
    {
      darwinConfigurations = {
        myhost = nix-darwin.lib.darwinSystem {
          modules = [
            {
              nixpkgs.hostPlatform = "aarch64-darwin";
              system.stateVersion = 5;

              nix.registry.nixpkgs.flake = nixpkgs;
            }
          ];
        };
      };
    };
}
```

No error for me 👍 The resulting registry contains the expected entry:

```json
{
  "flakes": [
    {
      "exact": true,
      "from": {
        "id": "nixpkgs",
        "type": "indirect"
      },
      "to": {
        "lastModified": 1725930920,
        "narHash": "sha256-RVhD9hnlTT2nJzPHlAqrWqCkA7T6CYrP41IoVRkciZM=",
        "path": "/nix/store/20yis5w6g397plssim663hqxdiiah2wr-source",
        "rev": "44a71ff39c182edaf25a7ace5c9454e7cba2c658",
        "revCount": 635017,
        "type": "path"
      }
    }
  ],
  "version": 2
}
```

emilazy:
Sorry for the dramatic reaction, I have just been repeatedly burned by Nixpkgs mismatch issues :)

I will try to explain what’s going on here and what issues there still are tomorrow.

emilazy:
Didn’t end up getting around to writing down all the moving parts here today; will try to do so in the next couple days. The tl;dr is that this PR won’t make anything worse and should be okay to merge, but the whole way this is set currently is quite incorrect in a way that can result in pretty bad outcomes (e.g. pinning a vulnerable version of Nixpkgs even though your system uses a good one, or even getting a different release branch entirely). I think the upstream NixOS version of this is also broken in a similar but less likely to trigger way (due to the fact that it’s in the Nixpkgs repository itself). I’m still thinking about the best way to resolve the issues; https://github.com/LnL7/nix-darwin/pull/1076 is not strictly related, but could potentially also use some rethinking depending on what ends up seeming best (and I should adjust it to fix https://github.com/LnL7/nix-darwin/issues/792 anyway).

Enzime (approved):

PR #1084, opened 2024-09-17
checks: show Sequoia migration commands for other installers

emilazy:
This is hopefully largely cosmetic, but should increase user confidence in the safety and effectiveness of the migration, and the Determinate Systems and Lix installer commands will also update the `/nix/receipt.json` files to match the changes made. They cannot properly handle the format of each other’s receipts, so we need to detect which was used.

### DIFF
--- a/modules/system/checks.nix
+++ b/modules/system/checks.nix
@@ -72,16 +72,24 @@ let
         printf >&2 'compatibility with macOS Sequoia 15. Your _nixbld1 user currently has\n'
         printf >&2 'UID %d rather than the new default of 351.\n' "$firstBuildUserID"
         printf >&2 '\n'
-        printf >&2 'You can automatically migrate your users using the following script\n'
-        printf >&2 'from the Nix repository:\n'
+        printf >&2 'You can automatically migrate the users with the following command:\n'
         printf >&2 '\n'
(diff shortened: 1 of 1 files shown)


modules/system/checks.nix
```diff
@@ -72,16 +72,24 @@ let
...
-        printf >&2 'Systems installer or are using Lix. If you are comfortable using the\n'
-        printf >&2 'script without review, you can run:\n'
-        printf >&2 '\n'
-        printf >&2 "    curl --proto '=https' --tlsv1.2 -sSf -L https://github.com/NixOS/nix/raw/master/scripts/sequoia-nixbld-user-migration.sh | bash -\n"
+        if [[ -e /nix/receipt.json ]]; then
+            if
+                ${pkgs.jq}/bin/jq --exit-status \
+                'try(.planner.settings | has("enable_flakes"))' \
+                /nix/receipt.json \
+                >/dev/null
```
Samasaur1, line 82: Should `stderr` be redirected to `/dev/null` as well?

emilazy: It shouldn’t print anything to standard error unless `/nix/receipt.json` contains malformed JSON. I figured that’s a messed‐up enough situation that you might want some signal of it.

Enzime (approved):
LGTM

PR #1088, opened 2024-09-22
launchd: only support setting `command` to a list

Enzime:
@emilazy should I just deprecate setting `command` to a string with a warning?

### DIFF
--- a/modules/launchd/default.nix
+++ b/modules/launchd/default.nix
@@ -45,8 +45,9 @@ let
         };
 
         command = mkOption {
-          type = types.either types.str types.path;
-          default = "";
+          type = types.listOf (types.either types.str types.path);
+          default = [ ];
+          apply = escapeShellArgs;
           description = "Command executed as the service's main process.";
         };
 
@@ -55,15 +56,6 @@ let
           default = "";
           description = "Shell commands executed as the service's main process.";
         };
-
-        # preStart = mkOption {
-        #   type = types.lines;
-        #   default = "";
-        #   description = ''
-        #     Shell commands executed before the service's main process
-        #     is started.
-        #   '';
-        # };
 
         serviceConfig = mkOption {
           type = types.submodule launchdConfig;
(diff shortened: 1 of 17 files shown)


emilazy:
Deprecating that seems reasonable, but I’m a bit worried that it means the obvious translation of `prog $a` to `[ "prog" "$a" ]` behaves differently because of the escaping. I guess those cases can use `script` instead? Maybe we should use another name for this variant and just deprecate `command` entirely?

Enzime:
I was thinking of changing the types to any options that set extra arguments to lists as well so that we can run them through `lib.escapeShellArgs` as well

modules/launchd/default.nix
```diff
@@ -80,11 +72,11 @@ let
       };
 
       config = {
-        command = mkIf (config.script != "") (pkgs.writeScript "${name}-start" ''
+        command = mkIf (config.script != "") [ (pkgs.writeScript "${name}-start" ''
```
emilazy, line 75: Should we add a `wait4path` in here? (I’m fine if you’d rather leave that to future work.)

modules/services/autossh.nix
```diff
@@ -91,7 +92,7 @@ in
               # How often AutoSSH checks the network, in seconds
               environment.AUTOSSH_POLL="30";
 
-              command = "${pkgs.autossh}/bin/autossh -M ${toString mport} ${s.extraArguments}";
+              command = [ (lib.getExe pkgs.autossh) "-M" "${toString mport}" ] ++ s.extraArguments;
```
emilazy, line 95: Looks like this is the only place where we pass arbitrary user arguments in this field. After this change users will have to adjust their configurations and wouldn’t be able to reference environment variables here any more. I’m okay with it if you are, but I’m guessing we might have services for which this is undesirable in future.

emilazy (commented):
Looks reasonable to me. There’s a test failure; do you know if it’s related?

modules/launchd/default.nix
Enzime: I don't think it's necessary as `serviceConfig.ProgramArguments` already does `wait4path /nix/store`

modules/services/autossh.nix
Enzime: In those situations, I think we should direct them to set `serviceConfig.ProgramArguments` directly (possibly with `lib.mkAfter`)

I did a quick GitHub Code search and didn't see any configs that used environment variables for `services.autossh.extraArguments`:

https://github.com/search?utf8=✓&q=path%3A*.nix+services+autossh+extraArguments&type=Code

Enzime:
Closing this as I don't have the time to keep working on this

PR #1089, opened 2024-09-23
Update nextdns to use `command`

niklasravnsborg:
Following up on #1052, I updated the nextdns service to use `command` in order for it to wait for the nix store to be available.


niklasravnsborg:
@Enzime Just wanted to bump this quickly... :)

Enzime (approved):

PR #1090, opened 2024-09-23
launchd: allow WatchPaths to contain `~`

Enzime:


Enzime:
@emilazy can you take a look at another one liner?

Issue #1092, opened 2024-09-24
Karabiner elements stopped working after flake update, with:  `virtual_hid_device_service_client connect_failed: No such file or directory`


archie-judd:
Hi,

I recently updated my Nix Flake, and Karabiner Elements stopped working. Darwin is pointing at pointing Nixpkgs 24.05.

I checked the log in the UI and see repeated lines of `[info] [grabber] virtual_hid_device_service_client_ connect_failed: No such file or directory` after initialisation. This is the location of the error in Karabiner's [source code](https://github.com/pqrs-org/Karabiner-Elements/blob/9583ba39436dbb8fb6feb06fa60f96e297edddb7/src/core/grabber/include/grabber/device_grabber.hpp#L104-L106) and here is my log in the Karabiner Elements UI:

```
[2024-09-24 13:31:52.855] [info] [session_monitor] session_monitor_receiver_client is started.
[2024-09-24 13:31:52.861] [info] [grabber] current_console_user_id: 501
[2024-09-24 13:31:52.861] [info] [session_monitor] on_console_changed: true
[2024-09-24 13:31:52.861] [info] [session_monitor] session_monitor_receiver_client is connected.
[2024-09-24 13:31:52.861] [info] [session_monitor] connected
[2024-09-24 13:31:52.862] [info] [grabber] receiver is terminated
[2024-09-24 13:31:52.862] [info] [grabber] receiver is initialized
[2024-09-24 13:31:52.862] [info] [grabber] receiver: bound
[2024-09-24 13:31:52.862] [info] [grabber] receiver: chown socket: 501
[2024-09-24 13:31:54.081] [error] [observer] grabber_client error: Connection reset by peer
[2024-09-24 13:31:54.081] [info] [observer] grabber_client is closed.
[2024-09-24 13:31:55.169] [info] [observer] grabber_client is connected.
[2024-09-24 13:31:55.169] [info] [observer] rescan devices...
[2024-09-24 13:31:58.774] [info] [console_user_server] on_console_changed: on_console:true
[2024-09-24 13:31:58.776] [info] [console_user_server] grabber_client is started.
[2024-09-24 13:31:58.777] [info] [console_user_server] grabber_client is connected.
[2024-09-24 13:31:58.826] [info] [grabber] karabiner_console_user_server is connected.
[...]

PR #1093, opened 2024-09-28
fix: karabiner elements virtualhiddeviceclient error

will-lol:
- command needed to be quoted
- resolves https://github.com/LnL7/nix-darwin/issues/1092
- works when tested on my machine. would be good if someone who had this issue could also give it a spin


Enzime (approved):
LGTM

Knew something like this would happen, definitely need to finish off #1088 soon

PR #1094, opened 2024-09-29
Adding option for slow-motion-allowed

rogeruiz:
This patch was inspired by a recent [Daring Fireball post][l] where I was
reminded about this feature and how it's missing from being able to configure it
in Nix-Darwin.

The command I had run to get this working is outlined in the post & below.

```sh
default write com.apple.dock slow-motion-allowed -bool YES;
```

[l]: https://daringfireball.net/linked/2024/09/28/hidden-pref-to-restore-slow-motion-dock-minimizing-on-macos

### DIFF
--- a/modules/system/defaults/dock.nix
+++ b/modules/system/defaults/dock.nix
@@ -147,6 +147,14 @@ in {
       '';
     };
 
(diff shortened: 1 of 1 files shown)


Enzime (approved):
LGTM

Thanks for the contribution

PR #1097, opened 2024-09-30
programs/bash: move to completion.*

isabelroses:
a port of https://github.com/NixOS/nixpkgs/pull/291552 for darwin

### DIFF
--- a/modules/programs/bash/default.nix
+++ b/modules/programs/bash/default.nix
@@ -7,6 +7,10 @@ let
 in
 
 {
+  imports = [
+    (mkRenamedOptionModule [ "programs" "bash" "enableCompletion" ] [ "programs" "bash" "completion" "enable" ])
+  ];
+
   options = {
 
     programs.bash.enable = mkOption {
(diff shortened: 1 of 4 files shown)


modules/programs/bash/default.nix
```diff
@@ -21,14 +25,18 @@ in
...
-      description = ''
-        Enable bash completion for all interactive bash shells.
+    programs.bash.completion = {
+      enable = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Enable bash completion for all interactive bash shells.
+
+          NOTE. This doesn't work with bash 3.2, which is the default on macOS.
```
Enzime, line 35: Is this still the default on macOS?

emilazy: macOS will never ship a newer Bash since Apple refuses to touch GPLv3, though it’s not the default *login shell* any more.

bestlem: No bash is not the default and has not been for 10? years. The default now is zsh

Enzime:
```suggestion
          NOTE: This doesn't work with bash 3.2, which is installed by default on macOS by Apple.
```

Enzime (changes requested):

Enzime (approved):
LGTM

Thanks for the contribution

PR #1101, opened 2024-10-09
defaults: don't output Dock PID

Enzime:


Enzime:
@domenkozar @emilazy if you could take a look at this one liner, that'd be great

emilazy (approved):
Whoops, this got lost in my inbox. LGTM

PR #1105, opened 2024-10-15
add FlashDateSeparators for time separator config

mmenanno:
This adds the option to flash the time separators. Found the details around it here https://macos-defaults.com/menubar/flashdateseparators.html#set-to-false-default-value and was able to get it working in my config under:

```
CustomUserPreferences = {
  "com.apple.menuextra.clock" = {
    FlashDateSeparators = true;
  };
};
```

With this change you should now be able to do:

```
menuExtraClock = {
  FlashDateSeparators = true;
};
```

### DIFF
--- a/modules/system/defaults/clock.nix
+++ b/modules/system/defaults/clock.nix
@@ -65,5 +65,13 @@ with lib;
       '';
     };
 
+    system.defaults.menuExtraClock.FlashDateSeparators = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        When enabled, the time separators will flash on and off each second.
+      '';
+    };
+
   };
 }


Craz1k0ek (approved):

mmenanno:
@Craz1k0ek any idea why two of the tests are failing for me? The error message is unclear to me.

Craz1k0ek:
@mmenanno clueless. Might have to do with the `default` value. You've set it to `false`, but `null` might be better

mmenanno:
Looks like that fixed it ✅ There is a non-required test that failed, but I believe that is just a flaky test as it is complaining about something unrelated.

Craz1k0ek (approved):

PR #1106, opened 2024-10-16
module: add aerospace service

NickHu:
This packages the [AeroSpace window manager](https://github.com/nikitabobko/AeroSpace) as a service.

### DIFF
--- a/modules/services/aerospace/default.nix
+++ b/modules/services/aerospace/default.nix
@@ -1,0 +1,156 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+with lib;
+
+let
+  cfg = config.services.aerospace;
+
+  format = pkgs.formats.toml { };
+  configFile = format.generate "aerospace.toml" cfg.settings;
+in
+
+{
+  options = with types; {
+    services.aerospace = {
+      enable = mkEnableOption "AeroSpace window manager";
+
+      package = mkOption {
+        type = types.path;
+        default = pkgs.aerospace;
+        description = "The AeroSpace package to use.";
+      };
+
+      settings = mkOption {
+        type = submodule {
+          freeformType = format.type;
+          options = {
+            start-at-login = mkOption {
+              type = addCheck bool (b: !false || !cfg.enable);
+              default = false;
+              description = "Do not start AeroSpace at login. (Managed by launchd instead)";
+            };
+            after-login-command = mkOption {
+              type = addCheck (listOf str) (l: l == [ ] || !cfg.enable);
+              default = [ ];
+              description = "Do not use AeroSpace to run commands after login. (Managed by launchd instead)";
+            };
+            after-startup-command = mkOption {
+              type = addCheck (listOf str) (l: l == [ ] || !cfg.enable);
+              default = [ ];
+              description = "Do not use AeroSpace to run commands after startup. (Managed by launchd instead)";
+            };
+            enable-normalization-flatten-containers = mkOption {
+              type = bool;
+              default = true;
+              description = "Containers that have only one child are \"flattened\".";
+            };
+            enable-normalization-opposite-orientation-for-nested-containers = mkOption {
+              type = bool;
+              default = true;
+              description = "Containers that nest into each other must have opposite orientations.";
+            };
+            accordion-padding = mkOption {
+              type = int;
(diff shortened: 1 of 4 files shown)


modules/services/aerospace/default.nix
```diff
@@ -0,0 +1,156 @@
...
+{
+  options = with types; {
+    services.aerospace = {
+      enable = mkEnableOption "AeroSpace window manager";
+
+      package = mkOption {
+        type = types.path;
+        default = pkgs.aerospace;
+        description = "The AeroSpace package to use.";
+      };
```
Enzime, line 26:
```suggestion
      package = lib.mkPackageOption pkgs "aerospace" { };
```

modules/services/aerospace/default.nix
```diff
@@ -0,0 +1,156 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+with lib;
```
Enzime, line 8:
```suggestion
```

modules/services/aerospace/default.nix
```diff
@@ -0,0 +1,156 @@
...
+        default = pkgs.aerospace;
+        description = "The AeroSpace package to use.";
+      };
+
+      settings = mkOption {
+        type = submodule {
+          freeformType = format.type;
+          options = {
+            start-at-login = mkOption {
+              type = addCheck bool (b: !false || !cfg.enable);
```
Enzime, line 33: Doesn't that function always equal `true`?

modules/services/aerospace/default.nix
```diff
@@ -0,0 +1,156 @@
...
+            after-login-command = mkOption {
+              type = addCheck (listOf str) (l: l == [ ] || !cfg.enable);
+              default = [ ];
+              description = "Do not use AeroSpace to run commands after login. (Managed by launchd instead)";
+            };
+            after-startup-command = mkOption {
+              type = addCheck (listOf str) (l: l == [ ] || !cfg.enable);
+              default = [ ];
+              description = "Do not use AeroSpace to run commands after startup. (Managed by launchd instead)";
+            };
```
Enzime, line 46: Why do we want to prevent users from configuring `aerospace` to run commands after logging in?

modules/services/aerospace/default.nix
Enzime: Instead of implementing this with `addCheck`

I think it would make more sense to implement with an assertion

modules/services/aerospace/default.nix
```diff
@@ -0,0 +1,156 @@
...
+    (mkIf (cfg.enable) {
+      environment.systemPackages = [ cfg.package ];
+
+      launchd.user.agents.aerospace.serviceConfig = {
+        ProgramArguments =
+          [ "${cfg.package}/Applications/AeroSpace.app/Contents/MacOS/AeroSpace" ]
+          ++ optionals (cfg.settings != { }) [
+            "--config-path"
+            "${configFile}"
+          ];
```
Enzime, line 150:
```suggestion
        command = "${cfg.package}/Applications/AeroSpace.app/Contents/MacOS/AeroSpace" + (lib.optionalString (cfg.settings != { }) " --config-path ${configFile}");
```

Using `command` over `ProgramArguments` as it ensures that this service doesn't run before the Nix Store is mounted

modules/services/aerospace/default.nix
```diff
@@ -0,0 +1,156 @@
...
+          AeroSpace configuration, see
+          <link xlink:href="https://nikitabobko.github.io/AeroSpace/guide#configuring-aerospace"/>
+          for supported values.
+        '';
+      };
+    };
+  };
+
+  config = mkMerge [
+    (mkIf (cfg.enable) {
```
Enzime, line 141:
```suggestion
  config = mkIf cfg.enable {
```

Enzime (changes requested):

modules/services/aerospace/default.nix
NickHu: These commands only run if aerospace starts itself (via the `start-at-login` option), as opposed to via `launchd`

NickHu:
@Enzime is there anything else to change?

Enzime (approved):
LGTM

Thanks for the contribution

PR #1107, opened 2024-10-17
Updated defaults

Craz1k0ek:
* Added some more defaults settings

### DIFF
--- a/modules/system/defaults/finder.nix
+++ b/modules/system/defaults/finder.nix
@@ -41,6 +41,15 @@ in
       '';
     };
 
+    system.defaults.finder.FXRemoveOldTrashItems = mkOption {
+      type = types.nullOr types.bool;
+      default = null;
+      description = ''
+        Remove items in the trash after 30 days.
+        The default is false.
+      '';
+    };
+
     system.defaults.finder.FXPreferredViewStyle = mkOption {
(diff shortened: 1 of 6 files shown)


Enzime (changes requested):
Can you update the tests?

modules/system/defaults/finder.nix
```diff
@@ -38,6 +38,15 @@ with lib;
       '';
     };
 
+    system.defaults.finder.FXRemoveOldTrashItems = mkOption {
+      type = types.nullOr types.bool;
+      default = null;
+      description = ''
+        Remove items in the bin after 30 days.
```
Samasaur1, line 45: Should the docs call it the trash instead of the bin?

Craz1k0ek:
> Can you update the tests?

@Enzime Not sure what I need to update to make the test succeed. Error message is incomprehensible to me

Enzime:
Can you rebase your PR on master? It should fix the failing test, and then can you add some more tests to test the new options you added?

Craz1k0ek:
@Enzime not sure why the tests are failing. Anything wrong in my syntax? The error message is so poor, I can't make anything of it

Enzime:
This has been superseded by #1163

PR #1110, opened 2024-10-21
Run `brew` as regular user instead of as root

### LINKED ISSUE #1047: Cannot use system.defaults when running darwin-rebuild switch as non-root user
I'm running into an issue where I get an error if I try to set any system defaults as a non-root user. For example, trying to set `system.defaults.dock.autohide = true` (this is the only default I'm setting) causes this error:
- `defaults[41585:2126857] Could not write domain com.apple.dock; exiting`

Unsurprisingly, this is same error returned if I were to run run `defaults write com.apple.dock autohide true` without sudo. I'm able to use `sudo darwin-rebuild switch` to apply the defaults. The real issue comes in when I try to use homebrew since it no longer supports being ran as root, so running `sudo darwin-rebuild switch` returns this error:
- `Error: Running Homebrew as root is extremely dangerous and no longer supported.
As Homebrew does not drop privileges on installation you would be giving all
build scripts full access to your system.`

I was able to isolate the conflict to homebrew and `system.defaults`:
- Commenting out my homebrew config and running `sudo darwin-rebuild switch` causes no issues and applies the defaults successfully. 
- Commenting out my defaults config instead and running `darwin-rebuild switch` causes no issues and installs the homebrew packages successfully. 

I'm able to manually run `sudo defaults write` and have granted Full Disk Access to my terminal application (kitty), but I also have the issue in Terminal and Alacritty when running `darwin-rebuild switch` without sudo.

Neved4:
Fixes #1047

This makes `brew` run commands as a normal user, dropping privileges and avoiding brew errors, see: https://github.com/LnL7/nix-darwin/issues/1047#issue-2485414447.

**Note**: Would appreciate guidance if there's a better way to write this using Nix, or if the current way is correct in the context of `nix-darwin`.

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -135,15 +135,35 @@ let
     };
 
     config = {
-      brewBundleCmd = concatStringsSep " " (
-        optional (!config.autoUpdate) "HOMEBREW_NO_AUTO_UPDATE=1"
-        ++ [ "brew bundle --file='${brewfileFile}' --no-lock" ]
-        ++ optional (!config.upgrade) "--no-upgrade"
-        ++ optional (config.cleanup == "uninstall") "--cleanup"
-        ++ optional (config.cleanup == "zap") "--cleanup --zap"
-        ++ config.extraFlags
-      );
-    };
+      let
+        runBrewCmd = ''
+          RunBrewCmd() {
+            case "\$(id -u)" in
+              0)
+                sudo -u "\$SUDO_USER" brew "\$@"
+                ;;
+              *)
+                brew "\$@"
+                ;;
+            esac
+          }
+
+          RunBrewCmd
+        '';
+      in
+      {
+        brewBundleCmd = concatStringsSep " " (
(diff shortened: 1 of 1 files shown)


emilazy:
This is already part of the per‐user activation scripts. You’re probably running `sudo darwin-rebuild`, which is not supported (although it will be supported once user activation is removed and then you’ll have to specify the user to run Homebrew as).

Neved4:
@emilazy Thx Emily! I suggest closing this one and https://github.com/LnL7/nix-darwin/issues/1047 until it's supported, cheers 💙

emilazy:
Looks like the non‐sudo part of https://github.com/LnL7/nix-darwin/issues/1047 may be an actual bug (that I don’t understand the cause of).

FWIW, here is the draft of The Plan™ around user activation: https://github.com/LnL7/nix-darwin/pull/1016#issuecomment-2285910068. I am focused on the Nixpkgs 24.11 release right now but hopefully there will be movement on it after that.

Thanks for the contribution in any case! We will hopefully match NixOS better here soon; `sudo darwin-rebuild switch` should be the way to go.

Neved4:
Thank you!

PR #1112, opened 2024-10-22
ci: don't run tests twice for PRs from forks

Enzime:
By only triggering on pushes to `master`, when users push to non-`master` branches on forks, this workflow won't get triggered and will only get triggered when users make a PR to the main repo.


emilazy (approved):
LGTM.

BTW `macos-12` is getting removed in a few months, we should really bump the OS version we’re using.

PR #1113, opened 2024-10-22
ci: update Nix to match versions in nixpkgs

Enzime:


emilazy (approved):
LGTM, thanks. Any idea what’s going on with that existing test failure?

Enzime:
I started investigating the test failure earlier but it wasn't super clear what the issue was

I suspect there's a commit that broke it in `nixpkgs-unstable`, I'll try to bisect to see if I can figure it out

Maybe it's due to the Nix 2.18 -> 2.24 bump

PR #1115, opened 2024-10-22
darwin-rebuild: Align usage description with implementation

PhilipBorgesen:
This change updates the usage description of `darwin-rebuild.sh` as follows:

 * Add options:

     * `--print-build-logs` / `-L`
     * `--no-registries`
     * `--commit-lock-file`
     * `--no-write-lock-file`
    * ` --override-input`

 * Specify that `--no-flake` and the flake related options are mutually exclusive.

 * Specify that `--keep-going` and `-k` are aliases.

 * Specify that `--keep-failed` and `-K` are aliases.

 * Correct the description of `--update-input`; only one value follows.

The `--no-build-hook` argument remains undocumented, like for `nix`.

### DIFF
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -9,11 +9,15 @@ showSyntax() {
   echo "               [--list-generations] [{--profile-name | -p} name] [--rollback]" >&2
   echo "               [{--switch-generation | -G} generation] [--verbose...] [-v...]" >&2
   echo "               [-Q] [{--max-jobs | -j} number] [--cores number] [--dry-run]" >&2
-  echo "               [--keep-going] [-k] [--keep-failed] [-K] [--fallback] [--show-trace]" >&2
-  echo "               [-I path] [--option name value] [--arg name value] [--argstr name value]" >&2
-  echo "               [--flake flake] [--no-flake] [--update-input input flake] [--impure]" >&2
-  echo "               [--recreate-lock-file] [--no-update-lock-file] [--refresh]" >&2
-  echo "               [--offline] [--substituters substituters-list] ..." >&2
+  echo "               [--keep-going | -k] [--keep-failed | -K] [--fallback] [--show-trace]" >&2
+  echo "               [--print-build-logs | -L] [--impure] [-I path]" >&2
+  echo "               [--option name value] [--arg name value] [--argstr name value]" >&2
+  echo "               [--no-flake | [--flake flake]" >&2
+  echo "                             [--commit-lock-file] [--recreate-lock-file]" >&2
+  echo "                             [--no-update-lock-file] [--no-write-lock-file]" >&2
+  echo "                             [--override-input input flake] [--update-input input]" >&2
+  echo "                             [--no-registries] [--offline] [--refresh]]" >&2
+  echo "               [--substituters substituters-list] ..." >&2
   exit 1
 }
 


pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -10,10 +10,12 @@ showSyntax() {
   echo "               [{--switch-generation | -G} generation] [--verbose...] [-v...]" >&2
   echo "               [-Q] [{--max-jobs | -j} number] [--cores number] [--dry-run]" >&2
   echo "               [--keep-going] [-k] [--keep-failed] [-K] [--fallback] [--show-trace]" >&2
+  echo "               [--print-build-logs] [-L]" >&2
```
Enzime, line 13:
```suggestion
  echo "               [{--print-build-logs | -L}]" >&2
```

pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -10,10 +10,12 @@ showSyntax() {
   echo "               [{--switch-generation | -G} generation] [--verbose...] [-v...]" >&2
   echo "               [-Q] [{--max-jobs | -j} number] [--cores number] [--dry-run]" >&2
   echo "               [--keep-going] [-k] [--keep-failed] [-K] [--fallback] [--show-trace]" >&2
```
Enzime, line 12:
```suggestion
  echo "               [{--keep-going, -k} {--keep-failed, -K} [--fallback] [--show-trace]" >&2
```

Enzime (changes requested):

pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -10,10 +10,12 @@ showSyntax() {
...
+  echo "               [--print-build-logs] [-L]" >&2
   echo "               [-I path] [--option name value] [--arg name value] [--argstr name value]" >&2
-  echo "               [--flake flake] [--no-flake] [--update-input input flake] [--impure]" >&2
-  echo "               [--recreate-lock-file] [--no-update-lock-file] [--refresh]" >&2
-  echo "               [--offline] [--substituters substituters-list] ..." >&2
+  echo "               [--no-flake | --flake flake] [--no-registries] [--commit-lock-file]" >&2
```
Enzime, line 15:
```suggestion
  echo "               [--no-flake | [--flake flake] [--no-registries] [--commit-lock-file] ...]" >&2
```

If you wanted to be even more correct, all the flake related flags could go inside there

PhilipBorgesen: I did so initially, but then I saw that `/etc/nix-darwin/flake.nix` (if it exists) will be used if `--flake` is not passed. That means that the flake related options do not depend on `--flake` being passed.

Enzime:
```suggestion
  echo "               [--no-flake | [[--flake flake] [--no-registries] [--commit-lock-file] ...]" >&2
```

With an extra set of square brackets around `--flake` to imply that it's optional

pkgs/nix-tools/darwin-rebuild.sh
PhilipBorgesen: You mean `[{--keep-going | -k}] [{--keep-failed | -K}]`, right?

Enzime: Actually, it seems like `-k` and `-K` only work for `nix-build` rather than `nix build`, so best not to change this

Enzime: see: https://github.com/NixOS/nix/issues/2105

pkgs/nix-tools/darwin-rebuild.sh
PhilipBorgesen: Please review my change, it's not your suggestion as-is but what I believe you meant.

pkgs/nix-tools/darwin-rebuild.sh
PhilipBorgesen: Well, that they are mutually exclusive does not mean that they always can be used.
The usage description does not detail which options may be used with which actions.

Similar would be the case for `--max-jobs` and `-j`, BTW.

Enzime: The syntax we're using for `[{ --long, -l }]` I'm understanding as equivalent flags, so I would rather just leave it as is

PhilipBorgesen: If compatibility is a concern between `nix build` and `nix-build`, then I suggest removing the usage description for the shorthands, to deprecate them. Only the long options can be used with both build commands.

Enzime: I'm happy to just leave them there as is and it actually matches `nixos-rebuild --help`

PhilipBorgesen: `nixos-rebuild --help` says
```
[{--keep-failed | -K}] [{--keep-going | -k}]
```
so I guess that means "no revert"?

To "leave them there as is" is ambiguous, meaning leave the change as is, or leaving the codebase as is.

PhilipBorgesen: Resolve the thread(s) if you are happy with what is.

Enzime: I just checked `nixos-rebuild --help` and I see:

```
[--cores  number] [--keep-going | -k] [--keep-failed | -K]
```

So I think let's match that

pkgs/nix-tools/darwin-rebuild.sh
Enzime: Yep that's good 👍

pkgs/nix-tools/darwin-rebuild.sh
PhilipBorgesen: Done

pkgs/nix-tools/darwin-rebuild.sh
```diff
@@ -9,11 +9,13 @@ showSyntax() {
...
+  echo "               [--keep-going | -k] [--keep-failed | -K] [--fallback] [--show-trace]" >&2
+  echo "               [--print-build-logs | -L]" >&2
   echo "               [-I path] [--option name value] [--arg name value] [--argstr name value]" >&2
-  echo "               [--flake flake] [--no-flake] [--update-input input flake] [--impure]" >&2
-  echo "               [--recreate-lock-file] [--no-update-lock-file] [--refresh]" >&2
-  echo "               [--offline] [--substituters substituters-list] ..." >&2
+  echo "               [--no-flake | [--flake flake] [--no-registries] [--commit-lock-file]" >&2
+  echo "               [--no-update-lock-file] [--no-write-lock-file] [--recreate-lock-file]" >&2
+  echo "               [--override-input input flake] [--update-input input]] [--impure]" >&2
+  echo "               [--refresh] [--offline] [--substituters substituters-list] ..." >&2
```
Enzime, line 18:
```suggestion
  echo "               [--refresh] [--offline] [--substituters substituters-list] ...]" >&2
```

Can you add a closing square bracket here for the flake options section?

PhilipBorgesen: The closing square bracket is currently after `[--update-input input]`.

I'm not familiar with the

* `--impure`
* `--refresh`
* `--offline`
* `--substituters`

`nix build` flags but I believe they are not flake-specific?

Enzime: `--offline` and `--refresh` are `nix build` only (running `nix-build` with them outputs `error: unrecognised flag`)

My mistake I didn't see the closing square bracket, could you reorder the flags so that those flags are with the rest of the flake specific flags?

It might also be a good idea to add start a new line after the closing square bracket for the flake flags section

PhilipBorgesen: Why if `--offline` treated as a build/metadata flag, but `--refresh` as a lock flag?

```
    --show-trace|--keep-going|--keep-failed|--verbose|-v|-vv|-vvv|-vvvv|-vvvvv|--fallback|--offline)
      extraMetadataFlags+=("$i")
      extraBuildFlags+=("$i")
      ;;
```
```
    -L|-vL|--print-build-logs|--impure|--recreate-lock-file|--no-update-lock-file|--no-write-lock-file|--no-registries|--commit-lock-file|--refresh)
      extraLockFlags+=("$i")
      ;;
```

PhilipBorgesen: I note that the `nix build` `--no-registries` option is deprecated in favor of `--no-use-registries`.

PhilipBorgesen: `--offline` and `--refresh` are now listed as flake options.

If you would like to make further changes, feel free to make the edit yourself.
You have permissions to make changes to the source branch.

Enzime (approved):
LGTM

Thanks for the contribution

Issue #1109, opened 2024-10-21
Update `security.pam.enableSudoTouchIdAuth` for Sonoma+


citizen428:
Since macOS 14 (Sonoma), one can modify `/etc/pam.d/sudo_local` which will survive OS upgrades. By default, there's a file called `/etc/pam.d/sudo_local.template` with the following contents:

```
# sudo_local: local config file which survives system update and is included for sudo
# uncomment following line to enable Touch ID for sudo
#auth       sufficient     pam_tid.so
```

This gets included in `/etc/pam.d/sudoers`:

```
auth       include        sudo_local
```

It seems like a good idea to use this file so Touch ID for sudo will be available also right after upgrades, without having to run `darwin-rebuild switch` first.

purcell:
I think this could be easily accomplished by making `file` [here](https://github.com/LnL7/nix-darwin/blob/a001f44cfc47164839eb61c6b1e7f4288813f7e8/modules/security/pam.nix#L20) conditional on `config.system.darwinVersion`.

citizen428:
Yes, that should work, I briefly played around with that idea earlier today. In the end I just settled for this and removed the setting from my `flake.nix`, which is GoogEnough™ for me.

```nix
environment.etc = {
  "pam.d/sudo_local".text = ''
    auth sufficient pam_tid.so
  '';
};
```

emilazy:
We should unconditionally generate `sudo_local` and retain only enough logic to add the `auth       include        sudo_local` line to `/etc/pam.d/sudo` when absent. See https://github.com/LnL7/nix-darwin/pull/1020 which I just realized I forgot to get around to reviewing again, oops :(

citizen428:
Ugh, I only searched issues, not PRs, sorry. But I like your suggestion @emilazy, I think that's the best way forward.

Enzime:
Duplicate of #784

PR #1116, opened 2024-10-23
fix: Use sudo_local instead of sudo when changing sudo settings

nipunlakshank:
As macOS resets the /etc/pam.d/sudo file after system updates, using /etc/pam.d/sudo_local file will avoid the need to rebuild nix-darwin everytime the system gets updated as it is not affected.

### DIFF
--- a/modules/security/pam.nix
+++ b/modules/security/pam.nix
@@ -17,7 +17,7 @@ let
   # should be deleted when the option is disabled.
   mkSudoTouchIdAuthScript = isEnabled:
   let
-    file   = "/etc/pam.d/sudo";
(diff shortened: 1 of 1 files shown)


emilazy:
This would break pre‐Sonoma versions of macOS; see https://github.com/LnL7/nix-darwin/issues/1109#issuecomment-2427106303.

Enzime:
Thanks for the contribution but this is a duplicate of #1020 which I am planning to review soon

nipunlakshank:
> Thanks for the contribution but this is a duplicate of #1020 which I am planning to review soon

Oh I didn't see that, sorry. Looking forward for it

PR #1117, opened 2024-10-23
users: prevent deleting the user calling `darwin-rebuild`

Enzime:
`sysadminctl -deleteUser` will only prevent you from deleting the current user if it's not the last admin and not the last secure token user, otherwise it will happily oblige.

### DIFF
--- a/modules/users/default.nix
+++ b/modules/users/default.nix
@@ -149,6 +149,15 @@ in
       echo "setting up users..." >&2
 
       deleteUser() {
+        # TODO: add `darwin.primaryUser` as well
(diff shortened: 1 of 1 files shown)


emilazy (dismissed):
This won’t make much sense once all activation is done hermetically as root, but seems good for now.

Enzime:
I think it will still make sense as most users will be running `sudo darwin-rebuild switch` so `$SUDO_USER` will still be set and may even be different to `darwin.primaryUser`

emilazy (approved):

PR #1118, opened 2024-10-24
users: ensure users' shells are installed

Enzime:


emilazy (approved):

WaytrueDeng:
still can't set fish as default shell

PR #1120, opened 2024-10-24
users: update properties for known users

Enzime:
I'm leaving `gid` and `isHidden` for later PRs

- [x] Don't allow changing home directory of current user and `root`

### DIFF
--- a/modules/system/activation-scripts.nix
+++ b/modules/system/activation-scripts.nix
@@ -86,6 +86,7 @@ in
       exit $_status
     '';
 
+    # FIXME: activationScripts.checks should be system level
     system.activationScripts.userScript.text = ''
       #! ${stdenv.shell}
       set -e
--- a/modules/system/shells.nix
+++ b/modules/system/shells.nix
@@ -14,9 +14,15 @@ in
       example = literalExpression "[ pkgs.bashInteractive pkgs.zsh ]";
       description = ''
         A list of permissible login shells for user accounts.
-        No need to mention `/bin/sh`
-        and other shells that are available by default on
-        macOS.
+
+        The default macOS shells will be automatically included:
+          - /bin/bash
+          - /bin/csh
+          - /bin/dash
+          - /bin/ksh
+          - /bin/sh
+          - /bin/tcsh
+          - /bin/zsh
       '';
       apply = map (v: if types.shellPackage.check v then "/run/current-system/sw${v.shellPath}" else v);
     };
--- a/modules/users/default.nix
+++ b/modules/users/default.nix
@@ -110,32 +110,44 @@ in
     # however currently that runs at user level activation as that runs before system level activation
     # TODO: replace `$USER` with `$SUDO_USER` when system.checks runs from system level
     system.checks.text = lib.mkAfter ''
-      requireFDA() {
-        fullDiskAccess=false
+      ensurePerms() {
+        homeDirectory=$(dscl . -read /Users/nobody NFSHomeDirectory)
+        homeDirectory=''${homeDirectory#NFSHomeDirectory: }
 
-        if cat /Library/Preferences/com.apple.TimeMachine.plist > /dev/null 2>&1; then
-          fullDiskAccess=true
-        fi
-
-        if [[ "$fullDiskAccess" != true ]]; then
-          printf >&2 '\e[1;31merror: users cannot be %s without Full Disk Access, aborting activation\e[0m\n' "$2"
-          printf >&2 'The user %s could not be %s as `darwin-rebuild` was not executed with Full Disk Access.\n' "$1" "$2"
-          printf >&2 '\n'
-          printf >&2 'Opening "Privacy & Security" > "Full Disk Access" in System Settings\n'
-          printf >&2 '\n'
-          # This command will fail if run as root and System Settings is already running
-          # even if System Settings was launched by root.
-          open "x-apple.systempreferences:com.apple.preference.security?Privacy_AllFiles"
-
+        if ! sudo dscl . -change /Users/nobody NFSHomeDirectory "$homeDirectory" "$homeDirectory" &> /dev/null; then
           if [[ -n "$SSH_CONNECTION" ]]; then
-            printf >&2 'Please enable Full Disk Access for programs over SSH by flipping\n'
-            printf >&2 'the switch for `sshd-keygen-wrapper`.\n'
+            printf >&2 '\e[1;31merror: users cannot be %s over SSH without Full Disk Access, aborting activation\e[0m\n' "$2"
+            printf >&2 'The user %s could not be %s as `darwin-rebuild` was not executed with Full Disk Access over SSH.\n' "$1" "$2"
+            printf >&2 'You can either:\n'
+            printf >&2 '\n'
+            printf >&2 '  grant Full Disk Access to all programs run over SSH\n'
+            printf >&2 '\n'
+            printf >&2 'or\n'
+            printf >&2 '\n'
+            printf >&2 '  run `darwin-rebuild` in a graphical session.\n'
+            printf >&2 '\n'
+            printf >&2 'The option "Allow full disk access for remote users" can be found by\n'
+            printf >&2 'navigating to System Settings > General > Sharing > Remote Login\n'
+            printf >&2 'and then pressing on the i icon next to the switch.\n'
+            exit 1
           else
-            printf >&2 'Please enable Full Disk Access for your terminal emulator by flipping\n'
-            printf >&2 'the switch in System Settings.\n'
+            # The TCC service required to change home directories is `kTCCServiceSystemPolicySysAdminFiles`
+            # and we can reset it to ensure the user gets another prompt
+            tccutil reset SystemPolicySysAdminFiles > /dev/null
+
+            if ! sudo dscl . -change /Users/nobody NFSHomeDirectory "$homeDirectory" "$homeDirectory" &> /dev/null; then
+              printf >&2 '\e[1;31merror: permission denied when trying to %s user %s, aborting activation\e[0m\n' "$2" "$1"
+              printf >&2 '`darwin-rebuild` requires permissions to administrate your computer,\n' "$1" "$2"
+              printf >&2 'please accept the dialog that pops up.\n'
+              printf >&2 '\n'
+              printf >&2 'If you do not wish to be prompted every time `darwin-rebuild updates your users,\n'
+              printf >&2 'you can grant Full Disk Access to your terminal emulator in System Settings.\n'
+              printf >&2 '\n'
+              printf >&2 'This can be found in System Settings > Privacy & Security > Full Disk Access.\n'
+              exit 1
+            fi
           fi
 
-          exit 1
         fi
       }
 
@@ -149,7 +161,7 @@ in
           exit 1
         fi
 
-        requireFDA "$1" deleted
+        ensurePerms "$1" delete
       }
 
       ${concatMapStringsSep "\n" (v: let
@@ -169,7 +181,7 @@ in
         u=$(id -u ${name} 2> /dev/null) || true
         if ! [[ -n "$u" && "$u" -ne "${toString v.uid}" ]]; then
           if [ -z "$u" ]; then
-            requireFDA ${name} created
+            ensurePerms ${name} create
           fi
 
           ${optionalString (v.home != null && v.name != "root") ''
@@ -211,7 +223,7 @@ in
           g=''${g#PrimaryGroupID: }
           if [[ "$g" -eq ${toString v.gid} ]]; then
             echo "deleting group ${v.name}..." >&2
-            dscl . -delete ${dsclGroup} 2> /dev/null
+            dscl . -delete ${dsclGroup}
           else
             echo "[1;31mwarning: existing group '${v.name}' has unexpected gid $g, skipping...[0m" >&2
           fi
@@ -245,7 +257,7 @@ in
         if [ -n "$g" ]; then
           if [ "$g" -gt 501 ]; then
             echo "deleting group ${name}..." >&2
-            dscl . -delete ${dsclGroup} 2> /dev/null
+            dscl . -delete ${dsclGroup}
           else
             echo "[1;31mwarning: existing group '${name}' has unexpected gid $g, skipping...[0m" >&2
           fi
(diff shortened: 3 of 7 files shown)


Enzime:
Don't allow changing home directory after creation for `$SUDO_USER` and `root`

Enzime:
@emilazy this is ready to review :)

modules/system/shells.nix
```diff
@@ -11,6 +11,17 @@ in
...
+        [
+          "/bin/bash"
+          "/bin/csh"
+          "/bin/dash"
+          "/bin/ksh"
+          "/bin/sh"
+          "/bin/tcsh"
+          "/bin/zsh"
+        ]
+      '';
```
emilazy, line 24: This is misleading since setting this value isn’t a NOP, should just clarify the description if necessary.

modules/users/default.nix
```diff
@@ -94,6 +94,12 @@ in
   };
 
   config = {
+    assertions = [
+      {
+        assertion = cfg.users ? root -> (cfg.users.root.home == null || cfg.users.root.home == "/var/root");
+        message = "`users.users.root.home` must be set to either `null` or `/var/root`.";
+      }
+    ];
```
emilazy, line 102: No objection to this, but any reason for this particular safety check? I guess it probably breaks things a bunch to change this?

modules/users/default.nix
```diff
@@ -245,10 +257,36 @@ in
...
             dscl . -create ${dsclUser} IsHidden ${if v.isHidden then "1" else "0"}
-            ${optionalString v.createHome "createhomedir -cu ${name}"}
           fi
-          # Always set the shell path, in case it was updated
-          dscl . -create ${dsclUser} UserShell ${lib.escapeShellArg (shellPath v.shell)}
+
+          # Always update all properties on users to keep them inline with configuration
+          dscl . -create ${dsclUser} PrimaryGroupID ${toString v.gid}
+          ${optionalString (v.description != null) "dscl . -create ${dsclUser} RealName ${lib.escapeShellArg v.description}"}
+          ${optionalString (v.shell != null) "dscl . -create ${dsclUser} UserShell ${lib.escapeShellArg (shellPath v.shell)}"}
```
emilazy, line 265: Is `dscl` not deprecated for these tasks? I’m a little worried about having this long‐term split where we’re mixing `sysadminctl` and `dscl` when they might behave differently.

modules/users/default.nix
```diff
@@ -245,10 +257,36 @@ in
...
+                printf >&2 'or remove it from your configuration.\n'
+                exit 1
+              fi
+            else
+              dscl . -create ${dsclUser} NFSHomeDirectory ${lib.escapeShellArg v.home}
+            fi
+          ''}
+          # This command doesn't do anything if the folder already exists which should be the case
+          # for $SUDO_USER and root, so it should be safe to run unconditionally.
+          ${optionalString v.createHome "createhomedir -cu ${name}"}
```
emilazy, line 289: I assume there’s no automated migration of files or anything, right? Do we want to be automating home directory changes, at least without a warning of some kind for this? We could split the difference by erroring out when the directory is wrong and printing commands to update it, which is an approach I like for avoiding automating “destructive” migrations while still making sure the activated configuration remains accurate.

modules/users/user.nix
```diff
@@ -65,10 +73,17 @@
...
+      default = null;
       example = literalExpression "pkgs.bashInteractive";
-      description = "The user's shell.";
+      description = ''
+        The user's shell. This defaults to `null`.
+
+        When this is set to `null`, the value is managed by macOS instead of
+        `nix-darwin`. This means if the user has not been created yet,
+        `sysadminctl` will be called without the `-shell` flag which means the
+        user will have a default shell of `/bin/zsh`.
```
emilazy, line 85: This is scary: it means that a previously‐secure configuration with the `/sbin/nologin` default can now result in a system that inadvertently lets a user log in. Can we make sure we use `/sbin/nologin` by default on creation?

emilazy (changes requested):

modules/users/default.nix
Enzime: https://ss64.com/mac/sysadminctl.html

`sysadminctl` only deals with creating, deleting users, changing passwords and secure token settings, as far as I can tell updating settings is still `dscl`'s responsibility

modules/users/user.nix
Enzime: Sure, I'll switch it to `/usr/bin/false` to match what macOS uses for its accounts

The end goal is the default will be determined by the type of user being created in a later PR

modules/system/shells.nix
Enzime: Fixed 👍

modules/users/default.nix
Enzime: The code that ensures `home` is the same between the config and the system is set to not run on `root`, so this is the only place that checks `root`'s `home` is correct

Enzime: I've added a comment to make the intention clearer and it should be more clear now that it's in a more relevant commit

modules/users/default.nix
Enzime: I've changed the PR to only run `createhomedir` on user creation (as `sysadminctl -addUser -home` won't create the home directory), I've removed the logic that updates `home` and added a check to ensure the `home` directory set in the config matches the system for all users.

Enzime:
@emilazy this should be ready to review again :)

The other change I've made is to move all the checking logic to `system.checks.text` to run first and avoid partial activations

Enzime:
Turns out `system.checks` runs in user activation (presumably because it needs to run first) not root activation...

So I've updated my checks code to handle running from user activation level

Enzime:
@emilazy should be good to review now :)

modules/system/activation-scripts.nix
```diff
@@ -86,6 +86,7 @@ in
       exit $_status
     '';
 
+    # FIXME: activationScripts.checks should be system level
```
emilazy, line 89:
```suggestion
    # FIXME: activationScripts should be system level
```
😆

modules/users/default.nix
```diff
@@ -94,10 +95,112 @@ in
...
+        message = "`users.users.root.home` must be set to either `null` or `/var/root`.";
+      }
+    ];
 
     users.gids = mkMerge gids;
     users.uids = mkMerge uids;
 
+    # NOTE: We put this in `system.checks` as we want this to run first to avoid partial activations
+    # however currently that runs at user level activation as that runs before system level activation
+    # TODO: replace `$USER` with `$SUDO_USER` when system.checks runs from system level
```
emilazy, line 112: FWIW I plan to make sure `$SUDO_USER` doesn’t leak through to avoid activating‐user dependencies creeping back in when user activation is gone. But that’s okay since we can just use `primaryUser` or I guess check currently logged‐in users or something.

modules/users/default.nix
```diff
@@ -94,10 +95,112 @@ in
...
+    system.checks.text = lib.mkAfter ''
+      requireFDA() {
+        fullDiskAccess=false
+
+        if cat /Library/Preferences/com.apple.TimeMachine.plist > /dev/null 2>&1; then
+          fullDiskAccess=true
+        fi
+
+        if [[ "$fullDiskAccess" != true ]]; then
+          printf >&2 '\e[1;31merror: users cannot be %s without Full Disk Access, aborting activation\e[0m\n' "$2"
```
emilazy, line 122: Wasn’t there a separate setting to manage users and groups? I avoid granting Full Disk Access to my terminal so this feels a little unfortunate, but better to check than not.

BTW, if we work on moving things out of activation scripts to launchd daemons as NixOS is doing we might be able to avoid this.

(Edit: Sorry, I didn’t realize all this was already present when reviewing! I should have made these comments when it was added. None of this is a blocker, given that.)

modules/users/default.nix
```diff
@@ -94,10 +95,112 @@ in
...
+          printf >&2 'The user %s could not be %s as `darwin-rebuild` was not executed with Full Disk Access.\n' "$1" "$2"
+          printf >&2 '\n'
+          printf >&2 'Opening "Privacy & Security" > "Full Disk Access" in System Settings\n'
+          printf >&2 '\n'
+          # This command will fail if run as root and System Settings is already running
+          # even if System Settings was launched by root.
+          open "x-apple.systempreferences:com.apple.preference.security?Privacy_AllFiles"
```
emilazy, line 129: A little magic for my tastes and will be weird on headless systems, probably better just to print instructions.

modules/users/default.nix
```diff
@@ -94,10 +95,112 @@ in
...
+
+          if [[ -n "$SSH_CONNECTION" ]]; then
+            printf >&2 'Please enable Full Disk Access for programs over SSH by flipping\n'
+            printf >&2 'the switch for `sshd-keygen-wrapper`.\n'
```
emilazy, line 133: This feels kind of dodgy, I’m not sure we should be recommending it without caveats.

modules/users/default.nix
```diff
@@ -94,10 +95,112 @@ in
...
+        fi
+
+        requireFDA "$1" deleted
+      }
+
+      ${concatMapStringsSep "\n" (v: let
+        name = lib.escapeShellArg v.name;
+        dsclUser = lib.escapeShellArg "/Users/${v.name}";
+      in ''
+        ${optionalString cfg.forceRecreate ''
```
emilazy, line 160: We decided we could just drop this for now, right? Might be best to do now to avoid adding more logic we have to verify the correctness of when we’d like to kill it off.

modules/users/default.nix
```diff
@@ -94,10 +95,112 @@ in
...
+              printf >&2 '\e[1;31merror: config contains the wrong home directory for %s, aborting activation\e[0m\n' ${name}
+              printf >&2 'nix-darwin does not support changing the home directory of existing users.
+              printf >&2 '\n'
+              printf >&2 'Please set:\n'
+              printf >&2 '\n'
+              printf >&2 '    users.users.%s.home = "%s";\n' ${name} "$homeDirectory"
+              printf >&2 '\n'
+              printf >&2 'or remove it from your configuration.\n'
+              exit 1
+            fi
```
emilazy, line 189: 💜

We could print instructions on how to manually update it? Not necessary for an MVP though.

modules/users/user.nix
```diff
@@ -65,10 +73,17 @@
     };
 
     shell = mkOption {
-      type = types.either types.shellPackage types.path;
-      default = "/sbin/nologin";
+      type = types.nullOr (types.either types.shellPackage types.path);
+      default = "/usr/bin/false";
```
emilazy, line 77: You can break your shell by starting to manage your user with nix-darwin unless you explicitly `null` this out, right? I was thinking we could just make `null` create users with `/usr/bin/false` but otherwise act the same way as the default. Not sure if that’s a good idea or not.

emilazy (approved):
Changes LGTM, thanks! Some comments but I’m happy for this to be merged as‐is.

modules/users/user.nix
Enzime: This is a good catch, I think this bug has been around since #818 and I'm not sure if anyone has been hit by it yet as it's currently gated behind `users.knownUsers` and my suspicion is that most users only add themselves to `users.knownUsers` when they want to update their shell as most other options are less useful.

I'll add the change to make `null` create users with `/usr/bin/false` and but still mean don't update this user's shell. This change should be superseded by adding a `type` option because then we could not pass the `-shell` flag to `sysadminctl` and it would set the correct shell based on whether `-roleAccount` was passed or not (alternatively we could determine the default passed on the `type` in nix-darwin as well)

modules/users/default.nix
Enzime: After further investigation, there's a TCC service called `kTCCServiceSystemPolicySysAdminFiles` which shows the message: `%s would like to administer your computer. Administration can include modifying passwords, networking, and system settings.`

I believe this service is a subset of Full Disk Access which is the reason why granting FDA will not prompt you for this TCC service.

It turns out this permission gets triggered when you try to change a user's home directory as that can allow bypassing TCC (as TCC has a per-user database in a user's home directory) see: https://wojciechregula.blog/post/change-home-directory-and-bypass-tcc-aka-cve-2020-27937/

The easiest way to trigger this prompt:

```
$ sudo dscl . -change /Users/_nixbld1 NFSHomeDirectory /var/empty /var/empty 
```

`-change` here is safe as it won't change the value unless the first argument matches the current value:

```
$ dscl . -read /Users/_nixbld1 NFSHomeDirectory
NFSHomeDirectory: /var/empty
$ sudo dscl . -change /Users/_nixbld1 NFSHomeDirectory /private/var/empty /tmp
<main> attribute status: eDSAttributeNotFound
<dscl_cmd> DS Error: -14134 (eDSAttributeNotFound)
$ dscl . -read /Users/_nixbld1 NFSHomeDirectory
NFSHomeDirectory: /var/empty
```

When running this command from a graphical session, you'll be able to see the prompt. However, if you run it from SSH, there will be no prompt and you'll get permission denied:

```
$ sudo dscl . -change /Users/_nixbld1 NFSHomeDirectory /var/empty 
<main> attribute status: eDSPermissionError
<dscl_cmd> DS Error: -14120 (eDSPermissionError)
```

On the `Remote Login` modal, there's a checkbox `Allow full disk access for remote users` which is equivalent to toggling FDA for `sshd-keygen-wrapper`. I chose to suggest enabling it in the FDA menu just because it's easier to open:

```
$ open "x-apple.systempreferences:com.apple.preferences.sharing?Services_RemoteLogin"
$ sleep 2
$ open "x-apple.systempreferences:com.apple.preferences.sharing?Services_RemoteLogin"
```

On newer versions of macOS, this deep link will work on the first try however on older versions it requires calling it twice with a sleep to make sure the Sharing page loaded before trying again.

Enzime: With this information, I think we can change the checking logic to request `kTCCServiceSystemPolicySysAdminFiles` and then if that fails, give the user a relevant message:

If they're running over SSH, tell them to enable FDA for SSH or that they'll need to not use SSH with `darwin-rebuild`.
If they're running locally, tell them to wait 1 minute before running `darwin-rebuild switch` again and make sure to accept the prompt and possibly mention turning on FDA for their terminal emulator.

modules/users/default.nix
Enzime: Updated the message to clarify what this permission does and give an alternative 👍

modules/users/default.nix
Enzime: I'll remove it in my next PR 👍

Enzime:
@emilazy this is ready for review :)

emilazy (approved):
Looks great, thank you for seeing this through! Feel free to merge.

My one remaining nit is that `sudo dscl . -change /Users/nobody NFSHomeDirectory "$homeDirectory" "$homeDirectory"` scares me a little and it’d be nice if we could just touch some meaningless `NixTestWhatever` field somewhere, maybe even not attached to a user at all. But I think it should probably be fine and I’m just being paranoid.

PR #1124, opened 2024-10-25
Cleanup release.nix

SuperSandro2000:


release.nix
```diff
@@ -4,12 +4,10 @@
 }:
 
 let
-  inherit (release) mapTestOn packagePlatforms pkgs all linux darwin;
+  inherit (release) mapTestOn packagePlatforms pkgs;
 
   system = "x86_64-darwin";
```
Enzime, line 9:
```suggestion
```

If you have the time or energy, would you be interested in removing the hardcoded `system` and running the tests on `aarch64-darwin` as well?

release.nix
```diff
@@ -61,7 +59,7 @@ let
     in
       buildFromConfig configuration (config: config.system.build.run-test);
 
-  release = import <nixpkgs/pkgs/top-level/release-lib.nix> {
+  release = import (nixpkgs + "/pkgs/top-level/release-lib.nix") {
```
Enzime, line 62: Out of curiosity, how are you accessing this from a flake/what are you using it for?

release.nix
SuperSandro2000: I can only run things on x86-64_linux and aarch64-linux 😅

release.nix
SuperSandro2000: I am trying to get hold of an options.json for https://search.nüschtos.de/ in https://github.com/NuschtOS/search.nuschtos.de/pull/22/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R98 but because of the hatdcoded x86-64_linux platform as noted above, I ended up copying the important bits from this file.

Enzime (approved):
Merging this with just the flake fix

Thanks for the contribution

SuperSandro2000:
The other code is completely unused and in a let in 😅

Enzime:
It’s okay, just leaving it there for when someone fixes tests not running on `aarch64-darwin`

PR #1125, opened 2024-10-25
Configure the folder that new Finder windows open

aschleck:


Enzime (changes requested):
Can you add tests?

aschleck:
Sure, done

modules/system/defaults/finder.nix
```diff
@@ -96,5 +96,13 @@ with lib;
       '';
     };
 
+    system.defaults.finder.NewWindowTarget = mkOption {
+      type = types.nullOr types.str;
```
Enzime, line 100:
```suggestion
      type = types.nullOr types.enum [ "PfHm" ... ];
```

If I understand correctly, only these values are possible:

https://gist.github.com/ChristopherA/98628f8cd00c94f11ee6035d53b0d3c6#file-macos-preferences-defaults-sh-L280-L286

modules/system/defaults/finder.nix
```diff
@@ -96,5 +96,13 @@ with lib;
...
+      default = null;
+      description = ''
+        Change the default folder shown in Finder windows. Use "PfHm" to default to the home folder.
+        The default is unset ("Recents").
+      '';
+    };
   };
```
Enzime, line 107: Could you also add `NewWindowTargetPath`? I'm assuming that's the `defaults` option when setting `NewWindowTarget` to `Other`

Enzime (changes requested):

modules/system/defaults/finder.nix
aschleck: Good call. I also added PfID which opens iCloud Drive

modules/system/defaults/finder.nix
aschleck: Done

modules/system/defaults/finder.nix
```diff
@@ -96,5 +96,42 @@ with lib;
...
+      type = types.nullOr types.str;
+      default = null;
+      description = ''
+        Sets the URI to open when NewWindowTarget is "PfLo". Spaces and similar characters must be
+        escaped.
+        Example: "file:///Users/foo/cats%20and%20dogs".
+        The default is the home folder.
+      '';
+    };
   };
```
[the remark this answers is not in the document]
aschleck: Thank you for making the suggested change here! Those make sense to me

modules/system/defaults/finder.nix
```diff
@@ -96,5 +96,42 @@ with lib;
...
+      type = types.nullOr (types.enum [
+        "PfAF"
+        "PfCm"
+        "PfDe"
+        "PfDo"
+        "PfHm"
+        "PfID"
+        "PfLo"
+        "PfVo"
+      ]);
```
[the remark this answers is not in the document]
aschleck: Had to explicitly check that key is non-null but this sounds good, thanks for suggesting

modules/system/defaults/finder.nix
```diff
@@ -2,6 +2,9 @@
 
 with lib;
```
Enzime, line 3:
```suggestion
```

modules/system/defaults/finder.nix
```diff
@@ -2,6 +2,9 @@
...
 
+let
+  cfg = config.system.defaults.finder;
+in
```
Enzime, line 7:
```suggestion
let
  inherit (lib) mkOption types ...;

  cfg = config.system.defaults.finder;
in
```

Enzime (changes requested):

modules/system/defaults/finder.nix
```diff
@@ -96,5 +99,54 @@ with lib;
...
+  };
+
+  config = {
+    assertions = [{
+      assertion = cfg.NewWindowTargetPath != null -> cfg.NewWindowTarget == "PfLo";
+      message = "`system.defaults.finder.NewWindowTarget` should be set to `Other` when `NewWindowTargetPath` is non-null.";
+    }
+    {
+      assertion = cfg.NewWindowTarget == "PfLo" -> cfg.NewWindowTargetPath != null;
+      message = "`system.defaults.finder.NewWindowTarget` should be non-null when `NewWindowTarget` is set to `Other`.";
```
dpomykala, line 149: Hey, I noticed a small typo in the message for the second assertion. I think it should be `NewWindowTargetPath` in the first part of the message (currently `NewWindowTarget` is mentioned twice).

```suggestion
      message = "`system.defaults.finder.NewWindowTargetPath` should be non-null when `NewWindowTarget` is set to `Other`.";
```

BTW. Great job with this feature. I'm looking forward to this PR!

modules/system/defaults/finder.nix
aschleck: Woah this is neat syntax, thanks for suggesting it

Enzime (approved):
LGTM

Thanks for the contribution

PR #1126, opened 2024-10-26
Add keepalive flag for emacs service

booxter:
Otherwise, the service will never recover after the daemon dies for any reason.


Enzime (approved):
LGTM

Thanks for the contribution

PR #1129, opened 2024-10-28
module: add prometheus-node-exporter service

mweinelt:
This PR adds a module for the Prometheus Node Exporter, which I want to use in the nixos.org infra.

I will test this shortly, but it is probably good enough for a review.

Options are similar to [NixOS](https://search.nixos.org/options?channel=24.05&show=services.prometheus.exporters.node.extraFlags&from=0&size=50&sort=relevance&type=packages&query=exporters.node), but I chose a different namespace, since what NixOS has is probably overkill.

### DIFF
--- a/modules/services/monitoring/prometheus-node-exporter.nix
+++ b/modules/services/monitoring/prometheus-node-exporter.nix
@@ -1,0 +1,117 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  inherit (lib)
+    concatStringsSep
+    escapeShellArgs
+    getExe
+    mkEnableOption
+    mkIf
+    mkOption
+    mkPackageOption
+    mkRemovedOptionModule
+    types
+  ;
+
+  cfg = config.services.prometheus.exporters.node;
+in {
+  imports = [
+    (mkRemovedOptionModule [ "services" "prometheus" "exporters" "node" "openFirewall" ] "No nix-darwin equivalent to this NixOS option.")
+    (mkRemovedOptionModule [ "services" "prometheus" "exporters" "node" "firewallFilter" ] "No nix-darwin equivalent to this NixOS option.")
+    (mkRemovedOptionModule [ "services" "prometheus" "exporters" "node" "firewallRules" ] "No nix-darwin equivalent to this NixOS option.")
+  ];
+
+  options = {
+    services.prometheus.exporters.node = {
+      enable = mkEnableOption "Prometheus Node exporter";
+
+      package = mkPackageOption pkgs "prometheus-node-exporter" { };
+
+      listenAddress = mkOption {
+        type = types.str;
+        default = "";
+        example = "0.0.0.0";
+        description = ''
+          Address where Node exporter exposes its HTTP interface. Leave empty to bind to all addresses.
+        '';
+      };
+
+      port = mkOption {
+        type = types.port;
+        default = 9100;
+        description = ''
+          Port where the Node exporter exposes its HTTP interface.
+        '';
+      };
+
+      extraFlags = mkOption {
+        type = types.listOf types.str;
+        default = [ ];
+        example = [ "--log.level=debug" ];
+        description = ''
+          Extra commandline options to pass to the Node exporter executable.
+        '';
+      };
+
+      enabledCollectors = mkOption {
+        type = types.listOf types.str;
+        default = [ ];
+        description = ''
+          Collectors to enable in addition to the ones that are [enabled by default](https://github.com/prometheus/node_exporter#enabled-by-default).
+        '';
+      };
+
+      disabledCollectors = mkOption {
+        type = types.listOf types.str;
+        default = [ ];
+        example = [ "boottime" ];
+        description = ''
+          Collectors to disable from the list of collectors that are [enabled by default](https://github.com/prometheus/node_exporter#enabled-by-default).
+        '';
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    users.users._prometheus-node-exporter = {
+      uid = config.ids.uids._prometheus-node-exporter;
+      gid = config.ids.gids._prometheus-node-exporter;
+      home = "/var/empty";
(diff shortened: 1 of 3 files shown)


modules/services/monitoring/prometheus-node-exporter.nix
```diff
@@ -0,0 +1,81 @@
...
+    mkIf
+    mkOption
+    mkPackageOption
+    types
+  ;
+
+  cfg = config.services.prometheus-node-exporter;
+in {
+  options = {
+    services.prometheus-node-exporter = {
```
emilazy, line 23: Let’s make this `services.prometheus.exporters.node` to align with NixOS and increase configuration portability. We’ll need `(lib.mkRemovedOptionModule [ "services" "prometheus" "exporters" "node" "openFirewall" ] "No nix-darwin equivalent to this NixOS option.")`, and the same for `firewallRules` and `firewallFilter`. No need to share implementation for now, I think, since NixOS has a bunch of layers of abstraction over their Prometheus exporters that we might not want to pull in immediately and it won’t be a breaking change if we do later.

modules/services/monitoring/prometheus-node-exporter.nix
```diff
@@ -0,0 +1,81 @@
...
+      enable = mkEnableOption "Prometheus Node exporter";
+
+      package = mkPackageOption pkgs "prometheus-node-exporter" { };
```
emilazy, line 26: No `package` option in NixOS (strangely?). I don’t consider the divergence a big deal, though, since NixOS should probably have it.

modules/services/monitoring/prometheus-node-exporter.nix
```diff
@@ -0,0 +1,81 @@
...
+
+      listenAddress = mkOption {
```
emilazy, line 28: I’m not sure why NixOS has both `port` and `listenAddress`. If one is (soft‐)deprecated then we can probably just keep the best practice. Otherwise maybe we should support both.

modules/services/monitoring/prometheus-node-exporter.nix
```diff
@@ -0,0 +1,81 @@
...
+        example = [ "boottime" ];
+        description = ''
+          Collectors to disable from the list of collectors that are [enabled by default](https://github.com/prometheus/node_exporter#enabled-by-default).
+        '';
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    launchd.daemons.prometheus-node-exporter = {
```
emilazy, line 65: I believe this will run as `root`. We need `user` and `group` from the NixOS module, and should default to something with a `_` in front to mark it as a role account. Unfortunately that means static allocation; take a look at `modules/services/hercules-ci-agent/default.nix` for an example. I can write the definitions for this if needed. cc @Enzime as well

emilazy (changes requested):

modules/services/monitoring/prometheus-node-exporter.nix
mweinelt: Generalization of the interface, but since node-exporter is written in Go it can work with specifications like `:9100`, which a normal BSD style sockaddr struct cannot.

modules/services/monitoring/prometheus-node-exporter.nix
mweinelt: Added user and group.

modules/services/monitoring/prometheus-node-exporter.nix
mweinelt: Kept the package option as is.

modules/services/monitoring/prometheus-node-exporter.nix
mweinelt: Done

mweinelt:
All requested changes

modules/services/monitoring/prometheus-node-exporter.nix
```diff
@@ -0,0 +1,115 @@
...
+      shell = "/usr/bin/false";
+      description = "System user for the Prometheus Node exporter";
+    };
+
+    users.groups._prometheus-node-exporter = {
+      gid = config.ids.gids._prometheus-node-exporter;
+      description = "System group for the Prometheus Node exporter";
+    };
+
+
```
zowoq, line 94: Users/groups need to be "known".

```suggestion

    users.knownGroups = [ "_prometheus-node-exporter" ];
    users.knownUsers = [ "_prometheus-node-exporter" ];

```

modules/services/monitoring/prometheus-node-exporter.nix
```diff
@@ -0,0 +1,115 @@
...
+    launchd.daemons.prometheus-node-exporter = {
+      script = concatStringsSep " "
+        ([
+          (getExe cfg.package)
+          "--web.listen-address"
+          "${cfg.listenAddress}:${cfg.port}"
```
zowoq, line 100:
```suggestion
          "${cfg.listenAddress}:${toString cfg.port}"
```

modules/services/monitoring/prometheus-node-exporter.nix
```diff
@@ -0,0 +1,117 @@
...
+    launchd.daemons.prometheus-node-exporter = {
+      script = concatStringsSep " "
+        ([
+          (getExe cfg.package)
+          "--web.listen-address"
+          "${cfg.listenAddress}:${toString cfg.port}"
+        ]
+        ++ (map (collector: "--collector.${collector}") cfg.enabledCollectors)
+        ++ (map (collector: "--no-collector.${collector}") cfg.disabledCollectors)
+      ) + escapeShellArgs cfg.extraFlags;
```
emilazy, line 106: Since we probably don’t want shell injection in the other arguments here, could just use `escapeShellArgs` for the whole thing.

emilazy (approved):
Looks great to me now. @Enzime, did you want to have another pass over this?

Only one non‐blocking nit.

Enzime (approved):
LGTM

PR #1130, opened 2024-10-28
Allow configuring the fn key action

aschleck:
See values here: https://macos-defaults.com/keyboard/applefnusagetype.html

Wasn't sure if it made sense to create a file for one option but seemed okay. All com.apple.HIToolbox available options I see are:
```
    "com.apple.HIToolbox" =     {
        AppleCurrentKeyboardLayoutInputSourceID = "com.apple.keylayout.US";
        AppleDictationAutoEnable = 0;
        AppleEnabledInputSources =         (
                        {
                InputSourceKind = "Keyboard Layout";
                "KeyboardLayout ID" = 0;
                "KeyboardLayout Name" = "U.S.";
            },
                        {
                "Bundle ID" = "com.apple.CharacterPaletteIM";
                InputSourceKind = "Non Keyboard Input Method";
            }
        );
        AppleFnUsageType = 2;
        AppleSelectedInputSources =         (
                        {
                InputSourceKind = "Keyboard Layout";
                "KeyboardLayout ID" = 0;
                "KeyboardLayout Name" = "U.S.";
            }
        );
    };
```

### DIFF
--- a/modules/system/defaults/hitoolbox.nix
+++ b/modules/system/defaults/hitoolbox.nix
@@ -1,0 +1,29 @@
+{ lib, ... }:
+
+{
+  options = {
+
+    system.defaults.hitoolbox.AppleFnUsageType = lib.mkOption {
+      type = lib.types.nullOr (lib.types.enum [
+        "Do Nothing"
+        "Change Input Source"
+        "Show Emoji & Symbols"
+        "Start Dictation"
+      ]);
+      apply = key: if key == null then null else {
+        "Do Nothing" = 0;
+        "Change Input Source" = 1;
+        "Show Emoji & Symbols" = 2;
+        "Start Dictation" = 3;
+      }.${key};
+      default = null;
+      description = ''
+        Chooses what happens when you press the Fn key on the keyboard. A restart is required for
+        this setting to take effect.
+
+        The default is unset ("Show Emoji & Symbols").
+      '';
+    };
+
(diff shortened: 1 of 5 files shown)


aschleck:
Also not sure if this should be system.defaults.HIToolbox instead of system.defaults.hitoolbox. Apple capitalization seems to be all over the place despite the defaults write command being case insensitive.

modules/system/defaults/hitoolbox.nix
```diff
@@ -0,0 +1,26 @@
+{ config, lib, ... }:
+
+with lib;
```
Enzime, line 3:
```suggestion
```

Enzime (changes requested):

modules/system/defaults/hitoolbox.nix
aschleck: It's required for mkOption and types so can't remove, unless you mean you prefer `lib.mkOption` and `lib.types`. Will leave out for now.

Enzime: Yeah, can you change them to `lib.mkOption` and `lib.types`?

aschleck: Changed the three uses to lib.*

Enzime (approved):
LGTM

Thanks for the contribution

PR #1133, opened 2024-10-30
karabiner-elements: allow use of custom package

bew:
This is necessary to be able to experiment changes to the package without completely copying the module.

cc @Enzime

### DIFF
--- a/modules/services/karabiner-elements/default.nix
+++ b/modules/services/karabiner-elements/default.nix
@@ -9,18 +9,19 @@ let
 in
 
 {
-  options = {
-    services.karabiner-elements.enable = mkEnableOption "Karabiner-Elements";
+  options.services.karabiner-elements = {
+    enable = mkEnableOption "Karabiner-Elements";
+    package = mkPackageOption pkgs "karabiner-elements" { };
   };
 
   config = mkIf cfg.enable {
-    environment.systemPackages = [ pkgs.karabiner-elements ];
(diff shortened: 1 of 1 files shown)


modules/services/karabiner-elements/default.nix
```diff
@@ -9,12 +9,18 @@ let
...
-  options = {
-    services.karabiner-elements.enable = mkEnableOption "Karabiner-Elements";
+  options.services.karabiner-elements = {
+    enable = mkEnableOption "Karabiner-Elements";
+    package = mkOption {
+      type = types.package;
+      default = pkgs.karabiner-elements;
+      defaultText = "pkgs.karabiner-elements";
+      description = "Karabiner-elements derivation to use.";
+    };
```
Enzime, line 19:
```suggestion
    package = mkPackageOption pkgs "karabiner-elements" { };
```

Enzime (changes requested):
There are other uses of `pkgs.karabiner-elements` in this file that hasn't been replaced with `cfg.package`

modules/services/karabiner-elements/default.nix
bew: you're totally right, I updated the PR with the changes

(I didn't have my dev environment yet on this mac 👀)

Enzime (approved):
LGTM

Thanks for the contribution

Issue #1003, opened 2024-07-11
Installing flake breaks nix on MacOS Sonoma completely


jjpe:
Below is my flake.
Installing it has broken **all** nix tooling completely on my MBP to the point that all `nix` CLI commands are now completely unavailable. I've done a factory reset because without nix I can't use nix to fix nix. After the factory reset, I tried to build my flake again, and again it *COMPLETELY* breaks nix!
What's the fix here? And why does nix-darwin break my system in the first place?

EDIT: 
I just found out that running `darwin-installer` makes the nix tooling available again.
But even then, I can't really build my flake as-is without breaking the system. So a fix is still necessary.

```nix
{
  description = "System Flake (cassio)";

  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
    nix-darwin.url = "github:LnL7/nix-darwin";
    nix-darwin.inputs.nixpkgs.follows = "nixpkgs";
    home-manager.url = "github:nix-community/home-manager";
    home-manager.inputs.nixpkgs.follows = "nixpkgs";
  };

  outputs = inputs@{ self, nix-darwin, nixpkgs, home-manager }:
  let
    nixpkgsConfig = {
      config.allowUnfree = true;
    };
    configuration = { config, pkgs, ... }: {
      users.users.j = {
        name = "j";
        home = "/Users/j";
      };

      # List packages installed in system profile. To search by name, run:
      # $ nix-env -qaP | grep wget
      environment.systemPackages = with pkgs; [
   #     aspell
   #     bandwhich # Bandwidth monitor
   #     bat
   #     bottom # replacement for top
   #     broot # a filterable hierarchical file viewer with a tree-based TUI
   #     # cargo-apk
   #     # cargo-asm
   #     cargo-audit
   #     cargo-bloat
   #     cargo-edit
   #     cargo-crev
   #     cargo-criterion
   #     cargo-expand # Rust macro expansion utility
   #     cargo-outdated # Find outdated Rust crate dependencies
   #     cargo-make # build tool on top of cargo
   #     cargo-msrv # Find the Minimum Supported Rust Version for a crate
   #     cargo-ndk # Android build support for Rust
   #     cargo-watch # Execute commands when Rust project files change
   #     cargo-workspaces
   #     ## clang # TODO: poisons rust builds
   #     ## chromium # TODO: broken on MacOS, at least for now
   #     colmena
   #     diesel-cli
[...]

Issue #922, opened 2024-04-04
Nix-Darwin x HM doesn't install paths in for zsh `/run/current-system/sw/bin`


Eveeifyeve:
The issue is when I install nix on my machine m2 Macbook pro 13inch it some how doesn't have this path exported which I need for `darwin-rebuild` for some reason only `nix run nix-darwin -- switch --flake ~/.dotfiles/` seems to work.

### PATH Echo Output
```
/Users/eveeifyeve/.nix-profile/bin:/nix/var/nix/profiles/default/bin:/usr/bin:/bin
```

### List /run/current-system/sw/bin
```
bash			darwin-option		info			nix-build		nix-daemon		nix-instantiate		pdftexi2dvi		texi2dvi
bashbug			darwin-rebuild		install-info		nix-channel		nix-env			nix-prefetch-url	pod2texi		texi2pdf
brew			darwin-uninstaller	makeinfo		nix-collect-garbage	nix-hash		nix-shell		sh			texindex
darwin-help		darwin-version		nix			nix-copy-closure	nix-info		nix-store		texi2any
```

## Info
- Nix: 2.21.1
- MacOS: Sonoma 14.4
- M2 Macbook pro 13inch with 8 gb of memory
- Nix Config: https://github.com/Eveeifyeve/Dotfiles

Eveeifyeve:
I have tried both the distributed systems and nixos.org nix and still not working so it's not a nix problem.

Enzime:
If you add `programs.zsh.enable = true;` does that fix it?

yswtrue:
same here I have set `programs.fish.enable = true;`, but there is no in$PATH
```
 ~  echo $PATH                                                  21.3s  Sat Apr  6 03:34:06 2024
/nix/var/nix/profiles/default/bin /opt/homebrew/bin /usr/local/bin /System/Cryptexes/App/usr/bin /usr/bin /bin /usr/sbin /sbin /var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin /var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin /var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin
```

Eveeifyeve:
If i add the programs.enabled.zsh no it doesn't work.

Eveeifyeve:
I have discussed in both distributed systems and nix discord and told me it's a nix-darwin issue.

[...]

Issue #177, opened 2019-12-01
zsh: command not found: darwin-rebuild


idmyn:
Hi, thanks for the project :)

I'm having two issues (which I expect are related). When I run `darwin-rebuild switch` I get the error `zsh: command not found: darwin-rebuild`. I also can't get packages listed in `~/.nixpkgs/darwin-configuration.nix` to install when I run `/run/current-system/sw/bin/darwin-rebuild switch`. I'm running Catalina and included `NIX_IGNORE_SYMLINK_STORE=1; export NIX_IGNORE_SYMLINK_STORE` in my `~/.zprofile` as per https://github.com/LnL7/nix-darwin/issues/166#issuecomment-558813578.

Output of `echo $PATH`:
```
/Users/david/.rvm/gems/ruby-2.6.3/bin:/Users/david/.rvm/gems/ruby-2.6.3@global/bin:/Users/david/.rvm/rubies/ruby-2.6.3/bin:/Users/david/.gem/ruby/2.6.3/bin:/Users/david/.rubies/ruby-2.6.3/lib/ruby/gems/2.6.0/bin:/Users/david/.rubies/ruby-2.6.3/bin:/Users/david/.pyenv/shims:/Users/david/.nix-profile/bin:/Users/david/.nix-profile/bin:/Users/david/.nvm/versions/node/v12.6.0/bin:/Users/david/.pyenv/shims:/Users/david/.gem/ruby/2.6.3/bin:/Users/david/.rubies/ruby-2.6.3/lib/ruby/gems/2.6.0/bin:/Users/david/.rubies/ruby-2.6.3/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/local/bin:/opt/local/sbin:/Library/TeX/texbin:/usr/local/MacGPG2/bin:/opt/X11/bin:/Library/Frameworks/Mono.framework/Versions/Current/Commands:/Users/david/.nvm/versions/node/v12.6.0/bin:/Users/david/.yarn/bin:/Users/david/.config/yarn/global/node_modules/.bin:/Users/david/go/bin:/Users/david/.local/bin:/Users/david/.rvm/bin
```

Output of `echo $NIX_PATH`:
```
darwin-config=/Users/david/.nixpkgs/darwin-configuration.nix:/nix/var/nix/profiles/per-user/root/channels:/Users/david/.nix-defexpr/channels:/Users/david/.nix-defexpr/channels
```

[...]

PR #1134, opened 2024-10-31
zsh: enable by default as zsh is the default shell on macOS

Enzime:
Historically this was a footgun because users would not always have this enabled leading to `darwin-rebuild` and other programs not being found.

See #177, #922 and #1003


emilazy (approved):
Makes no sense to enable Bash but not Zsh by default.

Issue #779, opened 2023-09-14
What is the correct way to set fish as the user's default shell?


dstcruz:
After digging around, I'm still not able to get the user's default shell to point to fish. Regardless of what I try, the following command always reports zsh:

```shell
> dscl . -read ~/ UserShell
UserShell: /bin/zsh

> echo $SHELL
/bin/zsh
```

Here are the things I have tried:

* Setting `environment.loginShell = "${pkgs.fish}/bin/fish -l";` after adding `pkgs.fish` to `environment.shells`, following the example provided in `modules/examples/lnl.nix`.
* Setting `environment.loginShell = pkgs.fish;` as I have found user configs in GitHub that are using that, but no dice.
* Setting `programs.fish.enable = true;` as this seems to work for `zsh` but not for `fish` (see #146).
* Setting `users.users.my-user.shell = pkgs.fish;` this seems to do nothing.
* Setting `programs.fish.enable = true;` in my `home-manager` section. Tried this just in case. Does not change `UserShell`. My understanding is that `home-manager` might not be able to set the user's shell as that requires root (??).

At this point, I have to manually use `chsh -s /run/current-system/sw/bin/fish`. If that's what is required it would be nice to have that documented somewhere. None of the `fish` related issues explicitly say that "that's the right way to do it", only that that's what people are using to force the change.

Also, while digging around, it seems that `environments.loginShell` is defined in line 79 of `modules/environment/default.nix`, but I could not find anywhere where it is actually used. Did I miss something?

iriscanavan:
> At this point, I have to manually use `chsh -s /run/current-system/sw/bin/fish`.

I have the same issue, but how would `chsh -s /run/current-system/sw/bin/fish` work here since it's a symlink to the Nix store?

ian-h-chamberlain:
> I have the same issue, but how would `chsh -s /run/current-system/sw/bin/fish` work here since `/run/current-system/sw/bin/fish` a symlink to the Nix store?

This worked for me, I had to manually append `/run/current-system/sw/bin/fish` to `/etc/shells` as well first. I guess the OS doesn't have a problem with symlink shells as long as they actually exist?

Enzime:
Updating `/etc/shells` manually or via `environment.shells` should no longer be necessary since #1118 as long as you set the shell you want to use with `users.users.<user>.shell = pkgs.fish;`

PR #1135, opened 2024-10-31
environment: remove misleading `environment.loginShell` option 

### LINKED ISSUE #361: What does loginShell do?
Inspecting the code, I could not find what `loginShell` is used for except for `tmux`. For context, I'm trying to make macOS load the nix-darwin `zsh` instead of the system one, with an existing user (not created through nix-darwin).

I have tried:

```
  programs.zsh = {
    enable = true;
  };
  environment.variables = {
    SHELL = "${pkgs.zsh}/bin/zsh";
  };
  environment.loginShell = "${pkgs.zsh}/bin/zsh -l";
```

The only way to make it work is to either `chsh` or something like whats done in `dscl . -create '/Users/${v.name}' UserShell '${v.shell}'`.
I think it would be great if https://github.com/LnL7/nix-darwin/blob/master/modules/users/default.nix#L152-L168 did set the shell even if the user already exists.

Enzime:
This is a source of confusion as this option was only used to configure tmux's default command, rather than `users.users.<user>.shell` which is what most people actually wanted which wasn't working very reliably prior to #1120 

I also chose to remove `programs.tmux.defaultCommand` as I suspect that most users want it to default to their login shell and very few need to change it.

Closes #361

See #151, #779

### DIFF
--- a/modules/programs/tmux.nix
+++ b/modules/programs/tmux.nix
@@ -41,6 +41,7 @@ in
 {
   imports = [
     (mkRenamedOptionModule [ "programs" "tmux" "tmuxConfig" ] [ "programs" "tmux" "extraConfig" ])
+    (mkRemovedOptionModule [ "programs" "tmux" "defaultCommand" ] "Use `programs.tmux.extraConfig` to configure the default command instead. If unset, tmux will default to using your system configured login shell.")
   ];
   options = {
     programs.tmux.enable = mkOption {
@@ -84,11 +85,6 @@ in
       description = "Cater to iTerm2 and its tmux integration, as appropriate.";
     };
 
-    programs.tmux.defaultCommand = mkOption {
-      type = types.either types.str types.package;
-      description = "The default command to use for tmux panes.";
-    };
-
     programs.tmux.tmuxOptions = mkOption {
       internal = true;
       type = types.attrsOf (types.submodule text);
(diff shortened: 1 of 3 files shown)


modules/programs/tmux.nix
```diff
@@ -41,6 +41,7 @@ in
 {
   imports = [
     (mkRenamedOptionModule [ "programs" "tmux" "tmuxConfig" ] [ "programs" "tmux" "extraConfig" ])
+    (mkRemovedOptionModule [ "programs" "tmux" "defaultCommand" ] "Use `programs.tmux.extraConfig` to configure the default command instead. If unset, tmux will default to using your system configured login shell.")
```
emilazy, line 44: Is there a reason to get rid of this?

emilazy (commented):
NixOS doesn’t have this, so seems reasonable to me.

modules/programs/tmux.nix
Enzime: I couldn't find any examples of anyone using this online and it seemed easy enough to configure without this option:

```nix
programs.tmux.extraConfig = ''
  set -g default-command "zsh -l"
'';
```

So I decided it was easier to just remove this option than add support for a `null` default (as we no longer have `environment.loginShell`)

Enzime: Happy to add it back in if you prefer

emilazy (approved):
LGTM

PR #1139, opened 2024-11-03
system: run `shellcheck` on `activate` and `activate-user` scripts 

Enzime:


Enzime:
@emilazy not sure if this is a good idea https://github.com/LnL7/nix-darwin/pull/1139/commits/b335c0538dfb5f23abcd183fccd1df526b799a61 I originally made this change as `createRun` runs before `check` and it seemed pointless to emit a `warning` when check would fail later and emit an error and print instructions

modules/system/base.nix
```diff
@@ -2,22 +2,43 @@
...
+          printf >&2 '$ sudo /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -B # For Catalina\n'
+          printf >&2 '$ sudo /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -t # For Big Sur and later\n' >&2
+          printf >&2 '\n'
+          printf >&2 'The current contents of /etc/synthetic.conf is:\n'
+          printf >&2 '\n'
+          sudo sed 's/^/    /' /etc/synthetic.conf >&2
+          printf >&2 '\n'
+          exit 1
+        fi
+      else
```
emilazy, line 31: This condition definitely seems strange and inverted now.

emilazy (commented):
Generally I think the reason we might check for some things twice is to avoid partial activations, and also because people or tools might mess up and run only one of `activate-user`/`activate`. But I’m not sure that applies here.

In this case, however, I actually have direct personal experience with the warning! I set nix-darwin up in a macOS 10.12 VM recently – before we needed all this synthetic mount stuff at all – and it (naturally) printed the `apfs.util` warning but activation succeeded anyway, since it could just make a regular symlink. Now, whether we care about 10.12 at all, well… certainly after 24.11 it’d be pointless to support things below 11.3 as Nixpkgs won’t either.

modules/system/base.nix
Enzime: I added a comment to explain further but the old logic didn't check whether `/etc/synthetic.conf` existed and would always try to create the symlink at `/run`, but as we know Nix will be installed (and will be installed at `/nix` most likely) then `/etc/synthetic.conf` should always exist and we can only create the symlink when we know we're on a system without `/etc/synthetic.conf` (<10.15)

Enzime: The `runLink` check did contain the `/etc/synthetic.conf` existence check

https://github.com/LnL7/nix-darwin/pull/1139/commits/4bc361a6379b33fac1d3fae2d2d5989ef40fa57d#diff-98c72e612838af210de471a5278ebb18d17588b45d0c7f83bae45e3cda942ad9L28

Enzime:
@emilazy ready for review :)

modules/system/base.nix
emilazy: In the previous version we had a fallback that did the symlink when `! test -L /run`; I think this is why it worked for me on pre‐Catalina. Now it’s in the else branch and we only try and set it up when `test -L /run` is true, i.e. it doesn’t need setting up. That seems to not make any sense to me.

I’d also be fine just dropping support for old versions, of course, given the direction Nixpkgs is heading in (though we should add a `sw_vers` check in that case).

Enzime: I believe the `else` matches the `if [[ -e /etc/synthetic.conf ]]` line but I can’t confirm as I’m on my phone now

Samasaur1 (approved):
This branch works fine on my config (in fact it actually helped me catch some warnings in custom activation scripts) and it looks good to me

modules/services/github-runner/service.nix
```diff
@@ -48,14 +48,19 @@ in
...
 
-          ${pkgs.coreutils}/bin/mkdir -p -m 0750 ${escapeShellArg (mkLogDir cfg)}
+          ${pkgs.coreutils}/bin/mkdir -p ${escapeShellArg (mkLogDir cfg)}
           ${pkgs.coreutils}/bin/chown ${user}:${group} ${escapeShellArg (mkLogDir cfg)}
         '' + optionalString (cfg.workDir == null) ''
-          ${pkgs.coreutils}/bin/mkdir -p -m 0750 ${escapeShellArg (mkWorkDir cfg)}
+          ${pkgs.coreutils}/bin/mkdir -p ${escapeShellArg (mkWorkDir cfg)}
           ${pkgs.coreutils}/bin/chown ${user}:${group} ${escapeShellArg (mkWorkDir cfg)}
+        '' + ''
+          umask -S "$oldUmask"
```
emilazy, line 63: `umask` is scary. Maybe we can just create the parent directory separately here, or at least isolate this into a subshell.

modules/system/base.nix
emilazy: You’re totally right, sorry!

modules/system/checks.nix
```diff
@@ -104,7 +88,7 @@ let
 
   buildUsers = ''
     buildUser=$(dscl . -read /Groups/nixbld GroupMembership 2>&1 | awk '/^GroupMembership: / {print $2}') || true
-    if [ -z $buildUser ]; then
+    if [ -z "$buildUser" ]; then
```
emilazy, line 91: Maybe we can just use `[[ … ]]` here.

modules/services/github-runner/service.nix
Enzime: Fixed 👍

modules/system/checks.nix
Enzime: Fixed 👍

modules/services/github-runner/service.nix
```diff
@@ -48,14 +48,23 @@ in
...
+            ${pkgs.coreutils}/bin/mkdir -p ${escapeShellArg (mkLogDir cfg)}
+            ${pkgs.coreutils}/bin/chown ${user}:${group} ${escapeShellArg (mkLogDir cfg)}
+
+            ${optionalString (cfg.workDir == null) ''
+              ${pkgs.coreutils}/bin/mkdir -p ${escapeShellArg (mkWorkDir cfg)}
+              ${pkgs.coreutils}/bin/chown ${user}:${group} ${escapeShellArg (mkWorkDir cfg)}
+            ''}
+
+            umask -S "$oldUmask"
+          )
```
emilazy, line 67: No need to save and restore the `umask` with a subshell.

emilazy (dismissed):
LGTM, one nit. I’m assuming that ShellCheck doesn’t noticeably slow down the system build.

modules/services/github-runner/service.nix
Enzime: Fixed 👍

emilazy (approved):

chroth7:
hi all - I think this does break my `darwin-rebuild switch`:

```
error: builder for '/nix/store/....drv' failed with exit code 1;
       last 7 log lines:
       >
       > In /nix/store/.../activate line 80:
       > while read src; do
       >       ^--^ SC2162 (info): read without -r will mangle backslashes.
       >
       > For more information:
       >   https://www.shellcheck.net/wiki/SC2162 -- read without -r will mangle backs...
       For full logs, run 'nix log /nix/store/....drv'.
```

       I will investigate some more, but seems it fails on the missing `-r` there

Enzime:
Please open the `activate` or `activate-user` script at the line specified and report which activation script is causing the issues for you. As everyone’s configs are different and have different things enabled, the line numbers aren’t the same for everybody

PRs to fix the `shellcheck` errors will be greatly appreciated

chroth7:
@Enzime here you go:
```
# Set up applications.
echo "setting up /Applications..." >&2
rm -rf /Applications/Nix\ Apps
mkdir -p /Applications/Nix\ Apps
find /nix/store/jr12kg0c0ivdj375ds8b0ph4cism2gdn-system-applications/Applications -maxdepth 1 -type l -exec readlink '{}' + |
while read src; do
  app_name=$(basename "$src")
  echo "copying $src" >&2
  /nix/store/pfg0rf4rhm8dwcgkpk9rga2m4ygprdnx-mkalias-1.0.0/bin/mkalias "$src" "/Applications/Nix Apps/$app_name"
done

```

it's the `while read src; do` that's missing the `-r`

Enzime:
@chroth7 That code is not from nix-darwin’s activation scripts, please check your config for any custom `activationScripts` you’ve set up

chroth7:
oh wow, you are so right @Enzime - sorry for wasting your time!
found it & fixed it!

thanks a lot

mrjson79:
Confirm what @chroth7 is writing as well. Also had an activation script and had to add -r to the `read src` part.
So it looks like this now `read -r src`

chroth7:
And we probably all got it from the same YouTube video 😀 I have added a comment there

PR #1140, opened 2024-11-04
release: fix tests not running on `aarch64-darwin`

Enzime:
I've cleaned up `release.nix` by removing package related jobs which aren't run as far as I'm aware and already live in [nixpkgs](https://github.com/NixOS/nixpkgs/blob/2d2a9ddbe3f2c00747398f3dc9b05f7f2ebb0f53/pkgs/top-level/release.nix#L90-L139).

I've also exposed the docs and tests in `flake.nix` under `packages` and `checks` respectively which means they should get tested with `nix flake check` and allows you to build individual tests with `nix build` i.e. `nix build nix-darwin#checks.aarch64-darwin.users-groups`


emilazy:
IIRC the last time I tried to do this there was some issue. Maybe the schema was wrong or something? I assume `nix flake check` works okay like this?

We could probably get rid of some of the existing test jobs as redundant to this, but maybe it’s fine.

Enzime:
As far as I'm aware, `nix flake check` expects `packages` to be flat and I'm guessing `checks` as well

Another schema change is in https://github.com/LnL7/nix-darwin/pull/1140/commits/bf3e8e10b56b1a9896536bc550c12d5f758ec2ee where previously we were generating `jobs.x86_64-darwin.examples.simple.x86_64-darwin`

I've removed the new test jobs and replaced the commands in the existing `test-{un,}stable` jobs as they should be the same

emilazy (approved):

PR #1141, opened 2024-11-04
uninstaller: improve

### LINKED ISSUE #730: Tracking issue for improving uninstalling
Just off the top of my head:
- [x] Needs to clean up `/run`, esp. after https://github.com/LnL7/nix-darwin/pull/687
- [x] Didn't delete `/Applications/Nix Apps`
- [x] Should restore pre-nix-darwin configuration like `nix.conf`, `bashrc` and `zshrc`
- [x] Change user shells that point to `/run/current-system`
- [x] Flake uninstall procedure needs documenting in README
- [x] Update documentation to try running the uninstaller from `$PATH`

Enzime:
Closes #730

- [ ] test this PR
- [ ] add check to ensure `/etc/bashrc`, `/etc/zshrc`, `/etc/shells` got restored correctly
- [ ] add check to ensure no users have shells pointing to `/run/*`
- [ ] run checks to let users know if uninstallation didn't work correctly and direct users to file an issue
- [ ] add check to ensure `run` symlink is no longer getting set up

### DIFF
--- a/pkgs/darwin-uninstaller/configuration.nix
+++ b/pkgs/darwin-uninstaller/configuration.nix
@@ -1,4 +1,4 @@
-{ lib, ... }:
+{ lib, pkgs, ... }:
 
 with lib;
 
@@ -33,18 +33,6 @@ with lib;
             sudo cp /nix/var/nix/profiles/default/Library/LaunchDaemons/org.nixos.nix-daemon.plist /Library/LaunchDaemons/org.nixos.nix-daemon.plist
             sudo launchctl load -w /Library/LaunchDaemons/org.nixos.nix-daemon.plist
         fi
-
-        if ! grep -q etc/profile.d/nix-daemon.sh /etc/bashrc; then
-            echo >&2 "Found no nix-daemon.sh reference in /etc/bashrc"
-            echo >&2 "add this snippet back to /etc/bashrc:"
-            echo >&2
-            echo >&2 "    # Nix"
-            echo >&2 "    if [ -e '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh' ]; then"
-            echo >&2 "      . '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'"
-            echo >&2 "    fi"
-            echo >&2 "    # End Nix"
-            echo >&2
-        fi
     fi
 
     # grep will return 1 when no lines matched which makes this line fail with `set -eo pipefail`
@@ -56,5 +44,9 @@ with lib;
 
       dscl . -create /Users/"$user" UserShell /bin/zsh
     done
+
+    while IFS= read -r -d "" file; do
+      mv "$file" "''${file%.*}"
+    done < <(find /etc -name '*.before-nix-darwin' -follow -print0)
   '';
 }
--- a/pkgs/darwin-uninstaller/configuration.nix
+++ b/pkgs/darwin-uninstaller/configuration.nix
@@ -46,5 +46,15 @@ with lib;
             echo >&2
         fi
     fi
+
+    # grep will return 1 when no lines matched which makes this line fail with `set -eo pipefail`
+    dscl . -list /Users UserShell | { grep "\s/run/" || true; } | awk '{print $1}' | while read -r user; do
+      shell=$(dscl . -read /Users/"$user" UserShell)
+      if [[ "$shell" != */bin/zsh ]]; then
+        echo >&2 "warning: changing $user's shell from $shell to /bin/zsh"
+      fi
+
+      dscl . -create /Users/"$user" UserShell /bin/zsh
+    done
   '';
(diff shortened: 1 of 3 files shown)


pkgs/darwin-uninstaller/configuration.nix
```diff
@@ -7,46 +7,44 @@ with lib;
...
-            echo >&2 "    if [ -e '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh' ]; then"
-            echo >&2 "      . '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'"
-            echo >&2 "    fi"
-            echo >&2 "    # End Nix"
-            echo >&2
-        fi
     fi
+
+    # grep will return 1 when no lines matched which makes this line fail with `set -eo pipefail`
+    dscl . -list /Users UserShell | { grep "\s/run/" || true; } | awk '{print $1}' | while read -r user; do
```
emilazy, line 39: Should maybe include `/nix` in this too, in case someone is pointing to a profile directory? Not sure.

pkgs/darwin-uninstaller/configuration.nix
```diff
@@ -7,46 +7,44 @@ with lib;
...
+      shell=$(dscl . -read /Users/"$user" UserShell)
+      if [[ "$shell" != */bin/zsh ]]; then
+        echo >&2 "warning: changing $user's shell from $shell to /bin/zsh"
+      fi
+
+      dscl . -create /Users/"$user" UserShell /bin/zsh
+    done
+
+    find /etc -name '*.before-nix-darwin' -follow | ${lib.getExe pkgs.parallel} --will-cite mv '{}' '{.}'
```
emilazy, line 48: We’re not actually going to cite GNU Parallel and I’ve seen enough on that topic from its author that I’d rather just avoid calling it entirely. GNU xargs supports `-P` so perhaps we can use that (and `-print0`/`-0`). In practice I suspect almost nobody has enough files that they need any parallelism here though; a classic `while read -d '' …; do …; done < <(find … -print0)` loop might suffice, if `xargs` doesn’t.

pkgs/darwin-uninstaller/default.nix
```diff
@@ -50,51 +34,61 @@ stdenv.mkDerivation {
...
+
     echo >&2
     echo >&2 "NOTE: The /nix/var/nix/profiles/system* profiles still exist and won't be garbage collected."
     echo >&2
     echo >&2 "Done!"
     echo >&2
-    exit
   '';
+}).overrideAttrs ({
+  passthru.check = writeShellApplication {
```
emilazy, line 72: You can use `derivationArgs.passthru` here instead of the `overrideAttrs` (and a name like e.g. `tests.uninstaller` would be more conventional).

pkgs/darwin-uninstaller/configuration.nix
Enzime: That should still work as Nix is still installed but also that seems very very unlikely

Maybe nix-darwin should be responsible for uninstalling Nix as well 🤔

pkgs/darwin-uninstaller/configuration.nix
Enzime: Yeah I'm not the biggest fan of the citation related stuff of Parallel, I mainly used it because it has the convenient `{.}` token

Enzime:
```
Academic tradition requires you to cite works you base your article on.
If you use programs that use GNU Parallel to process data for an article in a
scientific publication, please cite:

  Tange, O. (2024, July 22). GNU Parallel 20240722 ('Assange').
  Zenodo. https://doi.org/10.5281/zenodo.12789352

This helps funding further development; AND IT WON'T COST YOU A CENT.
If you pay 10000 EUR you should feel free to use GNU Parallel without citing.

More about funding GNU Parallel and the citation notice:
https://www.gnu.org/software/parallel/parallel_design.html#citation-notice

To silence this citation notice: run 'parallel --citation' once.
```

The one thing to note here is that as we're not doing anything related to scientific publications we don't need to cite it

pkgs/darwin-uninstaller/default.nix
Enzime: Updated 👍

pkgs/darwin-uninstaller/configuration.nix
Enzime: Updated to not use GNU Parallel anymore 👍

emilazy (dismissed):
LGTM; haven’t tested this myself so I’m not totally confident it won’t clobber somebody’s setup but the uninstaller is already prone to that and this seems like movement in the right direction.

emilazy:
Though CI is unhappy :)

emilazy (approved):

Issue #1085, opened 2024-09-19
`github-runner`: Labels are not passed correctly


srid:
After #1055, the github-runner's configure script contains bash that sets labels like this:

```bash
--labels "'sambar,aarch64-darwin,x86_64-darwin'"
```

Note the double quotation above.

PR #1055 did add the double quotes so I'm guessing that's the trigger?

When passing `--labels` like this, GitHub will consider that inner quote (`'`) as being part of the label text, leading to:

![image](https://github.com/user-attachments/assets/8e741cee-e416-4a06-8ca6-b191ebfb152d)

Ref: https://github.com/juspay/github-nix-ci/issues/10

PR #1143, opened 2024-11-04
github-runner: Fix labels for different nixpkgs versions

YorikSar:
Changes to escapeShellArg introduced in https://github.com/NixOS/nixpkgs/pull/333744 made different versions of nixpkgs behave differently. If current nix-darwin is used with nixpkgs before that change, labels end up having labels quoted twice (see https://github.com/LnL7/nix-darwin/issues/1085), but without changes from https://github.com/LnL7/nix-darwin/pull/1055, with new nixpkgs, labels end up not quoted at all, and ShellCheck ends up complaining that commas might have been used as array item separator (see https://www.shellcheck.net/wiki/SC2054).
Use the old version of escapeShellArg to always escape the list of labels and make nix-darwin work with both old and new versions of nixpkgs.

Fixes https://github.com/LnL7/nix-darwin/issues/1085

Cc @Gabriella439

### DIFF
--- a/modules/services/github-runner/service.nix
+++ b/modules/services/github-runner/service.nix
@@ -94,6 +94,10 @@ in
 
         script =
           let
+            # https://github.com/NixOS/nixpkgs/pull/333744 introduced an inconsistency with different
+            # versions of nixpkgs. Use the old version of escapeShellArg to make sure that labels
+            # are always escaped to avoid https://www.shellcheck.net/wiki/SC2054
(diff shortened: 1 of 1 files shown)


Enzime (approved):
LGTM

Thanks for the contribution

Enzime:
Can you rebase this on top of `master` to get the tests to run?

YorikSar:
Oof. I don't know what came over me. Fixed now.

Enzime (approved):
LGTM

Thanks for the contribution

PR #1146, opened 2024-11-05
prometheus-node-exporter: fix log permissions

emilazy:
The daemon won’t start as the assigned user doesn’t have permissions to create the log file.


mweinelt (approved):
Looks good to me.

Issue #726, opened 2023-07-10
Look into updating `services.activate-system`


emilazy:
The list of activation scripts has grown a lot. `services.activate-system`'s hasn't, which contributes to annoying configuration drift between system updates and the next explicit activation. We should figure out exactly which ones should be added or even if it's possible to just run the entire system activation script in the daemon.

LnL7:
It's not necessarily out of sync. I introduced that deamon because some things like the `/run` symlink would disappear after every boot as it's in a tempfs. So I only added the activation scripts for things that actually needed to be applied again.

That said it might not be a bad idea to just activate everything after boot in case system changes are overwritten by a macOS update, etc.

emilazy:
My preference would be to activate the whole system, yeah, although there may need to be some exceptions to that (e.g. when Homebrew Cask runs installers they probably expect to be in a graphical context and might need to prompt the user).

hraban:
Homebrew also needs network connectivity, and in general is not friendly to being run headless. I would personally like to separate homebrew from more clearly deterministic scripts in headless mode.

PR #1147, opened 2024-11-05
darwin-rebuild: Don't prompt for sudo multiple times

kabiroberai:
I've found that `darwin-rebuild switch` always asks me for sudo permissions twice — once when it runs `sudo nix-env --set` and the second time when it runs `sudo $systemConfig/activate` — even though I have a reasonable (default) `timestamp_timeout`.

In debugging this, I realized that the call to `brew bundle` within `activate-user` was somehow making it so that the next `sudo` invocation would require a password again. I'm honestly not sure of the root cause behind this strange behavior exhibited by `brew bundle`, but as a workaround I realized that if we 1) elevate privileges, 2) drop down to call `activate-user`, and 3) call `activate` in the elevated context, we don't need to `sudo` again since we're already root.

FWIW I'm open to input here, especially as a new nix-darwin user: is this issue specific to me or is everyone just used to authenticating twice? But Brew Bundle aside, this PR does work around the general issue where arbitrary user scripts in `activate-user` can result in sudo prompting again, so I think this might be worth merging regardless of the root cause.

### DIFF
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -225,6 +225,11 @@ fi
 
 if [ "$action" = switch ] || [ "$action" = activate ] || [ "$action" = rollback ]; then
   if [ "$USER" != root ]; then
+    # Running `$systemConfig/activate-user` causes subsequent sudo invocations to prompt
+    # for the password even if we previously ran sudo in the same session; indeed, by this
+    # point we've already run `nix-env --set` as sudo. To avoid prompting a second time,
+    # we become root *before* running activate-user and then drop down to the user to
+    # invoke it. This way, the call to activate doesn't require a password.
     sudo @shell@ -c "su $USER -c $systemConfig/activate-user && $systemConfig/activate"
   else
     "$systemConfig/activate-user"
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -224,11 +224,10 @@ if [ "$action" = switch ]; then
 fi
 
 if [ "$action" = switch ] || [ "$action" = activate ] || [ "$action" = rollback ]; then
-  "$systemConfig/activate-user"
-
   if [ "$USER" != root ]; then
-    sudo "$systemConfig/activate"
+    sudo @shell@ -c "su $USER -c $systemConfig/activate-user && $systemConfig/activate"
   else
+    "$systemConfig/activate-user"
     "$systemConfig/activate"
   fi
 fi
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -21,17 +21,16 @@ showSyntax() {
   exit 1
 }
 
-sudo() {
-  # REMOVEME when support for macOS 10.13 is dropped
-  # macOS 10.13 does not support sudo --preserve-env so we make this conditional
-  if command sudo --help | grep -- --preserve-env= >/dev/null; then
-    # We use `env` before our command to ensure the preserved PATH gets checked
-    # when trying to resolve the command to execute
-    command sudo -H --preserve-env=PATH --preserve-env=SSH_CONNECTION env "$@"
-  else
-    command sudo -H "$@"
-  fi
-}
+# REMOVEME when support for macOS 10.13 is dropped
+# macOS 10.13 does not support sudo --preserve-env so we make this conditional
+if /usr/bin/sudo --help | grep -- --preserve-env= >/dev/null; then
+  # We use `env` before our command to ensure the preserved PATH gets checked
+  # when trying to resolve the command to execute
+  sudo="/usr/bin/sudo -H --preserve-env=PATH --preserve-env=SSH_CONNECTION"
+else
+  sudo="/usr/bin/sudo -H"
+fi
+sudo() { $sudo "$@"; }
 
 # Parse the command line.
 origArgs=("$@")
@@ -230,7 +229,7 @@ if [ "$action" = switch ] || [ "$action" = activate ] || [ "$action" = rollback 
     # point we've already run `nix-env --set` as sudo. To avoid prompting a second time,
     # we become root *before* running activate-user and then drop down to the user to
     # invoke it. This way, the call to activate doesn't require a password.
-    sudo @shell@ -c "su $USER -c $systemConfig/activate-user && $systemConfig/activate"
+    sudo @shell@ -c "$sudo -u $USER $systemConfig/activate-user && $systemConfig/activate"
   else
     "$systemConfig/activate-user"
     "$systemConfig/activate"


Enzime (dismissed):
LGTM

@emilazy any thoughts on this PR?

emilazy:
I don’t suppose `brew bundle` is perhaps taking a long time for you (say, over the default 5 minute `sudo` timeout)?

My thoughts:

1. This drops the `env` from the `sudo` call, which seems probably unintentional and could be bad.
2. Most users with the default `sudo` timeout don’t seem to experience being prompted multiple times, or I’d expect to have heard about it sooner.
3. For users with a zero‐second `sudo` timeout (like me), the profile switch + activation dance will be two prompts even after this PR.
4. Given (2) and (3), this seems like it’s just working around some specific peculiarity of your setup. That’s not inherently a problem, of course, but it’d be good to understand what’s going on here – e.g. perhaps Homebrew is installing something that invokes `sudo` in such a way as to clear the grace period? But that doesn’t explain why it’d be happening all the time…
5. *However,* in this case, fixing that peculiarity is changing the activation environment slightly in a way that I don’t *think* should break anything but feel hard to be fully confident about.
6. And, after Nixpkgs 24.11 is released – so, in a month or two – I plan to finally deprecate the concept of a *user activation phase* entirely, per [The Plan™](https://github.com/LnL7/nix-darwin/pull/1016#issuecomment-2285910068). This would mean that everyone would consistently only get one `sudo` prompt, because activation would be done with `sudo darwin-rebuild switch`, like NixOS. The good news is that this will result in an activation environment more like this PR anyway. The bad news is that I think the backwards‐compatibility logic to accommodate this change would become slightly more complicated.

(1) is the only blocker here for me, but the factors combined make me wonder if it’s worth doing this right now. It’d be good to understand what’s causing this better, since it seems like most people don’t experience this problem – would you perhaps be able to share your nix-darwin configuration, or at least the Homebrew portion of it?

kabiroberai:
So I just did some more digging and discovered that this is a "feature" of homebrew:

https://github.com/Homebrew/brew/blob/451c3a8735eb9b3050e40802a771eddf7c23e568/bin/brew#L72

I've boiled it down to this MRE:
```bash
sudo ls # asks for password
brew --help
sudo ls # asks again!
```

@emilazy responding to some of your other points:

1: Dropping the `env` was indeed an oversight on my part, can fix that. On a related note, do we want to use `command sudo` or `/usr/bin/sudo`? I switched to an absolute path since I figured it's more deterministic but let me know if there's a reason we use `command` here.

2: Going by the fact that this is in the `brew` driver, perhaps other people have run into this and simply haven't reported it? Might be worth asking a few folks?

3: In that case could we go a step further and do the elevate/drop dance at the point where we _first_ invoke `sudo`? I recognize this would make it easier to violate the principle of least privilege without extra care — but indeed going by point 6 it sounds like this is already the intended direction.

6: It's good to hear that! I recall instinctively trying `sudo darwin-rebuild` at first and running into errors that made me slightly concerned that I'd broken something. Handling elevated privileges correctly also seems very exciting because it enables creating something like a launch daemon that calls `darwin-rebuild` — perhaps driven by a GUI app, properly validating the XPC audit token to be sure — without needing to reauthenticate. can't wait to unpack that LPE-shaped can of worms lmao

Let me know if you still want more details about my configuration — or whether you're able to repro the MRE ootb.

emilazy:
Re (1), I think I would prefer `command sudo` because maybe the user installed their own `sudo` and wants to use it (though that’s not really possible to do through nix-darwin at present). (2) is indeed puzzling – I’m no longer so puzzled by this behaviour now that I know Homebrew is intentionally doing it, but the Homebrew module is popular and this is definitely the kind of thing I’d expect to get reports about, but maybe people aren’t too fussy for once! I once tried to make (3) happen but if you trace out the logic of the script it’s pretty awkward to get the profile‐switching `sudo` to be the same as the activation `sudo` – sufficiently awkward that I decided to just wait to do away with the idea of running `darwin-rebuild` as a non‐privileged user entirely, like my similar inclination for this PR.

Moving more stuff from activation scripts into launchd daemons, in line with NixOS’s trajectory, is definitely a goal. We technically do already have a system activation daemon that runs as boot, but it barely activates anything… (see https://github.com/LnL7/nix-darwin/issues/726). Activation changes will also make it easier for us to support multi‐user setups and to mostly reuse existing NixOS deployment tools, which is pretty exciting, and part of the reason I’m hesitant to introduce more activation complexity that we’d have to preserve for a while for backwards compatibility.

khaneliman:
> Re (1), I think I would prefer `command sudo` because maybe the user installed their own `sudo` and wants to use it (though that’s not really possible to do through nix-darwin at present). (2) is indeed puzzling – I’m no longer so puzzled by this behaviour now that I know Homebrew is intentionally doing it, but the Homebrew module is popular and this is definitely the kind of thing I’d expect to get reports about, but maybe people aren’t too fussy for once!

I've seen this behavior for a long time, but never bothered reporting it or looking into it :P

z0al:
> Most users with the default sudo timeout don’t seem to experience being prompted multiple times, or I’d expect to have heard about it sooner.

For what it's worth, this also happens to me, but I didn't care enough to report an issue. My sudo timeout is 30 minutes.

The issue started a few months ago around the time the change to restart the dock on activation is introduced if I remember correctly

adamjhf:
I also get prompted for password ~3 times for darwin-rebuild with Homebrew and sudo timeout, across multiple machines. I've temporarily removed Homebrew to avoid this.

emilazy:
FWIW, https://github.com/LnL7/nix-darwin/pull/1341 should finally address this.

bestlem:
Aren't the mulitiple sudo asking for password  because sudo here is being run from two different users
The first sudo is run from your normal user But brew bundle is being run inside this sudo run so is being run from root.

Those two sudos will have separate timers.

From logging - with all sudo attempts being logged and my user name is mark
```
imestamp                       (process)[PID]
2025-02-21 16:22:10.698190+0000  localhost sudo[82749]:     mark : TTY=ttys005 ; PWD=/Users/mark/nix ; USER=root ; ENV=PATH=/nix/store/0xg0qimxdixcb1ivngckpzk6w0mwzx6k-nvd-0.2.4/bin:/nix/store/nmhqh8mhbb6vziaamyrajc2z851gnlv4-nix-output-monitor-2.1.4/bin:/nix/store/d3alkzaq4a0yy8alnn0yamz62aim50qy-python3-3.12.8/bin:/nix/store/xlaag98yv6fx15hwn3nzfdk2rkjm3m4n-python3.12-invoke-2.2.0/bin:/nix/store/p932iklga837vjhxsrm2827zm42v0dgz-pre-commit-4.0.1/bin:/nix/store/kz7pqrmhd6i77ysm6vz1l1fhy6jxv1xf-mwb-snowfall-dir/bin:/nix/store/7ppd0j3scxlgz9v106iqr4i05vs6dv8f-bash-interactive-5.2p37/bin:/path-not-set:/Users/mark/nix/.direnv/bin:/etc/profiles/per-user/mark/bin:/run/current-system/sw/bin:/nix/var/nix/profiles/default/bin:/Users/mark/.nix-profile/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/mark/.orbstack/binCOMMAND=/usr/bin/env nix build --no-link --profile /nix/var/nix/profiles/system /var/folders/xp/rprsp4f540d9pwxx5z6ry8x80000gp/T/nh-oseHihjS/result
2025-02-21 16:22:10.795175+0000  localhost sudo[82756]:     mark : TTY=ttys005 ; PWD=/Users/mark/nix ; USER=root ; COMMAND=/System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -t
2025-02-21 16:22:20.150708+0000  localhost sudo[83420]:     mark : TTY=ttys005 ; PWD=/Users/mark/nix ; USER=root ; ENV=PATH=/nix/store/0xg0qimxdixcb1ivngckpzk6w0mwzx6k-nvd-0.2.4/bin:/nix/store/nmhqh8mhbb6vziaamyrajc2z851gnlv4-nix-output-monitor-2.1.4/bin:/nix/store/d3alkzaq4a0yy8alnn0yamz62aim50qy-python3-3.12.8/bin:/nix/store/xlaag98yv6fx15hwn3nzfdk2rkjm3m4n-python3.12-invoke-2.2.0/bin:/nix/store/p932iklga837vjhxsrm2827zm42v0dgz-pre-commit-4.0.1/bin:/nix/store/kz7pqrmhd6i77ysm6vz1l1fhy6jxv1xf-mwb-snowfall-dir/bin:/nix/store/7ppd0j3scxlgz9v106iqr4i05vs6dv8f-bash-interactive-5.2p37/bin:/path-not-set:/Users/mark/nix/.direnv/bin:/etc/profiles/per-user/mark/bin:/run/current-system/sw/bin:/nix/var/nix/profiles/default/bin:/Users/mark/.nix-profile/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/mark/.orbstack/binCOMMAND=/usr/bin/env /var/folders/xp/rprsp4f540d9pwxx5z6ry8x80000gp/T/nh-oseHihjS/result/activate
2025-02-21 16:22:22.411682+0000  localhost sudo[83833]:     root : TTY=ttys005 ; PWD=/Users/mark/nix ; USER=mark ; COMMAND=/nix/store/nb82156vxf1407vassc2j85dgpjcfzj9-activation-mark
2025-02-21 16:22:23.911374+0000  localhost sudo[84147]:     mark : a password is required ; TTY=ttys008 ; PWD=/Users/mark/nix ; USER=root ; COMMAND=/usr/bin/true
```

PR #1149, opened 2024-11-07
flake: expose docs on Linux as well

Enzime:
Continuation of #1124 

This is useful for https://search.nüschtos.de/

@SuperSandro2000 let me know if this is sufficient or if more changes are necessary to be able to consume this package

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,6 +7,8 @@
 
     jobs = forAllSystems (system: import ./release.nix {
       inherit nixpkgs system;
+
+      nix-darwin = self;
     });
   in {
     lib = {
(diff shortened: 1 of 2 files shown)


SuperSandro2000:
Yep, that seems to work, thanks https://github.com/NuschtOS/search.nuschtos.de/pull/28

Enzime:
Since #1140 the `Update manual` workflow fails on any new commits that don't change the manual as `nix-darwin`'s revision information wasn't being passed in, I've added a commit that should fix that in this PR :)

emilazy (approved):

PR #1150, opened 2024-11-07
Allow customising the path to nix.conf

P-E-Meunier:
The determinate systems installer writes to /etc/nix/nix.conf, and nix-darwin sometimes also wants to write that file, but refuses to overwrite it.

This commit is way to make the two coexist, by customising the path in which nix-darwin wants to write its nix.conf (the determinate systems installer by default uses an `include` statement on `nix.custom.conf` in the same directory).

### DIFF
--- a/modules/nix/default.nix
+++ b/modules/nix/default.nix
@@ -174,6 +174,14 @@ in
         '';
       };
 
+      nixConfPath = mkOption {
+        type = types.string;
+        default = "nix/nix.conf";
+        description = ''
+          Path to the nix.conf file inside /etc. Customising this makes it possible to have extra configuration provided outside of nix options.
+        '';
+      };
+
       # Not in NixOS module
       useDaemon = mkOption {
         type = types.bool;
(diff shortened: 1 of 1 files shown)


emilazy:
Hi there, it seems like you installed the unofficial downstream “Determinate Nix” distribution, which is not the same thing as using the Determinate Systems installer. I believe this only happens if you pass `--determinate` when using their installer; you can use the Determinate Systems installer to install regular, upstream Nix, which shouldn’t have this problem. (Yes, the naming is really confusing.)

Unfortunately I don’t think this will work properly with Determinate Nix as the configuration file we generate may overwrite settings related to the Determinate Nix‐specific functionality, defeating the point of using their distribution. I am also not sure how it interacts with the proprietary daemon they install, which apparently may write to `nix.conf` itself. They ship a nix-darwin module for Determinate Nix which may or may not fix this issue, but it’s not really something we can support either as it monkey‐patches our activation script code.

I would recommend either reinstalling upstream Nix with the Determinate Systems installer, or contacting Determinate Systems to get support for their Determinate Nix software. If you didn’t intentionally install Determinate Nix and are still running into this error, though, let me know; it’s possible that the default behaviour of the installer has changed and we might need to work around that or address it in documentation.

P-E-Meunier:
Thanks! This fix actually works for me. The only reason I'm using Determinate Nix is the support for AWS EC2 instances, and as I'm only experimenting with it I wasn't aware of these things, thanks!

Nix 2024 could *determinately* be used as a tutorial of how to destroy a community in five easy steps.

PR #1152, opened 2024-11-08
nix: don't allow using `auto-optimise-store` as it can corrupt the store

### LINKED ISSUE #484: Nix fails to create some temporary symlinks
On aarch64-darwin, I get the following output from nix:
```
<N> ➜ darwin-rebuild -L --flake /Users/jamie/Git/nixfiles switch
error: cannot link '/nix/store/.tmp-link-64480-79859421' to '/nix/store/.links/0i6ypwi0pzkgpxxqprlpyik72zxzkvkzrj7ahqj5b6hmhv2ldxm0': File exists
error: cannot link '/nix/store/.tmp-link-64480-1103639509' to '/nix/store/.links/0dwrq1sy5fawdjjvm0na718awbd10xn6dg3i0xxg9f7r5d9aa968': File exists
...
```

I have to pass `--fallback` to `darwin-rebuild` to get it to work.

Enzime:
Upstream issue at https://github.com/NixOS/nix/issues/7273

Should be fixed by https://gerrit.lix.systems/c/lix/+/2100

Tested the version check with Nix, Lix `2.91.0` (`pkgs.lixVersions.latest`), Lix `2.92.0-devpre20241106_1ecfff9` (`github:lix-project/lix/1ecfff9c37f64871661d615e407ec3571489eff7`, doesn't include the fix) and Lix `2.92.0-devpre20241107_72cce7b` (`github:lix-project/lix`) which does include the change

Closes #484

### DIFF
--- a/modules/nix/default.nix
+++ b/modules/nix/default.nix
@@ -404,7 +404,7 @@ in
           { darwin-config = "${config.environment.darwinConfig}"; }
           "/nix/var/nix/profiles/per-user/root/channels"
         ];
-        
+
         defaultText = lib.literalExpression ''
           lib.optionals cfg.channel.enable [
             # Include default path <darwin-config>.
@@ -527,8 +527,10 @@ in
               description = ''
                 If set to true, Nix automatically detects files in the store that have
                 identical contents, and replaces them with hard links to a single copy.
-                This saves disk space. If set to false (the default), you can still run
-                nix-store --optimise to get rid of duplicate files.
+                This saves disk space. If set to false (the default), you can enable
+                {option}`nix.optimise.automatic` to run {command}`nix-store --optimise`
+                periodically to get rid of duplicate files. You can also run
+                {command}`nix-store --optimise` manually.
               '';
             };
 
@@ -761,6 +763,13 @@ in
         { assertion = elem "nixbld" config.users.knownGroups -> elem "nixbld" createdGroups; message = "refusing to delete group nixbld in users.knownGroups, this would break nix"; }
         { assertion = elem "_nixbld1" config.users.knownGroups -> elem "_nixbld1" createdUsers; message = "refusing to delete user _nixbld1 in users.knownUsers, this would break nix"; }
         { assertion = config.users.groups ? "nixbld" -> config.users.groups.nixbld.members != []; message = "refusing to remove all members from nixbld group, this would break nix"; }
+
+        {
+          # Should be fixed in Lix by https://gerrit.lix.systems/c/lix/+/2100
+          # As `isNixAtLeast "2.92.0" "2.92.0-devpre20241107" == false`, we need to explicitly check if the user is running Lix 2.92.0
+          assertion = cfg.settings.auto-optimise-store -> (cfg.package.pname == "lix" && (isNixAtLeast "2.92.0-devpre20241107" || cfg.package.version == "2.92.0"));
+          message = "`nix.settings.auto-optimise-store` is known to corrupt the Nix Store, please use `nix.optimise.automatic` instead.";
+        }
       ];
 
     # Not in NixOS module


modules/nix/default.nix
```diff
@@ -761,6 +763,12 @@ in
         { assertion = elem "nixbld" config.users.knownGroups -> elem "nixbld" createdGroups; message = "refusing to delete group nixbld in users.knownGroups, this would break nix"; }
         { assertion = elem "_nixbld1" config.users.knownGroups -> elem "_nixbld1" createdUsers; message = "refusing to delete user _nixbld1 in users.knownUsers, this would break nix"; }
         { assertion = config.users.groups ? "nixbld" -> config.users.groups.nixbld.members != []; message = "refusing to remove all members from nixbld group, this would break nix"; }
+
+        {
+          # Should be fixed in Lix by https://gerrit.lix.systems/c/lix/+/2100
+          assertion = cfg.settings.auto-optimise-store -> (cfg.package.pname == "lix" && isNixAtLeast "2.92.0-devpre20241107");
+          message = "`auto-optimise-store` is known to corrupt the Nix Store, please use `nix.gc.automatic` instead.";
```
emilazy, line 770: `nix.optimise`, rather.

modules/nix/default.nix
```diff
@@ -523,12 +523,14 @@ in
...
-              example = true;
               description = ''
                 If set to true, Nix automatically detects files in the store that have
                 identical contents, and replaces them with hard links to a single copy.
                 This saves disk space. If set to false (the default), you can still run
-                nix-store --optimise to get rid of duplicate files.
+                {command}`nix-store --optimise` to get rid of duplicate files.
+
+                NOTE: This is known to cause Nix Store corruption and using
+                {option}`{nix.gc.automatic} instead is recommended.
```
emilazy, line 533: Should probably be a `::: {.warning}`, but also if this is getting fixed then we might not need to warn explicitly against it in the `description` – the assertion should do a good enough job by itself.

modules/nix/default.nix
Enzime: Fixed 👍

khaneliman:
Thanks for adding this... it made me realize I had my options backwards on my mac...

modules/nix/default.nix
```diff
@@ -523,12 +523,13 @@ in
             auto-optimise-store = mkOption {
               type = types.bool;
               default = false;
-              example = true;
```
khaneliman, line 526: I don't think it being broken changes the example situation.

modules/nix/default.nix
Enzime: Removed, also `::: {.warning}` didn't do anything when I tested it (neither in the manpages or the HTML manual)

modules/nix/default.nix
```diff
@@ -761,6 +763,13 @@ in
         { assertion = elem "nixbld" config.users.knownGroups -> elem "nixbld" createdGroups; message = "refusing to delete group nixbld in users.knownGroups, this would break nix"; }
         { assertion = elem "_nixbld1" config.users.knownGroups -> elem "_nixbld1" createdUsers; message = "refusing to delete user _nixbld1 in users.knownUsers, this would break nix"; }
         { assertion = config.users.groups ? "nixbld" -> config.users.groups.nixbld.members != []; message = "refusing to remove all members from nixbld group, this would break nix"; }
+
+        {
+          # Should be fixed in Lix by https://gerrit.lix.systems/c/lix/+/2100
+          # As `isNixAtLeast "2.92.0" "2.92.0-devpre20241107" == false`, we need to explicitly check if the user is running Lix 2.92.0
+          assertion = cfg.settings.auto-optimise-store -> (cfg.package.pname == "lix" && (isNixAtLeast "2.92.0-devpre20241107" || cfg.package.version == "2.92.0"));
+          message = "`auto-optimise-store` is known to corrupt the Nix Store, please use `nix.optimise.automatic` instead.";
```
khaneliman, line 771: I haven't worked in this repo really, but I'm curious how these assertions are scoped in the output. Will it be obvious where the `auto-optimise-store` is? Or would it be beneficial to make it more explicit with `nix.settings.auto-optimise-store` or (nix.settings): `auto-optimise-store`? Just some thoughts.

Enzime: It shows up as:

```
       error:
       Failed assertions:
       - `auto-optimise-store` is known to corrupt the Nix Store, please use `nix.optimise.automatic` instead.
```

Changed it to be more explicit 👍

Enzime: I doubt it would be a problem in practice as users would be able to search their configs for `auto-optimise-store` but making it more explicit will probably be for the best

khaneliman (approved):
LGTM, thanks again!

emilazy (approved):

Issue #1155, opened 2024-11-08
Tracking issue for updating documentation


Enzime:
- [ ] Make flakes recommended
- [ ] Combine flakes and non-flake sections
  - [ ] Write configuration
  - [ ] Add `darwin` to `NIX_PATH`
  - [ ] Installation (`darwin-rebuild` for flakes, `darwin-installer` for non-flakes)
  - [ ] Using nix-darwin
  - [ ] Updating nix-darwin
  - [ ] Uninstalling

PR #1157, opened 2024-11-08
readme: small improvements

Enzime:
#1155

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,13 +20,6 @@ nix-build https://github.com/LnL7/nix-darwin/archive/master.tar.gz -A installer
 ./result/bin/darwin-installer
 ```
 
-> NOTE: the system activation scripts don't overwrite existing etc files, so files like `/etc/bashrc` and `/etc/zshrc` won't be
-> updated by default. If you didn't use the installer or skipped some of the options you'll have to take care of this yourself.
-> Either modify the existing file to source/import the one from `/etc/static` or remove it. Some examples:
-
-- `mv /etc/bashrc /etc/bashrc.before-nix-darwin`
-- `echo 'if test -e /etc/static/bashrc; then . /etc/static/bashrc; fi' | sudo tee -a /etc/bashrc`
-
 ## Updating
 
 The installer will configure a channel for this repository.
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # nix-darwin
 
-![Test](https://github.com/LnL7/nix-darwin/workflows/Test/badge.svg)
+[![Test](https://github.com/LnL7/nix-darwin/actions/workflows/test.yml/badge.svg)](https://github.com/LnL7/nix-darwin/actions/workflows/test.yml)
 
 Nix modules for darwin, `/etc/nixos/configuration.nix` for macOS.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -2,16 +2,20 @@
...
 
 This project aims to bring the convenience of a declarative system approach to macOS.
 nix-darwin is built up around [Nixpkgs](https://github.com/NixOS/nixpkgs), quite similar to [NixOS](https://nixos.org/).
 
-## Installing
+## Prerequisites
+
+- Nix
+
+We recommend using the Determinate Systems' [nix-installer](https://github.com/DeterminateSystems/nix-installer) as it supports uninstallation. Make sure you run it **without** the `--determinate` flag as using it with nix-darwin is unsupported.
```
emilazy, line 16:
```suggestion
- A Nix implementation

We support both Nix and Lix. As the official Nix installer does not include an automated uninstaller, and manual uninstallation on macOS is a complex process, we recommend instead using one of the following installers:

* The [Determinate Systems installer](https://github.com/DeterminateSystems/nix-installer?tab=readme-ov-file#determinate-nix-installer). **Make sure you run it without the `--determinate` flag,** as that will install their downstream Determinate Nix distribution, which does not work out of the box with nix-darwin, instead of upstream Nix.

* The [Lix installer](https://lix.systems/install/#on-any-other-linuxmacos-system).
```

Wording here could probably be refined.

README.md
```diff
@@ -2,16 +2,20 @@
 
 # nix-darwin
 
-![Test](https://github.com/LnL7/nix-darwin/workflows/Test/badge.svg)
+[![Test](https://github.com/LnL7/nix-darwin/actions/workflows/test.yml/badge.svg)](https://github.com/LnL7/nix-darwin/actions/workflows/test.yml)
```
emilazy, line 5: I suspect this badge does not really add any value. (But people love them, so…)

emilazy (commented):
Directionally LGTM, but I think we should clarify the installer suggestion and make it clear we support Lix too.

README.md
Enzime: The old badge added negative(?) value because it was always green even when CI failed 😆

README.md
Enzime: Updated 👍

emilazy (approved):

README.md
chama-chomo: Hello, as for the new nix user I find this quite confusing. I've installed Nix on MacOS using Determinate installer, but here I can read that it is not supported with '--determinate' flag? How should I translate that? Shall I reinstall Nix with no '--determinate' flag or is it possible to use some 'non' out of the box steps to make it working? Unfortunately it is not very well documented and I'm not sure what exactly I am losing if I don't use the flag. Thanks for your advice

emilazy: Yes, if you install without `--determinate` you will still get all the installation, repair, and uninstallation functionality, just without the [Determinate Nixd](https://docs.determinate.systems/determinate-nix/#determinate-nixd) functionality that is not compatible with nix-darwin out of the box (some of it can be replicated with native nix-darwin functionality, though). Unfortunately I don’t know if there is an easy way to convert a `--determinate` install into one of vanilla Nix, you’d have to ask Determinate Systems; an uninstall and reinstall may be the best optino.

I’d like clearer wording here, but unfortunately it is difficult since their branding is very confusing – it was officially called the “Determinate Nix Installer” before the “Determinate Nix” product existed, and it still installs upstream Nix rather than Determinate Nix by default. Most people refer to it as “the Determinate Systems installer” or similar so we matched that here to try and reduce the confusion a little, but it’s still tricky to get the wording right. If you have any suggestions on how we could help make this a little less puzzling I’m open to hearing them :)

PR #1159, opened 2024-11-09
docs: misc fixes

Enzime:


emilazy (commented):
LGTM except for the `release.nix` change, not the right place to do it and not sure if we should be doing it at all.

Enzime:
I made the change just to remove the `darwin` in the Version in the heading of the HTML manual, I didn't change it elsewhere as I wasn't thinking about whether we should drop it from the version string overall

Before
image

After
image

emilazy:
Yeah, but the way you’ve done it won’t affect the local `darwin-help`. I think we should just leave it as‐is for now since hopefully soon it’ll look like “Version 24.11”; the state version isn’t a very coherent basis for a versioning scheme anyway.

emilazy (approved):

PR #1164, opened 2024-11-11
nix-daemon: enable by default

Enzime:
Single user installs have been unsupported by the official Nix installer since 2.4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,6 @@ Configuration lives in `~/.nixpkgs/darwin-configuration.nix`. Check out
     ];
 
   # Auto upgrade nix package and the daemon service.
-  services.nix-daemon.enable = true;
   nix.package = pkgs.nix;
 }
 ```
--- a/README.md
+++ b/README.md
@@ -66,9 +66,6 @@ Configuration lives in `~/.nixpkgs/darwin-configuration.nix`. Check out
   environment.systemPackages =
     [ pkgs.vim
     ];
-
-  # Auto upgrade nix package and the daemon service.
-  nix.package = pkgs.nix;
 }
 ```
 
--- a/modules/examples/flake/flake.nix
+++ b/modules/examples/flake/flake.nix
@@ -16,9 +16,6 @@
         [ pkgs.vim
         ];
 
-      # Auto upgrade nix package and the daemon service.
-      # nix.package = pkgs.nix;
-
       # Necessary for using flakes on this system.
       nix.settings.experimental-features = "nix-command flakes";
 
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -67,9 +67,6 @@ Configuration lives in `~/.nixpkgs/darwin-configuration.nix`. Check out
     [ pkgs.vim
     ];
```
emilazy, line 69: Nit: Stray blank line left over. Also, maybe we should leave `nix.package = pkgs.nix;` in the examples so that people know how to override it?

emilazy (commented):
> Single user installs have been unsupported by the official Nix installer since 2.4.

I don’t think that’s true, is it? I did a new single‐user installation of 2.24 in a VM recently. Rather it’s that they don’t work (at least, with SIP enabled) on Catalina or newer. I think it’s okay to drop support for older macOS versions but we should add a `sw_vers` check if so; it’s not really about the Nix version.

That said, this just changes defaults without actually dropping support for anything so I think it’s *probably* fine.

Enzime:
I believe they were disabled in https://github.com/NixOS/nix/pull/4289

> disable single-user installs on macOS

I assume a dedicated enough user could do a single-user install on macOS (probably do a multi-user install, delete the users and group and `chown /nix`)

README.md
Enzime: I've readded the example but changed it to `nix.package = pkgs.nixVersions.nix_2_18;` to make it more clear what the line is, because I think before it wasn't clear that `pkgs.nix` was already the default and the line was unnecessary

modules/examples/flake/flake.nix
```diff
@@ -16,9 +16,8 @@
         [ pkgs.vim
         ];
 
-      # Auto upgrade nix package and the daemon service.
-      services.nix-daemon.enable = true;
-      # nix.package = pkgs.nix;
+      # Uncomment this if you'd like to use a different version of Nix.
+      # nix.package = pkgs.nixVersions.nix_2_18;
```
emilazy, line 20: Nix 2.18 is about to go out of upstream’s support policy at the end of the month so I’m a bit worried about this. Maybe we could just drop the line after all, or change it to `nixVersions.git` so it won’t go stale.

emilazy (commented):
FWIW I used the official installer and I think I only had to pass `--no-daemon`. LGTM other than nit.

modules/examples/flake/flake.nix
Enzime: Removed it 👍

emilazy (approved):

Enzime:
The PR above added these lines https://github.com/NixOS/nix/blob/master/scripts/install-nix-from-closure.sh#L61-L65 which disabled the `--no-daemon` flag

emilazy:
Oh, hmm. I guess I did a daemon install but without a separate volume, then.

PR #1165, opened 2024-11-11
module: add rosetta2-gc

mweinelt:
Provides a service to prune the Rosetta 2 JIT bytecode cache.


Enzime:
`newsyslog` runs on macOS and be configured with files in `/etc/newsyslog.d`

We currently don't use it although eventually it would be great to just have it on by default for all the launchd daemons and agents we manage and then we can just turn on logging by default for them all

modules/services/rosetta2-gc/default.nix
```diff
@@ -0,0 +1,61 @@
...
+
+  launchdTypes = import ../../launchd/types.nix { inherit config lib; };
+
+  logFile = "/var/log/rosetta2-gc.log";
+
+  cfg = config.rosetta2-gc;
+in
+
+{
+  options.rosetta2-gc = {
```
zowoq, line 22:
```suggestion
  cfg = config.services.rosetta2-gc;
in

{
  options.services.rosetta2-gc = {
```

Enzime: Maybe it would be better to move this under `virtualisation.rosetta.gc.*` to be more similar to NixOS

Enzime: Can you move the file to a virtualisation folder?

mweinelt: Updated. Also made the name say rosetta2, since what we're doing might be specific to that.

Enzime: Rosetta is preferable as Rosetta 1 refers to the software used to migrate from PowerPC to Intel Macs released in 2006.

mweinelt: Alright! Updated.

Enzime:
Can you also add a test?

emilazy:
Hmm, so what this service is actually doing is deleting all *purgeable space* from the disk, which AIUI is essentially cache data that is marked as inessential. It would I guess just purge the Rosetta 2 caches if the path to wherever those are is specified, but currently this has nothing to do with Rosetta 2 per se.

Nominally, it’s fine to do this indiscriminately, because it’s the FS equivalent of flushing disk cache from memory. But it makes me wonder why this service was added to the infra and if there is some known problem it addresses? As with disk cache, this space is meant to be removed and reused automatically if anything else on the system needs the space, and as every Linux user knows it’s not actually a good thing if `free -m` shows low values. It’s possible that Apple’s implementation here is imperfect, but I’m wondering if this service was added to address a specific issue or was just prophylactically introduced because it looked like disk space was low?

mweinelt:
I don't know what the source for the command we're using is. But this one does it similarly and links to a source which is unfortunately down.

https://gist.github.com/thimslugga/17adbee3c45713594c091e80ecd0cf69#file-01-ec2-mac-tuning-sh-L287-L294

Issue #1142, opened 2024-11-04
`services.aerospace.settings.on-window-detected` has the wrong type


Sabinu:
https://github.com/LnL7/nix-darwin/blob/189d2d422c773fa065cc9c72e6806f007ebb9be0/modules/services/aerospace/default.nix#L73

https://nikitabobko.github.io/AeroSpace/guide#on-window-detected-callback

it is not a list of strings; it is a set with atributes: { x = 1; y = 2; }

Enzime:
Could you make a PR?

teelahti:
Similarly, [workspace-to-monitor-force-assignment](https://nikitabobko.github.io/AeroSpace/guide#assign-workspaces-to-monitors) does not work even with the default configuration sample when inside services.aerospace.settings block. Nix complains about the data type:

```text
error: syntax error, unexpected INT_LIT
at /nix/store/2ya55va8svjxbz5hrr5c17iykwgaf2a2-source/nix/hosts/mbp/default.nix:157:7:
   156|     workspace-to-monitor-force-assignment = {
   157|       1 = "1";
```

If I had to guess, this needs to be raised to be a property of its own, with correct data types.

dbozbay:
Has anyone managed to get parse into `services.aerospace.settings.on-window-detected` correctly? It's the one setting I can't figure out. Would love to completely ditch Aerospace via Homebrew + config.file and migrate to services.aerospace, but this is holding me back.

thuvasooriya:
> Has anyone managed to parse into `services.aerospace.settings.on-window-detected` successfully? I can't figure out. Would love to completely ditch Aerospace via Homebrew +` aerospace.toml` and migrate to `services.aerospace` completely, but this is holding me back.

https://github.com/LnL7/nix-darwin/pull/1168
this one fixes the issue by allowing multiple types, not sure if it's the nix way to do things, but it works for now. I think they haven't gotten around to merging it yet

blindFS:
Unrelated question, I have a long aerospace.toml file, is it possible to load it as `services.aerospace.settings` ?
let expression seems not allowed in flake.nix.

blindFS:
> > I have a long aerospace.toml file, is it possible to load it as `services.aerospace.settings` ?
> 
> I'm not sure if this is what you are asking, you can import a toml like this
> 
> ```
>   services.aerospace = {
>     enable = true;
>     settings = pkgs.lib.importTOML ../../config/aerospace.toml;
>   };
> ```

[...]

PR #1168, opened 2024-11-14
aerospace: fix on-window-detected

talhaHavadar:
This fixes the issue #1142 

I am not proficient in nix so I tried to create a sort generic type, it might be too broad in this case but this fixes the Aerospace crash issue due to incorrect typing.


jnsgruk:
Would be nice to land this! Is there anything I can do to help?

talhaHavadar:
actually had no idea how to speed things up, this is my first time contributing this repo.

NikolayGalkin (approved):

modules/services/aerospace/default.nix
```diff
@@ -71,9 +85,17 @@ in
               description = "Default orientation for the root container.";
             };
             on-window-detected = lib.mkOption {
-              type = listOf str;
+              type = listOf (attrsOf (either callbackConf (listOf callbackConf)));
```
Enzime, line 88: I think it would be better with `types.listOf (types.attrsOf types.anything)`

Ideally we would be able to type this better, but I'm not familiar with Aerospace

Enzime (changes requested):

modules/services/aerospace/default.nix
talhaHavadar: I think this is cleaner indeed :) just pushed the changes

z0al:
fyi, this PR is older but https://github.com/LnL7/nix-darwin/pull/1168 is more complete

z0al:
@talhaHavadar my bad. I linked the wrong pr by mistake 😓 . Here is the correct pr https://github.com/LnL7/nix-darwin/pull/1208

talhaHavadar:
Closing this ticket since the fix landed with another PR

PR #1172, opened 2024-11-16
openssh: init module

Enzime:


modules/services/openssh.nix
```diff
@@ -0,0 +1,49 @@
...
+
+    services.openssh.useAppleServer = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      example = false;
+      description = ''
+        Whether to use Apple's built-in OpenSSH server.
+      '';
+    };
+  };
```
emilazy, line 26: Honestly I think we can just drop this if it defaults to `true`, since nobody is ever going to set it to `false` until/unless it’s implemented, and I guess time has mellowed me into thinking it might be too annoying to force people to explicitly toggle it to use this :)

modules/services/openssh.nix
```diff
@@ -0,0 +1,49 @@
...
+        launchctl enable system/com.openssh.sshd
+        launchctl bootstrap system /System/Library/LaunchDaemons/ssh.plist
+      fi
+    '' else ''
+      if [[ "$(systemsetup -getremotelogin | sed 's/Remote Login: //')" == "On" ]]; then
+        launchctl bootout system/com.openssh.sshd
+        launchctl disable system/com.openssh.sshd
+      fi
+    '');
+  };
```
emilazy, line 48: The logic for when `cfg.enable` is false worries me a little, because it seems like you might think your SSH server is off and you’re totally protected, but there’s actually a race condition between boot and the activation scripts running where the server can be accessed?

Do we know if there are any meaningful differences between what this does and what `systemsetup -setremotelogin` does? It may be better to just require FDA for this module, as it would certainly solve the race condition.

Enzime: There shouldn't be a race condition because the SSH server is disabled and won't get loaded again on reboot

Enzime: Currently we're adding the text to the `launchd` activation script, which is not one of the activation scripts that runs on boot (`etcChecks`, `etc` and `keyboard`) (Unrelated tangent: we'll probably want `users.users.launchd.agents` to allow us to do user activation when they log in as it might have interactive prompts)

When you `bootout` and `disable` the SSH server, it won't get reenabled on reboot, I've tested this on a VM by manually going into the Sharing menu, enabling Remote Login, doing a `darwin-rebuild switch` with `services.openssh.enable = false;`, rebooting and checking that Remote Login is still disabled

emilazy: Ah, so it actually toggles the visible preference in System Settings when you do this? That’s good enough for me then. Strange that they have a back door like this, though; I wouldn’t be surprised if we have to move to the proper command in future.

emilazy (approved):

PR #1178, opened 2024-11-17
fix: allow users to disable the homebrew check

isabelroses:
With the recent addition of https://github.com/LnL7/nix-darwin/pull/961, this unintentionally broke [nix-homebrew](https://github.com/zhaofengli/nix-homebrew) see [the issue](https://github.com/zhaofengli/nix-homebrew/issues/45). Thus this PR implments `system.checks.verifyHomebrewInstall` as a way to disable the check allowing nix-homebrew to function normally if users disable it.

### DIFF
--- a/modules/system/checks.nix
+++ b/modules/system/checks.nix
@@ -297,7 +297,7 @@ let
   '';
 
   homebrewInstalled = ''
-    if [[ ! -f ${escapeShellArg config.homebrew.brewPrefix}/brew ]]; then
+    if [[ ! -f ${escapeShellArg config.homebrew.brewPrefix}/brew && -z "''${INSTALLING_HOMEBREW:-}" ]]; then
         echo "[1;31merror: Using the homebrew module requires homebrew installed, aborting activation[0m" >&2
         echo "Homebrew doesn't seem to be installed. Please install homebrew separately." >&2
         echo "You can install homebrew using the following command:" >&2


emilazy:
Does `${config.homebrew.brewPrefix}/brew` not exist with nix-homebrew? We should probably fix the check or nix-homebrew if so.

isabelroses:
> Does `${config.homebrew.brewPrefix}/brew` not exist with nix-homebrew?

It should exist, however the issue is with first activation since it installs homebrew during an activation script so it cannot find homebrew. All following rebuilds behave normally.

ghostiam:
Doesn't exist, because brew installation fails due to the current check.
If set `homebrew.enable = false;` first, then brew will install using `nix-homebrew` and then everything will work. 
But on a clean system, the new check blocks installation with `homebrew.enable = true;`

Enzime:
Instead of adding an option to disable the check, what about adding to `nix-homebrew`:

```nix
{
  system.checks.text = mkBefore ''
    INSTALLING_HOMEBREW=1
  '';
}
```

Then we could modify the check in `nix-darwin` to see if that variable is set

isabelroses:
> Instead of adding an option to disable the check, what about adding to `nix-homebrew`:
> 
> ```nix
> {
>   system.checks.text = mkBefore ''
>     INSTALLING_HOMEBREW=1
>   '';
> }
> ```
> 
> Then we could modify the check in `nix-darwin` to see if that variable is set

I wasn't 100% sure what you ment so I hope [`2792516`](https://github.com/LnL7/nix-darwin/pull/1178/commits/27925163d403e163e7a29c0420a710ac5bd2f43b) is fine

modules/system/checks.nix
```diff
@@ -297,7 +297,7 @@ let
   '';
 
   homebrewInstalled = ''
-    if [[ ! -f ${escapeShellArg config.homebrew.brewPrefix}/brew ]]; then
+    if [[ ! -f ${escapeShellArg config.homebrew.brewPrefix}/brew && -z "$INSTALLING_HOMEBREW" ]]; then
```
Enzime, line 300:
```suggestion
    if [[ ! -f ${escapeShellArg config.homebrew.brewPrefix}/brew && -z "${INSTALLING_HOMEBREW:-}" ]]; then
```

Enzime (changes requested):

Enzime (approved):
LGTM

Thanks for the contribution

modules/system/checks.nix
```diff
@@ -297,7 +297,7 @@ let
   '';
 
   homebrewInstalled = ''
-    if [[ ! -f ${escapeShellArg config.homebrew.brewPrefix}/brew ]]; then
+    if [[ ! -f ${escapeShellArg config.homebrew.brewPrefix}/brew && -z "${INSTALLING_HOMEBREW:-}" ]]; then
```
Enzime, line 300:
```suggestion
    if [[ ! -f ${escapeShellArg config.homebrew.brewPrefix}/brew && -z "''${INSTALLING_HOMEBREW:-}" ]]; then
```

Enzime (changes requested):

Enzime (approved):

Issue #1054, opened 2024-09-03
Autoformatting PRs


dustypomerleau:
While I was preparing #1053, I noticed that `nixfmt` makes significant changes to the current formatting (I am currently running `nixfmt nixpkgs-unstable-2024-08-16`). Is `nix-darwin` interested in adopting the [RFC version](https://github.com/NixOS/rfcs/pull/166) of `nixfmt`? Should contributors use it (or explicitly avoid it)? It would be nice to have some sort of agreed approach to autoformatting.

Enzime:
I've been meaning to get around to adding an autoformatter for this codebase, I would definitely merge a PR that adds it to the codebase using https://github.com/cachix/git-hooks.nix or something similar

emilazy:
Yes, we should format with `nixfmt-rfc-style` and put the commit in `.git-blame-ignore-revs` and enforce it on all files with CI going forward. I don’t think we need an elaborate transition like Nixpkgs had. Hooks cannot be relied on for correctness since they are entirely opt‐in on the user machine, and I personally [strongly dislike pre‐commit hooks](https://matrix.to/#/!JHmxsrPoybcbBVDSjm:numtide.com/$gtoJy7R4jqS6VItUo6QRhzjkWoY5Xf_vs6ZcOnOCz_c?via=nixos.org&via=matrix.org&via=nixos.dev) as opposed to pre‐push hooks, which sadly there is less available automation around. So I would prefer to punt on hooks for now unless anyone feels like writing a pre‐push one.

Enzime:
The stage at which the hook runs at can be customized with the `hooks` option for the above framework, however testing will be need to check if the hooks support `pre-push` out of the box as they may not be able to handle multiple commits.

I don't agree that we should punt on hooks even if we could only have a `pre-commit` hook, because they're opt in and the code style will inevitably be enforced by CI, so my preference would still be to include the hook.

emilazy:
Part of the problem is that hooks/development shells are a security can of worms, because they come from the currently checked‐out commit but run outside of Nix builds. That means that they’re a very easy takeover vector for anyone who checks a PR out to review it, especially if they’re installed by default in an `.envrc`. See e.g. the discussion in https://github.com/NixOS/nixpkgs/pull/325793.

Enzime:
That sounds reasonable to me to not include `.envrc` and make it explicitly opt in to use the `devShell`

antoineco:
[...]

PR #1187, opened 2024-11-21
dock: allow setting spacer tiles

khaneliman:
You can create spacer tiles in the dock by passing empty tile-data with specific tile-types.

Used to do this in my old dotfiles and have been meaning to upstream a change to support doing this in nix-darwin. 
image

### DIFF
--- a/modules/system/defaults/dock.nix
+++ b/modules/system/defaults/dock.nix
@@ -134,10 +134,14 @@ in {
       description = ''
         Persistent applications in the dock.
       '';
-      apply = value:
-        if !(isList value)
-        then value
-        else map (app: { tile-data = { file-data = { _CFURLString = app; _CFURLStringType = 0; }; }; }) value;
+      apply =
+      let
+        tileTypes = ["spacer-tile" "small-spacer-tile"];
+        toSpecialTile = type: { tile-data = {}; tile-type = type; };
+        toAppTile = cfurl: { tile-data = { file-data = { _CFURLString = cfurl; _CFURLStringType = 0; }; }; };
+        toTile = s: if elem s tileTypes then toSpecialTile s else toAppTile s;
+      in
+      value: if isList value then map toTile value else value;
     };
 
     system.defaults.dock.persistent-others = mkOption {
--- a/modules/system/defaults/dock.nix
+++ b/modules/system/defaults/dock.nix
@@ -128,20 +128,72 @@ in {
     };
 
     system.defaults.dock.persistent-apps = mkOption {
-      type = types.nullOr (types.listOf (types.either types.path types.str));
+      type = let
+        taggedType = types.attrTag {
+              app = mkOption {
+                description = "An application to be added to the dock.";
+                type = types.str;
+              };
+              file = mkOption {
+                description = "A file to be added to the dock.";
+                type = types.str;
+              };
+              folder = mkOption {
+                description = "A folder to be added to the dock.";
+                type = types.str;
+              };
+              spacer = mkOption {
+                description = "A spacer to be added to the dock. Can be small or regular size.";
+                type = types.submodule {
+                  options.small = mkOption {
+                    description = "Whether the spacer is small.";
+                    type = types.bool;
+                    default = false;
+                  };
+                };
+              };
+            };
+
+        simpleType = types.either types.str types.path;
+        toTagged = path: { app = path; };
+        in
+      types.nullOr (types.listOf (types.coercedTo simpleType toTagged taggedType));
       default = null;
-      example = [ "/Applications/Safari.app" "/System/Applications/Utilities/Terminal.app" ];
+      example = [
+        { app = "/Applications/Safari.app"; }
+        { spacer = { small = false; }; }
+        { spacer = { small = true; }; }
+        { folder = "/System/Applications/Utilities"; }
+        { file = "/User/example/Downloads/test.csv"; }
+      ];
       description = ''
-        Persistent applications in the dock.
+        Persistent applications, spacers, files, and folders in the dock.
       '';
       apply =
       let
-        tileTypes = ["spacer-tile" "small-spacer-tile"];
-        toSpecialTile = type: { tile-data = {}; tile-type = type; };
-        toAppTile = cfurl: { tile-data = { file-data = { _CFURLString = cfurl; _CFURLStringType = 0; }; }; };
-        toTile = s: if elem s tileTypes then toSpecialTile s else toAppTile s;
+        toTile = item: if item ? app then {
+        tile-data.file-data = {
+          _CFURLString = item.app;
+          _CFURLStringType = 0;
+        };
+        } else if item ? spacer then {
+          tile-data = { };
+          tile-type = if item.spacer.small then "small-spacer-tile" else "spacer-tile";
+        } else if item ? folder then {
+          tile-data.file-data = {
+            _CFURLString = "file://" + item.folder;
+            _CFURLStringType = 15;
+          };
+          tile-type = "directory-tile";
+        } else if item ? file then {
+          tile-data.file-data = {
+            _CFURLString = "file://" + item.file;
+            _CFURLStringType = 15;
+          };
+          tile-type = "file-tile";
+        } else item;
       in
-      value: if isList value then map toTile value else value;
+      value: if value == null then null else map toTile value;
     };
 
     system.defaults.dock.persistent-others = mkOption {
--- a/tests/fixtures/system-defaults-write/activate-user.txt
+++ b/tests/fixtures/system-defaults-write/activate-user.txt
@@ -255,7 +255,7 @@ defaults write com.apple.dock 'persistent-apps' $'<?xml version="1.0" encoding="
 			<key>file-data</key>
 			<dict>
 				<key>_CFURLString</key>
-				<string>MyApp.app</string>
+				<string>/Applications/MyApp.app</string>
 				<key>_CFURLStringType</key>
 				<integer>0</integer>
 			</dict>
@@ -267,7 +267,7 @@ defaults write com.apple.dock 'persistent-apps' $'<?xml version="1.0" encoding="
 			<key>file-data</key>
 			<dict>
 				<key>_CFURLString</key>
-				<string>Cool.app</string>
+				<string>/Applications/Cool.app</string>
 				<key>_CFURLStringType</key>
 				<integer>0</integer>
 			</dict>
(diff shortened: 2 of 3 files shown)


modules/system/defaults/dock.nix
```diff
@@ -133,7 +133,13 @@ in {
       apply = value:
         if !(isList value)
         then value
-        else map (app: { tile-data = { file-data = { _CFURLString = app; _CFURLStringType = 0; }; }; }) value;
+        else map (app:
+          if app == "large-spacer"
+          then { tile-data = {}; tile-type = "spacer-tile"; }
+          else if app == "small-spacer"
+          then { tile-data = {}; tile-type = "small-spacer-tile"; }
```
khaneliman, line 140: Didn't know if another option would be to check against a list of `tile-types` and just forward the app as a tile-type into a single if block

khaneliman:
Sorry, forgot to update test fixture.

modules/system/defaults/dock.nix
```diff
@@ -131,9 +131,16 @@ in {
         Persistent applications in the dock.
       '';
       apply = value:
+        let
+          tileTypes = ["spacer-tile" "small-spacer-tile"];
+        in
```
MattSturgeon, line 136: Maybe this could be made even more readable with some named functions?

```nix
apply =
let
  tileTypes = ["spacer-tile" "small-spacer-tile"];
  toSpecialTile = type: { tile-data = {}; tile-type = type; };
  toAppTile = cfurl: { tile-data = { file-data = { _CFURLString = cfurl; _CFURLStringType = 0; }; }; };
  toTile = s: if elem s tileTypes then toSpecialTile s else toAppTile s;
in
value: if isList value then map toTile value else value;
```

However, seeing as this is being done to the option value, why not take advantage of `types.coercedTo` to a) allow users to define custom tile attrsets and b) simplify the coercion?

`coercedTo` also improves documentation by being explicit about the fact that coercion is done. Coercing manually in `apply` can surprise users who expect to be able to _read_ the same value types that they _define_.

```nix
type =
let
  # The same helpers as above
  specialTileTypes = ["spacer-tile" "small-spacer-tile"];
  toSpecialTile = type: { tile-data = {}; tile-type = type; };
  toAppTile = cfurl: { tile-data = { file-data = { _CFURLString = cfurl; _CFURLStringType = 0; }; }; };
  
  # The coercion function
  toTile = s: if elem s specialTileTypes then toSpecialTile s else toAppTile s;

  # The "tile" type
  # Or maybe a freeform submodule is better?
  tileType = (types.attrsOf anything).addCheck (attrs: attrs ? tile-data);

  # The "coerced" tile type
  coercedTileType = types.coercedTo (types.either types.path types.str) toTile tileType;
in
types.nullOr (types.listOf coercedTileType);
```

This is a change from the existing code though, so maybe out of scope for this PR?

khaneliman: > a) allow users to define custom tile attrsets
I think the type coercion looks nice, but is `addCheck` only available at the top level of `lib.types` ?

MattSturgeon: > but is addCheck only available at the top level of lib.types ?

Yeah I misremembered the exact usage.

It should be `addCheck (attrsOf anything) (hasAttr "attrname")`

Ideally the type would have an overridden description too, or be a submodule, but I don't know how far you want to go with verbose nix code 😅

khaneliman: @Enzime Any opinion on implementation options?

modules/system/defaults/dock.nix
```diff
@@ -134,10 +134,14 @@ in {
       description = ''
         Persistent applications in the dock.
       '';
-      apply = value:
-        if !(isList value)
-        then value
-        else map (app: { tile-data = { file-data = { _CFURLString = app; _CFURLStringType = 0; }; }; }) value;
+      apply =
```
Enzime, line 137: I think it would be good if you could change the type to be something more like:

```
[
  { type = "app"; path = "/Applications/Safari.app"; }
  { type = "spacer"; }
  { type = "spacer"; small = true; }
  { type = "folder"; path = "..."; }
]
```

I'll leave the exact types we should have (apps, files, spacers) and options up to you

Enzime (dismissed):

modules/system/defaults/dock.nix
MattSturgeon: Perhaps a submodule with an enum `type` option, alongside type-specific value options such as `path`?

```nix
types.submodule {
  options = {
    type = mkOption {
      type = types.enum [
        "app"
        # ...
      ];
      default = "app";
      description = "TODO";
    };
    path = mkOption { /* ... */ };
    # ...
  };
}
```

Convenience (and compatibility with the current option) could be achieved using `types.coercedTo`:

```nix
types.coercedTo
  types.str
  (path: {inherit path;})
  (submodule)
```

Enzime: https://nixos.org/manual/nixos/unstable/#sec-option-types-sums

I would use `types.attrTag`

khaneliman:
Took me too long to realize that I didn't update my flake's usage for spacer tiles.... I was so confused.

khaneliman:
@Enzime Alright, sorry I forgot about this for a bit. Updated with last suggested changes.

modules/system/defaults/dock.nix
```diff
@@ -128,18 +128,75 @@ in {
...
+              };
+              folder = mkOption {
+                description = "A folder to be added to the dock.";
+                default = null;
+                type = types.submodule {
+                  options.path = mkOption {
+                    description = "Path to the folder.";
+                    type = types.str;
+                  };
+                };
```
MattSturgeon, line 163: Don't hold me to this, but assuming an app/folder will _only_ ever need a path value, and it's unlikely this submodule will ever expand, I think we could ditch the submodule and just assign a path directly?


```nix
folder = mkOption {
  description = "A folder to be added to the dock.";
  type = types.str;
  # (No default)
};
```

```nix
example = [
  { app = "/Applications/Safari.app"; }
  { spacer.small = false; }
  { spacer = { }; }
  { folder = "/System/Applications/Utilities"; }
]
```

modules/system/defaults/dock.nix
```diff
@@ -128,18 +128,75 @@ in {
...
+        Persistent applications, spacers, and folders in the dock.
       '';
-      apply = value:
-        if !(isList value)
-        then value
-        else map (app: { tile-data = { file-data = { _CFURLString = app; _CFURLStringType = 0; }; }; }) value;
+      apply =
+      let
+        toTile = item: if item ? app then {
+          tile-data = { file-data = { _CFURLString = item.app.path; _CFURLStringType = 0; }; };
```
MattSturgeon, line 181: Maybe I've missed some nix-darwin style preference, but IMO this is more readable as:

```nix
tile-data.file-data = {
  _CFURLString = item.app.path;
  _CFURLStringType = 0;
};
```

modules/system/defaults/dock.nix
```diff
@@ -128,18 +128,75 @@ in {
...
     };
-
+    # apply =
+    # let
+    #   tileTypes = ["spacer-tile" "small-spacer-tile"];
+    #   toSpecialTile = type: { tile-data = {}; tile-type = type; };
+    #   toAppTile = cfurl: { tile-data = { file-data = { _CFURLString = cfurl; _CFURLStringType = 0; }; }; };
+    #   toTile = s: if elem s tileTypes then toSpecialTile s else toAppTile s;
+    # in
+    # value: if isList value then map toTile value else value;
```
MattSturgeon, line 199: Leftover debugging code?

modules/system/defaults/dock.nix
khaneliman: Yeah, I like this more... I was just trying to fulfill the requested form above. https://github.com/LnL7/nix-darwin/pull/1187#discussion_r1865152117

modules/system/defaults/dock.nix
khaneliman: I haven't seen any formatter configuration or documentation on formatting in this repo. So, I have no idea whether I should just format the whole file with nixfmt or just try to fit it into whats in here...

dustypomerleau: There was a discussion in #1054 that would be worth revisiting. I agree that a default format would be really helpful.

modules/system/defaults/dock.nix
MattSturgeon: I think a submodule makes sense if we expect there could be other keys to be set in future, e.g. if it was likely you may set `path` _and_ `name` or `icon` or whatever.

I'm not familiar with the `tile-data` schema so I don't know how likely that is.

khaneliman: TBH I'm not very familiar with it either and the documentation isn't great either https://developer.apple.com/documentation/devicemanagement/dock/staticitem/tile-data.

khaneliman:
@Enzime friendly ping

modules/system/defaults/dock.nix
khaneliman: Updated to app and folder being just a path string.

MattSturgeon:
@Enzime any chance of taking another look? :slightly_smiling_face:

modules/system/defaults/dock.nix
```diff
@@ -128,16 +128,61 @@ in {
...
+        tile-data.file-data = {
+          _CFURLString = item.app;
+          _CFURLStringType = 0;
+        };
+        } else if item ? spacer then {
+          tile-data = { };
+          tile-type = if item.spacer.small then "small-spacer-tile" else "spacer-tile";
+        } else if item ? folder then {
+          tile-data.file-data = {
+            _CFURLString = "file://" + item.folder;
```
emilazy, line 179: I think we need to use `lib.escapeURL` here, right?

modules/system/defaults/dock.nix
```diff
@@ -128,16 +128,61 @@ in {
...
+            _CFURLStringType = 15;
+          };
+          tile-type = if strings.hasInfix "." (last (splitString "/" item.folder)) then "file-tile" else "directory-tile";
```
emilazy, line 182: I think it would be best to make this an explicit optionrather than guessing based on the filename. Especially since “folder” is a weird name to use for a file tile.

modules/system/defaults/dock.nix
```diff
@@ -128,16 +128,61 @@ in {
...
+        } else item;
+      in
+      value: if isList value then map toTile value else value;
```
emilazy, line 185: Do we not need to handle `toTile` in the non‐list case?

emilazy (changes requested):

modules/system/defaults/dock.nix
khaneliman: Seems to break when using that. 
image
image

modules/system/defaults/dock.nix
```diff
@@ -128,16 +128,72 @@ in {
...
+          tile-type = "directory-tile";
+        } else if item ? file then {
+          tile-data.file-data = {
+            _CFURLString = "file://" + item.file;
+            _CFURLStringType = 15;
+          };
+          tile-type = "file-tile";
+        } else item;
+      in
+      value: if isList value then map toTile value else toTile value;
```
MattSturgeon, line 196: Looking at the type, `value` will always be null or a list.

Therefore shouldn't this be something like:
```suggestion
      value: if value == null then null else map toTile value;
```

MattSturgeon (approved):

emilazy (approved):
Looks good to me, thanks!

PR #1189, opened 2024-11-22
defaults: expose-group-by-app -> expose-group-apps

khaneliman:
Noticed option wasn't doing anything and saw it was renamed in previous macOS releases. 

Not sure why the rename trace is being triggered a couple times, I don't see it referenced anywhere.

### DIFF
--- a/modules/system/defaults-write.nix
+++ b/modules/system/defaults-write.nix
@@ -9,6 +9,9 @@ let
     "defaults write ${domain} '${key}' $'${strings.escape [ "'" ] (generators.toPlist { } value)}'";
 
   defaultsToList = domain: attrs: mapAttrsToList (writeDefault domain) (filterAttrs (n: v: v != null) attrs);
+  # Filter out options to not pass through
+  # dock has alias options that we need to ignore
+  dockFiltered = (builtins.removeAttrs cfg.dock ["expose-group-by-app"]);
 
   # defaults
   alf = defaultsToList "/Library/Preferences/com.apple.alf" cfg.alf;
@@ -21,7 +24,7 @@ let
   LaunchServices = defaultsToList "com.apple.LaunchServices" cfg.LaunchServices;
   NSGlobalDomain = defaultsToList "-g" cfg.NSGlobalDomain;
   menuExtraClock = defaultsToList "com.apple.menuextra.clock" cfg.menuExtraClock;
-  dock = defaultsToList "com.apple.dock" cfg.dock;
+  dock = defaultsToList "com.apple.dock" dockFiltered;
   finder = defaultsToList "com.apple.finder" cfg.finder;
   hitoolbox = defaultsToList "com.apple.HIToolbox" cfg.hitoolbox;
   magicmouse = defaultsToList "com.apple.AppleMultitouchMouse" cfg.magicmouse;
--- a/modules/system/defaults/dock.nix
+++ b/modules/system/defaults/dock.nix
@@ -6,6 +6,10 @@ let
   # Should only be used with options that previously used floats defined as strings.
   inherit (config.lib.defaults.types) floatWithDeprecationError;
 in {
+  imports = [
+    (mkRenamedOptionModule [ "system" "defaults" "dock" "expose-group-by-app" ] [ "system" "defaults" "dock" "expose-group-apps" ])
+  ];
+
   options = {
 
     system.defaults.dock.appswitcher-all-displays = mkOption {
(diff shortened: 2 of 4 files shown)


Enzime:
What version did it get renamed in?

modules/system/defaults/dock.nix
```diff
@@ -220,7 +224,7 @@ in {
         Magnified icon size on hover. The default is 16.
       '';
     };
-   
+
```
Enzime, line 227:
```suggestion
```

Enzime (changes requested):
Can you add this option to the tests?

khaneliman:
> Can you add this option to the tests?

This is showing  a peculiar behavior with the module. Adding the rename alias causes the test to generate both settings instead of just the option provided.. If I remove the alias we see just the setting passed in created.

modules/system/defaults-write.nix
```diff
@@ -10,6 +10,10 @@ let
 
   defaultsToList = domain: attrs: mapAttrsToList (writeDefault domain) (filterAttrs (n: v: v != null) attrs);
 
+  # Filter out options to not pass through
+  # dock has alias options that we need to ignore
+  dockFiltered = (builtins.removeAttrs cfg.dock ["expose-group-by-app"]);
```
Enzime, line 15: Without this, does that mean when you run it locally it sets both values?

Enzime: Can you use `mkRemovedOption` instead then?

khaneliman: I tried that and it functioned the same, except with an assertion instead of a warning.

Enzime: From the code you sent, it looks like there wouldn't be a value set for the old option name when you use `mkRemovedOptionModule`

khaneliman: There's a couple different open PRs to try and implement a better solution for trying to handle passing metadata around with options so eventually we don't have to do stuff like this https://github.com/NixOS/nixpkgs/pull/341199

khaneliman: > From the code you sent, it looks like there wouldn't be a value set for the old option name when you use `mkRemovedOptionModule`

I would have assumed, too. But, it throws the error that we're trying to use a removed option.

Enzime: How about refactoring the code to iterate through `options` and filter out hidden fields, so we don't have to maintain a separate list?

khaneliman: That's the original approach I was going to go through, but was convinced, during brainstorming with someone else, to go this route since a hidden value isn't specific to an alias rename. I can go that route if that's what you'd prefer in this repo.

Enzime: I'm curious what other situations would construct `hidden` values that we wouldn't want filtered out

khaneliman: Sometimes you have public values that a user can use to configure something and you use a hidden option to construct the final value.

Enzime: That should be fine as I don't think we would use any of those for `system.defaults`

I'm not sure if you're interested, one wishlist item I have for refactoring the `defaults` code is to properly integrate `CustomUserPreferences` and `CustomSystemPreferences`, by moving all the options to be under `system.defaults.system.<...>` and `system.defaults.user.<...>`, making all the existing options be live inside there, using the freeform settings type as well as submodules to allow each group to define its domain

Something like:

```nix
{
  options.system.defaults.system.alf.globalstate = mkOption {
   ...
  };

  config.system.defaults.system.alf.domain = "/Library/Preferences/com.apple.alf";
}
```

Enzime: I'm fine with leaving this as is as the list is currently just this one renamed option

tests/system-defaults-write.nix
```diff
@@ -46,6 +46,7 @@
   system.defaults.menuExtraClock.Show24Hour = false;
   system.defaults.menuExtraClock.ShowDayOfWeek = true;
   system.defaults.menuExtraClock.ShowDate = 2;
+  system.defaults.dock.expose-group-apps  = true;
```
Enzime, line 49:
```suggestion
  system.defaults.dock.expose-group-apps = true;
```

Enzime (changes requested):

modules/system/defaults-write.nix
khaneliman: > I'm fine with leaving this as is as the list is currently just this one renamed option

Sounds good, I started a branch for refactoring the defaults-write stuff for merging and using freeform type support. But, I am still working on some vimPlugins nixpkgs stuff atm that's keeping me mentally from being able to focus on other work atm.

Enzime (approved):
LGTM

Thanks for the contribution

PR #1198, opened 2024-11-26
Fix detection of /etc/synthetic.conf entry

### LINKED ISSUE #1197: /etc/synthetic.conf gets a new entry every run
I'm not sure if this is a regression from  https://github.com/LnL7/nix-darwin/pull/1177 (it doesn't substantially change the way this works) but after I updated my flake, now every time I run a `darwin-rebuild` I'm getting an extra line in `/etc/synthetic.conf`:

```
➜ sudo wc -l /etc/synthetic.conf
      33 /etc/synthetic.conf

➜ darwin-rebuild switch --flake ".#myhost"
setting up /run via /etc/synthetic.conf...
...

➜ sudo wc -l /etc/synthetic.conf
      34 /etc/synthetic.conf

➜ sudo tail -n 5 /etc/synthetic.conf
run     private/var/run
run     private/var/run
run     private/var/run
run     private/var/run
run     private/var/run

➜ sudo head -n 5 /etc/synthetic.conf
nix
run     private/var/run
run     private/var/run
run     private/var/run
run     private/var/run
run     private/var/run
run     private/var/run
run     private/var/run
run     private/var/run
run     private/var/run
```

Flake.lock relevant parts: (from https://github.com/ascandella/dotfiles/blob/4fcb2a10b8d09bc7b97c2df5a2de81b801acc002/flake.lock#L94)

```
      "locked": {
        "lastModified": 1732420287,
        "narHash": "sha256-CzvYF4x6jUh/+NEEIFrIY5t1W/N3IA2bNZJiMXu9GTo=",
        "owner": "LnL7",
        "repo": "nix-darwin",
        "rev": "3c52583b99666a349a6219dc1f0dd07d75c82d6a",
        "type": "github"
      },
```

```
$ ls -l /etc/synthetic.conf
-rw-------  1 root  wheel  664 Nov 25 15:52 /etc/synthetic.conf
```

I suspect we need to change this `grep` to `sudo grep` command here?
https://github.com/LnL7/nix-darwin/blob/3c52583b99666a349a6219dc1f0dd07d75c82d6a/modules/system/base.nix#L8-L11

ascandella:
This file is owned by root and mode 600 on my system, so the grep is failing and it's adding a new entry every run.

```sh
-rw-------  1 root  wheel  664 Nov 25 15:52 /etc/synthetic.conf
```

Fixes #1197 1197


modules/system/base.nix
```diff
@@ -5,7 +5,7 @@
     IFS="." read -r -a macOSVersion <<< "$(sw_vers -productVersion)"
```
Enzime, line 6: Could you add logic here to detect if we have more than one line containing `run\t` and to delete all of the lines?

Also, could you add logic to `chmod` the file to `644` if it is not already that?

Enzime (changes requested):

modules/system/base.nix
```diff
@@ -5,6 +5,14 @@
...
 
     if [[ ''${macOSVersion[0]} -gt 10 || ( ''${macOSVersion[0]} -eq 10 && ''${macOSVersion[1]} -ge 15 ) ]]; then
+      if [[ $(stat -c '%a' /etc/synthetic.conf) != "644" ]]; then
```
ascandella, line 8: One thing that was confusing to me here is I originally wrote this for macOS's `/usr/bin/stat` which has a different argument format. It appears that the environment that this runs in uses a gnu `stat` but please check me here.

modules/system/base.nix
```diff
@@ -5,6 +5,14 @@
...
+        echo "fixing permissions on /etc/synthetic.conf..."
+        sudo chmod 644 /etc/synthetic.conf
+      fi
```
Enzime, line 11:
```suggestion
      fi

```

modules/system/base.nix
```diff
@@ -5,6 +5,14 @@
...
+      if [[ $(grep -c '^run\b' /etc/synthetic.conf) -gt 1 ]]; then
+        echo "found duplicate run entries in /etc/synthetic.conf, removing..."
+        sudo sed -i "" -e '/^run\tprivate\/var\/run$/d' /etc/synthetic.conf
+      fi
```
Enzime, line 15:
```suggestion
      fi

```

modules/system/base.nix
Enzime: Yeah fairly sure `stat` comes from `coreutils` and the command works for me:

```
$ nix shell nixpkgs#coreutil -c stat -c '%a' /etc/synthetic.conf
600
```

modules/system/base.nix
Enzime: Most files in `/etc` are world readable and in my opinion this file should be as well, I don't think any of the information in it is secret

Enzime (changes requested):

modules/system/base.nix
ascandella: Ahh that makes total sense, thanks for clarifying!

ascandella:
> LGTM
> 
> Thanks for the contribution

Thanks for the insanely quick turnaround! Love this project and appreciate all your work, couldn't use my computer without it.

PR #1202, opened 2024-11-28
linux-builder: default `maxJobs` to amount of cores for Linux builder

Enzime:


Enzime:
@domenkozar could you take a look at this super small change? :)

domenkozar (approved):

PR #1208, opened 2024-12-02
aerospace: add `workspace-to-monitor-force-assignment` option and fix `on-window-detected` type

### LINKED ISSUE #1142: `services.aerospace.settings.on-window-detected` has the wrong type

thuvasooriya:
trying to fix #1142

this might be redundant or even extra since i found out later that we already have #1168
fixes the error for most of my use cases not sure if there are any other issues.

let me know if there is some changes that i should make for this to be merged.

### DIFF
--- a/modules/services/aerospace/default.nix
+++ b/modules/services/aerospace/default.nix
@@ -9,7 +9,31 @@ let
   cfg = config.services.aerospace;
 
   format = pkgs.formats.toml { };
-  configFile = format.generate "aerospace.toml" cfg.settings;
+  filterAttrsRecursive = pred: set:
+    lib.listToAttrs (
+      lib.concatMap (
+        name: let
+          v = set.${name};
+        in
+          if pred v
+          then [
+            (lib.nameValuePair name (
+              if lib.isAttrs v
+              then filterAttrsRecursive pred v
+              else if lib.isList v
+              then
+                (map (i:
+                  if lib.isAttrs i
+                  then filterAttrsRecursive pred i
+                  else i) (lib.filter pred v))
+              else v
+            ))
+          ]
+          else []
+      ) (lib.attrNames set)
+    );
+  filterNulls = filterAttrsRecursive (v: v != null);
+  configFile = format.generate "aerospace.toml" (filterNulls cfg.settings);
 in
 
 {
@@ -72,9 +96,84 @@ in
               description = "Default orientation for the root container.";
             };
             on-window-detected = lib.mkOption {
-              type = listOf str;
+              type = listOf (submodule {
+                options = {
+                  "if" = lib.mkOption {
+                    type = submodule {
+                      options = {
+                        app-id = lib.mkOption {
+                          type = nullOr str;
+                          default = null;
+                          description = "The application ID to match (optional).";
+                        };
+                        workspace = lib.mkOption {
+                          type = nullOr str;
+                          default = null;
+                          description = "The workspace name to match (optional).";
+                        };
+                        window-title-regex-substring = lib.mkOption {
+                          type = nullOr str;
+                          default = null;
+                          description = "Substring to match in the window title (optional).";
+                        };
+                        app-name-regex-substring = lib.mkOption {
+                          type = nullOr str;
+                          default = null;
+                          description = "Regex substring to match the app name (optional).";
+                        };
+                        during-aerospace-startup = lib.mkOption {
+                          type = nullOr bool;
+                          default = null;
+                          description = "Whether to match during aerospace startup (optional).";
+                        };
+                      };
+                    };
+                    default = { };
+                    description = "Conditions for detecting a window.";
+                  };
+                  check-further-callbacks = lib.mkOption {
+                    type = nullOr bool;
+                    default = null;
+                    description = "Whether to check further callbacks after this rule (optional).";
+                  };
+                  run = lib.mkOption {
+                    type = oneOf [str (listOf str)];
+                    example = ["move-node-to-workspace m" "resize-node"];
+                    description = "Commands to execute when the conditions match (required).";
+                  };
+                };
+              });
               default = [ ];
-              description = "Commands to run every time a new window is detected.";
+              example = [
+                {
+                  "if" = {
+                    app-id = "Another.Cool.App";
+                    workspace = "cool-workspace";
+                    window-title-regex-substring = "Title";
+                    app-name-regex-substring = "CoolApp";
+                    during-aerospace-startup = false;
+                  };
+                  check-further-callbacks = false;
+                  run = ["move-node-to-workspace m" "resize-node"];
+                }
+              ];
+              description = "Commands to run every time a new window is detected with optional conditions.";
+            };
+            workspace-to-monitor-force-assignment = lib.mkOption {
+              type = attrsOf (oneOf [int str (listOf str)]);
+              default = { };
+              description = ''
+                Map workspaces to specific monitors.
+                Left-hand side is the workspace name, and right-hand side is the monitor pattern.
+              '';
+              example = {
+                "1" = 1; # First monitor from left to right.
+                "2" = "main"; # Main monitor.
+                "3" = "secondary"; # Secondary monitor (non-main).
+                "4" = "built-in"; # Built-in display.
+                "5" = "^built-in retina display$"; # Regex for the built-in retina display.
+                "6" = ["secondary" "dell"]; # Match first pattern in the list.
+              };
             };
             on-focus-changed = lib.mkOption {
               type = listOf str;
--- a/tests/services-aerospace.nix
+++ b/tests/services-aerospace.nix
@@ -21,6 +21,32 @@ in
       alt-k = "focus up";
       alt-l = "focus right";
     };
+    on-window-detected = [
+      {
+        "if" = {
+          app-id = "Another.Cool.App";
+          during-aerospace-startup = false;
+        };
+        check-further-callbacks = false;
+        run = "move-node-to-workspace m";
+      }
+      {
+        "if".app-name-regex-substring = "finder|calendar";
+        run = "layout floating";
+      }
+      {
+        "if".workspace = "1";
+        run = "layout h_accordion";
+      }
+    ];
+    workspace-to-monitor-force-assignment = {
+        "1" = 1;
+        "2" = "main";
+        "3" = "secondary";
+        "4" = "built-in";
+        "5" = "^built-in retina display$";
+        "6" = [ "secondary" "dell" ];
+    };
   };
 
   test = ''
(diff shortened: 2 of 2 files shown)


thuvasooriya:
This one solves the issue #1142 and implements [workspace-to-monitor-force-assignment]
@Enzime

Enzime (changes requested):
Can you remove all the formatting changes?

Can you update the tests as well?

tests/services-aerospace.nix
```diff
@@ -31,6 +31,6 @@ in
       ${config.out}/user/Library/LaunchAgents/org.nixos.aerospace.plist`
```
Enzime, line 32: Can you update the settings above and use the new option you added and set `on-window-detected` maybe with the example value?

tests/services-aerospace.nix
```diff
@@ -31,6 +31,6 @@ in
...
 
     echo >&2 "checking config in $conf"
-    if [ `cat $conf | wc -l` -eq "27" ]; then echo "aerospace.toml config correctly contains 27 lines"; else return 1; fi
+    if [ `cat $conf | wc -l` -eq "29" ]; then echo "aerospace.toml config correctly contains 29 lines"; else return 1; fi
```
Enzime, line 34: Can you replace this check with something that actually checks the contents of `$conf` to see that it is setting the settings correctly like:

https://github.com/LnL7/nix-darwin/blob/c6b65d946097baf3915dd51373251de98199280d/tests/services-skhd.nix#L20

Enzime (changes requested):

tests/services-aerospace.nix
thuvasooriya: i was doubtful of the initial implementation of the testing as well, unfortunately I didn't know much about it, I'll try this weekend to replicate the skhd one.

z0al:
Works locally for me. Thanks @thuvasooriya for taking care of this

modules/services/aerospace/default.nix
```diff
@@ -72,9 +72,37 @@ in
               description = "Default orientation for the root container.";
             };
             on-window-detected = lib.mkOption {
-              type = listOf str;
+              type = types.listOf (types.attrsOf types.anything);
```
z0al, line 75: Is there any reason for `types.anything` instead of a proper sub module?

talhaHavadar: I did try to come up with something more restricted (not much really) in #1168 but don't have much knowledge about aerospace and possible values of this toml table so we kind of backed up to accept everything. Just leaving it here in case we want to play with it a bit.
```
  callbackConf =
    with lib.types;
    oneOf [
      bool
      int
      float
      str
      path
      attrs
    ]
    // {
      description = "AeroSpace config callback (bool, int, float, str, path or attrs)";
    };
...
            on-window-detected = lib.mkOption {
              type = listOf (attrsOf (either callbackConf (listOf callbackConf)));
...
```

z0al: > don't have much knowledge about aerospace and possible values of this toml table

Looking at the code & docs I believe: 
1.`run` command is an array of strings and it's required.
2. `if.app-id`, `if.workspace`, `if.window-title-regex-substring` and `if.app-name-regex-substring` are all optional and of type string
3. `if.during-aerospace-startup` is a boolean and optional 
4. `check-further-callbacks` is optional and of type boolean

I will leave it to @thuvasooriya if they want to add proper typing here. Otherwise, I can follow up with a PR once this is merged.

thuvasooriya: Thankyou for the valuable PR @talhaHavadar, and @z0al for taking the time to go through the docs. This was my first Nix contribution ever so I didn't want to mess around too much.
My initial solution before I saw #1168 was
```
              type = listOf (attrsOf (oneOf [
                bool
                str
                attrs
              ]));
```
which was not rigid in terms of types but then I changed it according to the suggestion made by @Enzime in #1168.
I'll see if I can come up with anything better. Thanks again.

thuvasooriya: https://github.com/thuvasooriya/nix-darwin/commit/49c1990eab619be2fc46528cc2d797316585edb3
let me know if this can be merged to this PR @Enzime @z0al
Tried to whip something up from the above doc description, not sure if it handles all the edge cases. but works with the test cases and my config

thuvasooriya: Here is some information, so that some smart people can figure out what's wrong quickly. Hope it helps. I've tried some methods to make the fields with default values disappear only to fail.

current PR generated config - **works properly**
```toml
[[on-window-detected]]
run = "layout floating"

[on-window-detected.if]
app-name-regex-substring = "telegram|finder"

[[on-window-detected]]
run = "move-node-to-workspace 0"

[on-window-detected.if]
app-name-regex-substring = "ripes"
```

my shoddy implementation of proper types in https://github.com/thuvasooriya/nix-darwin/commit/49c1990eab619be2fc46528cc2d797316585edb3 - **doesn't work**

```toml
[[on-window-detected]]
check-further-callbacks = true
run = "move-node-to-workspace 0"

[on-window-detected.if]
app-id = ""
app-name-regex-substring = "ripes"
during-aerospace-startup = false
window-title-regex-substring = ""
workspace = ""
```

Enzime: If you change the type of `app-id` to be `types.nullOr types.str` and set the default value to `null`, it should omit it from the output

thuvasooriya: > If you change the type of `app-id` to be `types.nullOr types.str` and set the default value to `null`, it should omit it from the output

Yeah that's one of the very first things i tried

```nix
                        app-id = lib.mkOption {
                          type = nullOr str;
                          default = null;
                          description = "The application ID to match (optional).";
                        };
```

```sh
error: builder for '/nix/store/6i98bgjlsmqsaqn5hz3id9p0ipchf3ln-aerospace.toml.drv' 
failed with exit code 1;
       last 1 log lines:
       > Error: Cannot convert data to TOML (null values are not supported)
```

z0al: You can try to [filter the config](https://github.com/NixOS/nixpkgs/blob/0850b8d575c01f01d52084419fa5932046b29d7a/nixos/modules/services/security/kanidm.nix#L45C18-L45C68) of null values before passing the settings to`toml.generate`

thuvasooriya: > You can try to [filter the config](https://github.com/NixOS/nixpkgs/blob/0850b8d575c01f01d52084419fa5932046b29d7a/nixos/modules/services/security/kanidm.nix#L45C18-L45C68) of null values before passing the settings to`toml.generate`

This is what i was looking for.. Thx!

thuvasooriya: unfortunately replicating the filter is somewhat troublesome for me.
my guess is that the list of submodules is not playing nicely with filterAttrs.
I tried the following and still get the above cannot convert data to TOML, meaning the null is not filtered.

I tried the default functions, filterAttrs, filterAttrsRecursive
Then I tried to implement a custom function to iterate over each item in the submodule list. Which doesn't work either. Maybe there is some obvious mistake that I'm overlooking. 
Any help or direction would be appreciated. @Enzime @z0al
```nix
  filterOptionalFields = value:
    if lib.isAttrs value then
      let
        filteredAttrs = lib.mapAttrs (_: v: filterOptionalFields v) value;
        nonNullAttrs = lib.filterAttrs (_: v: v != null) value;
      in
        filteredAttrs // nonNullAttrs
    else if lib.isList value then
      lib.filter (x: x != null) (map filterOptionalFields value)
    else
      value;
  # filterOptionalFields = lib.filterAttrsRecursive (_: v: v != null);
  configFile = format.generate "aerospace.toml" (filterOptionalFields cfg.settings);
```

z0al: Ah, I see. My bad. I didn't know that `filterAttrsRecursive` doesn't handle the case of lists.

Here is a custom implementation that works. I took the original `filterAttrsRecursive` code and added a condition for lists + changed the predicate signature to accept only values.

```nix
  # Code needs proper formatting
  filterAttrsRecursive =
    pred:
    set:
    lib.listToAttrs (
      lib.concatMap
        (name:
        let v = set.${name}; in
        if pred v then [
          (lib.nameValuePair name (
            if lib.isAttrs v then filterAttrsRecursive pred v
            else if lib.isList v then
              (
                map
                  (i:
                  if lib.isAttrs i then
                    filterAttrsRecursive pred i
                  else i)
                  (lib.filter pred v)
              )
            else v
          ))
        ] else [ ]
        )
        (lib.attrNames set)
    );
  filterNulls = filterAttrsRecursive (v: v != null);
  configFile = format.generate "aerospace.toml" (filterNulls cfg.settings);
```

Let's hope that Enzime has a trick up their sleeve so we don't have to keep this custom function :) 

P.S. Don't forget to change all optional attributes to default to null even things like `check-further-callbacks` as Aerospace behaves differently when it's set to `false` vs left out.

thuvasooriya: cool! it works! thanks.
merged the changes, and yes every optional is set to null by default.

tests/services-aerospace.nix
```diff
@@ -43,5 +65,20 @@ in
     grep 'alt-j = "focus down"' $conf
     grep 'alt-k = "focus up"' $conf
     grep 'alt-l = "focus right"' $conf
+
+    grep 'check-further-callbacks = false' $conf
+    grep 'run = "move-node-to-workspace m"' $conf
+    grep 'app-id = "Another.Cool.App"' $conf
+    grep 'during-aerospace-startup = false' $conf
```
z0al, line 72: It would be great if you also check that unspecified options don't appear in the config e.g. the `window-title-regex-substring`.

thuvasooriya: I don't think default null values have a way of somehow entering TOML files, since they have the concept of null = omitted. Hence I'm not sure how to feasibly test that in this specific implementation. maybe we check that `check-further-callbacks = false` is not present before the line `run = "layout floating"`

Enzime: I think it's fine to do a negative assertion for fields that don't show up in the config file, you can also be explicit about the value by setting them to `null` intentionally rather than relying on the default value of the option

You can see how I did them here: https://github.com/LnL7/nix-darwin/commit/fd6660cb9182fde5e593246e311dc3c2dd4b9d13

thuvasooriya: thanks for the hint!
I'll try to tackle that when I get some time.

thuvasooriya:
@Enzime just made a small change to detect `window-title-regex-substring` not being present, and added example configs for all the other callback examples. if we have to do this properly then we should check how the whole file is output in order to check the presence or absence of certain troublemakers.

This works flawless for now. Let me know if you want anything else for this to be merged.

yzx9:
Hello, just checking in—any updates? I really need this!

thuvasooriya:
@Enzime pinging again just in case you missed this. I hope this has no blockers for a merge.

Enzime:
Can you squash all the commits and rebase this on top of master

thuvasooriya:
@Enzime can you confirm this is what you requested? never done this sort of thing before

Enzime (approved):
LGTM

Thanks for the contribution

PR #1213, opened 2024-12-04
fix(aerospace): allow startup commands

z0al:
The `after-startup-command` is disabled by mistake. While it's true that the `after-login-command` doesn't work, the startup commands work just fine (see [relevant code](https://github.com/nikitabobko/AeroSpace/blob/de93349dc342175b0f8332d2db4c41c433378eb1/Sources/AppBundle/initAppBundle.swift#L30-L33))


z0al:
> Can you update the tests?

@Enzime done

Enzime (approved):
LGTM

Thanks for the contribution

PR #1215, opened 2024-12-05
fix(aerospace): remove unusable on-window-detected option

S1M1S:
Add extraConfig to replace this.

After a deep dive into pkgs.fomat.toml json2toml is incapable of handling the dot in `if.app-id`, so it must be added with plain text.

Also [on-window-changed] seems to have been replaced by [[on-window-changed]] in the aerospace config spec, so we're better off removing the option I removed.

Tests passing, and my config is working.

### DIFF
--- a/modules/services/aerospace/default.nix
+++ b/modules/services/aerospace/default.nix
@@ -9,7 +9,9 @@ let
   cfg = config.services.aerospace;
 
   format = pkgs.formats.toml { };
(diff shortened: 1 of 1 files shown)


z0al:
Have you seen https://github.com/LnL7/nix-darwin/pull/1208?

S1M1S:
Looks like my lack of understanding of TOML is showing! #1208 works for my needs.

PR #1218, opened 2024-12-07
github-runner: fix service not starting

Enzime:
https://github.com/LnL7/nix-darwin/commit/3b738c765de1bb4ecc4993fa092b27dd46d495ed made `/var/log/github-runner` have `o=`, however the launch daemon fails to redirect stdout/stderr if the parent directory doesn't have `o=rx`, even if the inner directory has `o=rx`

### DIFF
--- a/modules/services/github-runner/options.nix
+++ b/modules/services/github-runner/options.nix
@@ -3,7 +3,9 @@
 , ...
 }:
 
-with lib;
+let
+  inherit (lib) literalExpression mkOption mkPackageOption types;
+in
 {
   options.services.github-runners = mkOption {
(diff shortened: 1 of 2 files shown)


Enzime:
@emilazy @domenkozar can you review this fix as GitHub runners are currently broken

sandydoo:
@Enzime, I was deploying some runners today and ran into code 78 (permissions issues). I did some manual permissions hackery and, while I'm not entirely certain what made it work, I did apply `chmod o=rx` to `/var/lib/github-runners`. It seems likely that this directory also has overly restrictive permissions.

Issue #1027, opened 2024-08-03
Option to add `vscode` entries in Brewfile


ian-h-chamberlain:
As described in the `homebrew-bundle` README, `vscode` is a [supported option](https://github.com/Homebrew/homebrew-bundle?tab=readme-ov-file#usage) in Brewfiles.

This seems like it would be a convenient addition to the `homebrew` module - it's already possible to achieve something similar with `homebrew.extraConfig` but it seems reasonable enough to add this for parity with casks, whalebrew etc.

One question that comes to mind is whether any automatic installation of VSCode should be enabled by such an option, like how `whalebrew` does:
> When this option is used, "whalebrew" is automatically added to [homebrew.brews](https://daiderd.com/nix-darwin/manual/index.html#opt-homebrew.brews).

Some people might be using `programs.vscode` or other methods to install VSCode, but in that case they're probably declaring extensions that way, so maybe it would make sense to install the `visual-studio-code` cask (although some people may want `visual-studio-code@insiders` instead).
I'd probably lean towards just documenting that VSCode must be installed _somehow_, since activation would probably just fail anyway if it wasn't: https://github.com/Homebrew/homebrew-bundle/blob/e97d33cc31c64fdd0c7412c182668783a2d706d6/lib/bundle/vscode_extension_installer.rb#L22

Would a PR to add this functionality be accepted? I have a working implementation already so I'm happy to upstream if so.

PR #1222, opened 2024-12-15
add support for installing vscode extensions via brew

fcjr:


ethancedwards8 (approved):

ethancedwards8:
@Samasaur1 any reason why this couldn't be merged?

fcjr:
Hey @Enzime just thought I'd check in here and see if there is anything else required on my end to get this past the finish line.

Please let me know, thanks!

ctcjab:
As a user who would also like to see this PR merged, I'm also wondering if there's anything I can do to help here.

ctcjab:
Any other maintainers interested in getting this nice feature landed? CC @Samasaur1 @malob et al.

malob:
I'm not very familiar with Homebrew for installing Visual Studio Code plugins, so it's difficult for me to weigh in on this directly. But I had Claude Code do a review of this PR for me, here's what it concluded (which seems right to me).

---

## PR #1222 Review: Add VS Code Extensions via `brew bundle`

### Summary

The PR adds a `homebrew.vscode` option that generates `vscode "extension-id"` lines in the Brewfile, enabling `brew bundle` to manage VS Code extensions. It follows the same pattern as the existing `whalebrews` option.

### The Good

1. **Real `brew bundle` feature**: The `vscode` directive is a first-class feature of `brew bundle` (now merged into `Homebrew/brew` core). This is a legitimate gap in the nix-darwin homebrew module.

2. **Code quality is solid**: The implementation follows existing patterns closely -- the Brewfile generation (`vscode "golang.go"`) is correct, the option type (`listOf str`) matches what you'd expect, and the test covers it.

3. **Community demand**: There are multiple people asking for this to land (the PR is from Dec 2024 with comments through Feb 2026). Issue #1027 originally requested it.

4. **Cleanup support**: Homebrew's cleanup command already handles VS Code extensions -- if `onActivation.cleanup` is set to `"uninstall"` or `"zap"`, extensions not in the Brewfile will be removed. This "just works" with the generated Brewfile entries.

### The Problem: Auto-adding `visual-studio-code` cask

This is the one significant design flaw. The PR adds this line:

```nix
homebrew.casks = optional (cfg.vscode != [ ]) "visual-studio-code";
```

This forces the regular `visual-studio-code` cask into your cask list whenever any extensions are declared. This is problematic because:

**Homebrew itself supports 4 editors** (from [`bundle.rb:which_vscode`](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/bundle.rb)):

```ruby
def self.which_vscode
  @which_vscode ||= which("code", ORIGINAL_PATHS)
  @which_vscode ||= which("codium", ORIGINAL_PATHS)
  @which_vscode ||= which("cursor", ORIGINAL_PATHS)
  @which_vscode ||= which("code-insiders", ORIGINAL_PATHS)
end
```

So users of **VSCodium**, **Cursor**, or **VS Code Insiders** would get regular VS Code force-installed alongside their preferred editor.

**Homebrew already auto-installs VS Code if needed** (from [`vscode_extension_installer.rb:preinstall!`](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/bundle/vscode_extension_installer.rb)):

```ruby
def self.preinstall!(name, no_upgrade: false, verbose: false)
  if !Bundle.vscode_installed? && Bundle.cask_installed?
    Bundle.brew("install", "--cask", "visual-studio-code", verbose:)
  end
  # ...
end
```

So the nix-darwin auto-add is **redundant** with Homebrew's own behavior, and **harmful** for users of alternative editors.

The original issue author (#1027) anticipated this exact problem and recommended against auto-installing:

> I'd probably lean towards just documenting that VSCode must be installed _somehow_

### The `whalebrews` Analogy Doesn't Quite Hold

The PR mirrors the `whalebrews` pattern (which auto-adds `"whalebrew"` to `brews`), but the analogy breaks down:

| | `whalebrew` | VS Code |
|---|---|---|
| Variants | One CLI tool, no alternatives | 4 supported editors |
| Auto-install risk | None -- there's only one `whalebrew` | Could install wrong variant |
| Without auto-install | `whalebrew` entries would silently fail | Homebrew handles it itself |

### Recommendation

I'd suggest requesting these changes before merging:

1. **Remove the auto-cask line** (`homebrew.casks = optional (cfg.vscode != [ ]) "visual-studio-code";`). Homebrew already handles auto-installing VS Code when processing `vscode` entries, so this is redundant. Let the user add their preferred editor cask (`visual-studio-code`, `visual-studio-code@insiders`, `vscodium`, etc.) to `casks` themselves.

2. **Update the option description** to mention that a compatible editor must be installed (or will be auto-installed by Homebrew), and list the supported variants:

   ```
   List of Visual Studio Code extensions to install using Homebrew Bundle.

   A compatible editor (Visual Studio Code, VSCodium, Cursor, or VS Code Insiders)
   must be available. If none is found, Homebrew will attempt to install
   visual-studio-code automatically.
   ```

3. **Minor**: The PR description's claim "When this option is used, `visual-studio-code` is automatically added to `homebrew.casks`" should be removed from the option docs if the auto-add is dropped.

### Verdict

**Good addition, needs one change.** The core feature (generating `vscode` lines in the Brewfile) is correct and useful. The auto-adding of `visual-studio-code` to casks should be removed because it's both redundant with Homebrew's own behavior and harmful to users of alternative VS Code-compatible editors.

fcjr:
@malob thanks for the review, made the change and confirmed that brew bundle now has the behavior of auto-installing if no existing vscode version is found

malob:
Sweet. Looks good to me :)

@emilazy, @Enzime, FWIW, I'm happy for this to be merged.

PR #1231, opened 2024-12-18
services/yabai: cleanup and add logFile

khaneliman:
Similar to https://github.com/LnL7/nix-darwin/pull/840 for allowing one to monitor the application's output in their logs.

### DIFF
--- a/modules/services/yabai/default.nix
+++ b/modules/services/yabai/default.nix
@@ -72,6 +72,13 @@ in
       '';
       description = "Extra arbitrary configuration to append to the configuration file";
     };
+
+    logFile = mkOption {
+      type = types.path;
+      default = "/var/tmp/yabai.log";
+      example = "/Users/khaneliman/Library/Logs/yabai.log";
+      description = "Path to the yabai log file";
+    };
   };
 
   config = lib.mkMerge [
@@ -91,6 +98,8 @@ in
           EnvironmentVariables = {
             PATH = "${cfg.package}/bin:${config.environment.systemPath}";
           };
+          StandardOutPath = cfg.logFile;
+          StandardErrorPath = cfg.logFile;
         };
       };
     })
@@ -102,6 +111,8 @@ in
         serviceConfig = {
           RunAtLoad = true;
           KeepAlive.SuccessfulExit = false;
+          StandardOutPath = "/var/log/yabai-sa.out.log";
+          StandardErrorPath = "/var/log/yabai-sa.err.log";
         };
       };
 
--- a/modules/services/yabai/default.nix
+++ b/modules/services/yabai/default.nix
@@ -1,39 +1,41 @@
-{ config, lib, pkgs, ... }:
-
-with lib;
-
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
 let
+  inherit (lib) mkOption types;
+
   cfg = config.services.yabai;
 
-  toYabaiConfig = opts:
-    concatStringsSep "\n" (mapAttrsToList
-      (p: v: "yabai -m config ${p} ${toString v}")
-      opts);
+  toYabaiConfig =
+    opts:
+    lib.concatStringsSep "\n" (lib.mapAttrsToList (p: v: "yabai -m config ${p} ${toString v}") opts);
 
-  configFile = mkIf (cfg.config != { } || cfg.extraConfig != "")
-    "${pkgs.writeScript "yabairc" (
-      (if (cfg.config != {})
-       then "${toYabaiConfig cfg.config}"
-       else "")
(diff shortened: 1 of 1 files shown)


modules/services/yabai/default.nix
```diff
@@ -62,41 +64,59 @@ in
...
+    (lib.mkIf (cfg.enableScriptingAddition) {
       launchd.daemons.yabai-sa = {
         script = "${cfg.package}/bin/yabai --load-sa";
-        serviceConfig.RunAtLoad = true;
-        serviceConfig.KeepAlive.SuccessfulExit = false;
+        serviceConfig = {
+          RunAtLoad = true;
+          KeepAlive.SuccessfulExit = false;
+          StandardOutPath = "/var/log/yabai-sa.out.log";
+          StandardErrorPath = "/var/log/yabai-sa.err.log";
```
Samasaur1, line 115: Here you use separate files for standard output and standard error

modules/services/yabai/default.nix
```diff
@@ -62,41 +64,59 @@ in
...
+              "-c"
+              configFile
+            ];
+          KeepAlive = true;
+          RunAtLoad = true;
+          EnvironmentVariables = {
+            PATH = "${cfg.package}/bin:${config.environment.systemPath}";
+          };
+          StandardOutPath = cfg.logFile;
+          StandardErrorPath = cfg.logFile;
```
Samasaur1, line 102: Here you use the same file for standard output and standard error

Samasaur1 (changes requested):
I noticed some inconsistencies about the log files. Intuitively (and based on #1226 which includes the upstream plist) I'd say we probably want separate standard output and standard error logs by default

Samasaur1:
Otherwise this PR looks great to me

modules/services/yabai/default.nix
```diff
@@ -62,41 +64,66 @@ in
...
+      example = "/Users/khaneliman/Library/Logs/yabai.log";
+      description = "Path to the yabai error log file";
+    };
+
+    outLogFile = mkOption {
+      type = types.path;
+      default = "/var/tmp/yabai.out.log";
+      example = "/Users/khaneliman/Library/Logs/yabai.log";
+      description = "Path to the yabai stdout log file";
+    };
```
Samasaur1, line 88: Is there a reason you picked `/var/tmp` over `/tmp` or `/var/log`? Upstream just puts them in `/tmp`, but `/var/log` (like you did for the scripting additions) also makes sense to me

Samasaur1:
Not gonna comment on the other two PRs but I have the same question about log location

modules/services/yabai/default.nix
khaneliman: User agents dont have access to /var/log but the global daemon does.

Samasaur1: That makes sense. What about `/tmp` instead of `/var/tmp`? I don't really care either way, I've just never seen `/var/tmp` used before and didn't actually know that existed

khaneliman: /tmp could be purged on reboots, /var/tmp wouldn't be. So depends on what you think the default behavior should be.

Samasaur1: Ack sorry something is up with my notifications so I missed this. My personal preference is `/tmp` because it's what upstream appears to use, I think it's more visible, and I think it's fine if these logs don't persist across a reboot.

That said, I don't have permissions to merge this, so we could leave it up to @Enzime

0rphee:
Just chiming in, the default yabai plist (as I'm writing this) can be found here: https://github.com/koekeishiya/yabai/blob/f4ada8d557aeb9c8ba0d031ee6f645acffff50fd/src/misc/service.h#L8-L42

As a user, I think its better to keep most of it as-is, to avoid differences in behavior from a 'normal' yabai installation. However, it would be nice to allow the option to override the launch agent configuration.

Samasaur1:
Yeah, just the fct that the option exists means that users will be able to override it. I think we should match upstream and change the default log location to under `/tmp`

Samasaur1 (approved):

spikespaz (changes requested):
This PR seems to have changed the formatter, which makes for a noisy diff.

Could you please format the original file with your formatter, and standardize it for nix-darwin, and then rebase this PR to produce a clean diff, with only changes necessary for the stated goal?

Either that, or try to make this change without a formatter.

khaneliman:
> This PR seems to have changed the formatter, which makes for a noisy diff.
> 
> Could you please format the original file with your formatter, and standardize it for nix-darwin, and then rebase this PR to produce a clean diff, with only changes necessary for the stated goal?
> 
> Either that, or try to make this change without a formatter.

It's already 2 commits to separate the formatting from the function.

0rphee:
Is this PR dead?

PR #1232, opened 2024-12-18
services/skhd: cleanup and add logFile

khaneliman:
Similar to https://github.com/LnL7/nix-darwin/pull/840 and https://github.com/LnL7/nix-darwin/pull/1231 for allowing one to monitor the application's output in their logs.

### DIFF
--- a/modules/services/skhd/default.nix
+++ b/modules/services/skhd/default.nix
@@ -1,26 +1,29 @@
-{ config, lib, pkgs, ... }:
-
-with lib;
-
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
 let
   cfg = config.services.skhd;
-in
 
+  inherit (lib) mkOption types;
+in
 {
-  options = {
-    services.skhd.enable = mkOption {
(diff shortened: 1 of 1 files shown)


Samasaur1 (changes requested):
same comment as #1231, I assume we want separate standard output and standard error log files. otherwise good.

modules/services/skhd/default.nix
```diff
@@ -1,46 +1,71 @@
...
+      description = "Path to the error log file.";
+    };
+
+    outLogFile = mkOption {
+      type = types.path;
+      default = "/var/tmp/skhd.out.log";
+      example = "/Users/khaneliman/Library/Logs/skhd.log";
+      description = "Path to the stdout log file.";
+    };
+  };
```
Samasaur1, line 46: I think these should also match upstream and should default to being under `/tmp`

Samasaur1 (changes requested):

Samasaur1 (approved):

PR #1238, opened 2024-12-26
programs.ssh: add extraConfig option

Mic92:
Same interface as in NixOS. This is useful to apply configuration for remote builders.


Enzime (approved):
LGTM

Thanks for the contribution

PR #1239, opened 2024-12-26
add flashScreen option

that70schris:


that70schris:
nvm, this isn't the right setting

PR #1242, opened 2024-12-28
defaults: add screencapture.target option

filipkania:
Adds the ability to set `com.apple.screencapture.target` default, which tells screencapture to output screenshot to e.g. clipboard. I prefer to put my screenshots directly into my clipboard, and there wasn't an option to do that through nix-darwin, so i've made the PR :)

Tested it using different valid values on Sequoia, valid values were observed by setting different "Save to" settings and reading using `defaults read com.apple.screencapture target`:

untitled

### DIFF
--- a/modules/system/defaults/screencapture.nix
+++ b/modules/system/defaults/screencapture.nix
@@ -48,5 +48,20 @@ with lib;
         Show thumbnail after screencapture before writing to file. The default is true.
       '';
     };
+
+    system.defaults.screencapture.target = mkOption {
+      type = types.nullOr types.str;
+      default = null;
+      description = ''
+        Target to which screencapture should save screenshot to. The default is "file".
+        Valid values include:
+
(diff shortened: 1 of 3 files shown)


Enzime:
Can you update the tests?

filipkania:
sure!

modules/system/defaults/screencapture.nix
```diff
@@ -48,5 +48,20 @@ with lib;
         Show thumbnail after screencapture before writing to file. The default is true.
       '';
     };
+
+    system.defaults.screencapture.target = mkOption {
+      type = types.nullOr types.str;
```
Enzime, line 53:
```suggestion
      type = types.nullOr (types.enum [ "file" "clipboard" ... ]);
```

Enzime (changes requested):

Enzime (commented):
LGTM

Thanks for the contribution

Enzime (approved):

PR #1246, opened 2024-12-31
tailscale: fix broken DNS on IPv6 only tailnets

nolith:
When a tailnet has the disableIPv4 settings it will not deploy IPv4, resolving any ts.net address is broken because 100.100.100.100 is not reachable.

https://tailscale.com/kb/1337/acl-syntax#disableipv4


modules/services/tailscale.nix
```diff
@@ -61,10 +61,10 @@ in
       };
     };
 
-    networking.dns = mkIf cfg.overrideLocalDns [ "100.100.100.100" ];
+    networking.dns = mkIf cfg.overrideLocalDns [ "100.100.100.100" "fd7a:115c:a1e0::53" ];
 
     # Ensures Tailscale MagicDNS always works even without adding 100.100.100.100 to DNS servers
-    environment.etc."resolver/ts.net".text = "nameserver 100.100.100.100";
+    environment.etc."resolver/ts.net".text = "nameserver 100.100.100.100\nnameserver fd7a:115c:a1e0::53";
```
Samasaur1, line 67:
```suggestion
    environment.etc."resolver/ts.net".text = ''
      nameserver 100.100.100.100
      nameserver fd7a:115c:a1e0::53
    '';
```

Samasaur1: at this point I'd probably swap to a multi-line string literal

Samasaur1 (changes requested):

Samasaur1:
oops i screwed up that review a little. anyway that's a minor nit and this otherwise looks good. I grepped for `100.100.100.100` and the only other matches were in comments

Samasaur1:
ack it won't let me leave an inline comment on an unchanged part of the file, but the comment on line 36 should be updated as well:

https://github.com/LnL7/nix-darwin/blob/71a3a075e3229a7518d76636bb762aef2bcb73ac/modules/services/tailscale.nix#L36

modules/services/tailscale.nix
```diff
@@ -41,7 +41,7 @@ in
 
   config = mkIf cfg.enable {
     assertions = [{
-      assertion = !cfg.overrideLocalDns || config.networking.dns == [ "100.100.100.100" ];
+      assertion = !cfg.overrideLocalDns || config.networking.dns == [ "100.100.100.100" "fd7a:115c:a1e0::53" ];
```
Enzime, line 44:
```suggestion
      assertion = cfg.overrideLocalDns -> (builtins.elem config.networking.dns "100.100.100.100" || builtins.elem config.networking.dns "fd7a:115c:a1e0::53");
```

Enzime (changes requested):

nolith:
Thank you all, I made all the suggested edit.

I'll test it tomorrow on my other machine that is using this option.

modules/services/tailscale.nix
```diff
@@ -33,15 +33,15 @@ in
...
+        As this option sets 100.100.100.100 and fd7a:115c:a1e0::53 as your sole DNS servers, if the requirements above are not met,
         all non-MagicDNS queries WILL fail.
       '';
     };
   };
 
   config = mkIf cfg.enable {
     assertions = [{
-      assertion = !cfg.overrideLocalDns || config.networking.dns == [ "100.100.100.100" ];
+      assertion = cfg.overrideLocalDns -> (builtins.elem config.networking.dns "100.100.100.100" || builtins.elem config.networking.dns "fd7a:115c:a1e0::53");
```
Enzime, line 44:
```suggestion
      assertion = cfg.overrideLocalDns -> (builtins.any (x: !builtins.elem x [ "100.100.100.100" "fd7a:115c:a1e0::53" ]) config.networking.dns);
```

Enzime (changes requested):

modules/services/tailscale.nix
nolith: Hello, I'm not an expert of nix but I don't think this suggestion has the same behaviour of the existing implementation.

The original implementation was `!cfg.overrideLocalDns || config.networking.dns == [ "100.100.100.100" ]` and we are swapping it with an arrow operator.

`b1 -> b2` == `!b1 || b2`

```nix
nix-repl> networking_dns = [ "100.100.100.100" "fd7a:115c:a1e0::53" ]
nix-repl> networking_dns ==  [ "100.100.100.100" "fd7a:115c:a1e0::53" ]
true

nix-repl> builtins.elem "100.100.100.100" networking_dns || builtins.elem "fd7a:115c:a1e0::53" networking_dns
true

nix-repl> builtins.any (x: x != "100.100.100.100" && x != "fd7a:115c:a1e0::53") networking_dns
false

nix-repl> builtins.elem "100.100.100.100" networking_dns && builtins.elem "fd7a:115c:a1e0::53" networking_dns
true

nix-repl> builtins.any (x: x != "100.100.100.100" || x != "fd7a:115c:a1e0::53") networking_dns
true
```

Enzime: `x != 5 || x != 6` is always `true`

The original condition we’re looking to verify is that at least one element in the list is neither `100.100.100.100` nor `fd7a:115c:a1e0::53` so we want to use `&&`

nolith: > `x != 5 || x != 6` is always true

🤦 

> The original condition we’re looking to verify is that at least one element in the list is neither `100.100.100.100` nor `fd7a:115c:a1e0::53` so we want to use `&&`

But this is not what the original condition was doing `!cfg.overrideLocalDns || config.networking.dns == [ "100.100.100.100" ]`.

If we analyse `cfg.overrideLocalDns -> (builtins.any (x: x != "100.100.100.100" && x != "fd7a:115c:a1e0::53") config.networking.dns)`:
- when `cfg.overrideLocalDns == false` the assertion returns `true` - **same as before**
-  when `cfg.overrideLocalDns == true` we return the result of `(builtins.any (x: x != "100.100.100.100" && x != "fd7a:115c:a1e0::53") config.networking.dns)`, but this is checking a different condition compared to `config.networking.dns == [ "100.100.100.100" ]`. When using magicDNS you want to set a global nameserver in the admin panel, not adding another nameserver to `/etc/resolv.conf`. I think the original intention of that `assertion` was to make sure you are not adding an extra nameserver that may result in unexpected results depending on which server is selected for the resolution.

nolith: hello @Enzime, what do you think of my last comment?

nolith: @Enzime did you had any chance to read my comment?

I'd like to close this PR and get rid of the override in my flake.

nolith:
I'm pretty sure the intersection of nix-darwing users and tailscale users with an ipv6 only tailnet is very small (likely only myself).

But dear reader, if you are like me, this is how I circumvented this bug without loading a forked flake of nix-darwin.

```
  # overriding DNS doesn't work with IPv6-only networks - https://github.com/nix-darwin/nix-darwin/pull/1246
  services.tailscale.overrideLocalDns = false;
  environment.etc."resolver/ts.net".text = lib.mkForce "nameserver fd7a:115c:a1e0::53";
```

PR #1248, opened 2024-12-31
ci, readme: update stable nixpkgs to 24.11

Samasaur1:
This will also let me remove a Python 3.11 workaround in #1205 :)


Enzime (approved):
LGTM

Thanks for the contribution

PR #1253, opened 2025-01-05
Add --ignore-dependencies option for casks

jelmansouri:
Some casks have extrenal dependencies managed by brew, neovide for examples declares neovim as a dependency, a problem arises when you want to use a nix managed neovim instead


Enzime (approved):
LGTM

Thanks for the contribution

PR #1259, opened 2025-01-08
Make release branches a thing

emilazy:
Finally!

Once this is approved, I will cut a `nix-darwin-24.11` branch and PR the appropriate changes.

Left as future work:

* Building a seperate manual for each release, moving content from `README.md` and `CHANGELOG` to the manual, and linking an index page with the manuals for each version from `README.md`.

* Infrastructure for automated backports.

* Writing a checklist for the release process (or better, automating it).

* Perhaps having a bot automatically update our `flake.lock` so we can just tell people to use nix-darwin’s Nixpkgs input rather than bringing their own (and risking mismatches).

Closes: #727

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,42 +6,42 @@ on:
       - master
 
 env:
-  CURRENT_STABLE_CHANNEL: nixpkgs-24.11-darwin
+  NIXPKGS_BRANCH: nixpkgs-unstable
+  NIX_DARWIN_BRANCH: master
+  NIX_VERSION: 2.24.11
 
 jobs:
+  # The `test-stable` and `install-against-stable` job names are
+  # load‐bearing, despite their inaccuracy on the unstable branch, as
+  # they are set as required checks in the repository configuration,
+  # which only repository admins can change.
+  #
+  # TODO: Change them once the repository configuration is updated.
+
   test-stable:
     runs-on: macos-13
     steps:
     - uses: actions/checkout@v4
-    - name: Install nix corresponding to latest stable channel
+    - name: Install Nix
       uses: cachix/install-nix-action@v30
       with:
-        install_url: https://releases.nixos.org/nix/nix-2.18.8/install
-    - run: nix flake check --override-input nixpkgs nixpkgs/${{ env.CURRENT_STABLE_CHANNEL }}
-
-  test-unstable:
-    runs-on: macos-13
-    steps:
-    - uses: actions/checkout@v4
-    - name: Install nix from current unstable channel
-      uses: cachix/install-nix-action@v30
-      with:
-        install_url: https://releases.nixos.org/nix/nix-2.24.9/install
-    - run: nix flake check --override-input nixpkgs nixpkgs/nixpkgs-unstable
+        install_url: https://releases.nixos.org/nix/nix-${{ env.NIX_VERSION }}/install
+    - run: nix flake check --override-input nixpkgs nixpkgs/${{ env.NIXPKGS_BRANCH }}
 
   install-against-stable:
     runs-on: macos-13
+    timeout-minutes: 30
     steps:
     - uses: actions/checkout@v4
-    - name: Install nix corresponding to latest stable channel
+    - name: Install Nix
       uses: cachix/install-nix-action@v30
       with:
-        install_url: https://releases.nixos.org/nix/nix-2.18.8/install
-        nix_path: nixpkgs=channel:${{ env.CURRENT_STABLE_CHANNEL }}
+        install_url: https://releases.nixos.org/nix/nix-${{ env.NIX_VERSION }}/install
+        nix_path: nixpkgs=channel:${{ env.NIXPKGS_BRANCH }}
     - name: Install channels
       run: |
-        nix-channel --add https://github.com/LnL7/nix-darwin/archive/master.tar.gz darwin
-        nix-channel --add https://nixos.org/channels/${{ env.CURRENT_STABLE_CHANNEL }} nixpkgs
+        nix-channel --add https://github.com/LnL7/nix-darwin/archive/${{ env.NIX_DARWIN_BRANCH }}.tar.gz darwin
+        nix-channel --add https://nixos.org/channels/${{ env.NIXPKGS_BRANCH }} nixpkgs
         nix-channel --update
     - name: Install nix-darwin
       run: |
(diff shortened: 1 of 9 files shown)


modules/system/version.nix
```diff
@@ -59,11 +59,17 @@ in
       description = "Label to be used in the names of generated outputs.";
     };
 
+    system.darwinRelease = mkOption {
+      type = types.str;
+      default = (lib.importJSON ../../version.json).release;
+      description = "The nix-darwin release (e.g. `24.11`).";
+    };
```
Samasaur1, line 66: This should be read-only, right?

modules/examples/flake/flake.nix
```diff
@@ -3,7 +3,7 @@
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
-    nix-darwin.url = "github:LnL7/nix-darwin";
+    nix-darwin.url = "github:LnL7/nix-darwin/master";
```
bew, line 6: Hello,
Why this change? Isn't it already the default 🤔

emilazy: Yeah. The `eval-config.nix` error tells people to ensure their input looks like `github:LnL7/nix-darwin/NIX-DARWIN-BRANCH`, and it felt simpler to just align the examples with that. I don’t love how it looks but I couldn’t think of a way to account for it in the `eval-config.nix` error that wouldn’t confuse people or be even more verbose. But I’m open to alternative suggestions.

My ideal would be to rename our `master` branch to `nix-darwin-unstable` to match `nixpkgs-unstable`, but that requires repo admin privileges so it’s not easy to do immediately.

eval-config.nix
```diff
@@ -1,10 +1,75 @@
+let
+  nixDarwinVersion = builtins.fromJSON (builtins.readFile ./version.json);
+
+  checkRelease = lib:
+    # Avoid breaking configurations when the unstable Nixpkgs version
```
Enzime, line 5: Is this something we need to avoid? I assume we would just want to cut a release whenever NixOS makes a release, although I'm not sure how much effort that will be, probably best to make a guide when you do it for the first time with 24.11

eval-config.nix
```diff
@@ -1,10 +1,75 @@
...
+    else
+      lib.versionAtLeast lib.trivial.release nixDarwinVersion.release;
+in
+
 { lib
 , modules
 , baseModules ? import ./modules/module-list.nix
 , specialArgs ? { }
 , check ? true
+, enableNixpkgsReleaseCheck ? true
```
Enzime, line 22: Do we want to introduce a new argument here? It's probably not a big deal, but I wonder if any users would end up disabling this for whatever reasons, if we want to keep it maybe we could rename it to something like `allowUnsupportedNixpkgsVersions`

eval-config.nix
```diff
@@ -1,10 +1,75 @@
...
 
+assert enableNixpkgsReleaseCheck -> checkRelease lib || throw ''
+
+  nix-darwin now uses release branches that correspond to Nixpkgs
+  releases. You are currently using nix-darwin ${nixDarwinVersion.release} with
+  Nixpkgs ${lib.trivial.release}.
+
+  Check <https://status.nixos.org/> for the currently supported
+  releases of Nixpkgs. nix-darwin’s release support policy follows that
+  of Nixpkgs.
```
Enzime, line 33:
```suggestion
  releases of Nixpkgs. nix-darwin only supports using the same version of Nixpkgs corresponding to the nix-darwin release.
```

eval-config.nix
```diff
@@ -1,10 +1,75 @@
...
+      $ sudo nix-channel --list
+      nixpkgs https://nixos.org/channels/NIXPKGS-BRANCH
+      darwin https://github.com/LnL7/nix-darwin/archive/NIX-DARWIN-BRANCH.tar.gz
+      …
+      $ nix-channel --list
+      …
+
+  If `darwin` or `nixpkgs` are present in `nix-channel --list` (without
+  `sudo`), you should delete them with `nix-channel --remove NAME`.
+  These can contribute to version mismatch problems.
```
Enzime, line 59: I think this should live in its own check

.github/workflows/test.yml
```diff
@@ -6,42 +6,35 @@ on:
...
-    steps:
-    - uses: actions/checkout@v4
-    - name: Install nix from current unstable channel
-      uses: cachix/install-nix-action@v30
-      with:
-        install_url: https://releases.nixos.org/nix/nix-2.24.9/install
-    - run: nix flake check --override-input nixpkgs nixpkgs/nixpkgs-unstable
-
-  install-against-stable:
+  install:
```
Enzime, line 24:
```suggestion
  install-against-stable:
```

Until we can get LnL to change the required checks

.github/workflows/test.yml
```diff
@@ -6,42 +6,35 @@ on:
...
 
 env:
-  CURRENT_STABLE_CHANNEL: nixpkgs-24.11-darwin
+  NIXPKGS_BRANCH: nixpkgs-unstable
+  NIX_DARWIN_BRANCH: master
+  NIX_VERSION: 2.24.11
 
 jobs:
-  test-stable:
+  test:
```
Enzime, line 14:
```suggestion
  test-stable:
```

modules/system/version.nix
emilazy: Yes, `system.nixos.release`, which is read only (so I’ve fixed that). The inconsistent naming here is unfortunate, but I thought it better to stick with the existing convention throughout the file and address all of them at once at some later date. (Though I do admit to feeling a little bad about introducing a new option with a name I don’t love.)

eval-config.nix
emilazy: The problem with unconditionally using `==` is that as soon as 24.11 gets branched off and `master` is updated to 25.05, we’d start throwing an error on our `master` branch. We would have to bump the release in perfect lockstep with Nixpkgs (actually, in perfect lockstep with both `nixpkgs-unstable` and `nixos-unstable`, which is not necessarily possible – one can become 25.05 before the other). That would be more possible with automation, but I’ve punted on the automation for now to get this out the door.

This is a perennial problem with Home Manager, which only makes it a warning – every branch‐off, it starts printing spurious warnings. Just doing a conservative check for now seemed like the best thing; we can bump to 25.11 after both `nixpkgs-unstable` and `nixos-unstable` get it.

eval-config.nix
emilazy: I was attempting to convey “we don’t support or update our release branches corresponding to unmaintained Nixpkgs branches”, but maybe that’s not really important to convey. If you want us to be more explicit about requiring the version match here, maybe I could strengthen the first paragraph of the error message?

eval-config.nix
emilazy: As in, add an additional check to make sure the channels are all in order? I think we need to mention it here because having multiple `nixpkgs` channels that are out of sync with versions is a common source of issues; otherwise people might update one of the channels and still run into the error without ever being able to get to the check that would tell them what’s wrong.

I’m not opposed to having a separate check for this that triggers even if the versions are all in order but it should maybe be a separate PR? This will become a bigger deal as I push out stage 1 of The Plan, since people will be running `darwin-rebuild` as `root` and (I think? not sure) not have access to their regular user’s channels any more, especially since our installation instructions previously recommended adding `darwin` to the user’s channels (while `nixpkgs` gets automatically added to `root`’s channels). But in that case maybe the check should be in `darwin-rebuild` itself.

(The version check itself is deliberately outside of the module system and as early as possible in evaluation, because there can be changes to Nixpkgs that would prevent an entire configuration from evaluating and building in the first place.)

.github/workflows/test.yml
emilazy: Yep, done.

Samasaur1 (approved):

eval-config.nix
Enzime: Yeah that would be good, I think for us we care more about making sure users aren't using mismatched versions, as the support surface for us is lower I don't mind backporting fixes to super old release branches for anyone who really wants to use 21.05 or something

modules/system/version.nix
Enzime: It would be good to just add a comment to document the equivalent option

eval-config.nix
Enzime: Yeah, I was referring to checking that the channels are owned by root not by the current user

I think we should try to avoid additional nix-darwin specific login in `darwin-rebuild`, instead we could put it in the `apply` script as a check just to ensure that `root` has `nixpkgs` channel and that no other users have a `nixpkgs` channel

modules/system/version.nix
emilazy: This is just another entry in the various version options (`system.darwinLabel`, `system.darwinVersion`, `system.darwinVersionSuffix`, etc.) that all mirror various NixOS options. Do you want comments on all of those too? Maybe it should be in a different PR?

eval-config.nix
emilazy: I’ve clarified the wording here and dropped the sentence about support; let me know what you think.

(We definitely shouldn’t cut branches for releases older than 24.11, I don’t think. Backporting down to 21.05 would mean going through seven stable releases along the way, unless we reserve some features only for specific ancient branches. That all sounds like a nightmare and I don’t plan to do any work for unmaintained Nixpkgs branches that become insecure and unsafe to use almost immediately, but I guess as long as I’m not doing the backports or replying to the support requests it doesn’t really matter to me.)

eval-config.nix
emilazy: I don’t think it could live in an `apply` script, because the problem I anticipate is `sudo darwin-rebuild` not being able to find the `darwin` channel to build the configuration at all, thanks to our historical installation instructions. But I’ll have to do some testing. Anyway, seems more in scope for the The Plan PR than this one?

modules/system/version.nix
Enzime: Sure a different PR is fine

Enzime (approved):
LGTM

Enzime (approved):

emilazy:
Here goes nothing.

PR #1260, opened 2025-01-10
Add support for `AppleAccentColor`

developer-velo:
Introduces the option to configure the system-wide accent color (`AppleAccentColor`) in `NSGlobalDomain`.

### DIFF
--- a/modules/system/defaults/NSGlobalDomain.nix
+++ b/modules/system/defaults/NSGlobalDomain.nix
@@ -47,6 +47,25 @@ in {
       '';
     };
 
+    system.defaults.NSGlobalDomain.AppleAccentColor = mkOption {
+      type = types.nullOr (types.enum [ -1 0 1 2 3 4 5 6 ]);
+      default = null;
+      description = ''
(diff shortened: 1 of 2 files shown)


modules/system/defaults/NSGlobalDomain.nix
```diff
@@ -47,6 +47,25 @@ in {
       '';
     };
 
+    system.defaults.NSGlobalDomain.AppleAccentColor = mkOption {
+      type = types.nullOr (types.enum [ -1 0 1 2 3 4 5 6 ]);
```
Enzime, line 51:
```suggestion
      type = types.nullOr (types.enum [ "graphite" "red" ... ]);
      apply = v: ({ graphite = -1; red = 0; ... }).${v};
```

Enzime (changes requested):
Can you also update the test fixtures?

You can run the tests locally with:

```
$ nix build ~/nix-darwin#checks.aarch64-darwin.system-defaults-write
```

PR #1261, opened 2025-01-10
Add tests for `AppleInterfaceStyle`

developer-velo:
Adds tests to ensure proper functionality of the `AppleInterfaceStyle` option.


Enzime (changes requested):
You'll need to update the test fixtures in `tests/fixtures/system-defaults-write`

PR #1262, opened 2025-01-10
Add support for additional window tiling options

belak:
This adds support for the following defaults:

- com.apple.WindowManager.EnableTilingByEdgeDrag
- com.apple.WindowManager.EnableTopTilingByEdgeDrag
- com.apple.WindowManager.EnableTilingOptionAccelerator

They match the top 3 settings in the screenshot under the "Desktop & Dock" page of system settings.

Screenshot 2025-01-10 at 3 57 13 PM

### DIFF
--- a/modules/system/defaults/WindowManager.nix
+++ b/modules/system/defaults/WindowManager.nix
@@ -56,11 +56,36 @@ with lib;
         Hide items in Stage Manager.
       '';
     };
+
+    system.defaults.WindowManager.EnableTilingByEdgeDrag = mkOption {
+      type = types.nullOr types.bool;
+      default = null;
+      description = ''
+        Enable dragging windows to screen edges to tile them. The default is true.
+      '';
+    };
+
+    system.defaults.WindowManager.EnableTopTilingByEdgeDrag = mkOption {
+      type = types.nullOr types.bool;
+      default = null;
+      description = ''
+        Enable dragging windows to the menu bar to fill the screen. The default is true.
+      '';
+    };
+
+    system.defaults.WindowManager.EnableTilingOptionAccelerator = mkOption {
+      type = types.nullOr types.bool;
+      default = null;
+      description = ''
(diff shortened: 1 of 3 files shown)


modules/system/defaults/WindowManager.nix
```diff
@@ -56,11 +56,36 @@ with lib;
...
+        Enable holding alt to tile windows.
+      '';
+    };
+
     system.defaults.WindowManager.EnableTiledWindowMargins = mkOption {
       type = types.nullOr types.bool;
       default = null;
       description = ''
-        Enable Window Margins. The default is true.
+        Enable window margins when tiling windows.
```
Enzime, line 88:
```suggestion
        Enable window margins when tiling windows. The default is true.
```

Enzime (changes requested):
It looks like tests are failing, you can run them locally with:

```
$ nix build ~/nix-darwin#checks.aarch64-darwin.system-defaults-write
```

modules/system/defaults/WindowManager.nix
belak: I can add this back if you'd like, but none of the other options in this file follow the convention of defaulting to null and listing the OS default so I dropped it from the description.

Enzime: Could you add it to all the options in this file?

belak: Sure, you'd like the system defaults added, not "null"?

Enzime: Added to the description but the value is `null`, this is the convention we generally follow

belak: Alright, done! I used `defaults delete` manually to make sure I had the correct values for the defaults in the description.

Enzime (approved):
LGTM

Thanks for contribution

PR #1265, opened 2025-01-13
checks: check for macOS ≥ 11.3

emilazy:
To be merged after https://github.com/LnL7/nix-darwin/pull/1259 to align with Nixpkgs 25.05’s support.

### DIFF
--- a/modules/system/base.nix
+++ b/modules/system/base.nix
@@ -2,60 +2,35 @@
 
 {
   system.activationScripts.createRun.text = ''
-    IFS="." read -r -a macOSVersion <<< "$(sw_vers -productVersion)"
+    if [[ $(stat -c '%a' /etc/synthetic.conf) != "644" ]]; then
+      echo "fixing permissions on /etc/synthetic.conf..."
+      sudo chmod 644 /etc/synthetic.conf
+    fi
 
-    if [[ ''${macOSVersion[0]} -gt 10 || ( ''${macOSVersion[0]} -eq 10 && ''${macOSVersion[1]} -ge 15 ) ]]; then
-      if [[ $(stat -c '%a' /etc/synthetic.conf) != "644" ]]; then
-        echo "fixing permissions on /etc/synthetic.conf..."
-        sudo chmod 644 /etc/synthetic.conf
-      fi
+    if [[ $(grep -c '^run\b' /etc/synthetic.conf) -gt 1 ]]; then
+      echo "found duplicate run entries in /etc/synthetic.conf, removing..."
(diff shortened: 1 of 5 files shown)


modules/system/base.nix
```diff
@@ -2,60 +2,35 @@
...
+    fi
 
-      if [[ ! -L /run ]]; then
-        printf >&2 '[1;31merror: failed to symlink /run, aborting activation[0m\n'
-        printf >&2 'To create a symlink from /run to /var/run, please run:\n'
-        printf >&2 '\n'
-        printf >&2 '$ sudo ln -sfn private/var/link /run\n'
-        exit 1
-      fi
+    sudo /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -t || true
```
Enzime, line 20:
```suggestion
    sudo /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -t
```

The `|| true` was because we were running both `-B` and `-t` indiscriminately

modules/system/base.nix
```diff
@@ -2,60 +2,35 @@
...
+    if [[ ! -L /run ]]; then
+      printf >&2 '[1;31merror: apfs.util failed to symlink /run, aborting activation[0m\n'
+      printf >&2 'To create a symlink from /run to /var/run, please run:\n'
+      printf >&2 '\n'
+      printf >&2 "$ printf 'run\tprivate/var/run\n' | sudo tee -a /etc/synthetic.conf\n"
+      printf >&2 '$ sudo /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -t\n'
+      printf >&2 '\n'
+      printf >&2 'The current contents of /etc/synthetic.conf is:\n'
+      printf >&2 '\n'
+      sudo sed 's/^/    /' /etc/synthetic.conf >&2
```
Enzime, line 31:
```suggestion
      sed 's/^/    /' /etc/synthetic.conf >&2
```

Enzime (changes requested):

modules/system/base.nix
emilazy: Done.

emilazy:
It appears the `|| true` is in fact load‐bearing.

modules/system/base.nix
Enzime: I just had a quick test and from what I can tell it gives exit code 253 on success and 252 on failure, so maybe we could change this to assert that the exit code is 253

emilazy: We check for `[[ ! -L /run ]]` after, which seems like a check we’d probably want to keep anyway just in case something really weird happens, right? So I think the failure case is already covered.

Enzime (approved):
LGTM

PR #1266, opened 2025-01-13
checks: check `/etc/nix/nix.custom.conf` hash

emilazy:
After https://github.com/DeterminateSystems/nix-installer/pull/1379, the Determinate Systems installer will write custom user configuration to `nix.custom.conf` and include it from `nix.conf`. We need to check for its presence and error out if it contains any settings so as to not inadvertently clobber user configuration. The error message here could maybe be improved but it’s better than not handling it at all.

Drafted until that PR is ready to ship and we know what the hash of the `nix.conf` and empty `nix.custom.conf` will be.

This should land before https://github.com/LnL7/nix-darwin/pull/1259 if it’s ready in time, or else get backported.

---

**Note:** This work was funded by Determinate Systems. The design is my own.


doc/known-files/53712b4335030e2dbfb46bb235f8cffcac83fea404bd32dc99417ac89e2dd7c5
```diff
@@ -0,0 +1,14 @@
+
+# Generated by https://github.com/DeterminateSystems/nix-installer.
+# See `/nix/nix-installer --version` for the version details.
+
+!include nix.custom.conf
```
Enzime, line 5: Shouldn't this be at the bottom to be able to override their defaults?

emilazy:
A new version with the relevant PR has been released; I’ve tested the code and fixed it to use `etcChecks` and a separate removal step (curse you, user activation). It handles both the stock file and user customizations correctly.

doc/known-files/53712b4335030e2dbfb46bb235f8cffcac83fea404bd32dc99417ac89e2dd7c5
emilazy: Probably, yeah. This is what’s shipped in the wild in v0.33.0, though, so I think we have to handle this version of the file anyway.

emilazy:
Updated for the new hashes in v0.34.0; confirmed that it still works.

modules/nix/default.nix
```diff
@@ -823,8 +823,58 @@ in
       ]);
     users.knownGroups = mkIf cfg.configureBuildUsers [ "nixbld" ];
 
+    # The Determinate Systems installer puts user‐specified settings in
+    # `/etc/nix/nix.custom.conf` since vX.YY.Z. Supplement the
```
Enzime, line 829: I'm guessing this is meant to have an actual version number

Enzime (dismissed):
LGTM

modules/nix/default.nix
emilazy: Whoops, fixed.

Enzime (approved):

PR #1268, opened 2025-01-14
nix-darwin 24.11

emilazy:
The first release branch, and presumably the only one that will go back in time.


Enzime (approved):

emilazy:
Test failure seems to be because install CI tests against the branch base for some reason.

PR #1272, opened 2025-01-16
defaults-write: fix activation script conditionalization

emilazy:
These are lists, not attribute sets, so the condition was always succeeding. As far as I can tell this bug has been present since the function was added.

### DIFF
--- a/modules/system/defaults-write.nix
+++ b/modules/system/defaults-write.nix
@@ -42,7 +42,7 @@ let
   CustomSystemPreferences = flatten (mapAttrsToList (name: value: defaultsToList name value) cfg.CustomSystemPreferences);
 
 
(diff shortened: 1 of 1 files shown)


Enzime (approved):
LGTM

Enzime:
Out of curiosity, does this make any behavioural difference?

emilazy:
It doesn’t print the message if you don’t print any settings :)

I’m not sure why it’s conditionalized like this in the first place, but I’ve been avoiding touching the logic of the defaults stuff while working on The Plan (since otherwise it’d turn into more major surgery); I just ran into this logic error while adapting things for it, and since you’ll have to set a primary user to set user defaults it seemed cleaner to have the whole phase conditionalized away properly for when that isn’t the case.

Issue #1182, opened 2024-11-19
Add a `colima` module


lloeki:
This would be nice to have:

- for declaratively managing `colima` on macOS
- to support `virtualisation.docker.enable = true`: https://github.com/LnL7/nix-darwin/issues/112

Note that there seems to be a potential issue with sandboxing: https://github.com/abiosoft/colima/issues/490

There's this one upstream which requests a way to have a `LaunchAgent` or `LaunchDaemon` set up by `colima` but that would not be declarative: https://github.com/abiosoft/colima/issues/262

lloeki:
Turns out, this worked fine:

```
  launchd.agents."colima.default" = {
    command = "${pkgs.colima}/bin/colima start --foreground";
    serviceConfig = {
      Label = "com.colima.default";
      RunAtLoad = true;
      KeepAlive = true;

      # not sure where to put these paths and not reference $HOME; `/var/log`?
      StandardOutPath = "/Users/loic.nageleisen/.colima/default/daemon/launchd.stdout.log";
      StandardErrorPath = "/Users/loic.nageleisen/.colima/default/daemon/launchd.stderr.log";

      # not using launchd.agents.<name>.path because colima needs the system ones as well
      EnvironmentVariables = {
        PATH = "${pkgs.colima}/bin:${pkgs.docker}/bin:/usr/bin:/bin:/usr/sbin:/sbin";
      };
    };
  };
```

```
$ darwin-rebuild switch
$ launchctl load -w /Library/LaunchAgents/com.colima.default.plist
$ launchctl print gui/$(id -u)/com.colima.default
$ colima status
```

Note that one may have some sandbox allowance popups the first time you access some directories from within the VM.

edilmedeiros:
Seems that Apple is preparing a native container framework that could be used in the future instead of `colima`: https://developer.apple.com/videos/play/wwdc2025/346/

lloeki:
... which gives you [Apple's container framework via same-named command](https://github.com/apple/container) (which is more like firecracker microVMs), not dockerd/containerd/k8s which is what Colima provides e.g as a drop-in replacement for Docker for Mac and the likes.

PR #1275, opened 2025-01-16
colima: init

bryanhonof:
Adds https://github.com/abiosoft/colima as a system service so that users on macOS can use `docker` and other container tools like they'd do on Linux.
Creates a new user called `colima` to host the service on, and optionally creates a socket under `/var/run/docker.sock` so that the Docker CLI can interact with it from any user who's in the `_colima` group. This is similar to how Docker works on Linux.

This could be a gateway into implementing `virtualisation.docker.enable` in nix-darwin.

Refs: https://github.com/LnL7/nix-darwin/issues/1182

### DIFF
--- a/modules/services/colima/default.nix
+++ b/modules/services/colima/default.nix
@@ -1,0 +1,192 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+with lib;
+let
+  cfg = config.services.colima;
+  user = config.users.users."colima";
+  group = config.users.groups."_colima";
+in
+{
+  options.services.colima = {
+    enable = mkEnableOption "Colima, a macOS container runtime";
+
+    enableDockerCompatability = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Create a symlink from Colima's socket to /var/run/docker.sock, and set
+        its permissions so that users part of the _colima group can use it.
+      '';
+    };
+
+    package = mkPackageOption pkgs "colima" { };
+
+    stateDir = lib.mkOption {
+      type = types.path;
+      default = "/var/lib/colima";
+      description = "State directory of the Colima process.";
+    };
+
+    logFile = mkOption {
+      type = types.path;
+      default = "/var/log/colima.log";
+      description = "Combined stdout and stderr of the colima process. Set to /dev/null to disable.";
+    };
+
+    groupMembers = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      description = ''
+        List of users that should be added to the _colima group.
+        Only has effect with enableDockerCompatability enabled.
+      '';
+    };
+
+    runtime = mkOption {
+      type = types.enum [
+        "docker"
+        "containerd"
+        "incus"
+      ];
+      default = "docker";
+      description = "The runtime to use with Colima.";
+    };
+
+    architectue = mkOption {
+      type = types.enum [
+        "x86_64"
+        "aarch64"
+        "host"
+      ];
+      default = "host";
+      description = "The architecture to use for the Colima virtual machine.";
+    };
+
+    extraFlags = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      example = [ "--vz-rosetta" ];
+      description = "Extra commandline options to pass to the colima start command.";
+    };
+
+    vmType = mkOption {
+      type = types.enum [
+        "qemu"
+        "vz"
+      ];
+      default = "vz";
+      description = "Virtual machine type to use with Colima.";
+    };
+  };
+
+  config = mkMerge [
+    (mkIf cfg.enableDockerCompatability {
+      assertions = [
+        {
+          assertion = !cfg.enable;
+          message = "services.colima.enableDockerCompatability doesn't make sense without enabling services.colima.enable";
+        }
+      ];
+
+      launchd.daemons.colima-docker-compat = {
+        script = ''
+          # Wait for the docker socket to be created. This is important when
+          # we enabled Colima and Docker compatability at the same time, for
+          # the first time. Colima takes a while creating the VM.
+          until [ -S ${cfg.stateDir}/.colima/default/docker.sock ]
+          do
+            sleep 5
+          done
+
+          chmod g+rw ${cfg.stateDir}/.colima/default/docker.sock
+          ln -sf ${cfg.stateDir}/.colima/default/docker.sock /var/run/docker.sock
+        '';
+
+        serviceConfig = {
+          RunAtLoad = true;
+          EnvironmentVariables.PATH = "/usr/bin:/bin:/usr/sbin:/sbin";
+        };
+      };
+
+      users.groups."_colima".members = cfg.groupMembers;
+
+      environment.systemPackages = [
+        pkgs.docker
+      ];
+    })
+
+    (mkIf cfg.enable {
+      launchd.daemons.colima = {
+        script =
+          concatStringsSep " " [
+            "exec"
+            (getExe cfg.package)
+            "start"
+            "--foreground"
+            "--runtime ${cfg.runtime}"
+            "--arch ${cfg.architectue}"
+            "--vm-type ${cfg.vmType}"
+          ]
+          + escapeShellArgs cfg.extraFlags;
+
+        serviceConfig = {
+          KeepAlive = true;
+          RunAtLoad = true;
+          StandardErrorPath = cfg.logFile;
+          StandardOutPath = cfg.logFile;
+          GroupName = group.name;
+          UserName = user.name;
+          WorkingDirectory = cfg.stateDir;
+          EnvironmentVariables = {
+            PATH = "${pkgs.colima}/bin:${pkgs.docker}/bin:/usr/bin:/bin:/usr/sbin:/sbin";
+            COLIMA_HOME = "${cfg.stateDir}/.colima";
+          };
+        };
+      };
+
+      system.activationScripts.preActivation.text = ''
+        touch '${cfg.logFile}'
+        chown ${toString user.uid}:${toString user.gid} '${cfg.logFile}'
+      '';
+
+      users = {
+        knownGroups = [
+          "colima"
+          "_colima"
+        ];
+        knownUsers = [
+          "colima"
+          "_colima"
+        ];
+      };
+
+      users.users."colima" = {
+        uid = config.ids.uids.colima;
+        gid = config.ids.gids._colima;
+        home = cfg.stateDir;
+        # The username isn't allowed to have an underscore in the beginning of
+        # its name, otherwise the VM will fail to start with the following error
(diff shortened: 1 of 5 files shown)


Samasaur1 (changes requested):

jessevdp:
Just by pure coincidence I happen to really need Colima today to get docker running through nix-darwin on my new MacBook. So I just grabbed this module, added it to my config, and enabled it. Worked wonders! Super happy I spotted this PR! 😆

Since I'm an absolute noob I did have a tiny bit of trouble.. I expected it to be as simple as `services.colima.enable` bing, bang, boom -> docker works. So I was a bit surprised that I needed some explicit config.

Here's my `darwin/docker.nix`:

```nix
{
  pkgs,
  ...
}: {
  environment.systemPackages = with pkgs; [
    docker
  ];

  services.colima = {
    enable = true;
    createDockerSocket = true;
    groupMembers = [
      "my-username"
    ];
  };
}
```

All that to say that yes: I do want to use Colima for docker (by putting the socket in place). And yes: my user is explicitly allowed to access this socket.

There's probably a really good reason for this! In the end this (presumably) all has to do with:

> This could be a _**gateway**_ into implementing virtualisation.docker.enable in nix-darwin.

Still, I thought I'd share my experience as a novice user.

I'd love to see this get merged. Happy hacking! 😎

modules/services/colima/default.nix
```diff
@@ -0,0 +1,159 @@
...
+    createDockerSocket = mkEnableOption ''
+      Create a symlink from Colima's socket to /var/run/docker.sock, and set
+      it's permissions so that users part of the _colima group can use it.
+    '';
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.colima;
+      defaultText = literalExpression "pkgs.colima";
+    };
```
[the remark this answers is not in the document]
bryanhonof: I did not know of its existence t.b.h. :,)

modules/services/colima/default.nix
```diff
@@ -0,0 +1,159 @@
...
+          sleep 5
+        done
+
+        chmod g+rw ${user.home}/.colima/default/docker.sock
+        ln -sf ${user.home}/.colima/default/docker.sock /var/run/docker.sock
+      '';
+
+      serviceConfig.RunAtLoad = cfg.createDockerSocket;
+      serviceConfig.EnvironmentVariables.PATH = "/usr/bin:/bin:/usr/sbin:/sbin";
+    };
```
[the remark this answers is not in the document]
bryanhonof: I did it this way, so a user could still manually run `sudo launchctl kickstart -kp org.nixos.colima-create-docker-socket-and-set-permissions`. But maybe that's not necessary.

modules/services/colima/default.nix
```diff
@@ -0,0 +1,159 @@
...
+      type = types.package;
+      default = pkgs.colima;
+      defaultText = literalExpression "pkgs.colima";
+    };
+
+    logFile = mkOption {
+      type = types.path;
+      default = "/var/log/colima.log";
+      description = "Stdout and sterr of the colima process.";
+    };
```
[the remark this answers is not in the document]
bryanhonof: I believe I copied this part from another service in here. I wouldn't mind having separate stdout and stderr. I think a user could "disable" the logs by pointing it at `/dev/null`?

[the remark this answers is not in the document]
Samasaur1: okay, then I'd just split this into two options, `outLogFile` and `errorLogFile`, that default to `/var/log/colima.out.log` and `/var/log/colima.err.log`, and then instead update `serviceConfig` to match.

modules/services/colima/default.nix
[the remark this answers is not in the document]
Samasaur1: If that sounds like something that might be useful, then I'm happy leaving it this way

modules/services/colima/default.nix
```diff
@@ -0,0 +1,155 @@
...
+    users.users."colima" = {
+      uid = mkDefault 400;
+      gid = mkDefault group.gid;
+      home = mkDefault "/var/lib/colima";
+      # The username isn't allowed to have an underscore in its name, the VM
+      # will fail to start with the following error otherwise
+      #   > "[hostagent] identifier \"_colima\" must match ^[A-Za-z0-9]+(?:[._-](?:[A-Za-z0-9]+))*$: invalid argument" fields.level=fatal
+      name = "colima";
+      createHome = true;
+      shell = "/bin/bash";
```
bestlem, line 139: In macOS don't set the shell to /bin/bash as that is a very old shell.
Set it to /bin/sh or if you need bash use one from nixpkgs .  /bin/zsh would also work but not fully under nix's control.

Samasaur1: I would say definitely don't set it to nixpkgs bash. I suppose `/bin/sh` is slightly better than `/bin/bash` but I don't think it really matters

modules/services/colima/default.nix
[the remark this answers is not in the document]
bryanhonof: Hmm, looking at other services in this repo, it seems like most combine stderr and stdout into a single file. So, I'll adhere to the convention for consistency's sake.

modules/services/colima/default.nix
bryanhonof: @bestlem 

> Is there a way of letting macOS generate the UID if you are using ids above 500 so as not to conflict with non nix systems although 53x would be usial to be actually found so might only be a theoretical issue.?

Not that I know of, but I believe we're doing it this way due to how [the users module](https://github.com/LnL7/nix-darwin/blob/master/modules/users/default.nix) works. I remember looking into it a while ago, but it was a bit too complex for me back then. So my intuition might be wrong here.

> In macOS I would of thought that the shell should not be set to /bin/bash as that is a very old shell.

Yes, it is. However, this user is a so-called "system user", and not one the user should be logging into regularly. Having copied this section from other modules in this repo, I'd prefer to keep it set to `/bin/bash` to stay consistent.

bryanhonof:
@jessevdp thanks for the kind words.

> All that to say that yes: I do want to use Colima for docker (by putting the socket in place). And yes: my user is explicitly allowed to access this socket.

I made it also include the Docker CLI when you enable the compatibility now. Adding the user, I believe should still be a separate step/config. I could have it throw a warning though, but that might pollute the output.

bryanhonof:
Wrong button :)

modules/services/colima/default.nix
bestlem: I would say that being consistent is not necessarily correct.  We are already inconsistent in the users there bing a module to contain all the uids and then modules that do not use it.

I think the other modules using /bin/bash and also the single logs are wrong and they are the ones to change.

bryanhonof:
Actually, I think I just ran into an issue whilst trying to mount stuff with `docker run --volume ...`. Perhaps I _have_ to run the VM as root instead of a seperate user...

bryanhonof:
Haha, 

```
"must not run as the root"
```

bryanhonof:
Adding the `colima` user to the `staff` group seems to help. It allows me to use the `--mount` flag in `colima start` to mount other directories outside of the colima user.

jessevdp:
I think I just ran into this bind-mount issue too!

Keep in mind that I'm still using the setup duplicated from when this PR was initially openend.. I'm not sure much has changed in that regard.

But... I tried to run a VSCode devcontainer setup that attempts to use a bind mount. The folder I mounted is empty inside the container, and when I attempt `touch foo.txt` when attached to the container I get "Permission denied".

Any clue how to make this seamless? Can we configure that staff group membership from this module or is that a bad idea?

bestlem:
Isn't staff a bad idea - as that is the group for actual users.

Why does this have to be setup as a central user - colima works in home-manager that is as per the current user.

jessevdp:
I’m also curious about that! As somewhat of a noob I just accepted that part of this module as being “a best practice”. But I’m interested to learn why it would or wouldn’t be a good idea?

I guess because this is a darwin module you’d want the service to be shared and usable by multiple users?

jessevdp:
@bryanhonof why did you choose `uid` and `gid` 400? They seem to be taken already on my (relatively new) Mac?

```
dscacheutil -q group

-- snip --

name: com.apple.access_remote_ae
password: *
gid: 400

-- snip --
```

modules/services/colima/default.nix
```diff
@@ -0,0 +1,192 @@
...
+      type = types.enum [
+        "docker"
+        "containerd"
+        "incus"
+      ];
+      default = "docker";
+      description = "The runtime to use with Colima.";
+    };
+
+    architectue = mkOption {
```
niklaskorz, line 60:
```suggestion
    architecture = mkOption {
```

modules/services/colima/default.nix
```diff
@@ -0,0 +1,192 @@
...
+let
+  cfg = config.services.colima;
+  user = config.users.users."colima";
+  group = config.users.groups."_colima";
+in
+{
+  options.services.colima = {
+    enable = mkEnableOption "Colima, a macOS container runtime";
+
+    enableDockerCompatability = mkOption {
```
niklaskorz, line 18:
```suggestion
    enableDockerCompatibility = mkOption {
```

niklaskorz: And of course all other occurences of "compatability" -> "compatibility"

modules/services/colima/default.nix
niklaskorz: And all other occurences of "architectue" -> "architecture"

wra-bradshaw:
I imagine things are potentially stalled here because of Apple's release of the containerisation framework? 
In my research, colima and tools like it are going to remain relevant for a while, since the framework does not support the docker engine API (or any standard container API).
[Attempts to bridge the Apple containerisation framework and docker api have begun](https://github.com/socktainer/socktainer) but judging by [this issue](https://github.com/socktainer/socktainer/issues/14), it is going to be quite a while before we have anything resembling feature parity. 

For this reason, I think colima in nix-darwin is still a noble goal and I'm willing to pick up this PR. What do maintainers think? Worth pursuing?

bryanhonof:
Hey @will-lol feel free to pick up this PR, I ended up persueing a setup based on podman, since this was taking a bit too much time to my liking. The last thing I remember running into was Colima not allowing certain mount points to exist, or at least being very picky about which ones you can write to freely.

If you end up creating a new PR, please reference this one so I can close it. And if you end up having question, feel free to ping me as well. :)

wra-bradshaw:
Hey @bryanhonof ! I see this issue. I had a bit of success running colima as a launchd agent, but at that point I thought it would be better off in home manager.
[This](https://github.com/will-lol/.dotfiles/blob/main/home/extensions/colima.nix) is the home manager module I'm running at the moment, which is working quite well! I don't have a linux machine on deck to try it with, so I'll have to fire up a VM to test before filing a PR on the hm side.

wra-bradshaw:
https://github.com/nix-community/home-manager/pull/7913

This is the PR that merged recently in home manager. You can now do `services.colima.enable = true` :)

PR #1277, opened 2025-01-17
{environment,nix-tools}: correct default `$PATH` ordering to match macOS

emilazy:


Samasaur1 (commented):
confirmed that this is the `PATH` ordering that macOS uses on macOS 12, 14, and 15 (contents of `/etc/paths`).

I do see that at least on macOS 14 and above, there's also a `/System/Cryptexes/App/usr/bin` directory in `PATH`. Do we want to include this directory in the default `PATH` as well?
```
/usr/local/bin
/System/Cryptexes/App/usr/bin
/usr/bin
/bin
/usr/sbin
/sbin
```
(contents of `/etc/paths` on macOS 14.7)

emilazy:
Yeah, plausibly. I think if they put out a rapid security response update that fixed a vulnerability in `/usr/bin` or something the replacement would go there? I’m not sure. There’s also more stuff in `/etc/paths.d/10-cryptex` and `/etc/paths.d/100-rvictl` that gets added by default. Adding those seems like a reasonable idea, just not one I’m as certain about as this change. (But I’d probably approve a PR to do so.)

Enzime (approved):
LGTM

Did you run into an issue with the ordering?

emilazy:
Nah, I was just checking it against a stock installation while doing other work on these files and noticed the discrepancy.

PR #1279, opened 2025-01-18
darwin-uninstaller: remove `darwin` channel from `root` too

emilazy:
Handle this now that we recommend people use `sudo nix-channel --add`. I’m not sure `~/.nix-defexpr` is always used (e.g. with XDG directories?), so we just remove it unconditionally in both user and system activation and hope the warning isn’t too bothersome, but maybe we could do something smarter here.

@Enzime I don’t suppose you have any idea what the `[[ "$(shasum -a 256 /Library/LaunchDaemons/org.nixos.nix-daemon.plist | awk '{print $1}')" == "$(shasum -a 256 /Library/LaunchDaemons/org.nixos.nix-daemon.plist | awk '{print $1}')" ]]` thing from 3a89b614321ab8dad3962d79fc3a29bace9a8486 is about?

### DIFF
--- a/pkgs/darwin-uninstaller/configuration.nix
+++ b/pkgs/darwin-uninstaller/configuration.nix
@@ -16,12 +16,12 @@ with lib;
   nix.useDaemon = mkForce false;
 
   system.activationScripts.postUserActivation.text = mkAfter ''
-    if [[ -L ~/.nix-defexpr/channels/darwin ]]; then
-        nix-channel --remove darwin || true
-    fi
+    nix-channel --remove darwin || true
   '';
 
   system.activationScripts.postActivation.text = mkAfter ''
+    nix-channel --remove darwin || true
+
     if [[ -L /Applications/Nix\ Apps ]]; then
         rm /Applications/Nix\ Apps
     fi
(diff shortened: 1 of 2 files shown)


Samasaur1 (approved):

pkgs/darwin-uninstaller/configuration.nix
```diff
@@ -16,12 +16,12 @@ with lib;
   nix.useDaemon = mkForce false;
 
   system.activationScripts.postUserActivation.text = mkAfter ''
-    if [[ -L ~/.nix-defexpr/channels/darwin ]]; then
-        nix-channel --remove darwin || true
-    fi
+    nix-channel --remove darwin || true
```
Enzime, line 19:
```suggestion
    nix-channel --remove darwin
```

Surely not this one as well…

Enzime (approved):

Enzime:
If I remember correctly that check was meant to be is the installed version of nix-daemon the same as the one in your current system closure but I’m not entirely sure

emilazy:
Makes sense. I guess it can wait until another time to figure out what it should be doing :)

Though shouldn’t it be testing the opposite, i.e. the Nix from the system closure is no longer being used…?

Enzime:
Or possibly it was to check that it matches the one from your system profile (`/nix/var/nix/profiles/…`) which is the one that gets restored when you uninstall I’m not sure

PR #1280, opened 2025-01-18
{environment,readme}: default configuration path to `/etc/nix-darwin`

emilazy:
Split off from The Plan stack, where it becomes more important, but a good idea anyway given our general direction and to align the default between flake‐ and channel‐based setups. This is not a breaking change thanks to the `system.stateVersion` conditional (unless anyone is looking at `environment.darwinConfig` on a flake‐based setup, which I guess is possible although seems inadvisable), so we can do it now.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,13 +43,11 @@ jobs:
         nix-channel --update
     - name: Install nix-darwin
       run: |
-        export NIX_PATH=$HOME/.nix-defexpr/channels
-
-        mkdir -p ~/.config/nix-darwin
-        cp modules/examples/simple.nix ~/.config/nix-darwin/configuration.nix
+        sudo mkdir -p /etc/nix-darwin
+        sudo cp modules/examples/simple.nix /etc/nix-darwin/configuration.nix
 
         nixConfHash=$(shasum -a 256 /etc/nix/nix.conf | cut -d ' ' -f 1)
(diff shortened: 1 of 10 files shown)


emilazy:
This is running into the thing where we test installation against the branch base. Not sure if we should fix that or what. (Testing upgrades seems like a good thing, but in this case the way it’s done is kind of annoying.)

Samasaur1 (commented):
I definitely agree with the idea of the PR.

I don't feel completely confident in my understanding of the interaction of all these moving parts, though; what I read through seems good but I'm wondering if people will run into issues due to this change given the note on the `environment.darwinConfig` option:

> NOTE: Changing this requires running `darwin-rebuild switch -I darwin-config=/path/to/configuration.nix` the first time to make darwin-rebuild aware of the custom location.

Given the `stateVersion` condition, I imagine nobody will run into issues unless they update their `stateVersion`, in which case perhaps they're just on their own?

emilazy:
Yeah if you bump `system.stateVersion` without reading the changelog it's expected that stuff will explode.

emilazy:
Realized that we were telling people to install 24.11 with the unstable `darwin-rebuild`. Whoops! Fixed.

Enzime (dismissed):
LGTM

Thanks for making this change, been on my list for so long 👍

Enzime (approved):

PR #1281, opened 2025-01-18
darwin-rebuild: pass `${extraBuildFlags[@]}` to `nix-instantiate`

emilazy:
Theoretically required for `darwin-rebuild edit -I darwin-config=…`, I guess. We also shouldn’t run it for flake‐based setups.


Samasaur1 (approved):
I'm not sure when this would be necessary (mostly because I have never used `darwin-rebuild edit` but given that you can pass any Nix options you want to the script, we might as well pass them along

Enzime (approved):

PR #1282, opened 2025-01-18
nix-tools: drop `nixPackage`, overwrite `$PATH` rather than prepending, set `$NIX_PATH`

emilazy:
I wonder if we actually want nix-darwin specific things like `/etc/profiles/per-user/$USER/bin` in the default path here. Probably it doesn’t hurt to, especially for the uninstaller, but I guess it could cause issues with reinstalling after a weird half‐uninstalled nix-darwin setup or something. It’ll refer to `root`’s packages soon enough anyway…

### DIFF
--- a/pkgs/nix-tools/default.nix
+++ b/pkgs/nix-tools/default.nix
@@ -5,7 +5,22 @@
 , substituteAll
 , stdenv
 , profile ? "/nix/var/nix/profiles/system"
-, systemPath ? "$HOME/.nix-profile/bin:/etc/profiles/per-user/$USER/bin:/run/current-system/sw/bin:/nix/var/nix/profiles/default/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
+, # This should be kept in sync with the default
+  # `environment.systemPath`. We err on side of including conditional
+  # things like the profile directories, since they’re more likely to
+  # help than hurt, and this default is mostly used for fresh
+  # installations anyway.
+  systemPath ? lib.concatStringsSep ":" [
+  "$HOME/.nix-profile/bin"
+  "/etc/profiles/per-user/$USER/bin"
+  "/run/current-system/sw/bin"
+  "/nix/var/nix/profiles/default/bin"
+  "/usr/local/bin"
+  "/usr/bin"
+  "/bin"
+  "/usr/sbin"
+  "/sbin"
+]
 }:
 
 let
--- a/pkgs/nix-tools/default.nix
+++ b/pkgs/nix-tools/default.nix
@@ -5,12 +5,11 @@
 , substituteAll
 , stdenv
 , profile ? "/nix/var/nix/profiles/system"
-, nixPackage ? "/nix/var/nix/profiles/default"
 , systemPath ? "$HOME/.nix-profile/bin:/etc/profiles/per-user/$USER/bin:/run/current-system/sw/bin:/nix/var/nix/profiles/default/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
 }:
 
 let
-  extraPath = lib.makeBinPath [ nixPackage coreutils jq git ];
+  extraPath = lib.makeBinPath [ coreutils jq git ];
 
   writeProgram = name: env: src:
     substituteAll ({
--- a/pkgs/nix-tools/default.nix
+++ b/pkgs/nix-tools/default.nix
@@ -21,6 +21,11 @@
   "/usr/sbin"
   "/sbin"
 ]
+, # This should be kept in sync with the default `nix.nixPath`.
+  nixPath ? lib.concatStringsSep ":" [
+  "darwin-config=/etc/nix-darwin/configuration.nix"
+  "/nix/var/nix/profiles/per-user/root/channels"
+]
 }:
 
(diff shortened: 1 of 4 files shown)


emilazy:
Realized a bit more of the Plan stack could safely go into this.

Samasaur1 (approved):
I'm not sure I totally grasp the wider implications of these changes but the specific changes look good to me

README.md
```diff
@@ -175,16 +175,16 @@ The documentation is also available as manpages by running `man 5 configuration.
 
 ## Uninstalling
 
-To run the latest version of the uninstaller, you can run the following command:
+To use the uninstaller, you can run the following command:
```
Enzime, line 178: I’m not sure how I feel about swapping the recommendation preference as the uninstaller is still in a half finished state so the idea behind recommending the uninstaller from latest `master` is for when we do improve the uninstaller that they can get those benefits even if they’re running an old nix-darwin version

pkgs/nix-tools/default.nix
```diff
@@ -5,12 +5,11 @@
 , substituteAll
 , stdenv
 , profile ? "/nix/var/nix/profiles/system"
-, nixPackage ? "/nix/var/nix/profiles/default"
 , systemPath ? "$HOME/.nix-profile/bin:/etc/profiles/per-user/$USER/bin:/run/current-system/sw/bin:/nix/var/nix/profiles/default/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
 }:
 
 let
-  extraPath = lib.makeBinPath [ nixPackage coreutils jq git ];
+  extraPath = lib.makeBinPath [ coreutils jq git ];
```
Enzime, line 32: From my understanding, the reason we did it this way is so to we evaluate the system using its Nix

I believe this was to match how `nixos-rebuild` works but I’m not 100% sure

emilazy: That would be `/nix/var/nix/profiles/system`, not `/nix/var/nix/profiles/default`; the latter is usually going to be whatever Nix the user initially installed rather than anything we manage. There are two cases here:

1. The user is running `darwin-rebuild` directly from the flake; `nixPackage` is the default, so we add `/nix/var/nix/profiles/default/bin` to the front of the path. But `/nix/var/nix/profiles/default/bin` is already in the path; we will only get Nix from elsewhere if one is present in `$HOME/.nix-profile/bin`, `/etc/profiles/per-user/$USER/bin`, or `/run/current-system/sw/bin`. Usually a user doing this will be performing a fresh install and none of them will exist, so the behaviour will be the same before. But if the user has configured a custom Nix with nix-darwin, then it’ll be in `/run/current-system/sw/bin` and it doesn’t seem right for `/nix/var/nix/profiles/default/bin` to take precedence. In any case the default `$PATH` they get after installation will match this value, so if they have a Nix in one of those other paths then it’ll start getting used as soon as `darwin-rebuild` finishes running. I think the behaviour is less confusing after the change.

2. The user is running `darwin-rebuild` from their build system. `systemPath` will be derived from `environment.systemPath`, and `nixPackage` would have been derived from `nix.package`. However, the Nix module already adds `nix.package` to `environment.systemPackages`, so that package will already be in `/run/current-system/sw/bin`. `darwin-rebuild` will use whatever `nix` the user normally gets at the command line, as expected.

I think that in the vast majority of cases this change will make no difference, and in the cases where it does change behaviour I think the current behaviour is more confusing.

By matching `nixos-rebuild`, I guess you mean the default behaviour without `--fast` where it first builds the Nix from the configuration to be built and then uses that to build the rest of the system and activate it? That’s distinct from using the Nix from the system to be built rather than the currently active system, and I believe was added back when Nix evaluation semantics changes / Nixpkgs requirements for Nix versions advanced somewhat quicker than they do now, and was possibly kept as somewhat of a hack in the era of `nixFlakes`. We’d need additional code for that, and I’m not convinced it’s worth it. IIRC `nixos-rebuild-ng` doesn’t bother implementing that code path at all.

README.md
emilazy: Fair enough. The downside is that it won’t use their configured `environment.systemPath`, but I guess ideally the uninstaller should be hermetic enough that that shouldn’t matter? I can drop this change if you’d prefer.

pkgs/nix-tools/default.nix
Enzime: I had a look at https://github.com/LnL7/nix-darwin/commit/5288a723540fdb25e34364518115e2ef32ed19ad#diff-9245d764b3bb791a65d179021689da4227c0989b700778261c7e3fb56c4524f7 which is my original commit that introduced this behaviour as I switched `darwin-rebuild` from always being built from a nix-darwin system to being able to be built independently

I think I was trying to find a good default value to replace `config.nix.package` not realizing that it wasn't necessary as `/nix/var/nix/profiles/default/bin` was already in `PATH` and would lead to `/run/current-system/sw/bin` would be shadowed

This change makes sense to me now 👍

I'm fine with not implementing the 'slow' behaviour for now to match with `nixos-rebuild-ng`, I'm curious if they'll ever end up implementing it again

Enzime (approved):

Enzime (approved):
LGTM

PR #1283, opened 2025-01-18
checks: remove `runLink`

emilazy:
Now that we create `/run` as part of activation, this is no longer useful for anaything.

### DIFF
--- a/modules/system/checks.nix
+++ b/modules/system/checks.nix
@@ -47,14 +47,6 @@ let
     fi
   '';
 
-  runLink = ''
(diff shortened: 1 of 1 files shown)


Samasaur1 (approved):
This looks fine to me.

I noticed that the part that creates `/run` at activation is run under user activation instead of system activation (due to ordering). Will the Plan cover making this be part of system activation?

emilazy:
Yeah. It all moves over. Checks in general are a user activation thing that now.

Enzime (approved):

PR #1285, opened 2025-01-20
modules/nebula: add nebula overlay network

siriobalmelli:
Added on a similar pattern as netbird; this is in production on our systems locally.

### DIFF
--- a/modules/services/nebula.nix
+++ b/modules/services/nebula.nix
@@ -1,0 +1,259 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  inherit (lib)
+    filterAttrs
+    literalExpression
+    mapAttrsToList
+    mkEnableOption
+    mkIf
+    mkMerge
+    mkOption
+    mkPackageOption
+    optionalAttrs
+    recursiveUpdate
+    warnIf
+    types
+    ;
+
+  cfg = config.services.nebula;
+  enabledNetworks = filterAttrs (n: v: v.enable) cfg.networks;
+
+  format = pkgs.formats.yaml { };
+
+  resolveFinalPort =
+    netCfg:
+    if netCfg.listen.port == null then
+      if (netCfg.isLighthouse || netCfg.isRelay) then 4242 else 0
+    else
+      netCfg.listen.port;
+
+in
+{
+  # Interface
+
+  options.services.nebula = {
+    package = mkPackageOption pkgs "nebula" { };
+    networks = mkOption {
+      description = "Nebula network definitions.";
+      default = { };
+      type = types.attrsOf (
+        types.submodule {
+          options = {
+            enable = mkOption {
+              type = types.bool;
+              default = true;
+              description = "Enable or disable this network.";
+            };
+
+            ca = mkOption {
+              type = types.path;
+              description = "Path to the certificate authority certificate.";
+              example = "/etc/nebula/ca.crt";
+            };
+
+            cert = mkOption {
+              type = types.path;
+              description = "Path to the host certificate.";
+              example = "/etc/nebula/host.crt";
+            };
+
+            key = mkOption {
+              type = types.oneOf [
+                types.nonEmptyStr
+                types.path
+              ];
+              description = "Path or reference to the host key.";
+              example = "/etc/nebula/host.key";
+            };
+
+            staticHostMap = mkOption {
+              type = types.attrsOf (types.listOf types.str);
+              default = { };
+              description = ''
+                The static host map defines a set of hosts with fixed IP addresses on the internet (or any network).
+                A host can have multiple fixed IP addresses defined here, and nebula will try each when establishing a tunnel.
+              '';
+              example = {
+                "192.168.100.1" = [ "100.64.22.11:4242" ];
+              };
+            };
+
+            isLighthouse = mkEnableOption "lighthouse functionality for this node";
+
+            isRelay = mkEnableOption "relay functionality for this node";
+
+            lighthouses = mkOption {
+              type = types.listOf types.str;
+              default = [ ];
+              description = ''
+                List of IPs of lighthouse hosts this node should report to and query from. This should be empty on lighthouse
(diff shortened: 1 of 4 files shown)


Samasaur1 (approved):

modules/services/nebula.nix
```diff
@@ -0,0 +1,265 @@
...
+
+  options.services.nebula = {
+    package = mkPackageOption pkgs "nebula" { };
+    networks = mkOption {
+      description = "Nebula network definitions.";
+      default = { };
+      type = types.attrsOf (
+        types.submodule {
+          options = {
+            enable = mkOption {
```
Enzime, line 47:
```suggestion
            enable = mkEnableOption {
```

modules/services/nebula.nix
```diff
@@ -0,0 +1,265 @@
...
+              description = ''
+                When tun is disabled, a lighthouse can be started without a local tun interface (and therefore without root).
+              '';
+            };
+
+            tun.device = mkOption {
+              type = types.nullOr types.str;
+              default = null;
+              description = "Name of the tun device. Defaults to nebula.\${networkName}.";
+            };
```
Enzime, line 148:
```suggestion
            tun.device = mkOption {
              type = types.str;
              default = "nebula.\${networkName}";
              description = "Name of the TUN device.";
            };
```

You'll probably also need to set `defaultText`

modules/services/nebula.nix
```diff
@@ -0,0 +1,265 @@
...
+                You will likely experience connectivity issues: https://nebula.defined.net/docs/config/listen/#listenport
+              ''
+              settings
+          );
+        in
+        {
+          "nebula-${netName}" = {
+            script = ''
+              exec ${cfg.package}/bin/nebula -config ${configFile}
+            '';
```
Enzime, line 251:
```suggestion
            command = "${cfg.package}/bin/nebula -config ${configFile}";
```

Enzime (changes requested):

modules/services/nebula.nix
Enzime, line 47:
```suggestion
            enable = mkEnableOption {
```

Can you replace all the booleans that default to `false` and just enable logic with `mkEnableOption`?

Enzime (changes requested):

modules/services/nebula.nix
siriobalmelli: I'm not sure this is a good idea.

It appears that on Darwin the tunneling interface currently must be `utun[0-9]+`.

The parameter is not required, so I don't see an advantage to setting a default.

See https://nebula.defined.net/docs/config/tun/#tundev

siriobalmelli:
Apologies for the delay, I was on leave.

Thanks for the review and suggestions, added @Enzime as co-author

modules/services/nebula.nix
siriobalmelli: Implementing this change yields errors like:

```
error: cannot coerce a set to a string: { type = «thunk»; description = "Whether this node is a lighthouse."; }
```

I've reverted this for now, as it seems to be a stylistic rather than operational point - and I am using this in prod.

If there are in fact compelling reasons for this change, which I'm not tracking with, please help point me in the right direction 🙏

modules/services/nebula.nix
siriobalmelli: done

siriobalmelli:
Bump. @Enzime I reworked based on your points above already on 6 March. Any chance you could examine again please?

siriobalmelli:
@Enzime not sure if I'm doing something wrong here? I addressed each of your points, but left them for you to look at and mark resolved. Are you expecting me to mark them resolved? Are you expecting me to do something else? Could I get some feedback please?

siriobalmelli:
Bumping. Merging is blocked, but to my understanding I've done everything right and handled all queries. Not getting any feedback from @Enzime ... @LnL7 may I please have your feedback?

Super happy to go back and fix anything I missed, but I'm sitting in a mystery as to why this is hanging fire here.

Do you prefer I rebase? Do you prefer I leave as-is? Help pls 🙏

modules/services/nebula.nix
siriobalmelli: @Enzime any feedback would be appreciated pls 🙏

modules/services/nebula.nix
siriobalmelli: @Enzime any feedback appreciated. 🙏

siriobalmelli:
Rebased onto master and addressed all review feedback:

- `isLighthouse` and `isRelay` now use `mkEnableOption`
- `tun.disable` renamed to `tun.disabled` (matches nebula YAML key; `mkEnableOption` avoided due to inverted semantics)
- `tun.device` fallback removed — `dev` key is omitted when null, letting nebula choose the platform-appropriate default (fixes the Darwin `utun` constraint I mentioned earlier)
- Removed dead code (`nameToId`/`networkId`)
- Added `tests/services-nebula.nix` and `release.nix` entry — test covers both basic and lighthouse configs, validates generated YAML (port defaults, tun.dev absence)

@Enzime @Samasaur1 — would you be able to re-review? All changes address the original review comments.

kylosus:
Thanks for the PR. It would be very useful to have this merged :)

I tested the module on my machine and it works as expected. I'd suggest renaming `tun.disabled` back to `tun.disable` to keep parity with the NixOS module but besides that the configs seem interchangeable.

siriobalmelli:
> Thanks for the PR. It would be very useful to have this merged :)
> 
> I tested the module on my machine and it works as expected. I'd suggest renaming `tun.disabled` back to `tun.disable` to keep parity with the NixOS module but besides that the configs seem interchangeable.

Thank you, I appreciate it.

The nebula option itself however is `tun.disabled` and in this instance I think compatibility with the options naming of the package itself should take precedence: https://nebula.defined.net/docs/config/tun/#tundisabled

siriobalmelli:
This has been stranded 19 months since @Enzime requested changes and has declined to answer my follow-ups.

Dropping.

PR #1287, opened 2025-01-20
Add new option for double clicking the window title bar

Amwam:
Couldn't find this option already. 
Found the option using `defaults read` and exporting the results to a file, then comparing the different values after changing the setting
This was done on macOS Sequoia 15.2

Screenshot 2025-01-20 at 21 04 40

### DIFF
--- a/modules/system/defaults/NSGlobalDomain.nix
+++ b/modules/system/defaults/NSGlobalDomain.nix
@@ -7,6 +7,21 @@ let
   inherit (config.lib.defaults.types) floatWithDeprecationError;
 in {
   options = {
+    system.defaults.NSGlobalDomain.AppleActionOnDoubleClick = mkOption {
+      type = types.nullOr (
+        types.enum [
+          "Minimize"
+          "Zoom"
+          "Fill"
+          "None"
+        ]
+      );
+      default = null;
+      description = ''
+        Behaviour of double clicking a windows title bar.
+      '';
+    };
+
     system.defaults.NSGlobalDomain.AppleShowAllFiles = mkOption {
       type = types.nullOr types.bool;
       default = null;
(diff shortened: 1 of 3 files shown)


modules/system/defaults/NSGlobalDomain.nix
```diff
@@ -7,6 +7,21 @@ let
   inherit (config.lib.defaults.types) floatWithDeprecationError;
 in {
   options = {
+    system.defaults.NSGlobalDomain.AppleActionOnDoubleClick = mkOption {
+      type = types.nullOr (
+        types.enum [
+          "Minimize"
+          "Zoom"
```
Samasaur1: For me (macOS 15.5), even though System Settings shows "Zoom", the actual value used by defaults is `Maximize`
```suggestion
          "Maximize"
```

modules/system/defaults/NSGlobalDomain.nix
```diff
@@ -7,6 +7,21 @@ let
...
+          "Fill"
+          "None"
+        ]
+      );
+      default = null;
+      description = ''
+        Behaviour of double clicking a windows title bar.
```
Samasaur1: Please add a note in here that the System Settings option "Zoom" corresponds to the defaults value "Maximize".

also, nit: "windows" should be "window's"

Samasaur1 (changes requested):
Please also rebase on top of the latest master

Samasaur1:
If you'd prefer to match the options used in System Settings, you can instead mirror what we do with some Finder defaults:

https://github.com/nix-darwin/nix-darwin/blob/19346808c445f23b08652971be198b9df6c33edc/modules/system/defaults/finder.nix#L151-L177

modules/system/defaults/NSGlobalDomain.nix
Enzime: I think it would be good to use `apply` here to map from `Zoom` to `Maximise`

Amwam:
Thanks for the review. I've made the changes, however I'm unable to test them myself. I'm still pretty new to Nix and so I've been struggling through understanding how to set everything up.

For now I'm stepping away from Nix/nix-darwin, so won't continue to maintain this PR. If someone wants to take over, feel free and I'll close this one.

Thanks!

PR #1289, opened 2025-01-21
[24.11] Add support for additional window tiling options

belak:
This adds support for the following defaults:

- com.apple.WindowManager.EnableTilingByEdgeDrag
- com.apple.WindowManager.EnableTopTilingByEdgeDrag
- com.apple.WindowManager.EnableTilingOptionAccelerator

(cherry picked from commit 4075a3c23aa7996acc960e61df9f21038136d08a)

### DIFF
--- a/modules/system/defaults/WindowManager.nix
+++ b/modules/system/defaults/WindowManager.nix
@@ -56,11 +56,36 @@ with lib;
         Hide items in Stage Manager.
       '';
     };
+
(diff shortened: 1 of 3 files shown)


emilazy (approved):
Thanks!

FWIW if you can think of a way to clarify the version mismatch error to avoid the branch name confusion you had I’m all ears; it’s hard to write messages that are useful and simple to correct for end‐users without saying anything misleading 😅

PR #1296, opened 2025-01-25
Add known hash for determinate-nixd 0.3.0

onnimonni:
Hey,

I bumped into the same issue as in #1044 but for the `determinate-nixd` instead of the DetSys installer.

How to repeat:

1. Install Determinate nixd from: https://docs.determinate.systems/getting-started/
2. Try to install a nix-darwin from a flake:
```
$ nix run --extra-experimental-features "nix-command flakes" nix-darwin -- switch --flake .
warning: 'https://cache.flakehub.com' does not appear to be a binary cache
error: Unexpected files in /etc, aborting activation
The following files have unrecognized content and would be overwritten:

  /etc/nix/nix.conf

Please check there is nothing critical in these files, rename them by adding .before-nix-darwin to the end, and then try again.
```
3. Fork this repository and clone it locally:
```
$ git clone https://github.com/LnL7/nix-darwin/
$ cd nix-darwin
```
4. Copy the /etc/nix/nix.conf from the `determinate-nixd`:
```
cp /etc/nix/nix.conf ./doc/known-files/$(sha256sum /etc/nix/nix.conf | cut -d " " -f1)
```
5. Add the known sha256 hash into `modules/nix/default.nix`
6. Commit and Create the pr:
```
$ git add . && git commit -am "Add known hash for determinate-nixd $(determinate-nixd --version)"
$ gh pr create
```

### DIFF
--- a/doc/known-files/8a9ebee68a25f8e1a74f510059e4690f6c2c36842e9bad065060cacb178b3670
+++ b/doc/known-files/8a9ebee68a25f8e1a74f510059e4690f6c2c36842e9bad065060cacb178b3670
@@ -1,0 +1,21 @@
+# DETERMINATE NIX CONFIG
+# do not modify! this file will be replaced!
+# user modification can go in nix.custom.conf
+
+!include nix.custom.conf
+
+max-jobs = auto
+bash-prompt-prefix = (nix:$name)\040
+extra-experimental-features = nix-command flakes
+
+netrc-file = /nix/var/determinate/netrc
+post-build-hook = /nix/var/determinate/post-build-hook.sh
+
+always-allow-substitutes = true
+extra-substituters = https://cache.flakehub.com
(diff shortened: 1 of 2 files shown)


onnimonni:
I tested this approach a bit and the only problem with it seems to be that relaunching the determinate nix daemon will overwrite the `/etc/nix/nix.conf`:
```
$ darwin-rebuild switch --flake .
$ sha256sum /etc/nix/nix.conf
c95ec3bbffe13f0d1c0f4964e97650f9bbb711754503525f137b7f0f54d5ccbc  /etc/nix/nix.conf
$ sudo launchctl kickstart -k system/systems.determinate.nix-daemon
$ sha256sum /etc/nix/nix.conf
8a9ebee68a25f8e1a74f510059e4690f6c2c36842e9bad065060cacb178b3670  /etc/nix/nix.conf
```

onnimonni:
So it would be great if we could find a way to write the added nix configs to the included file:
```sh
$ head /etc/nix/nix.conf -n 4
# DETERMINATE NIX CONFIG
# do not modify! this file will be replaced!
# user modification can go in nix.custom.conf

!include nix.custom.conf
```

emilazy:
Unfortunately this PR wouldn’t work well, as `determinate-nixd` takes ownership of `/etc/nix/nix.conf` and overwrites it whenever the daemon starts. nix-darwin’s Nix management logic also wants to take ownership of the Nix installation (in many ways, not just the configuration) and can’t interoperate well with another service doing similar tasks. https://github.com/LnL7/nix-darwin/pull/1313 should provide a better option for Determinate users once it’s merged, where Nix management can be disabled on the nix-darwin side and the rest of nix-darwin’s functionality can be used as normal.

PR #1299, opened 2025-01-26
activation-scripts: add `/usr/sbin` and `/sbin` for user activation

Enzime:
Some binaries inside these folders can be run without `sudo` like `softwareupdate`.


emilazy:
Addressed *en passant* by https://github.com/LnL7/nix-darwin/pull/1313, for what it’s worth.

PR #1300, opened 2025-01-26
ci: use the PR head as `<darwin>` for install test

emilazy:
This was installing against the base branch, which causes annoying issues when things like the state version are bumped. It’s possible this was intentional to test the upgrade path, but it’s inconsistent with the flake installation test and keeps breaking stuff on my PRs. If we want to test the upgrade path then we could re‐add it again later in a more sophisticated manner (e.g. using the example config from the previous version).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,6 @@ on:
 
 env:
   NIXPKGS_BRANCH: nixpkgs-unstable
-  NIX_DARWIN_BRANCH: master
   NIX_VERSION: 2.24.11
 
 jobs:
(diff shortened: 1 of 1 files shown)


Enzime (approved):
Would be good to readd a better upgrade test in the future

Or alternatively have the permissions to merge even when the test fails

emilazy:
Yeah, a proper upgrade test would be nice. But we don’t even have one for flakes right now and the idea of making that work for the Plan where the `sudo` use would change before and after gives me a headache, so seems like it’d be better to re‐add it after the current batch of knotty PRs.

Enzime (approved):

PR #1301, opened 2025-01-27
Add support for extraModules input

oliverwiegers:
NixOS offers the possibility to pass a list of extraModules to lib.nixosSystem which later get merged with the modules list and baseModules.
This offers the possibility to read additional modules from environment files or a custom module-list.nix.

NixOS implementation is
[here](https://github.com/NixOS/nixpkgs/blob/8601973cec591ab18357eb7c26c52769d990d1ed/nixos/lib/eval-config.nix#L33).

Dor nix-darwin an implementation was started in #592. I restarted the implementation and removed the usage of `NIXOS_EXTRA_MODULE_PATH` env variable because NixOS deprecated this one.

This is my first PR. Anything missing?


oliverwiegers:
@emilazy @Enzime I can't request a review. Is just opening the PR and waiting the way to go?

Let  me know if you need anything from me :)

emilazy:
I’d be okay doing this just on general NixOS interface compatibility grounds, but I’m curious what value it adds for you without the environment variable, since there’s no difference between setting `extraModules` and just adding things to `modules`?

oliverwiegers:
@emilazy you're right it does not add any additional functionality without the environment variable. I would be happy to keep it, but I wanted to make it as close to the NixOS implementation as possible.

My main goal (should have stated this earlier) is to provide a way to separate different modules when invoking `darwinSystem` to make the function call more readable.
As an example: By having `extraModules` as an input I can put my `configuration.nix` and third party modules in the `modules` input and my custom modules in `extraModules`. 

In short: Make ones personal config easier to read. 

Does that make sense?

P.S. Using the variable would be still usable in somewhat readable way:

```nix
lib.darwinSystem {
  ...
  modules = [configuration.nix];
  extraModules = [
    { imports = [ (builtins.getEnv "NIXOS_EXTRA_MODULE_PATH") ]; }
  ];
  ...
}
```

Which I think is more readable that putting both in `modules`

oliverwiegers:
@emilazy if this isn't something you would like to add you can close the PR 😊

PR #1305, opened 2025-01-28
checks: prepare for user activation removal

emilazy:


emilazy:
Closing in favour of https://github.com/LnL7/nix-darwin/pull/1313 which addresses the need for most of these checks more holistically.

PR #1311, opened 2025-01-30
checks: fix sw_vers parameter for macOSVersion

tarc:
In `modules/system/checks.nix`, `sw_vers` is called with `--productVersion`, but the correct parameter syntax is `-productVersion`.


emilazy (approved):
Thanks! This check is specifically aimed at people on OS versions, so doing the maximally‐compatible thing here makes sense. I should have tested this on 10.12, really.

PR #1320, opened 2025-02-03
users: use `launchctl managername` to determine session type

emilazy:
See commit messages for details.

### DIFF
--- a/modules/users/default.nix
+++ b/modules/users/default.nix
@@ -147,7 +147,7 @@ in
         homeDirectory=''${homeDirectory#NFSHomeDirectory: }
 
         if ! sudo dscl . -change /Users/nobody NFSHomeDirectory "$homeDirectory" "$homeDirectory" &> /dev/null; then
-          if [[ -n "$SSH_CONNECTION" ]]; then
+          if [[ "$(launchctl managername)" != Aqua ]]; then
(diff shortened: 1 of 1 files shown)


modules/users/default.nix
```diff
@@ -146,8 +146,8 @@ in
         homeDirectory=$(dscl . -read /Users/nobody NFSHomeDirectory)
         homeDirectory=''${homeDirectory#NFSHomeDirectory: }
 
-        if ! sudo dscl . -change /Users/nobody NFSHomeDirectory "$homeDirectory" "$homeDirectory" &> /dev/null; then
-          if [[ -n "$SSH_CONNECTION" ]]; then
+        if ! dscl . -change /Users/nobody NFSHomeDirectory "$homeDirectory" "$homeDirectory" &> /dev/null; then
```
Enzime, line 149: https://github.com/LnL7/nix-darwin/blob/ae406c04577ff9a64087018c79b4fdc02468c87c/modules/system/activation-scripts.nix#L99-L104

Checks still currently runs as part of user activation

Enzime (changes requested):

modules/users/default.nix
emilazy: Ah, of course. I missed that this is part of the checks phase. Moved that commit back to the Plan stack, thanks!

Enzime (approved):

Issue #1314, opened 2025-01-30
homebrew: add `masPackage` option


sellout:
When `homebrew.masApps` isn’t empty, the `homebrew` module adds `"mas"` to `homebrew.brews`.

Unfortunately, the `mas` formula is failing for me on an old Mac (stuck on Big Sur / macOS 11). But I can install `pkgs.mas` from Nixpkgs just fine, and `brew` will use that no problem.

The output is a bit funny. First Homebrew fails to install `mas`
```
Homebrew bundle...
Installing mas
Warning: You are using macOS 11.
We (and Apple) do not provide support for this old version.
It is expected behaviour that some formulae will fail to build in this old version.
It is expected behaviour that Homebrew will be buggy and slow.
Do not create any issues about this on Homebrew's GitHub repositories.
Do not create any issues even if you think this message is unrelated.
Any opened issues will be immediately closed without response.
Do not ask for help from Homebrew or its maintainers on social media.
You may ask for help in Homebrew's discussions but are unlikely to receive a response.
Try to figure out the problem yourself and submit a fix as a pull request.
We will review it but may or may not accept it.

mas: A full installation of Xcode.app 14.2 is required to compile
this software. Installing just the Command Line Tools is not sufficient.

Xcode 14.2 cannot be installed on macOS 11.
You must upgrade your version of macOS.
Error: mas: An unsatisfied requirement failed this build.
Installing mas has failed!
```
but then it later happily installs apps using `mas`.
```
Installing 1Password for Safari
Installing Deliveries
Installing Harvest
Installing reMarkable
Homebrew Bundle failed! 1 Brewfile dependency failed to install.
```
The most unfortunate bit is that this success-despite-failure isn’t enough to get nix-darwin activation to complete, so I need to do _something_ to get the module to not install `mas` via `brew`.

I was thinking an option like `homebrew.masPackage`. If it’s set to a derivation, just add that to `environment.systemPackages` (or maybe scope it to `brew`?). Otherwise, do the current behavior. I lean toward the default value being `pkgs.mas` (I try to only use Homebrew if I can’t get the package otherwise), but I also understand if `null` (i.e., the current behavior) is preferred.

emilazy:
I just ran into this while working on activation `$PATH`s, funnily enough.

[...]

PR #1329, opened 2025-02-08
homebrew: use `mas` from Nixpkgs

emilazy:
Currently, setting `homebrew.masApps` causes `"mas"` to automatically be added to `homebrew.brews`. Users who want to use Homebrew only for managing App Store apps and casks, like me, can override this by setting `homebrew.brews = lib.mkForce [ ];` and adding Nixpkgs’ `mas` to their `environment.systemPackages`.

When the activation script path no longer depends on the built configuration’s `environment.systemPackages`, this will no longer work. Since this was originally added before `mas` was packaged in Nixpkgs and we now have a perfectly serviceable binary package, we can add it to the `$PATH` when invoking Homebrew and skip the automatic formula installation.

As the Homebrew `bin` directory still comes first, users who specifically want the formula can restore the previous behaviour by explicitly adding `"mas"` to `homebrew.brews`.

Closes: #1314

---

This should probably wait until https://github.com/NixOS/nixpkgs/pull/379858 is merged since the Nixpkgs `mas` package is quite old right now (although it’s been working fine for me). I’ll take a look at that PR in the next couple days if nobody else gets around to it.

cc @sellout

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -708,9 +708,6 @@ in
       description = ''
         Applications to install from Mac App Store using {command}`mas`.
 
-        When this option is used, `"mas"` is automatically added to
-        [](#opt-homebrew.brews).
-
         Note that you need to be signed into the Mac App Store for {command}`mas` to
         successfully install and upgrade applications, and that unfortunately apps removed from this
         option will not be uninstalled automatically even if
@@ -768,8 +765,7 @@ in
     ];
 
     homebrew.brews =
-      optional (cfg.masApps != { }) "mas"
-      ++ optional (cfg.whalebrews != [ ]) "whalebrew";
+      optional (cfg.whalebrews != [ ]) "whalebrew";
 
     homebrew.brewfile =
       "# Created by `nix-darwin`'s `homebrew` module\n\n"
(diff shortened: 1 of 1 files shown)


modules/homebrew.nix
```diff
@@ -789,7 +785,8 @@ in
       # Homebrew Bundle
       echo >&2 "Homebrew bundle..."
       if [ -f "${cfg.brewPrefix}/brew" ]; then
```
sellout, line 787: I don’t think this is very likely, but what if `brew` is not in `brewPrefix`[^1]? Then `mas` won’t be added to the `PATH`.

[^1]: If `brew` isn’t on the `PATH`, then there are bigger issues, but it could just be in a non-standard location, but still on the `PATH`.

sellout (approved):
This is great, thanks.

> Users who want to use Homebrew only for managing App Store apps and casks, like me, can override this by setting `homebrew.brews = lib.mkForce [ ];` and adding Nixpkgs’ `mas` to their `environment.systemPackages`.

This did work for me, but I initially thought it wouldn’t because I pull my configuration together from multiple repos, and I had packages in `homebrew.brews` (because they weren’t in Nixpkgs, or weren’t working on darwin via Nixpkgs). But it turned out that there was only one for that particular host, and it had since been fixed in Nixpkgs, so I removed it.

tl;dr – it’s not always fine to `mkForce []` the list (and I wish there were better ways for manipulating lists, rather than just replacing them).

> This should probably wait until [NixOS/nixpkgs#379858](https://github.com/NixOS/nixpkgs/pull/379858) is merged since the Nixpkgs `mas` package is quite old right now (although it’s been working fine for me).

I have had some issues with Nixpkgs current `mas` when using `uninstall`, which is annoying as I have an activation script that uninstalls App Store packages that aren’t included in my nix-darwin configuration.

modules/homebrew.nix
emilazy: The `else` condition is just an error message, so that seems okay?

Enzime (approved):

modules/homebrew.nix
sellout: Whoops – I overlooked that. Yeah, totally fine, sorry.

PR #1336, opened 2025-02-13
readme: fix `sed` expression

DamienCassou:


Samasaur1:
My guess is that you're using GNU sed as opposed to the BSD sed that comes with macOS. This command works fine with the macOS sed

Samasaur1:
We could also update the command to be
```
sed -i'' -e "s/simple/$(scutil --get LocalHostName)/" flake.nix
```
which works with both BSD and GNU sed

DamienCassou:
@Samasaur1: I didn't install any version of sed specifically. I'm just switching from linux to os x today and I'm still in the process of installing stuff. Anyway, I changed my commit to follow your suggestion.

README.md
```diff
@@ -53,7 +53,7 @@ nix flake init -t nix-darwin/master
 # To use Nixpkgs 24.11:
 nix flake init -t nix-darwin/nix-darwin-24.11
 
-sed -i '' "s/simple/$(scutil --get LocalHostName)/" flake.nix
+sed -i'' "s/simple/$(scutil --get LocalHostName)/" flake.nix
```
Samasaur1, line 56:
```suggestion
sed -i'' -e "s/simple/$(scutil --get LocalHostName)/" flake.nix
```

Samasaur1 (changes requested):
You're still missing the `-e` option

README.md
Enzime: @DamienCassou I think it's still missing the `-e`

emilazy: `-i''` is incorrect for macOS’s default BSD sed. It needs to be `-i ''` or you get a `flake.nix-e` file.

DamienCassou:
I don't see any way to make a command that works in all cases. I am therefore closing the PR. Feel free to reopen if you see a way forward.

PR #1338, opened 2025-02-14
[24.11] nix: place `extra`-prefixed settings after their non-prefixed variants

Enzime:
Backport of #1337


emilazy:
LGTM, but this should probably use `git cherry-pick -x` like Nixpkgs does to include the original commit being backported (I still need to find a nice Jujutsu‐native way of doing that).

emilazy (approved):

PR #1341, opened 2025-02-16
The Plan, phase 1

emilazy:
It’s finally time.

This PR eliminates user activation, addressing https://github.com/LnL7/nix-darwin/issues/96 by following the first phase of the plan from https://github.com/LnL7/nix-darwin/pull/1016#issuecomment-2285910068, fulfilling a goal I started drafting patches for in 2023 before I was even a nix-darwin maintainer.

I strongly recommend reviewing this one commit‐by‐commit. See the commit messages for much more detail and for things I’m still uncertain about. I’ve been running this on my own system for a while now, but it could definitely use a staged roll‐out before we actually hit the merge button. We should also probably also reach out to authors of third‐party modules that will be affected once it’s mostly baked, though at least anything that breaks should mostly break *loudly*. (Adding text to our existing `activationScripts.{checks,userDefaults,userLaunchd,homebrew}.text` definitions is an exception – those will run as `root` now. But I can’t think of a particularly graceful way to handle that case, and anyway it’s not exactly something we fully support in the first place.)

Closes: #96


modules/system/primary-user.nix
```diff
@@ -0,0 +1,60 @@
+{
+  lib,
+  options,
+  config,
+  ...
+}:
+
+{
+  options = {
+    system.primaryUser = lib.mkOption {
```
spikespaz, line 10: Can we have an option to disable primary user, so that dogfooding finds problems with not having a primary user sooner?

spikespaz (commented):
> in practice, it may be useful for us to
nest this into `users.users.*.homebrew` instead, at the expense of
being an unsupported setup if used to its full potential. Since
that would be a breaking change to the inteface anyway, I think
adding `homebrew.user` for now is acceptable.

I understand the motive here, but are you sure it's not better to make more breaking changes in one go? I'm worried that going through another great migration would agitate some.

modules/system/primary-user.nix
spikespaz: I just realized the default is `null`, so never mind.

emilazy: Yeah, leaving it at the default will ensure your system doesn’t depend on any primary user and give a hopefully‐helpful error message if you set anything that requires it.

emilazy:
I previously thought that too – that we should figure out the correct interfaces and write good migration documentation and tools and get all the breaking changes over with in one go – but the end result was that this work stalled out for years because there’s a lot of API design and implementation work to get things to a good state: we need to decide which things should be global or per‐user; we need to decide what should stay in nix-darwin and what should be deduplicated with Home Manager; for things like our `system.defaults` options, they’re currently not fully declarative in the way that Nix‐managed options should be, so if we’re breaking everyone’s configuration then we ought to make that important breaking change at the same time; and so on.

If we tried to do it all at once, this work would have spent even longer in development hell and be unreviewably big. For the longest time i wasn’t sure how to resolve that, but after discussing on Matrix I realized that introducing a `system.primaryUser` option, though strange from an interface design perspective, would allow us to minimize the breaking changes in the first phase: the only required change for the vast majority of users should be having to set `system.primaryUser = "username";` in their configuration, which I think is an acceptable level of fuss for the improvement in internal architecture and coherent model it gives us, and allows people who want a truly user‐agnostic system manager to opt for that world now at the expense of reduced functionality. That takes the pressure off any major interface renovations and lets us work on them independently and batch them up as we see fit.

In the case of Homebrew, `homebrew.user` defaults to the primary user, so it’s purely a bonus for people who want to use Homebrew without setting `system.primaryUser` (or to use a different user entirely for Homebrew, I suppose).

hraban:
Most of the system.defaults stuff should really move to home-manager anyway, as you said. It's a natural fit and it works well.

mirkolenz:
I've been running this branch on my machine for the past week and had zero issues with it, thanks for these great improvements!

khaneliman:
Been running this for a few days since I got back from vacation. Noticing issues with nix store on reboots, haven't pinned down what's going on atm. Not sure if I need to do something different, I just tested copying the branch and rebasing it on master. Running my normal rebuild mentions the sudo requirement, doing a sudo rebuild mentions I need a primary user, adding a primary user and then it rebuilds.

 Just noting symptoms:

- Dock icons missing that point to nix store
- Firefox error about profile missing on boot (assuming same issue with dock about files being available on login)
- launchd services in a retry backoff until i unload them and load them manually
 
```bash
2025-03-08 14:57:39.712350 <Warning> Could not find and/or execute program specified by service: 2: No such file or directory: /nix/store/fa3jr2wxgbiyk37h0imzhhj13c3p6gqv-yabai-7.1.10/bin/yabai
2025-03-08 14:57:39.712351 <Error> Service could not initialize: access(/nix/store/fa3jr2wxgbiyk37h0imzhhj13c3p6gqv-yabai-7.1.10/bin/yabai, X_OK) failed with errno 2 - No such file or directory, error 0x6f - Invalid or missing Program/ProgramArguments
2025-03-08 14:57:39.712352 <Error> initialization failure: 24D70: xpcproxy + 36768 [1088][E4C3A3C3-0D57-3E4D-8388-880BC7F5F19E]: 0x6f
2025-03-08 14:57:39.712353 <Notice> Service setup event to handle failure and will not launch until it fires.
2025-03-08 14:57:39.712354 <Error> Missing executable detected. Job: 'org.nixos.yabai' Executable: '/nix/store/fa3jr2wxgbiyk37h0imzhhj13c3p6gqv-yabai-7.1.10/bin/yabai'

**Comment by emilazy** [comment]
That’s very confusing. I’m not sure how that could result from these changes. Are you sure you get the same result if you run the base commit that you rebased on top of? Were there any conflicts during the rebase that you had to resolve?

**Comment by khaneliman** [comment]
> > ```shell
> > 2025-03-08 14:57:39.712350 <Warning> Could not find and/or execute program specified by service: 2: No such file or directory: /nix/store/fa3jr2wxgbiyk37h0imzhhj13c3p6gqv-yabai-7.1.10/bin/yabai
> > 2025-03-08 14:57:39.712351 <Error> Service could not initialize: access(/nix/store/fa3jr2wxgbiyk37h0imzhhj13c3p6gqv-yabai-7.1.10/bin/yabai, X_OK) failed with errno 2 - No such file or directory, error 0x6f - Invalid or missing Program/ProgramArguments
> > 2025-03-08 14:57:39.712352 <Error> initialization failure: 24D70: xpcproxy + 36768 [1088][E4C3A3C3-0D57-3E4D-8388-880BC7F5F19E]: 0x6f
> > 2025-03-08 14:57:39.712353 <Notice> Service setup event to handle failure and will not launch until it fires.
> > 2025-03-08 14:57:39.712354 <Error> Missing executable detected. Job: 'org.nixos.yabai' Executable: '/nix/store/fa3jr2wxgbiyk37h0imzhhj13c3p6gqv-yabai-7.1.10/bin/yabai'
> > ```
> 
> This is the exact reason `nix-darwin` uses `/bin/wait4path`. You can replace the binary inside of the service with a `/bin/sh` that calls `wait4path` with the yabai executable or try copying it outside of the store

This is using the nix-darwin service. I do see that it doesn't contain `wait4path` like it used to. Thanks for reminding me about that. I'll see why it's missing now.

File: modules/system/launchd.nix [hunk 2]
```diff
@@ -132,14 +134,16 @@ in
...
 
-    system.activationScripts.userLaunchd.text = ''
+    system.activationScripts.userLaunchd.text = let
+      user = config.system.primaryUser;
+    in mkIf (config.launchd.user.envVariables != { } || userLaunchAgents != [ ]) ''
       # Set up user launchd services in ~/Library/LaunchAgents
       echo "setting up user launchd services..."
 
-      ${concatStringsSep "\n" (launchdVariables config.launchd.user.envVariables)}
+      ${concatStringsSep "\n" (launchdVariables "sudo -u ${user} --" config.launchd.user.envVariables)}
```
Samasaur1, line 143: Using `sudo -u ${user}` does feel like one of those weird corner-cases where launchd will just fail to do what you expect it to — I like the idea of using `launchctl asuser`, especially considering it hasn't been replaced with anything else (other than using a domain target for the given user), but I do notice the man page says

>  This executes the given command in as similar an execution context as possible to that of the target user's bootstrap. Adopted attributes include the Mach bootstrap namespace, exception server and security audit session. **It does not modify the process' credentials (UID, GID, etc.) or adopt any user-specific environment variables.** It affects only the Mach bootstrap context and directly- related attributes.

(emphasis mine). This might mean that the way to do it is
```
sudo -u ${user} -- launchctl asuser $(id -u ${user})
```
which is ugly but seems like it carries over more of the appropriate context.

That said, if this is working as-is in multi-user setups I'm also inclined to leave it as is until we swap to the new launchctl commands. My concern is mostly around activation for users who aren't graphically logged-in. This is already sometimes a pain point for me when I darwin-rebuild over SSH, and so I imagine that we may run into similar issues here if user A (logged in graphically) runs darwin-rebuild and the launchd activation doesn't work properly for user B (maybe not logged in, maybe logged in in a separate graphical session, etc.)

modules/system/defaults-write.nix
```diff
@@ -9,6 +9,10 @@ let
     "defaults write ${domain} '${key}' $'${strings.escape [ "'" ] (generators.toPlist { } value)}'";
 
   defaultsToList = domain: attrs: mapAttrsToList (writeDefault domain) (filterAttrs (n: v: v != null) attrs);
+  userDefaultsToList = domain: attrs: map
+    (cmd: "sudo --user=${escapeShellArg config.system.primaryUser} -- ${cmd}")
```
Samasaur1, line 13: Is there a reason you use `sudo -u ${user}` for `launchctl` but `sudo --user=${user}` here?

modules/environment/default.nix
```diff
@@ -175,6 +175,16 @@ in
 
   config = {
 
+    # This is horrible, sorry.
+    system.requiresPrimaryUser = mkIf (
+      config.nix.enable
+      && !config.nixpkgs.flake.setNixPath
+      && config.system.stateVersion < 6
+      && options.environment.darwinConfig.highestPrio == (mkOptionDefault {}).priority
```
Samasaur1, line 183: I hate that I can't think of a better way to check that this option wasn't changed from the default

modules/system/primary-user.nix
```diff
@@ -19,6 +19,13 @@
       '';
     };
 
+    system.primaryUserHome = lib.mkOption {
+      internal = true;
+      type = lib.types.str;
+      default =
+        config.users.users.${config.system.primaryUser}.home or "/Users/${config.system.primaryUser}";
+    };
```
Samasaur1, line 27: Will this break anything if `config.system.primaryUser` is set to `null` (the default)? If I'm reading it correctly, it's only an issue if `primaryUserHome` is read by something, in which case `primaryUser` needs to be set anyway, but I wonder what the error message would look like. Maybe it's caught by `nix.nixPath` requiring `primaryUser` to be set? Not sure that works for `environment.darwinConfig` though

pkgs/darwin-uninstaller/configuration.nix
```diff
@@ -15,13 +15,17 @@ with lib;
...
-  system.activationScripts.postUserActivation.text = mkAfter ''
-    nix-channel --remove darwin || true
-  '';
-
   system.activationScripts.postActivation.text = mkAfter ''
     nix-channel --remove darwin || true
 
+    ${optionalString (config.system.primaryUser != null) ''
+      sudo \
+        --user=${config.system.primaryUser} \
```
Samasaur1, line 23: not shell escaping the primary user

modules/services/activate-system/default.nix
```diff
@@ -65,7 +65,7 @@ in
           ln -sfn /run/current-system /nix/var/nix/gcroots/current-system
         fi
 
-        ${config.system.activationScripts.etcChecks.text}
+        ${config.system.activationScripts.checks.text}
```
Samasaur1, line 68: The one thing I am concerned about here is if `checks` has stuff that needs to be run in a graphical context. Based on a quick search through our in-tree checks it seems like this change would cause errors for the activation daemon from `modules/users/default.nix` in some cases, it might cause errors for our other checks, and might for out-of-tree checks if people were assuming that `checks` would only run in a graphical session.

modules/system/activation-scripts.nix
```diff
@@ -113,8 +113,8 @@ in
 
       ${cfg.activationScripts.preActivation.text}
 
-      ${cfg.activationScripts.createRun.text}
       ${cfg.activationScripts.checks.text}
+      ${cfg.activationScripts.createRun.text}
```
Samasaur1, line 117: This is also a potential point where out-of-tree checks could break, if people assume that `/run` will already exist. I think this is still a worthwhile change to make, but worth pointing out.

Samasaur1 (commented):
It's finally happening! This looks really good in general — left a couple notes and questions but I think many of these changes may just have to be breaking.

modules/environment/default.nix
khaneliman: This is the only way I've done it, too.

tests/homebrew.nix
```diff
@@ -15,6 +15,8 @@ in
 {
   homebrew.enable = true;
 
+  homebrew.user = "test-homebrew-user";
```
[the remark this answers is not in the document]
emilazy: The test systems don’t actually get activated; we just build the system and check the resulting package. (Sadly.) So this is fine.

modules/system/launchd.nix
emilazy: Right. I haven’t tested this extensively; if someone can confirm that it seems to work as‐is then I’d be inclined to just leave it until we start using the recommended commands rather than adding more deprecated ones. I’m certainly curious how this goes with actiavtion over SSH.

modules/system/defaults-write.nix
emilazy: No reason at all. I’ve moved to `--user=` across the board.

modules/environment/default.nix
emilazy: Yeah it’s awful and I hate it. But I hate not giving good error messages when doing complicated migrations even more.

modules/system/primary-user.nix
emilazy: It’ll be an evaluation error if anything depends on it. But, yes, that stuff should set `system.requiresPrimaryUser` (as `environment.darwinConfig` does) and therefore hopefully give a helpful error message before anything that uses it would get evaluated. `"${null}"` is also an error so I’m not sure allowing it to be `null` would improve anything. Maybe this could be a more descriptive `throw` in that case? But I haven’t triggered it so far in testing.

modules/services/activate-system/default.nix
emilazy: The activation daemon only activates system configurations that were previously current. So in theory it should never be trying to delete users and so on. What other checks do you think might be problematic here?

tests/system-defaults-write.nix
```diff
@@ -1,6 +1,8 @@
 { config, pkgs, lib, ... }:
 
 {
+  system.primaryUser = "test-defaults-user";
```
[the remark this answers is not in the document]
emilazy: (Same answer.)

modules/system/primary-user.nix
Samasaur1: Maybe
```nix
lib.warnIf (config.system.primaryUser == null) "`system.primaryUser` is unset but `system.primaryUserHome` is being read" config.users.users.${config.system.primaryUser}.home or "/Users/${config.system.primaryUser}";
```
?

On the other hand, it seems like this problem will only be hit due to a future mistake on our part (reading `primaryUserHome` in an option without adding that option to `requiresPrimaryUser`), so we could just leave this as is.

modules/system/activation-scripts.nix
Samasaur1: I think this is definitely an improvement that we should make, just noting the potential issue here in case we see it in the future.

modules/services/activate-system/default.nix
Samasaur1: Ah, good point. I don't have any problematic checks I can think of, and they should be read-only anyway, so I'm good with this.

modules/system/launchd.nix
Samasaur1: @mirkolenz @khaneliman and anyone else who's running this branch: any chance you manage any multi-user nix-darwin systems? Have you noticed any strangeness with launchd activation, especially if you've happened to run activations as one user while a different user is actively signed in?

Samasaur1: Or activation over SSH

Samasaur1: It's also occurring to me that even if this doesn't work perfectly, it should continue to work at least as well as the situation before this PR. I'd say that if anyone reports an issue with this, we can try and track it down, but otherwise we can just leave it as you have it here

Samasaur1 (approved):
I think that covers everything I saw! Another pair of eyes certainly wouldn't hurt but this LGTM

modules/system/launchd.nix
mirkolenz: All systems I tested this on are single-user only, sorry...

modules/system/primary-user.nix
emilazy: Making it a `throwIf` might be a good idea, and seems pretty harmless. I’ll think about it. The main thing is that I wish we could make incorrect uses of `config.system.primaryUser` fail loudly in the same way, but that’s harder.

modules/system/launchd.nix
emilazy: As in, because people will usually be doing `sudo darwin-rebuild …` as their primary user, and so the `sudo sudo --user=primary-user` will cancel out the UID stuff and the Mach bootstrap context will be inherited from their graphical session? I suppose so. But ideally this should work even if you set a different primary user, too, so it’d be good to know one way or another. I just haven’t had the time to set up a VM to test a separate admin user, or SSH activation, or whatever. (I tested all this on my development machine because I love danger.)

retrry:
Hey, I've just not long ago moved to flakes and don't have a lot of experience here. Could someone give me rough instructions how I would test it with flake setup? I would like to help 🙂 based on what I am seeing in the PR already started migrating away from services and config options which require primaryUser.

mirkolenz:
You can test it by adding the following snippet to your flake:

```nix
inputs.nix-darwin = {
  url = "github:lnl7/nix-darwin/pull/1341/merge";
  inputs.nixpkgs.follows = "nixpkgs"; # if you want to lock it to the same rev as your main nixpkgs instance
};
```

retrry:
So it is a bit strange for me with launch agents 🙂 I've had couple launch agents defined via `launchd.user.agents`. Those options require primaryUser set (these are only two options, which I've had left, which require primaryUser). I've tried two things:
1. Just remove those options and not set primaryUser. The problem is that those launchAgents do not get removed.
2. Set primaryUser, rebuild, remove launchd options, rebuild - those launch agents are still not removed.
3. Reverted to master config, rebuild, remove those options, rebuild - now they get removed.

Samasaur1:
3 sounds like expected behavior. 1, unfortunately, also sounds like expected behavior, since the code to handle removing those LaunchAgents has itself been removed. 2 is surprising to me though, unless the LaunchAgent plists are put at a different path?

emilazy:
Rebased. I’m expecting this to probably not land before we do the repository move, although it’s still my highest‐priority change to the actual codebase.

It would be great if someone had the time to test `launchd.user.agents` reloads in a multi‐user setup / over SSH, per https://github.com/LnL7/nix-darwin/pull/1341#discussion_r1989887502. I would consider that the main blocker for merge, other than people trying it out on their systems. (Do user agent reloads even work properly currently…? The activation script prints out some eyebrow‐raising messages when it tries, IIRC.)

I’m wondering if it sense to rename the activation script options while doing this, to pave the way for safe custom activation scripts that never overlap with NixOS (see the tragically‐delayed https://github.com/LnL7/nix-darwin/pull/664) and to avoid situations where someone has something in `system.activationScripts.homebrew` or whatever that really isn’t safe to run as `root` (although that kind of monkey‐patching is in warranty‐voiding territory anyway). I think that would be a pain for nix-homebrew, though (cc @zhaofengli), and maybe cross‐OS modules like Home Manager too, so maybe it’s not worth it.

@retrry Yes, this generally sounds like the typical unfortunate behaviour we currently have whenever we stop tracking files. I am not sure if there is much elegant we can do to avoid that right now. In the future @Samasaur1’s nice file tracking stuff would remove the files in (1) there. Maybe it would make sense to land that functionality before this, although I expect that the majority of people will just set a primary user at first rather than trying to de‐user‐ify their configuration upfront. I am indeed surprised by (2) though and maybe we should figure that out. Were you setting the same primary user as the one that was running `darwin-rebuild` beforehand?

retrry:
@emilazy sorry for taking so long to get back, but you know - life 😄 But today I've tested it again and yeah, launch agents do not get removed for me in a specific case 🙁 

My flow:
1. Update flake.nix
2. Run `darwin-rebuild switch`
3. Run `sudo darwin-rebuild switch`
4. Add launch agent and set primaryUser
5. Run `sudo darwin-rebuild switch -- agent gets added.
6. Comment out agent code
7. Run `sudo darwin-rebuild switch`
8. Agent is still there...

My launch agent is pretty simple:
```
  launchd.user.agents = {
    docker_host = {
      serviceConfig = {
        ProgramArguments = [
          "launchctl"
          "setenv"
          "DOCKER_HOST"
          "unix:///Users/${username}/.config/colima/default/docker.sock"
        ];
        RunAtLoad = true;
        KeepAlive = false;
      };
    };
  };
  ```
After testing more - I think the problem is when you fully remove launch agents. Then nothing is being run. 
  
I've create two agents docket_host and docker_host_2. If I remove one - it gets removed. If I remove both - it doesn't get removed 🙂 

Running `sudo darwin-rebuild switch` key log parts:
One removed, but at least 1 launch agent still exists:
```
setting up /etc...
removing user service org.nixos.docker_host
reloading nix-daemon...
```

Both removed and no agents exist:
```
setting up /etc...
reloading nix-daemon...
```

So it seems that if that config options is not defined - `setting up user launchd services...` just doesn't run at all.

I hope that helps 🤷

File: modules/system/checks.nix [hunk 11]
```diff
@@ -31,6 +31,18 @@ let
     fi
   '';
 
+  primaryUser = ''
+    primaryUser=${escapeShellArg config.system.primaryUser}
```
Enzime, line 35:
```suggestion
    primaryUser="${config.system.primaryUser}"
```

```
In /nix/store/726gp7jnx7blz46azrds8l28a5zqdqgg-darwin-system-25.05.0653ce2/activate line 63:
primaryUser=admin
^---------^ SC2209 (warning): Use var=$(command) to assign output (or quote to assign string).
 
For more information:
  https://www.shellcheck.net/wiki/SC2209 -- Use var=$(command) to assign outp...
```

Enzime (changes requested):

modules/system/checks.nix
Enzime: `escapeShellArg` will add some single quotes which will mess with the value of `primaryUser` I believe so best to leave it out here if we're manually quoting

Alternatively we could use `escapeShellArgAlways`:

https://github.com/nix-darwin/nix-darwin/blob/43975d782b418ebf4969e9ccba82466728c2851b/modules/services/github-runner/service.nix#L105-L108

Samasaur1: Could we also just do `escapeShellArg` and disable this warning? Although I don't know if that would be tangibly different from just doing what you suggested here.

hraban: Feels like the linter is wrong here because it doesn't know the value came from a proper escape function. my +1 on silencing the linter. For what that's worth :)

modules/users/default.nix
```diff
@@ -155,7 +155,7 @@ in
         homeDirectory=$(dscl . -read /Users/nobody NFSHomeDirectory)
```
Enzime, line 155: The comment above this is no longer necessary 👍

tests/networking-hostname.nix
```diff
@@ -9,6 +9,5 @@
     grep "scutil --set ComputerName 'EVE’s MacBook Pro'" ${config.out}/activate
     grep "scutil --set LocalHostName ${lib.escapeShellArg "EVE"}" ${config.out}/activate
     grep "scutil --set HostName ${lib.escapeShellArg "EVE"}" ${config.out}/activate
-    echo checking defaults write in ${config.out}/activate-user >&2
```
Enzime, line 12: Minor nit: could be squashed into the previous commit

README.md
```diff
@@ -186,13 +186,13 @@ The documentation is also available as manpages by running `man 5 configuration.
 To run the latest version of the uninstaller, you can run the following command:
 
 ```
-nix --extra-experimental-features "nix-command flakes" run nix-darwin#darwin-uninstaller
+sudo nix --extra-experimental-features "nix-command flakes" run nix-darwin#darwin-uninstaller
```
Enzime, line 189: Not a blocker, but this might lead to some issues with `system.primaryUser` not being set

Enzime (approved):
Apart from all my comments above, this PR looks good to me

I'm running successfully on 3 machines so far :)

Isolaryn:
Have been using this without issue so far.

kfiz:
> > > ```shell
> > > 2025-03-08 14:57:39.712350 <Warning> Could not find and/or execute program specified by service: 2: No such file or directory: /nix/store/fa3jr2wxgbiyk37h0imzhhj13c3p6gqv-yabai-7.1.10/bin/yabai
> > > 2025-03-08 14:57:39.712351 <Error> Service could not initialize: access(/nix/store/fa3jr2wxgbiyk37h0imzhhj13c3p6gqv-yabai-7.1.10/bin/yabai, X_OK) failed with errno 2 - No such file or directory, error 0x6f - Invalid or missing Program/ProgramArguments
> > > 2025-03-08 14:57:39.712352 <Error> initialization failure: 24D70: xpcproxy + 36768 [1088][E4C3A3C3-0D57-3E4D-8388-880BC7F5F19E]: 0x6f
> > > 2025-03-08 14:57:39.712353 <Notice> Service setup event to handle failure and will not launch until it fires.
> > > 2025-03-08 14:57:39.712354 <Error> Missing executable detected. Job: 'org.nixos.yabai' Executable: '/nix/store/fa3jr2wxgbiyk37h0imzhhj13c3p6gqv-yabai-7.1.10/bin/yabai'
> > > ```
> > 
> > 
> > This is the exact reason `nix-darwin` uses `/bin/wait4path`. You can replace the binary inside of the service with a `/bin/sh` that calls `wait4path` with the yabai executable or try copying it outside of the store
> 
> This is using the nix-darwin service. I do see that it doesn't contain `wait4path` like it used to. Thanks for reminding me about that. I'll see why it's missing now.

Did you find out why `/bin/wait4path` was missing? My guess is you were not using the `script` option but ` ProgramArguments` or ` Program` in the `serviceConfig` option. Btw.  `home-manager` doesn't have a `script` option in it's launchd module if I see this correctly. This currently stops me from migrating there for user launchd services.

modules/system/launchd.nix
Enzime: From my testing, the command @Samasaur1 posted should be the other way around:

```
$ ssh root@target
root# defaults read com.apple.dock
2025-05-12 13:14:57.756 defaults[32347:591299]
Domain com.apple.dock does not exist
root# sudo -u enzime defaults read com.apple.dock
2025-05-12 13:15:16.122 defaults[32350:591473]
Domain com.apple.dock does not exist
root# launchctl asuser $(id -u enzime) defaults read com.apple.dock
2025-05-12 13:15:27.168 defaults[32353:591575]
Domain com.apple.dock does not exist
root# sudo -u enzime -- launchctl asuser $(id -u enzime) defaults read com.apple.dock
Could not switch to audit session 0x18701: 1: Operation not permitted
root# launchctl asuser $(id -u enzime) sudo -u enzime defaults read com.apple.dock
{
    loc = "en_AU:AU";
}
```

modules/system/defaults-write.nix
Enzime: It looks like `launchctl asuser $(id -u ${user}) sudo --user=${user}` will be necessary here

On a new machine:

```
ssh root@machine darwin-rebuild switch
user defaults...
+ sudo --user=enzime -- defaults write com.apple.dock persistent-apps '...'
2025-05-12 13:10:15.274 defaults[32224:588660] Could not write domain com.apple.dock; exiting
```

Enzime:
For debugging issues in this PR without modifying nix-darwin you can add:

```nix
{
  system.activationScripts.preActivation.text = ''
    set -x
  '';
}
```

modules/system/launchd.nix
Samasaur1: I absolutely believe you that it's backwards; I'm not sure I even tested it. I think (if I try to get back in the mindset of me from two months ago) that I was mostly just talking about combining the two in whichever order worked.

kfiz:
> > > > > ```shell
> > > > > 2025-03-08 14:57:39.712350 <Warning> Could not find and/or execute program specified by service: 2: No such file or directory: /nix/store/fa3jr2wxgbiyk37h0imzhhj13c3p6gqv-yabai-7.1.10/bin/yabai
> > > > > 2025-03-08 14:57:39.712351 <Error> Service could not initialize: access(/nix/store/fa3jr2wxgbiyk37h0imzhhj13c3p6gqv-yabai-7.1.10/bin/yabai, X_OK) failed with errno 2 - No such file or directory, error 0x6f - Invalid or missing Program/ProgramArguments
> > > > > 2025-03-08 14:57:39.712352 <Error> initialization failure: 24D70: xpcproxy + 36768 [1088][E4C3A3C3-0D57-3E4D-8388-880BC7F5F19E]: 0x6f
> > > > > 2025-03-08 14:57:39.712353 <Notice> Service setup event to handle failure and will not launch until it fires.
> > > > > 2025-03-08 14:57:39.712354 <Error> Missing executable detected. Job: 'org.nixos.yabai' Executable: '/nix/store/fa3jr2wxgbiyk37h0imzhhj13c3p6gqv-yabai-7.1.10/bin/yabai'
> > > > > ```
> > > > 
> > > > 
> > > > This is the exact reason `nix-darwin` uses `/bin/wait4path`. You can replace the binary inside of the service with a `/bin/sh` that calls `wait4path` with the yabai executable or try copying it outside of the store
> > > 
> > > 
> > > This is using the nix-darwin service. I do see that it doesn't contain `wait4path` like it used to. Thanks for reminding me about that. I'll see why it's missing now.
> > 
> > 
> > @khaneliman Did you find out why `/bin/wait4path` was missing? My guess is you were not using the `script` option ([034c45d](https://github.com/nix-darwin/nix-darwin/commit/034c45dd0cac806b527e64c143020676e1070769)) but ` ProgramArguments` or ` Program` in the `serviceConfig` option. Btw. `home-manager` doesn't have a `script` option in its launchd module if I see this correctly. This currently stops me from migrating there for user launchd services.
> 
> Nah, I don't use any of those options. I just added the `launchd.user.agents.*.serviceConfig.KeepAlive.PathState` to get around the issue.

ok. So there is another way to specify an executable apart from Program/ProgramArguments and/or script? Interesting.

modules/system/launchd.nix
Enzime: I've created a PR downstream in home-manager as they'll need this as well: https://github.com/nix-community/home-manager/pull/7051

modules/system/checks.nix
Enzime: Silenced the linter 👍

modules/system/launchd.nix
Enzime: PR downstream's been merged, thanks @khaneliman 👍

This has been incorporated into this PR now too :)

modules/system/defaults-write.nix
Enzime: Done 👍

modules/users/default.nix
Enzime: Fixed 👍

modules/nix/nix-darwin.nix
```diff
@@ -5,6 +5,7 @@ let
     inherit (config.system) profile;
     inherit (config.environment) systemPath;
     nixPath = lib.optionalString config.nix.enable (lib.concatStringsSep ":" config.nix.nixPath);
+    nixPackage = if config.nix.enable then config.nix.package else null;
```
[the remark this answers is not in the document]
Enzime: Done 👍

emilazy:
Thanks to @Enzime for helping push this over the finish line. I think we should be ready to merge this and branch off 25.05 once a couple critical downstreams (nix-homebrew, nh, maybe others?) merge support for it.

Enzime (approved):
Should be good to go now 👍

khaneliman:
> Thanks to @Enzime for helping push this over the finish line. I think we should be ready to merge this and branch off 25.05 once a couple critical downstreams (nix-homebrew, nh, maybe others?) merge support for it.

My PR to `nh` was merged not too long ago, so we are good there.

emilazy:
Okay let’s do this. Sorry if this breaks your system.

NotAShelf:
nh 4.0+ is in the unstable branch, but it was not backported to stable due to a small regression in Darwin. I'll probably finish the PR that fixes the Darwin regression and does some internal cleanup shortly after 25.05 releases, and make a tag for 4.1.0.

PR #1347, opened 2025-02-21
Control Center: Add missing Options, and Fix some

iamanaws:


Iamanaws:
I just have doubt about the options that change their type, I know there are ways, but here I'm not sure how to add a warning or heads up, to indicate the allowed values had changed

modules/system/defaults/controlcenter.nix
```diff
@@ -1,100 +1,397 @@
...
+        Apple menu > System Preferences > Control Center > Focus
+
+        Options:
+          whenActive  - Show When Active (2)
+          hide        - Don't Show in Menu Bar (8)
+          always      - Always Show in Menu Bar (18)
+      '';
+    };
 
-            Show a Focus control in menu bar. Default is null.
```
bew, line 79: (for all options)
I think we should keep the `Show a Focus control in menu bar` line in the description (or something similar), just having the path to the option in settings doesn't give much information..
It's still not very informative though since it's not clear what a 'focus control' is, but it's better than nothing 🤔

modules/system/defaults/controlcenter.nix
```diff
@@ -1,100 +1,397 @@
...
+        true = 18;
+        false = 24;
+      };
+      description = ''
+        Apple menu > System Preferences > Control Center > Bluetooth
+
+        Show a bluetooth control in menu bar. Default is null.
+
+        18 = Show in Menu Bar
+        24 = Don't Show in Menu Bar
```
bew, line 130:
```suggestion
        true = Show in Menu Bar (18)
        false = Don't Show in Menu Bar (24)
```
Although, for a simple bool I'm not sure it's very useful to describe the option like this 👀 The description is enough for me.

modules/system/defaults/controlcenter.nix
```diff
@@ -1,100 +1,397 @@
...
+    system.defaults.controlcenter.BatteryShowEnergyMode = lib.mkOption {
+      type = lib.types.nullOr lib.types.bool;
+      default = null;
+      description = ''
+        Apple menu > System Preferences > Control Center > Battery
+
+        Show a battery energy mode. Default is null.
+
+        false - "When Active"
+        true  - "Always"
```
bew, line 103: Maybe don't use a boolean for this? an enum `whenActive` / `always` would be clearer.
OR rename the option like `...AlwaysShowBatteryEnergyMode` so the boolean makes sense 🤔

modules/system/defaults/controlcenter.nix
```diff
@@ -1,100 +1,397 @@
...
+        Apple menu > System Preferences > Control Center > AirDrop
 
-            Show a sound control in menu bar . Default is null.
+        Show a AirDrop control in menu bar. Default is null.
 
-            18 = Display icon in menu bar
-            24 = Hide icon in menu bar
-        '';
+        18 = Show in Menu Bar
+        24 = Don't Show in Menu Bar
```
bew, line 63:
```suggestion
        true = Show in Menu Bar
        false = Don't Show in Menu Bar
```
Or remove this section of the description?

modules/system/defaults/controlcenter.nix
```diff
@@ -1,100 +1,397 @@
...
+        true = 18;
+        false = 24;
+      };
+      description = ''
+        Apple menu > System Preferences > Control Center > Wi-Fi
+
+        Show a Wi-Fi control in menu bar. Default is null.
+
+        18 = Show in Menu Bar
+        24 = Don't Show in Menu Bar
```
bew, line 392:
```suggestion
        true = Show in Menu Bar (18)
        false = Don't Show in Menu Bar (24)
```

modules/system/defaults/Spotlight.nix
```diff
@@ -0,0 +1,18 @@
...
+    system.defaults.Spotlight.MenuItemHidden = lib.mkOption {
+      type = lib.types.nullOr lib.types.bool;
+      default = null;
+      description = ''
+        Apple menu > System Preferences > Control Center > Menu Bar Only > Spotlight
+
+        Show Spotlight in menu bar. Default is null.
+
+        false - "Don't Show in Menu Bar"
+        true  - "Show in Menu Bar"
```
bew, line 14:
```suggestion
        false - Don't Show in Menu Bar
        true  - Show in Menu Bar
```

bew (commented):
A few comments on the options ;)

For the type change, we _could_ have backward compatibility with warning by adding `true` & `false` in the `mkXXXXApply` mapping functions like:
```nix
mkEnumApply {
  always = 21,
  hide = 42,
  true = lib.warn ''something something use "always" now'' 21,
  false = lib.warn ''something something use `"hide"` now'' 42,
}
```

modules/system/defaults/controlcenter.nix
Iamanaws: I added descriptions for all options to make it more clear. Also, I leave the Setting paths, and try to make it more clear what it is mapping to.

```nix
description = ''
        Show an Accessibility Shortcuts control in menu bar.

        This option mirrors the setting found in:
          System Preferences > Control Center > AccessibilityShortcuts

        Available settings:
          both           - Show in both the Menu Bar and Control Center (3)
          menuBar        - Show in the Menu Bar only (6)
          controlCenter  - Show in Control Center only (9)
          hide           - Don't Show (12)
'';
```
 
 
 We could move this to the bottom of the description to make it aside, but they were there before, and I think they still have value to know what are you doing.

```nix
description = ''
        Show an Accessibility Shortcuts control in menu bar.
        
        Available settings:
          both           - Show in both the Menu Bar and Control Center (3)
          menuBar        - Show in the Menu Bar only (6)
          controlCenter  - Show in Control Center only (9)
          hide           - Don't Show (12)
        
       This option mirrors the setting found in:
          System Preferences > Control Center > AccessibilityShortcuts
 '';
```

Iamanaws: @bew 

Great, it's done

bew:
Starting to look good I think!

While looking at the test file I noticed something that might be improved:
```nix
  system.defaults.controlcenter.Sound = "always";
  system.defaults.controlcenter.StageManager = true;
  system.defaults.controlcenter.UserSwitcher = "menuBar";
  system.defaults.controlcenter.WiFi = true;
```
Those lines show how inconsistent the values are for seemingly similar options: some are `true`, others are `"always"` or `"menuBar"`..
👉 Could we avoid having multiple values for the same thing? 🤔

Iamanaws:
> Thanks for the changes, looks nice!
> 
> While looking at the test file I noticed something that might be improved:
> 
> ```nix
>   system.defaults.controlcenter.Sound = "always";
>   system.defaults.controlcenter.StageManager = true;
>   system.defaults.controlcenter.UserSwitcher = "menuBar";
>   system.defaults.controlcenter.WiFi = true;
> ```
> 

I updated the test file to use the same values for options that share the same type. 
So now we only apply "both", "always", and true.

Is that what you meant?


> Those lines show how inconsistent the values are for seemingly similar options: some are `true`, others are `"always"` or `"menuBar"`.. 👉 Could we avoid having multiple values for the same thing? 🤔

I leave as bool the ones that are more simple, but I could change them to reflect better the options. I grouped similar stuff and leave just 3 different patterns.

iamanaws:
It's been a few months, but I cleaned it up!

HuaDeity:
There should still be a weather option here.

iamanaws:
Added AirplayReceiverEnabled, TimeMachine, Weather, and TextInputMenuAgent

iamanaws:
@Samasaur1 just a friendly ping to bring this up

PR #1348, opened 2025-02-21
security: pam: use sudo_local rather than activation text

spikespaz:
This changes the `config` for `security.pam.enableSudoTouchIdAuth` to take control of `/etc/pam.d/sudo_local` instead of using `sed` to modify `/etc/pam.d/sudo` in the activation script.

I have tried to ensure that this still removes the line added by `nix-darwin` before this commit, but I'm not sure if this is robust enough.

Please, I welcome feedback if there is a better way.

### DIFF
--- a/modules/security/pam.nix
+++ b/modules/security/pam.nix
@@ -2,42 +2,9 @@
 
 with lib;
 
-let
-  cfg = config.security.pam;
+let cfg = config.security.pam;
 
-  # Implementation Notes
-  #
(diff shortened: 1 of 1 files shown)


emilazy:
This is the same as https://github.com/LnL7/nix-darwin/pull/1344 but without compatibility for macOS < 14, right?

Enzime:
Closing this as a duplicate of #1344

PR #1350, opened 2025-02-25
flake: make lib an option of darwinSystem

rhoriguchi:
With `nixosSystem` you do this, so you can add your own custom functions to lib.
```
let lib = (_: super: { custom = import ./lib.nix { lib = super; }; });
in nixosConfigurations.host1 = lib.nixosSystem { ... }
```

This allows you to pass `lib` to `darwinSystem` so you can use the custom functions in `nix-darwin`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -19,7 +19,7 @@
       evalConfig = import ./eval-config.nix;
 
       darwinSystem = args@{ modules, ... }: self.lib.evalConfig (
-        { inherit (nixpkgs) lib; }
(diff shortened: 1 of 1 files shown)


emilazy:
This should already work if you pass `lib = …` to `darwinSystem`. Though I don’t recommend it; mismatching `lib` and `pkgs` can cause all kinds of headaches. It’s better to use `specialArgs` to inject your own separate library instead.

rhoriguchi:
I don't know how I missed this, I guess I had some other issue as well.

PR #1353, opened 2025-02-26
Revert "Add networking.hosts and .hostFiles from nixos "

booxter:
Reverts LnL7/nix-darwin#939


booxter:
Testing if this solves my issues with localhost resolution for MacOS ssh client.

booxter:
Yes, it does resolve my issues. With the offending patch:

```
$ /usr/bin/ssh -4 localhost -p 31022
ssh: Could not resolve hostname localhost: nodename nor servname provided, or not known
```

With this revert:

```
/usr/bin/ssh localhost -p 31022
The authenticity of host '[localhost]:31022 ([127.0.0.1]:31022)' can't be established.
...
```

emilazy (dismissed):
Seems best to do this for now given the issues people are running into, but cc @Enzime.

booxter:
Oh we may now have to restore the original file that was moved to `/etc/hosts.before-nix-darwin`, or handle the recovery in some form.

emilazy:
Right. You can add `mv /etc/hosts{.before-nix-darwin,}` to an activation script.

booxter:
FYI I may not get to implementing an activation script in the next ~8h so if someone would like to take this over, be my guest. 🙏

Enzime (approved):

booxter:
Hold off; I'm testing it. Will reopen for review when done.

booxter:
How I tested:

1. Removed `/etc/hosts`. (Moved it to `/etc/hosts.bak`.)
2. Ran darwin switch.
3. Confirmed `restoring /etc/hosts...` message in the output.
4. Confirmed that `/etc/hosts` is restored.
5. Confirmed that `/etc/hosts.before-nix-darwin` no longer exists.

booxter:
@emilazy @Enzime this is ready to go. 🙏

emilazy (approved):
Seems like an appropriate quick fix for now.

Issue #294, opened 2021-02-18
users.users.<name>.name broken


roberth:
users/default.nix confuses attribute keys and names.

I've annotated the offending code with types and an inline comment where the error occurs. `<name>` corresponds to `ConfigName` and `name` corresponds to `Name` below)

```
  # isCreated : [ConfigName] -> ConfigName -> Bool
  isCreated = listOfKnown: name: elem name listOfKnown;
  # isDeleted : Map *ConfigName *Config -> Name -> Bool
  isDeleted = attrs: name: ! elem name (mapAttrsToList (n: v: v.name) attrs);

  # gids : Map GIDString Name
  gids = mapAttrsToList (n: toGID) (filterAttrs (n: v: isCreated cfg.knownGroups /* expecting ConfigName but actual is Name */ v.name) cfg.groups);
  # uids : Map UIDString Name
  uids = mapAttrsToList (n: toUID) (filterAttrs (n: v: isCreated cfg.knownUsers /* expecting ConfigName but actual is Name */ v.name) cfg.users);

  # createdGroups : [GroupConfig]
  createdGroups = mapAttrsToList (n: v: cfg.groups."${v}") cfg.gids;
  # createdGroups : [UserConfig]
  createdUsers = mapAttrsToList (n: v: cfg.users."${v}") cfg.uids;
  # deletedGroups : [GroupConfigName]
  deletedGroups = filter (n: isDeleted cfg.groups n) cfg.knownGroups;
  # deletedUsers : [UserConfigName]
  deletedUsers = filter (n: isDeleted cfg.users n) cfg.knownUsers;
```

Enzime:
Fixed in #1355

PR #1355, opened 2025-02-27
fix: use correct username for profile

toodeluna:
This pull request fixes an inconsistency I've found between NixOS and Nix Darwin.

```nix
{
  users.users.myUser = {
    name = "actual_name";
  };
}
```

On NixOS this code will create the folder `/etc/profiles/per-user/actual_name` whereas on nix-darwin it creates `/etc/profiles/per-user/myUser`. This causes problems with certain modules in home-manager for example, which expects the `actual_name` folder to exist.

### DIFF
--- a/modules/users/default.nix
+++ b/modules/users/default.nix
@@ -11,14 +11,14 @@ let
   group = import ./group.nix;
   user = import ./user.nix;
 
-  toGID = v: { "${toString v.gid}" = v.name; };
-  toUID = v: { "${toString v.uid}" = v.name; };
+  toGID = n: v: { "${toString v.gid}" = n; };
+  toUID = n: v: { "${toString v.uid}" = n; };
 
   isCreated = list: name: elem name list;
   isDeleted = attrs: name: ! elem name (mapAttrsToList (n: v: v.name) attrs);
 
-  gids = mapAttrsToList (n: toGID) (filterAttrs (n: v: isCreated cfg.knownGroups v.name) cfg.groups);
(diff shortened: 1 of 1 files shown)


Enzime:
I think this isn’t the only place it is wrong, if you could fix all the places it is wrong that would be great 

See https://github.com/LnL7/nix-darwin/issues/294

toodeluna:
The username should now be correct everywhere. I tested it with my own system flake and it all seems to work as expected now, but let me know if I've missed any other places.

modules/users/default.nix
```diff
@@ -331,7 +331,7 @@ in
     environment.systemPackages = systemShells;
 
     environment.etc = mapAttrs' (name: { packages, ... }: {
-      name = "profiles/per-user/${name}";
+      name = "profiles/per-user/${config.users.users.${name}.name or name}";
```
Enzime, line 334: What's the reasoning behind having the `or`?

toodeluna: Not sure honestly 😅 
I must've thought it was necessary in case the `name` attribute wasn't explicitly set, but it seems it will take the value of the attribute name by default so this isn't necessary. I'll change it to just use the `name` attribute.

toodeluna:
Did some more testing for this and there's some cases where it's still not working entirely, so I'm currently working on fixing those issues.

toonn (approved):
LGTM

Enzime (approved):
LGTM

Thanks for the contribution

PR #1356, opened 2025-02-28
homebrew: toggle analytics (off by default)

hraban:
Is there an easy way to add "test groups" or somehow modularize tests? I wanted to add another test where the setting is explicitly toggled off but it would become a bit messy to have three separate homebrew tests top level

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -136,7 +136,8 @@ let
 
     config = {
       brewBundleCmd = concatStringsSep " " (
-        optional (!config.autoUpdate) "HOMEBREW_NO_AUTO_UPDATE=1"
+        optional (cfg.global.noAnalytics) "HOMEBREW_NO_ANALYTICS=1"
+        ++ optional (!config.autoUpdate) "HOMEBREW_NO_AUTO_UPDATE=1"
         ++ [ "brew bundle --file='${brewfileFile}'" ]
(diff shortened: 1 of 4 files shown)


hraban:
@Enzime could I bother you for a review?

malob:
Thanks for adding this. A couple notes:
* To keep with the convention of this module, could you change name to be `analytics` as opposed to `noAnalytics`, and invert the behavior/default accordingly.
* Could you also update the description to be consistent with the other options, including adding an "implementation note".

hraban:
@malob ping for re-review :)

modules/homebrew.nix
```diff
@@ -218,14 +219,27 @@ let
...
+
+        See "https://docs.brew.sh/Analytics".  Setting this to `false` (default)
+        will turn the analytics off.  Setting this to `true` will turn them on.
+
+        Implementation note: when disabled, this option sets the
+        `HOMEBREW_NO_ANALYTICS` environment variable, by adding it to
+        [](#opt-environment.variables).
+
+      '';
+
```
Samasaur1, line 233: To match the other global options, can you change this to `mkOption` and set `type`, `default`, and `description` manually?

Samasaur1 (changes requested):
One minor nit; otherwise this looks good

PR #1362, opened 2025-03-04
homebrew: remove `--no-lock` flag

stepbrobd:
`darwin-rebuild` failed due to `Error: invalid option: --no-lock`

`--no-lock` flag removed in https://github.com/Homebrew/homebrew-bundle/pull/1630

PR #1364, opened 2025-03-04
--no-lock no longer in homebrew

gachikuku:
When rebuilding system an error is encountered:

`Error: invalid option: --no-lock`


gachikuku:
already someone spotted it! https://github.com/LnL7/nix-darwin/pull/1362

Issue #1349, opened 2025-02-25
The Determinate installer has "Determinate" now as the default in their installer!


oderwat:
We just had to search for the problem why a fresh nix-darwin did not work for one of our team members. It came down to that the nix installer from determinate now has "Determinate" as the default when installing Nix without the `--determinate` flag. I think you need to update your instructions.

oderwat:
Not exactly. They ask you if you want to use the Determinate option and the default is "yes". It was "no" before. The instructions for nix-darwin says, to use the defaults (and our internal manual too). I think we need to tell the user that they need to watch out for that option and not to choose the default. I did not test if there would be a "--no-determinate" like option for the command.

DavSanchez:
I see, thanks!

isker:
> I did not test if there would be a "--no-determinate" like option for the command.

There is not 😬.

adamcstephens:
This seems like a report that should be made to nix-installer. I don't think nix-darwin can do much about the changes.

oderwat:
> This seems like a report that should be made to nix-installer. I don't think nix-darwin can do much about the changes.

I posted this here, so that the installation instructions can get updated. Nobody (at least not me) expect that nix-darwin fixes something that is a pre-requisite for the installation of it.

Enzime:
https://github.com/DeterminateSystems/nix-installer?tab=readme-ov-file#determinate-nix-installer

Maybe their README is out of date as it still says the default is without `--determinate`

Samasaur1:
The source code of both [v0.37.2](https://github.com/DeterminateSystems/nix-installer/blob/9e27e381138df7f574917faf08a4be3eed2d4b11/src/settings.rs#L62-L71) and on [main](https://github.com/DeterminateSystems/nix-installer/blob/main/src/settings.rs#L62-L71) both show that it defaults to not installing Determinate, but I did find where it *does* prompt:

https://github.com/DeterminateSystems/nix-installer/blob/bedb14209c34988581785f432b49aa9ea1f5821a/src/cli/subcommand/install/determinate.rs#L31

[...]

PR #1367, opened 2025-03-05
readme: update instructions as Determinate Nix is now the default

Enzime:
Closes https://github.com/LnL7/nix-darwin/issues/1349


emilazy (approved):
This keeps coming up so agreed that we need to get a readme update pushed out ASAP to reduce user confusion rather than waiting on upstream to add flags or the like; hopefully we can refine later.

aspauldingcode:
> This keeps coming up so agreed that we need to get a readme update pushed out ASAP to reduce user confusion rather than waiting on upstream to add flags or the like; hopefully we can refine later.

The way it is worded brought more confusion.

kantum:
Why am I still reading this today? Just guess

Issue #1360, opened 2025-03-04
New nix-darwin install - "error: file 'darwin' was not found in the Nix search path (add it using $NIX_PATH or -I)"


victorhooi:
I'm trying to setup Nix on a new macOS machine (running macOS Sequoia 15.3.1), and I'm hitting some PATH errors trying to get nix-darwin to install. (I'm a Nix newbie - previously I was using the starter templates from [here](https://github.com/dustinlyons/nixos-config), but I figured I wanted to try some of it myself).

Firstly, I installed Nix using the [Determinate Nix installer](https://github.com/DeterminateSystems/nix-installer):
```
victorhooi@MacBook-Pro ~ % curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | \
  sh -s -- install
info: downloading installer https://install.determinate.systems/nix/tag/v0.38.1/nix-installer-aarch64-darwin
 INFO nix-installer v0.38.1
`nix-installer` needs to run as `root`, attempting to escalate now via `sudo`...
 INFO nix-installer v0.38.1
 INFO For a more robust Nix installation, use the Determinate package for macOS: https://dtr.mn/determinate-nix

Install Determinate Nix?

Proceed? ([Y]es/[n]o/[e]xplain): yes
Nix install plan (v0.38.1)
Planner: macos

Configured settings:
* determinate_nix: true

Planned actions:
* Install Determinate Nixd
* Create an encrypted APFS volume `Nix Store` for Nix on `disk3` and add it to `/etc/fstab` mounting on `/nix`
* Extract the bundled Nix (originally from /nix/store/1q9hil0gipfxndz3pq6jg4ggmn0syia8-nix-binary-tarball-2.26.2/nix-2.26.2-aarch64-darwin.tar.xz)
* Create a directory tree in `/nix`
* Synchronize /nix/var ownership
* Move the downloaded Nix into `/nix`
* Synchronize /nix/store ownership
* Create build users (UID 351-382) and group (GID 350)
* Configure Time Machine exclusions
* Setup the default Nix profile
* Place the Nix configuration in `/etc/nix/nix.conf`
* Configure the shell profiles
* Configuring zsh to support using Nix in non-interactive shells
* Create a `launchctl` plist to put Nix into your PATH
* Configure the Determinate Nix daemon
* Remove directory `/nix/temp-install-dir`


Proceed? ([Y]es/[n]o/[e]xplain): yes
[...]

PR #1368, opened 2025-03-05
readme: recommend `/etc/nix-darwin` more strongly

Enzime:
See https://github.com/LnL7/nix-darwin/issues/1360#issuecomment-2696316339

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ As the official Nix installer does not include an automated uninstaller, and man
 
 Despite being an experimental feature in Nix currently, nix-darwin recommends that beginners use flakes to manage their nix-darwin configurations.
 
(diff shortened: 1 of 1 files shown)


TechAssistance (commented):
lgtm - would recommend adding something along the lines of "you may use the `--flake $PATH-TO-NIX-CONFIG` as well to use locations such as ~/.config/nix/

emilazy:
We could more explicitly talk about linking a directory in your home to `/etc/nix-darwin`, which lets people have their config in `~/` while still using `darwin-rebuild switch` without additional flags. Our installation instructions do this but only as an unexplained command.

higuoxing:
I'm encountering the similar issue and this PR solved my problem.

PR #1374, opened 2025-03-06
Docs: Fix wrong reference to systemd

dwt:
Darwin has LaunchD.

### DIFF
--- a/modules/services/autossh.nix
+++ b/modules/services/autossh.nix
@@ -55,7 +55,7 @@ in
 
         default = [];
         description = ''
-          List of AutoSSH sessions to start as systemd services. Each service is
+          List of AutoSSH sessions to start as launchd daemon. Each service is
           named 'autossh-{session.name}'.
         '';
 


modules/services/autossh.nix
```diff
@@ -55,7 +55,7 @@ in
 
         default = [];
         description = ''
-          List of AutoSSH sessions to start as systemd services. Each service is
+          List of AutoSSH sessions to start as launchd daemon. Each service is
```
emilazy, line 58:
```suggestion
          List of AutoSSH sessions to start as launchd daemons. Each service is
```

emilazy (changes requested):

dwt:
@emilazy I did apply your suggestion - does that resolve your concerns?

emilazy:
Looks good, thanks! Could you squash the commits into one e.g. `autossh: fix incorrect reference to systemd`?

The next “service” should probably be changed to “daemon” too. (Sorry for missing that…)

dwt:
@emilazy I took the liberty to also update the formatting to nixpkgs current standard. Hope that's ok.

emilazy:
Thanks! We’re waiting on https://github.com/NixOS/nixpkgs/pull/380990 to do a treewide reformat as the commits need to go in `.git-blame-ignore-revs` to not mess up `git blame` and keeping in line with NixOS formatting makes backports easier. Would you mind dropping the reformatting commit for now? We’ll be doing a mass reformat around the time we move to the new repo.

dwt:
@emilazy Here you go

dwt:
When you're on in to change tall the things, maybe even renaming `master` to `main` could be on the table…

emilazy (approved):
Thanks!

I’ve actually thought that we might want to rename the main branch to `nix-darwin-unstable`, to extend the `nix-darwin-24.11` ↔ `nixpkgs-24.11-darwin` release branch correspondence we have. Not sure if that would be too weird though…

PR #1375, opened 2025-03-06
Docs: Fix references to NixOS

dwt:
This option applies to nix-darwin.


emilazy (approved):
Thanks. A lot of these old modules were copied straight from NixOS :)

PR #1377, opened 2025-03-08
Fix merging of system.defaults.CustomUserPreferences

zhaofengli:
Without this, custom user preferences defined in different modules for the same domain aren't merged.

For example, the following config results in only `UniversalSearchEnabled` defined:

```
{
  imports = [
    {
      system.defaults.CustomUserPreferences = {
        "com.apple.Safari"."NSUserKeyEquivalents"."Quit Safari" = "@~q";
      };
    }
    {
      system.defaults.CustomUserPreferences = {
        "com.apple.Safari"."UniversalSearchEnabled" = false;
      };
    }
  ];
}
```

### DIFF
--- a/modules/system/defaults/CustomPreferences.nix
+++ b/modules/system/defaults/CustomPreferences.nix
@@ -2,10 +2,25 @@
 
 with lib;
 
-{
+let
+  valueType = with lib.types; nullOr (oneOf [
+    bool
+    int
+    float
+    str
+    path
+    (attrsOf valueType)
+    (listOf valueType)
+  ]) // {
+    description = "plist value";
+  };
+  defaultsType = types.submodule {
+    freeformType = valueType;
+  };
+in {
   options = {
     system.defaults.CustomUserPreferences = mkOption {
-      type = types.attrs;
+      type = defaultsType;
       default = { };
       example = {
         "NSGlobalDomain" = { "TISRomanSwitchState" = 1; };
@@ -20,7 +35,7 @@ with lib;
     };
 
     system.defaults.CustomSystemPreferences = mkOption {
-      type = types.attrs;
+      type = defaultsType;
       default = { };
       example = {
         "NSGlobalDomain" = { "TISRomanSwitchState" = 1; };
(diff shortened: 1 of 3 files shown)


modules/system/defaults/CustomPreferences.nix
```diff
@@ -5,7 +5,7 @@ with lib;
 {
   options = {
     system.defaults.CustomUserPreferences = mkOption {
-      type = types.attrs;
+      type = types.attrsOf types.anything;
```
Enzime, line 8: Would freeform settings work here as well?

zhaofengli: Seems like a better option. Let's use that.

modules/system/defaults/CustomPreferences.nix
```diff
@@ -2,10 +2,14 @@
 
 with lib;
 
-{
+let
+  defaultsType = types.submodule {
+    freeformType = types.anything;
```
Enzime, line 7: Is there a better type than `types.anything` here? Maybe something related to `plist`?

tests/system-defaults-write.nix
```diff
@@ -1,6 +1,25 @@
 { config, pkgs, lib, ... }:
 
 {
+  imports = [
```
Enzime, line 4: Does this test fail without the `type` change above?

modules/system/defaults/CustomPreferences.nix
zhaofengli: I added a stricter type check that we could add to nixpkgs as `lib.formats.plist` later.

tests/system-defaults-write.nix
zhaofengli: Yes, and only the `WebKit2DeveloperExtrasEnabled` key shows up in the final config. I pushed only the test changes [to a separate branch](https://github.com/zhaofengli/nix-darwin/actions/runs/13862641673/job/38794597899).

Enzime (approved):
LGTM

Thanks for the contribution 🎉

PR #1379, opened 2025-03-10
nix-daemon: use PathState to detect /nix/store presence

booxter:
This allows to avoid granting Full Disc Access to /bin/{sh,wait4path}.

This is re-inventing the other PR posted for `nix` here:
https://github.com/NixOS/nix/pull/5698

Note: one has to grant Full Disc Access to nix-daemon for every new
derivation. In the future, we may want to consider copying `nix-daemon`
to a well-known location, which will allow the user to grant permissions
once only.

Signed-off-by: Ihar Hrachyshka <ihar.hrachyshka@gmail.com>


booxter:
Just hit a nix-daemon not started after reboot with this PR on my machine. Means this approach here won't work.

PR #1386, opened 2025-03-16
Back out "github-runner: replace `mkdir -p -m` with `umask`"

Enzime:
This backs out commit 3b738c765de1bb4ecc4993fa092b27dd46d495ed.

Setting a `umask` made the parent directory have too conservative of permissions making it so `_github-runner` couldn't access the child directories.


emilazy (approved):
I’ll trust you :)

PR #1396, opened 2025-03-22
Application 'linking' done right

### LINKED ISSUE #214: How to add applications so they show up in Applications (or Launchpad search)
It seems if I add applications in the user configuration file, they end up in ~/Applications rather than in the system Applications. I there a way to add packages to the system so the symlinks will show up in the root Applications, or at least will be searchable with Launchpad?

I couldn't find any "Getting Started" documentation with nix-darwin that would answer this question, but I'm very interested in getting more hands-on experience with the nix ecosystem

dwt:
This is a draft, because I do not yet have an option to actually test these changes, and I don't want switch my personal machine to nix-darwin@master.

I hope however, that this can be a start to get the ball rolling on a better story to install GUI applications with nix-darwin.

# What I did and considered

All existing attempts fell short.
So we fall back to plain old copying applications over.

Problems with alternatives:

- Symlinking: Spotlight doesn't index symlinks. Therefore one cannot use
Spotlight to find or open the apps. Also they don't show up in
LaunchPad.

- Trampolines: Apples Security & Privacy doesn't get the concept and
shows them with the wrong name. Having an app open during an update will
also make it show up twice in the Dock.

- Aliasses: Require either AppleScript (a permission we don't want to
have, as it easily bypasses Apples TCC) or extra tools (that would be
ok), but also Aliasses are not categorized as 'Application' by
SpotLight.

- Directory Hardlinks are not supported by APFS - but also wouldn't work
cross volume.

- clonefile also doesn't work cross-volume.

Which all leads us back to Don Copine and Pastone. *sigh*

Fixes #214

### DIFF
--- a/modules/system/applications.nix
+++ b/modules/system/applications.nix
@@ -33,12 +33,33 @@
         fi
       ''}
 
-      if [ ! -e '/Applications/Nix Apps' ] \
-         || ourLink '/Applications/Nix Apps'; then
-        ln -sfn ${config.system.build.applications}/Applications '/Applications/Nix Apps'
-      else
-        echo "warning: /Applications/Nix Apps is not owned by nix-darwin, skipping App linking..." >&2
+      targetFolder='/Applications/Nix Apps'
+
+      # Clean up old style symlink to nix store
+      if [ -e "$targetFolder" ] && ourLink "$targetFolder"; then
+        rm "$targetFolder"
       fi
+
+      mkdir -p "$targetFolder"
+
+      rsyncFlags=(
+        # mtime is standardized in the nix store, which would leave only file size to distinguish files.
+        # Thus we need checksums, despite the speed penalty.
+        --checksum
+        # Converts all symlinks pointing outside of the copied tree (thus unsafe) into real files and directories.
+        # This neatly converts all the symlinks pointing to application bundles in the nix store into
+        # real directories, without breaking any relative symlinks inside of application bundles.
+        # This is good enough, because the make-symlinks-relative.sh setup hook converts all $out internal
+        # symlinks to relative ones.
+        --copy-unsafe-links
+        --archive
+        --delete
+        --chmod=-w
+        --no-group
+        --no-owner
+      )
+
+      ${lib.getExe pkgs.rsync} "''${rsyncFlags[@]}" ${config.system.build.applications}/Applications/ "$targetFolder"
     '';
   };
 }
--- a/modules/system/applications.nix
+++ b/modules/system/applications.nix
@@ -1,19 +1,12 @@
-{ config, lib, pkgs, ... }:
-
-with lib;
-
-let
-
-  cfg = config.system;
-
-in
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
 
 {
-  options = {
-  };
-
   config = {
-
     system.build.applications = pkgs.buildEnv {
       name = "system-applications";
       paths = config.environment.systemPackages;
@@ -42,11 +35,10 @@ in
 
       if [ ! -e '/Applications/Nix Apps' ] \
          || ourLink '/Applications/Nix Apps'; then
-        ln -sfn ${cfg.build.applications}/Applications '/Applications/Nix Apps'
+        ln -sfn ${config.system.build.applications}/Applications '/Applications/Nix Apps'
       else
         echo "warning: /Applications/Nix Apps is not owned by nix-darwin, skipping App linking..." >&2
       fi
     '';
-
   };
 }
--- a/modules/system/applications.nix
+++ b/modules/system/applications.nix
@@ -7,6 +7,55 @@
 
 {
   config = {
+    system.checks.text = lib.mkAfter ''
+      ensureAppManagement() {
+        for appBundle in /Applications/Nix\ Apps/*.app; do
+          if [[ -d "$appBundle" ]]; then
+            if ! touch "$appBundle/.DS_Store" &> /dev/null; then
+              return 1
+            fi
+          fi
+        done
+
+        return 0
+      }
+
+      if ! ensureAppManagement; then
+        if [[ "$(launchctl managername)" != Aqua ]]; then
+          # It is possible to grant the App Management permission to `sshd-keygen-wrapper`, however
+          # there are many pitfalls like requiring the primary user to grant the permission and to
+          # be logged in when `darwin-rebuild` is run over SSH and it will still fail sometimes...
+          printf >&2 '\e[1;31merror: permission denied when trying to update apps over SSH, aborting activation\e[0m\n'
+          printf >&2 'Apps could not be updated as `darwin-rebuild` requires Full Disk Access to work over SSH.\n'
+          printf >&2 'You can either:\n'
+          printf >&2 '\n'
+          printf >&2 '  grant Full Disk Access to all programs run over SSH\n'
+          printf >&2 '\n'
+          printf >&2 'or\n'
+          printf >&2 '\n'
+          printf >&2 '  run `darwin-rebuild` in a graphical session.\n'
+          printf >&2 '\n'
+          printf >&2 'The option "Allow full disk access for remote users" can be found by\n'
+          printf >&2 'navigating to System Settings > General > Sharing > Remote Login\n'
+          printf >&2 'and then pressing on the i icon next to the switch.\n'
+          exit 1
+        else
+          # The TCC service required to modify notarised app bundles is `kTCCServiceSystemPolicyAppBundles`
+          # and we can reset it to ensure the user gets another prompt
+          tccutil reset SystemPolicyAppBundles > /dev/null
+
+          if ! ensureAppManagement; then
+            printf >&2 '\e[1;31merror: permission denied when trying to update apps, aborting activation\e[0m\n'
+            printf >&2 '`darwin-rebuild` requires permission to update your apps, please accept the notification\n'
+            printf >&2 'and grant the permission for your terminal emulator in System Settings.\n'
+            printf >&2 '\n'
+            printf >&2 'If you did not get a notification, you can navigate to System Settings > Privacy & Security > App Management.\n'
+            exit 1
+          fi
+        fi
+      fi
+    '';
+
     system.build.applications = pkgs.buildEnv {
       name = "system-applications";
       paths = config.environment.systemPackages;


reckenrode:
If it’s any help, [this][1] is the module I’ve been using for the last few years.

[1]: https://github.com/reckenrode/nixos-configs/blob/main/modules/by-name/co/copy-apps/module.nix

modules/system/applications.nix
```diff
@@ -37,11 +42,30 @@ in
...
+      if [ -e "$targetFolder" ] && ! ourFolder "$targetFolder"; then
+        echo "warning: $targetFolder is not owned by nix-darwin, skipping App linking..." >&2
       else
-        echo "warning: /Applications/Nix Apps is not owned by nix-darwin, skipping App linking..." >&2
+        # create and mark folder
+        mkdir -p "$targetFolder"
+        xattr -w org.nix-darwin generated "$targetFolder"
+
+        # sync applications over resolving the first symlinks, but retaining all internal ones
+        find ${cfg.build.applications}/Applications -maxdepth 1 -type l -exec rsync -avL --delete {} "$targetFolder" \;
```
reckenrode, line 66: `rsync` needs to use `--checksum`. Otherwise, it only compares based on mtime and size. The former will (or should) always be zero in the store, which means it effectively only compares by size. If a new file has the same size as the old one, the old one will be left in place, which can cause problems.

modules/system/applications.nix
```diff
@@ -37,11 +42,30 @@ in
...
-      if [ ! -e '/Applications/Nix Apps' ] \
-         || ourLink '/Applications/Nix Apps'; then
-        ln -sfn ${cfg.build.applications}/Applications '/Applications/Nix Apps'
+      ourfolder () {
+        local marker
+        marker=$(xattr -p org.nix-darwin "$1" 2>/dev/null)
+        [ "$marker" = 'generated' ]
+      }
+
+      targetFolder='/Applications/Nix Apps'
```
reckenrode, line 51: This is more of a future direction, but it would be nice if the `targetFolder` could be customized on a per-app basis. That would allow 1Password to work via nix-darwin. It currently expects to be in `/Applications` and doesn’t work anywhere else. (And if it could be used by Home Manager ….)

dwt:
> If it’s any help, [this](https://github.com/reckenrode/nixos-configs/blob/main/modules/by-name/co/copy-apps/module.nix) is the module I’ve been using for the last few years.

Currently reading your module, looks really good. Any reason this could not be upstreamed? About the only thing that I keep thinking about is that your approach probably will convert all symlinks used inside of application packages into real files? Not quite sure about that anymore after researching `--copy-unsafe-links`.

reckenrode:
I wonder if https://github.com/LnL7/nix-darwin/pull/1205 might be useful, particularly the copying part.

modules/system/applications.nix
dwt: Yeah, I want that too, as I have a fairly elaborate hierarchy in ~/Applications where I sort my apps by topic. Otherwise I wouldn't be able to find them again.

To keep this pull request focussed though, I would really like to push that out into future work.

reckenrode:
> Currently reading your module, looks really good. Any reason this could not be upstreamed?

The only reason I never tried to upstream it is that there was no consensus or agreement about what to do. Currently, I don’t have the bandwidth for development, so if someone is interested in working on this, they’re welcome to take whatever is useful from my module. The license is MIT, which is what nix-darwin uses.

> About the only thing that I keep thinking about is that your approach probably will convert all symlinks used inside of application packages into real files? Not quite sure about that anymore after researching `--copy-unsafe-links`.

My approach preserves internal symlinks while copying those files that point outside the bundle. They’re not common, but they can happen. In particular, [`_makeSymlinksRelative`hook][1] turns all symlinks in an output into relative symlinks. If a bundle links to a file outside of the bundle but in the same output, it will be turned into a relative reference, which will break when the bundle is copied to another location on the filesystem.

[1]: https://github.com/NixOS/nixpkgs/blob/674a159bc2df935bd6dfe1ce56092f4db1e33157/pkgs/build-support/setup-hooks/make-symlinks-relative.sh

dwt:
@reckenrode I have adapted the code from your module, as that is simpler and faster. Did I miss anything?

reckenrode:
> I have adapted the code from your module, as that is simpler and faster. Did I miss anything?

Doesn’t look like you did. 👍🏻

dwt:
Ok, finally got my VM ready to test this, and of course I missed a slash in the `rsync` call… 🤦

modules/system/applications.nix
emilazy: We should ideally look at the store path the app is from to avoid the expense of recomputing the hash on activation.

@Samasaur1 Will your files module work make this kind of thing easier? If so it might be better to land that before this PR so it can utilize the logic.

reckenrode: How would we know the store path of the app if it’s been copied out of the store? It’s probably not worth the extra complexity of trying to track it unless we use something like @Samasaur1’s PR. One issue is we won’t know what files to copy until activation. Does his PR have activation helpers this one could use?

dwt: @emilazy This diff is slightly outdated, I since understood from @reckenrode that we just need one rsync to achieve what we want. Maybe I misunderstood you, but that should mean that Applications change only when they actually change on disk. That _should_ prevent hash recompilation I think?

emilazy: > How would we know the store path of the app if it’s been copied out of the store?

Well, we’re already writing `xattr`s, right? But I agree that it might make sense to land this without and port it to the file mechanism later. I just worry about the speed of checksumming, since Mac apps can be pretty big.

> @emilazy This diff is slightly outdated, I since understood from @reckenrode that we just need one rsync to achieve what we want. Maybe I misunderstood you, but that should mean that Applications change only when they actually change on disk. That _should_ prevent hash recompilation I think?

I think that currently, `rsync` will always hash every single file in `${cfg.build.applications}/Applications` and `$targetFolder` to determine whether any of them changed.

dwt: That is correct. The filesystem could provide an optimization for this, as apfs does know if anything in a subtree has changed. I do not know wether `rsync` has any way to take advantage of that though.

modules/system/applications.nix
```diff
@@ -37,11 +38,40 @@ in
         rm ~/Applications/'Nix Apps'
       fi
 
-      if [ ! -e '/Applications/Nix Apps' ] \
-         || ourLink '/Applications/Nix Apps'; then
-        ln -sfn ${cfg.build.applications}/Applications '/Applications/Nix Apps'
+      ourfolder () {
```
emilazy, line 41: This is used as `ourFolder` later on, so this would break activation at present.

modules/system/applications.nix
```diff
@@ -37,11 +38,40 @@ in
...
+        local marker
+        marker=$(xattr -p org.nix-darwin "$1" 2>/dev/null)
```
emilazy, line 43: `xattr -p org.nix-darwin '/Applications/Nix Apps'` returns 1 for me, but I guess that’s fine, since we want the function to fail in that case anyway.

(FWIW I don’t know if the complexity we do around detecting ownership is really worth it here; we don’t do it for fonts and probably nobody expects a directory called “Nix Apps” to be owned by anything other than nix-darwin? But I don’t mind keeping it for now, since probably the files interface will change how this works anyway.)

modules/system/applications.nix
```diff
@@ -37,11 +38,40 @@ in
...
+        [ "$marker" = 'generated' ]
+      }
+
+      targetFolder='/Applications/Nix Apps'
+
+      # Clean up for links created at the old location in /Applications
```
emilazy, line 49: nit: Not the old location, right?

modules/system/applications.nix
```diff
@@ -37,11 +38,40 @@ in
...
+          --chmod=-w
+          --copy-unsafe-links
+          --delete
+          --no-group
+          --no-owner
+        )
+
+        # sync applications
+        ${lib.getBin pkgs.rsync}/bin/rsync \
+            -v "''${rsyncFlags[@]}" \
```
emilazy, line 73: Isn’t `-v` going to print a line for every file in every app? That seems way too noisy.

emilazy (changes requested):
LGTM in principle; some blocking errors right now, I think we should really get a little real‐world testing of this before merging. Could probably use a `CHANGELOG` entry since it seems likely it’ll break someone’s workflow, even though I think it’ll be an improvement overall.

If @reckenrode has been using this approach for years and doesn’t find the checksum performance to be a bother, then I’m willing to take that on faith.

modules/system/applications.nix
dwt: I was thinking about getting rid of it. I mainly did it to stay in the spirit of the code above. I would be glad to get rid of this check.

modules/system/applications.nix
dwt: Hm, that could be clearer. What do you think of 'Clean up old style symlink to nix store'

modules/system/applications.nix
dwt: 🤦✅ fixed

dwt: That actually did not kill the evaluation but instead just created output.

modules/system/applications.nix
```diff
@@ -37,11 +38,40 @@ in
...
+          --checksum
+          --chmod=-w
+          --copy-unsafe-links
+          --delete
+          --no-group
+          --no-owner
+        )
+
+        # sync applications
+        ${lib.getBin pkgs.rsync}/bin/rsync \
```
afh, line 72: Could `lib.getExe` be used here?
```suggestion
        ${lib.getExe pkgs.rsync} \
```

modules/system/applications.nix
```diff
@@ -37,11 +38,40 @@ in
...
+
+        rsyncFlags=(
+          --archive
+          --checksum
+          --chmod=-w
+          --copy-unsafe-links
+          --delete
+          --no-group
+          --no-owner
+        )
```
afh, line 69: What's the benefit of using an array for options here instead of simply adding the options one per line with a line continuation below in the actual `rsync` command?

afh (commented):
Great initiative, @dwt 👏 

Please find below a few nits

mathjiajia:
A small question here: is it possible to include those GUI apps specified via home-manager?
for example, kitty (but using home-manager).

reckenrode:
Home Manager would have to accept a PR duplicating this logic. My impression of https://github.com/nix-community/home-manager/issues/1341 is that there’s not a consensus on what to do, especially not on copying.

modules/system/applications.nix
reckenrode: That originated from my module, so I’ll explain why I did it: I generally don’t like splitting commands over multiple lines. I’d rather declare an array with the flags and keep the command itself on one line. Arrays also work better with comments (e.g., should why it’s using `--copy-unsafe-links` be documented?).

modules/system/applications.nix
dwt: In my tests, this takes around 2.5 Seconds of wall time to copy around 800mb of data (ut1999 for me) initially and around 1.3s when source and target are identical. That is not nothing, but also not too bad.

modules/system/applications.nix
dwt: Good point, I just didn't think of it. ✅ Changed

modules/system/applications.nix
dwt: I did remove it, though not the existing code that checks for the symlink. What do you think @emilazy Should that also go?

modules/system/applications.nix
afh: Thanks for the explanation, @reckenrode, much appreciated. Getting more context from comments on the used flags is very, very helpful; thanks @dwt on adding those 💯

Samasaur1:
re: the questions about my `system.file` PR (#1205):

The PR as is only supports linking files (and directories). I'm planning future PRs to support copying files, removing files, known hashes, etc., but they aren't part of the current PR and are at various levels of implementation. So I'd say we shouldn't hold up this PR on mine — we can always swap over the implementation later.

> One issue is we won’t know what files to copy until activation. Does his PR have activation helpers this one could use?

@reckenrode No, my PR sticks a JSON file in the system at build time, so you need to know what files are to be copied at build time as well. Although I don't see why we won't know what files to copy until activation — can't we just continue reusing `system.build.applications` and copy that whole directory? The linker/copier script will take care of intelligently only copying modified files. (is using `system.build.applications` technically IFD because of the `buildEnv` step?)

> Will your files module work make this kind of thing easier? If so it might be better to land that before this PR so it can utilize the logic.

@emilazy Yeah, my PR should make this easier — the goal in my head is to completely (or as best we can) eliminate this kind of problem. You should just be able to reach for `system.file` and it should link/copy and track automatically. Although as I said above we'd need the currently open PR and at least one follow-up PR, so it may be better not to wait

dwt:
Friendly ping at @emilazy - is there anything I can do to move this forward?

mathjiajia:
seems each time when `darwin-rebuild` it will copy paste the applications to `Nix` folder?

dwt:
@mathjiajia Due to the use of rsync, this should already be the case. In a followup pull request, we could make this even better by using batter to save source paths and only look at those that are not already up to date. That could perhaps be integrated with the `files` work currently underway, thus we didn't want to do it here.

dwt:
Friendly ping at @emilazy & @Enzime  - is there anything I can do to move this forward?

As far as I can tell, multiple well known community members have used code very similar to this for years now - for example @reckenrode and @Samasaur1. In my tests the performance tradeoff was ok, and will hopefully be further improved by @Samasaur1's work later on.

christian-burger:
Over at [hraban/mac-app-util](https://github.com/hraban/mac-app-util) a [new branch `link-contents` was developed](https://github.com/hraban/mac-app-util/tree/link-contents) that seems to be a viable solution for "application linking" without copying the folder of each app. Trampolines aren't used any more in this branch … Maybe before merging your "copy and paste" approach, you could test this?

The basic idea is to just symlink `Contents` and is [discussed in this issue here](https://github.com/hraban/mac-app-util/issues/27).

Apps linked this way
1. would be indexed by Spotlight.
2. Could be put into the dock.
3. There is no duplicate symbol of the "trampoline" — the trampoline basically is gone.
4. And if the app is open during an update with `darwin-rebuild`, you can close the app after the update and just click on the symbol in the dock again and the new version opens — as one would expect.
5. Even works with parameters like `open -nb "com.jetbrains.pycharm.ce" --args "/Users/user/PycharmProjects/PythonProject"`.
6. Works with context menu "Open with …" as well.
7. Doesn't doubles the space requirement (because of copy and paste) for the app.
8. And performance impact during post-update phase should be low.

Sorry, for being late to the party, but this new branch is just a week old. It looks good to me and I'm successfully using it for ten different apps — amongst them "kitty", "PyCharm Community" and "Visual Studio Code". I tried "MacVim" (mentioned in this PR), but it would not build on my system. Since so far only three people have tested it. Maybe a few more people could evaluate the viability of this approach for solving the "application linking" problem? And if it works, maybe this could be considered to be used instead of "copy and paste"? 🤷

PS: The only problem I encountered so far, is that `sbcl` is used in `mac-apps-util` and this is problematic on MacBooks with OCLP — I found a work-around, though. But `sbcl` is no requirement and could be easily replaced, if this is implemented in nix-darwin.

reckenrode:
I thought we tried that in the big discussion at https://github.com/nix-community/home-manager/issues/1341. I posted some experiences with various apps I use at https://github.com/nix-community/home-manager/issues/1341#issuecomment-882215413, and many of them were variously broken, but I think I created the `Contents` folder instead of symlinking it. Does symlinking it really work in most/all cases?

I also assume that apps with symlinked `Contents` still need to be copied into `/Applications/Nix Apps`. Are there still issues with scripts referring to the store path in that case?

hraban:
I've been test driving the link-contents branch for a while and it's hard to get a clear "go/nogo" signal but it looks ok so far. Had one issue that seemed solved by a reboot.

I followed that original thread closely, in fact I think that thread (or the HM one?) was why I originally developed [`mac-app-util`](https://github.com/hraban/mac-app-util) in the first place, but I don't recall seeing people try to symlink specifically the Contents directory, while creating new top-level .app directories.

For now basically the `link-contents` branch is nothing more than a mkdir for every .app, and a ln -s for every Contents inside of it. I definitely recommend giving it a try.

I guess this PR titled "done right" is also the place to point out something more fundamental: none of this can solve the issue of attestation of Nix apps. E.g. Nix-built Firefox will never be able to connect to the system-wide 1Password because the latter requires attestation of any browser with which it integrates. This is a bummer and a bit of a roadblock to full Nix management of GUI apps on darwin. I _think_ the only way around this is for any nix user to have a developer ID account with apple (obviously not ideal).

Thaks to @christian-burger for the idea btw 🙌

christian-burger:
> I also assume that apps with symlinked `Contents` still need to be copied into `/Applications/Nix Apps`. Are there still issues with scripts referring to the store path in that case?

Symlinking `Contents/` from `Nix Trampolines` is done directly into the Nix store:

```console
christian@chabl7:~ % ls -l /Applications/Nix\ Trampolines/Visual\ Studio\ Code.app 
total 0
lrwxr-xr-x  1 root  admin  103 Apr 13 19:17 Contents -> /nix/store/18i88ihy20334pkbbs0lfqiy0d58xqd0-vscode-1.98.2/Applications/Visual Studio Code.app//Contents
christian@chabl7:~ % 
```

`Nix Apps/` is just used by mac-app-utils to gather the apps, when populating `Nix Trampolines/`. It’s more of a proof of concept in that regard. I’m sure the indirection with collecting the apps from `Nix Apps/` could be solved easily, so that only one folder remains in `Applications/`, namely `Nix Apps/`.

> Does symlinking it really work in most/all cases? 😮

We won't know until more people test it. For the few people that tried, it seemed to work so far. 🤷

christian-burger:
> Well, the list of test cases can follow from the list of problems outlined here: [#1396 (comment)](https://github.com/nix-darwin/nix-darwin/pull/1396#issue-2940375657) - if your solution solves all of these, it's better.

I think it does. [The list of 8 pros before](https://github.com/nix-darwin/nix-darwin/pull/1396#issuecomment-2800067238) has taken the initial description of this PR into consideration. I forgot to mention:

9. LaunchPad works as well.

Though there is one tiny caveat:

For each instance (Nix store entry) of an app there is an entry for in „Privacy & Security“ — this is a screenshot of "Local Network" for demonstration:

Screenshot 2025-04-14 at 19 51 48

I think, it happened when I declined the initial request. I remember distinctly being asked a second time later on. Might be no problem, when the request is accepted. 🤷 Funny thing is, the switches are in sync — if I flip one the other flips, too. And if I right-click and „Open in Finder“ I am lead to the same folder `/Applications/Nix Trampolines/` both times.

If I take a look at `/Library/Preferences/com.apple.networkextension.plist`, this might be because macOS figures what the real path actually is when using symbolic links.

```
    114 => "org.whispersystems.signal-desktop"
    115 => "/nix/store/6y838zgdnvyi73qkzdqjw101h90nmjdz-signal-desktop-7.44.0/Applications/Signal.app/Contents/MacOS/Signal"
    […]
    117 => "org.whispersystems.signal-desktop"
    118 => "/nix/store/q8679n67k7mrg3g62kfn3kjsmi90vkv5-signal-desktop-7.46.0/Applications/Signal.app/Contents/MacOS/Signal"
```
Basically, it seems cosmetic to me. 🤷 The actual bundle name above seems to tie those entries together. Further down the line, one could think about cleaning up the obsolete entries or managing them.

BTW: I searched the mentioned thread for „contents“ and this is the only comment I came up with proposing to link `Contents/`: https://github.com/nix-community/home-manager/issues/1341#issuecomment-761045074

This wasn’t followed up because of an unspecified issue with an app named „Anki“ — but from a quick test I did, Anki runs with it … 🤷

joneshf:
FWIW, I moved over to using [an implementation without a trampoline that only makes one directory in `~/Applications`](https://github.com/joneshf/home-manager/blob/8e7a72395881668c41b7219cccdc95e13504236b/modules/copy-application-bundles/module.nix#L13-L62), and I don't have duplicate entries in my "Privacy & Security" settings. So if this idea of symlinking only the `Contents` directory is done, maybe removing the trampoline would also remove the duplicates?

I don't use `nix-darwin` to test this idea on, only `home-manager`. So maybe there's something else different about it being in `/Applications` vs. `~/Applications`.

reckenrode:
> Though there is one tiny caveat:
> 
> For each instance (Nix store entry) of an app there is an entry for in „Privacy & Security“ — this is a screenshot of "Local Network" for demonstration:
> 
> Screenshot 2025-04-14 at 19 51 48
> I think, it happened when I declined the initial request. I remember distinctly being asked a second time later on. Might be no problem, when the request is accepted. 🤷 Funny thing is, the switches are in sync — if I flip one the other flips, too. And if I right-click and „Open in Finder“ I am lead to the same folder `/Applications/Nix Trampolines/` both times.

This prompted a few questions. What happens when an application provides services using the new services pattern with embedded launch agents? Do multiple agents run? How do file associations work? Do they associate with the app in the store or in the Applications folder?

reckenrode:
I tried modifying my copy apps script to do some testing. I’ve already run into a few problems.

* Linking `Contents` doesn’t appear to work with apps that use `desktopToDarwinBundle`. The icon for my [FFXIV package](https://github.com/reckenrode/nix-packages/tree/6eeba2dd1336bfde0bf0769f1b9b353d4638d30c/pkgs/by-name/ff/ffxiv) is broken. When I try to run it, I get a message that the app is damaged or incomplete.
* Secretive thinks it’s not in the applications folder. It does when it’s copied. There are also multiple Secretive agents running now.

https://github.com/reckenrode/nixos-configs/blob/6acbc24aa5b8d7641f7da0527ead46e8740a3b36/modules/by-name/co/copy-apps/module.nix

christian-burger:
> I tried modifying my copy apps script to do some testing.

Thanks for trying it out.

> What happens when an application provides services using the new services pattern with embedded launch agents? Do multiple agents run?

Not sure what „new pattern“ you mean. I can only guess, here. If it’s done by Nix activation, I think only one agent will run — the active one. The not-selected/inactive service from the Nix store should be deactivated. At least if it’s managed correctly by nix-darwin, which is what I assume and have seen so far.

> How do file associations work? Do they associate with the app in the store or in the Applications folder?

Sadly, that’s not that easy. The default app, yes, it associates with the `/Application` folder. If I remove the app, macOS uses another available compatible app.

But, the recent option (under „Open With >“): No; it starts the app from `/nix/store`, it even lingers when I garbage collect the app there and fails to start … — but the latter seems more of a macOS issue.

> * Linking `Contents` doesn’t appear to work with apps that use `desktopToDarwinBundle`. The icon for my [FFXIV package](https://github.com/reckenrode/nix-packages/tree/6eeba2dd1336bfde0bf0769f1b9b353d4638d30c/pkgs/by-name/ff/ffxiv) is broken. When I try to run it, I get a message that the app is damaged or incomplete.

I tried Supersonic which does use `desktopToDarwinBundle` as well. This problem seems to be the relative symbolic in here: 

```shell
christian@chabl7:/nix/store/x1lk8k8rp2rp5m6lwdli5f7m6iyk05l2-supersonic-0.13.1/Applications/Supersonic.app/Contents/MacOS 3 % ls -l
total 0
lrwxr-xr-x  1 root  nixbld  26 Jan  1  1970 Supersonic -> ../../../../bin/supersonic
```

If I change it to an absolute link
```shell
christian@chabl7:/nix/store/x1lk8k8rp2rp5m6lwdli5f7m6iyk05l2-supersonic-0.13.1/Applications/Supersonic.app/Contents/MacOS 3 % ls -l
total 0
lrwxr-xr-x  1 root  nixbld  76 Apr 17 22:14 supersonic -> /nix/store/x1lk8k8rp2rp5m6lwdli5f7m6iyk05l2-supersonic-0.13.1/bin/supersonic
```
the app works. Does the `rsync` command used to copy the app for the „copy ’n paste“ approach modify the link appropiately? Or might it be another problem and not the symbolic link?

> * Secretive thinks it’s not in the applications folder. It does when it’s copied. There are also multiple Secretive agents running now.

Since the agent is installed by the app on its own. One solution for the multiple agents issue might be to install and manage the agent during Nix activation phase — like any other service/agent/daemon.

But, the running app knows it’s stored in `/nix/store/`. Nothing to be done about that, I think.

christian-burger:
Soooo … I have another solution, no one’s probably gonna like. But still, I’m throwing it out here: 

```
mount localhost:/nix/store/2ks2kvg4vp35r2m20q4jaxvravx2l0kv-secretive-2.4.1/Applications/Secretive.app/Contents /Applications/Secretive.app/Contents
```

macOS has no `mount -o bind`, but it has [NFS local mounts](https://apple.stackexchange.com/questions/197029/how-do-you-mount-bind-a-local-directory). I had to add the Nix store to `/etc/exports`

```
/nix/store -alldirs -mapall=christian localhost
```

Now Secretive thinks everything is okay.
Screenshot 2025-04-18 at 10 33 23

There is still a problem with symlinks coming from `desktopToDarwinBundle`… which could be solved by using APFS cloning feature instead: `copy -c <source> <target>`.

But another tiny problem is: If you want to update an app, you cannot switch the mount while the app is running and the mount point is in use. You can mount over it, though. But, this makes a garbage collect later on kind of difficult, if the system was not rebooted or the mounts corrected.

Which makes me wonder: What happens, when you upgrade an app put into `/Applications/` with „copy ’n paste“ while it is running?

reckenrode:
> > What happens when an application provides services using the new services pattern with embedded launch agents? Do multiple agents run?
> 
> Not sure what „new pattern“ you mean. I can only guess, here. If it’s done by Nix activation, I think only one agent will run — the active one. The not-selected/inactive service from the Nix store should be deactivated. At least if it’s managed correctly by nix-darwin, which is what I assume and have seen so far.

macOS 13 introduces new APIs to allow apps to manage services. They put their launch agents in the bundle.

https://developer.apple.com/documentation/servicemanagement/smappservice

As an aside, @Samasaur1 has explored generating stub applications for service management in nix-darwin. The primary advantage of this approach for nix-darwin (and Home Manager) is it allows agents to be grouped, have an icon, and have a nice name (other than `sh` or `nix`).


> > How do file associations work? Do they associate with the app in the store or in the Applications folder?
> 
> Sadly, that’s not that easy. The default app, yes, it associates with the `/Application` folder. If I remove the app, macOS uses another available compatible app.
> 
> But, the recent option (under „Open With >“): No; it starts the app from `/nix/store`, it even lingers when I garbage collect the app there and fails to start … — but the latter seems more of a macOS issue.

That’s not ideal. I assume if an application (like Firefox) offers to make itself the default, it will also use the store path (meaning an upgrade would continue launching the old version).

> > * Linking `Contents` doesn’t appear to work with apps that use `desktopToDarwinBundle`. The icon for my [FFXIV package](https://github.com/reckenrode/nix-packages/tree/6eeba2dd1336bfde0bf0769f1b9b353d4638d30c/pkgs/by-name/ff/ffxiv) is broken. When I try to run it, I get a message that the app is damaged or incomplete.
> 
> I tried Supersonic which does use `desktopToDarwinBundle` as well. This problem seems to be the relative symbolic in here:
> 
> ```shell
> christian@chabl7:/nix/store/x1lk8k8rp2rp5m6lwdli5f7m6iyk05l2-supersonic-0.13.1/Applications/Supersonic.app/Contents/MacOS 3 % ls -l
> total 0
> lrwxr-xr-x  1 root  nixbld  26 Jan  1  1970 Supersonic -> ../../../../bin/supersonic
> ```
> 
> If I change it to an absolute link
> 
> ```shell
> christian@chabl7:/nix/store/x1lk8k8rp2rp5m6lwdli5f7m6iyk05l2-supersonic-0.13.1/Applications/Supersonic.app/Contents/MacOS 3 % ls -l
> total 0
> lrwxr-xr-x  1 root  nixbld  76 Apr 17 22:14 supersonic -> /nix/store/x1lk8k8rp2rp5m6lwdli5f7m6iyk05l2-supersonic-0.13.1/bin/supersonic
> ```
> 
> the app works. Does the `rsync` command used to copy the app for the „copy ’n paste“ approach modify the link appropiately? Or might it be another problem and not the symbolic link?

It’s the relative link. The copying approach uses `--copy-unsafe-links` to convert any links outside the `Applications` folder to copies.

That means it also breaks applications that expect to find resources relative to their `bin`, but hopefully those applications are rare. (The only one that comes to mind is LLVM, which has problems with multiple outputs because of that.

> > * Secretive thinks it’s not in the applications folder. It does when it’s copied. There are also multiple Secretive agents running now.
> 
> Since the agent is installed by the app on its own. One solution for the multiple agents issue might be to install and manage the agent during Nix activation phase — like any other service/agent/daemon.
> 
> But, the running app knows it’s stored in `/nix/store/`. Nothing to be done about that, I think.

I tried a while ago, but IIRC manually managing the agent was brittle because Secretive expects to manage it on its own. You can end up with two agents running accidentally. I think it was also hard to kill because Secretive wants to manage the agent, and just killing it outside of Secretive will cause it to restart.

> macOS has no `mount -o bind`

The extremely annoying this is that macOS _does_ support bind mounts. They’re called nullfs mounts because they come from FreeBSD. You can see the [sources][nullfs_source] to `mount_nullfs` in the XNU sources. It’s possible to build it, but using it requires an Apple-private [entitlement][nullfs_entitlement]. ☹️

[nullfs_source]: https://github.com/apple-oss-distributions/xnu/tree/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/miscfs/nullfs
[nullfs_entitlement]: https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/miscfs/nullfs/null_vfsops.c#L81

reckenrode:
What would be really nice is if we could create a firmlink to the location in the store. Firmlinks solve the `realpath` problem that symlinks have. But no, that’s also an Apple-private thing we can’t use. 😕

Samasaur1:
> As an aside, @Samasaur1 has explored generating stub applications for service management in nix-darwin. The primary advantage of this approach for nix-darwin (and Home Manager) is it allows agents to be grouped, have an icon, and have a nice name (other than `sh` or `nix`).

I did end up doing a full drop-in replacement for home-manager (https://github.com/nix-community/home-manager/pull/5992), but you weren't able to register launch agents unless doing so from a graphical context, so I closed the PR. And (iirc) you couldn't activate bundled launch daemons without authenticating within System Settings.

dwt:
To be fair, this all sounds like the copying approach of this pull request brings a ton of real improvements that we could later build upon to make it even better.

Am I missing something here? The other approaches presented all seem to have shortcomings we would need to figure out later.

christian-burger:
@reckenrode, I appreciate very much that you are taking the time to explain.

> To be fair, this all sounds like the copying approach of this pull request brings a ton of real improvements that we could later build upon to make it even better.

I have to agree, from what I can gather, the copy 'n paste approach seems the most robust. And the only drawback is lost space. Which is huge for Apps using Electron like Signal, Element or Visual Studio Code, but so be it. 

Though I still prefer the "link-contents" approach for my use case. Because of that, would it be possible to make it „overridable“? So „mac-app-util“ and any other program have a way to implement their own way, without the `/Applications/Nix Apps/` folder?

I remember faintly, that there was something about blocking modules? Is that an option that could be used here by another flake/module to suppress `modules/system/applications.nix` from being executed? Would that work?

Samasaur1:
> Though I still prefer the "link-contents" approach for my use case. Because of that, would it be possible to make it „overridable“? So „mac-app-util“ and any other program have a way to implement their own way, without the `/Applications/Nix Apps/` folder?
> 
> I remember faintly, that there was something about blocking modules? Is that an option that could be used here by another flake/module to suppress `modules/system/applications.nix` from being executed? Would that work?

You'll be able to override this new default behavior in the same way you override the current default behavior (using `lib.mkForce system.activationScripts.applications.text ""`)

christian-burger:
Oh, very nice, thank you. Thought `types.lines` was exclusively sortable with `mkOrder`. Did not realize, that I can mix `mkOrder` with `mkOverride`. For some reason I assumed that it was either or and never tried. 😅

modules/system/applications.nix
Enzime: Sounds fine to me 👍

modules/system/applications.nix
Enzime: What existing code that checks for the symlink are you referring to? What's currently here looks fine

Enzime (approved):
Tested this locally and it works 👍

modules/system/applications.nix
```diff
@@ -37,12 +38,33 @@ in
...
-      else
-        echo "warning: /Applications/Nix Apps is not owned by nix-darwin, skipping App linking..." >&2
+      targetFolder='/Applications/Nix Apps'
+
+      # Clean up old style symlink to nix store
+      if [ -e "$targetFolder" ] && ourLink "$targetFolder"; then
+        rm "$targetFolder"
       fi
+
+      mkdir -p "$targetFolder"
```
Samasaur1, line 48: If there's (somehow) a file at this path, this command will fail, aborting activation. It'd be nice to avoid this because we're already past the "write boundary" (to use the home-manager terminology). I think we could probably add something to `system.checks` to catch this before we start running the actual activation scripts.

Samasaur1 (approved):
tested this and the "good path" (swapping from a symlink to this new rsync behavior) works flawlessly, as does swapping from Randy's app copying module.

There are a couple cases where problems could potentially occur. I left a comment on one (if `/Applications/Nix Apps` is a file). The other is that if `/Applications/Nix Apps` is not owned by us, we copy into it anyway. I'm making a note of these because I noticed them, but I think we're good to merge this anyway: I think it's unlikely that anyone will have anything at `/Applications/Nix Apps` that wasn't put there by nix-darwin, and even if there is something there, 1) it requires admin privileges to do so, so it's not really a security hole; 2) [every change breaks someone's workflow](https://xkcd.com/1172/)

modules/system/applications.nix
WeetHet: I would also add a check that the current app has App Management permissions if possible, because otherwise the activation would fail as well

Enzime: Do you know if the App Management permission is necessary?

This command worked without `sudo`, Full Disk Access or App Management:

```
$ cp -r /Applications/1Password.app /Applications/1Password-test.app
```

Enzime (changes requested):
I think we should make this an option that's enabled by default

I would like to opt out of this behaviour as my setup works quite well with symlinked applications from inside the store

Enzime:
@dwt will you work on this PR soon or should I pick up this PR?

modules/system/applications.nix
WeetHet: Overriding existing apps that were opened at least once requires it since they get an xattr

dwt: Not sure I have the right thing, [I assume you are talking about these permissions?](https://lapcatsoftware.com/articles/AppManagement.html) (sorry, I wasn't able to find apples documentation on the topic).

If I understand you right, you are considering that the nix-daemon needs this permission? (Or perhaps the terminal from which it is called?)

WeetHet: Yeah pretty much. macOS wants me to grant App Management permission to the app from which the rebuild is executed, not the nix-daemon

dwt:
I would welcome your contribution to this effort, as I'm not entirely sure what you want.

modules/system/applications.nix
dwt: I don't think there is anything we can do here to facilitate that in nix-darwin, or am I missing something?

WeetHet: > I don't think there is anything we can do here to facilitate that in nix-darwin, or am I missing something?

You can query the TCC database to see if the app you are running in has access:
```sh
$ sudo sqlite3 ~/Library/Application\ Support/com.apple.TCC/TCC.db \
        "SELECT client, auth_value FROM access WHERE service='kTCCServiceSystemPolicyAppBundles';"
```

WeetHet: Okay, so you can go through the PPID tree, get the top one's command, use `PlistBuddy` to retrieve `CFBundleIdentifier`, and check that the `auth_value` in the `TCC` database is `2`

dwt: @WeetHet I don't quite understand - what would be the goal? I can probably write a shell function that tries that, but what do I do then? If the app does not have those permissions, it will ask the user anyway, no?

WeetHet: > @WeetHet I don't quite understand - what would be the goal? I can probably write a shell function that tries that, but what do I do then? If the app does not have those permissions, it will ask the user anyway, no?

The activation would fail after the write boundary, not halted until the user gives the permissions, which is something we are trying to avoid. So, imho, we should put this into `system.checks` and abort the activation before the writeBoundary is reached

WeetHet: Idk if all macOS versions we support have an `App Management` permission though, cc @emilazy

emilazy:
Sorry for my absence. I would still like to move this PR forward.

I oppose making it optional – the issues with linking are real and I don’t think we should support two separate paths with different infelicities indefinitely. Monkey‐patching the activation script is always an option for people who are intent on keeping symlinks into the store, but as the main downside is just a little disk space and having multiple versions of packages around on disk is inherent to the Nix experience, I think taking the cue from Homebrew and switching to a pure copying approach is best.

I don’t have time to fully re‐review this right now though, sorry. Please do ping if this stalls again.

dwt:
@emilazy From my side this pull request seems mergeable. It is good enough in the sense that the copying works.

I would also like to give my vote towards not retaining the old behavior via an option, as while it is fast, it has unfixable bugs. (At least missing spotlight indexing, which also means one cannot open these apps from the command-space shortcut, but also these apps not showing up in Launchpad.)

There are some comments by @WeetHet that I am not sure yet I understand, but the code seems to hold and works well. (also it is already 'in production' by virtue of being used by @reckenrode and @Samasaur1 for a long time).

I still think this can be a good basis to build upon to later either restrict trying to copy only the actually updated binaries for speed, or integrating this logic into @Samasaur1 file work.

modules/system/applications.nix
dwt: @WeetHet The abort would only happen if the user does not consent to the Application Management, right? So if I get you correct, you want the check and then abort the switch action earlier? So the system cannot get into an indeterminate state?

Wouldn't that prevent the user getting asked if he wants to grant that permission? If so, we would have to communicate in the error message how to enable this permission and how to find out which app it needs to be enabled for, which is not super simple, as these settings can be named surprisingly different over all macOS languages.

dwt: @WeetHet After re-reading: are you suggesting that we should check the permission and if it is explicitly revoked, we fail the activation. If it is not yet set, we try and let the user choose to permit for the better user experience.

WeetHet: I don't know. I've implemented a failsafe in my configuration as follows:
https://git.sr.ht/~weethet/nix-dotfiles-flakes/commit/496bab721531049730da872b0bf87dbf3a5eafd7

I'm not sure if it's even always correct, so I thinks it's fine to proceed without it for now, but keep in mind that we should probably implement it at one point

WeetHet: Oh, that's conditioned on Full Disc Access, not App Management

WeetHet:
I drop my comments since I couldn't find a working way to check for App Management permissions

modules/system/applications.nix
Samasaur1: > @Samasaur1 You wanted a check that there was not a file already at this location `"$targetFolder"`, that would mean that a file `/Applications/Nix Apps` already exists. I don't know, wouldn't the error message from `mkdir` be already good enough in this case to figure out what the problem is? If not, what kind of assertion / error message would you prefer?

It's not that the error message is bad, but that we'd leave the system in an indeterminate state. Having a check in `system.checks` will check this before modifying users/custom activation scripts. For the message, I'd probably just have it say something like "Unexpected file at `/Applications/Nix Apps`; aborting activation" — you could model it after the message printed by `etcChecks`.

Samasaur1: That said, if this check is any more complicated than something like
```bash
if ! test -L "/Applications/Nix Apps" && test -f "/Applications/Nix Apps"; then
   echo error
   exit 1
fi
```
I'm fine not including a check, because I really don't see how someone could accidentally end up with a file there

Enzime: I think that checking the TCC database is the wrong approach because it requires Full Disk Access. Instead I would rather see a noop command that would request App Management permissions that we could add to `system.checks` similar to how we currently request `kTCCServiceSystemPolicySysAdminFiles`:

https://github.com/nix-darwin/nix-darwin/blob/32f0cf2140af6a852f8c8b6c8f15e4855d461b87/modules/users/default.nix#L113-L152

see https://github.com/nix-darwin/nix-darwin/pull/1120#discussion_r1817836999 for more context

dwt: I'm inclined to say that we should perhaps do the permission check in a separate pull request. Reason being, that the permission problem is already there, so this pull request should not make it worse? I hope that splitting this out into its own pull request could enhance the focus in this one, to get it merged. What do you think?

Enzime: I'm hesitant to merge this without the pre-switch check because after testing this PR I ran into this exact issue with having not granted App Management (which can't be done over SSH) which can lead to partial activations

Without this PR, updating your Applications doesn't lead to activation failing potentially in the middle

@emilazy what do you think?

dwt: @Enzime If this pull request triggers the permission where it wasn't triggered before, that seems quite a valid reason to add the permission check. Do you happen to have an idea why that wasn't triggered before? Adding Symlinks to `/Application/Nix Apps` seems like it should ask just as much for a permission than copying applications there?

I have tried to replicate this finding on a vm, and it seems I could not, with this pull request enabled, nix-darwin only asked for full disk access, but not application management permission. Can you give me a hint what I might be missing?

dwt: That sounds sensible. I'll see that I can add that check to check-phase. As I haven't done anything there before, is there anything that I should keep in mind for there?

Isolaryn:
Copying has definitely been working a lot better than linking for me. I know Mac storage comes at a premium so I would hesitate to only have copying as an option. 

The same code seems to work fine for home-manager applications as well.
```nix 
    home-manager.sharedModules = [
      ({
        lib,
        config,
        ...
      }: {
        # Disable old style linking of applications in home-manager
        targets.darwin.linkApps.enable = lib.mkForce false;
        home.activation.copyApplications = let
          targetDir = "${config.home.homeDirectory}/Applications/Home Manager Apps";
          homeApps = pkgs.buildEnv {
            name = "home-applications";
            paths = config.home.packages;
            pathsToLink = "/Applications";
          };
        in
          lib.hm.dag.entryAfter ["writeBoundary"] ''
            # Set up home applications.
            echo "setting up ${targetDir}..." >&2

            # Clean up old style symlinks
            if [ -e "${targetDir}" ] && [ -L "${targetDir}" ]; then
              rm "${targetDir}"
            fi
            mkdir -p "${targetDir}"

            rsyncFlags=(
              --checksum
              --copy-unsafe-links
              --archive
              --delete
              --chmod=-w
              --no-group
              --no-owner
            )
            ${lib.getExe pkgs.rsync} "''${rsyncFlags[@]}" ${homeApps}/Applications/ "${targetDir}"
          '';
      })
    ];
```

modules/system/applications.nix
dwt: @Enzime Even with the Element.app ([this one right?](https://github.com/NixOS/nixpkgs/blob/nixos-25.05/pkgs/by-name/el/element-desktop/package.nix#L153)) and opening it once again before switching again - I get no permission errors. With the Terminal and the integrated Terminal of Zed. While none of these apps have permission for either full disk access nor application management permissions. I seem to be missing something here.

Enzime: I haven’t actually tested `Element.app`, the one app I know for sure that should trigger it is `1Password.app`, you shouldn’t need an account to test it

To check app signing and notarisation you can use `Apparency.app`, it’s already packaged in Nixpkgs

dwt: @Enzime Maybe you could take a look at the latest commit. I am setting the check to something that seems sensible. But there surely must be a simple oversight somewhere, because I cannot see that check become part of the check script.

I.e. I do not get any response from trying to evaluate it with

```shell
nix eval .\#darwinConfigurations.nix-darwin-dev.config.system.activationScripts.checks.text l unvis | grep isMissingPermission
```

zhaofengli: Also I don't think we should suggest granting "Full Disk Access" - just "App Management" should be enough.

> Another thing I've found from testing is that touch /Applications/1Password.app/.DS_Store fails with different errors

"Permission denied" is for regular Unix file permissions, and "Operation not permitted" is from some higher-level policy (TCC, EndpointSecurity, some other MAC hook, ...). In a normal system,  apps installed through the App Store and `.ipa`s sideloaded via IPA Installer will be owned by root, hence the regular "Permission denied" error when you try to modify them without root.

Enzime: Looks like it's added correctly to `system.checks.text` for me:

```
$ nix eval --raw github:Enzime/dotfiles-nix#darwinConfigurations.hyperion-macos.config.system.activationScripts.checks.text --override-input nix-darwin github:nix-darwin/nix-darwin/pull/1396/head | grep isMissingPermissionToModifyApps
warning: not writing modified lock file of flake 'github:Enzime/dotfiles-nix':
• Updated input 'nix-darwin':
    'github:nix-darwin/nix-darwin/44a7d0e687a87b73facfe94fba78d323a6686a90?narHash=sha256-sNUUP6qxGkK9hXgJ%2Bp362dtWLgnIWwOCmiq72LAWtYo%3D' (2025-05-27)
  → 'github:nix-darwin/nix-darwin/44fa58393bcb1e8af3230ec5057c79a705629d32?narHash=sha256-a2Sbhd0/A8s3Qe4keshf82/K/KV/WH3NJ95r8EovA3o%3D' (2025-05-28)
isMissingPermissionToModifyApps() {
if isMissingPermissionToModifyApps; then
```

```
$ nix eval --raw github:Enzime/dotfiles-nix#darwinConfigurations.hyperion-macos.config.system.checks.text --override-input nix-darwin github:nix-darwin/nix-darwin/pull/1396/head | tail -n 25
warning: not writing modified lock file of flake 'github:Enzime/dotfiles-nix':
• Updated input 'nix-darwin':
    'github:nix-darwin/nix-darwin/44a7d0e687a87b73facfe94fba78d323a6686a90?narHash=sha256-sNUUP6qxGkK9hXgJ%2Bp362dtWLgnIWwOCmiq72LAWtYo%3D' (2025-05-27)
  → 'github:nix-darwin/nix-darwin/44fa58393bcb1e8af3230ec5057c79a705629d32?narHash=sha256-a2Sbhd0/A8s3Qe4keshf82/K/KV/WH3NJ95r8EovA3o%3D' (2025-05-28)
# touch can show multiple errors
# Operation not permitted: Likely means that you do not have application management permission yet
# Permission denied: Likely means that you do not have full disk access permission or do not have enough privileges
isMissingPermissionToModifyApps() {
  for appBundle in /Applications/Nix\ Apps/*.app; do
      if [ -d "$appBundle" ]; then
          touch "$appBundle/.DS_Store" || return 0
      fi
  done
  return 1
}
if isMissingPermissionToModifyApps; then
    printf >&2 '\e[1;31merror: Missing permission to modify apps, aborting activation\e[0m\n'
    printf >&2 'The current Application you are using to execute darwin-rebuild switch\n'
    printf >&2 'is missing one of the required permissions to modify the Applications in\n'
    printf >&2 '/Applications/Nix Apps.\n'
    printf >&2 '\n'
    printf >&2 'This means that you will either have to accept the permission prompt\n'
    printf >&2 'by MacOS, or add the permission manually in System Preferences.\n'
    printf >&2 '\n'
    printf >&2 'To add the necessary permission manually, go to System Preferences,\n'
    printf >&2 'Security & Privacy, Privacy tab, and add the current Application to\n'
    printf >&2 '"Full Disk Access" and "App Management".\n'
    exit 2
fi
```

`system.checks.text` will always run as `root` so we don't have to worry about getting `Permission denied`, if it fails it'll almost certainly be due to `Operation not permitted` and not being granted the `App Management` permission

dwt: @Enzime Would you recommend reusing that error message and instructions for app management? I have mostly written something off the top of my head to get started.

Enzime: Yeah I would recommend using the same flow and message

dwt: @Enzime Looking at that check, I'm at a loss how I would abstract that to also fit this use case. If that is what you are looking at, I think I cannot provide. I'm not even sure I understand the specific checks that are done there and how they would apply to this check? Could you advise, perhaps by providing a rough version as a change proposal here on GitHub?

My understanding is, that this is trying to check whether you are logged in via `ssh` and in that case checks that you need full disk access, while anything not `ssh` just needs application management permissions.

Is that even correct?

Enzime:
I've pushed a new version of the check, can people test this PR again?

modules/system/applications.nix
```diff
@@ -1,18 +1,60 @@
...
+}:
 
-  cfg = config.system;
+{
+  config = {
+    system.checks.text = lib.mkAfter ''
+      ensureAppManagement() {
+        for appBundle in /Applications/Nix\ Apps/*.app; do
+          if [[ -d "$appBundle" ]]; then
+            if ! touch "$appBundle/.DS_Store" &> /dev/null; then
```
Samasaur1, line 14: does this break code signatures?

Enzime: Just tested this with Apparency, code signatures and notarization are still fine after touching `.DS_Store`

dwt (commented):
LGTM

Enzime:
I currently don't have the time to test the check works

If someone can test it by running a system with an old version of Element in `environment.systemPackages` and then opening Element and ensure their terminal hasn't been granted App Management and Full Disk Access then trying to switch to a newer generation with a newer version of Element using this PR and ensuring that it correctly shows an error, then I'm happy to merge this PR

dwt:
I might be doing something wrong. In a VM I did go back to nix-darwin master, and an old version of nixpkgs in a flake, then called `nix flake update` to sync the lock file. After that I switched to that flake and got an old version of element installed.

Then going back to the current version of nixpkgs and this pull request, I again synced the lock file with `nix flake update` and installed the newer version of [element](https://search.nixos.org/packages?channel=unstable&show=element-desktop&from=0&size=50&sort=relevance&type=packages&query=element-desktop).

This installed cleanly, but without any errors, even though nothing in that vm has full disk access, or app management permissions.

Do I need to go back to this pr, but without the permission check patch to trigger it asking for permission?

Enzime:
I just checked Element from Nixpkgs using Apparency and it says it's not even signed so you'll probably have to test with a different app like 1Password

You'll want to make sure you're always running the PR when testing as without the PR then it won't be copying the apps to `/Applications/Nix Apps` so then it won't require the App Management permission to overwrite files

dwt:
I was now able to test this by sticking to the newest version of this pull request, and modifying the nixpkgs version to simulate version upgrades.

This worked fine.
- Going from the old (current master) to this pull request does not trigger a permission warning when switching
- Updating Applications does trigger the warning, if a notarized app that has been open is updated
- The warning is triggered if the app/terminal in question does not have application management permission
- Once that permission is set, no further warning is displayed.

@Enzime That looks good to me, would you want additional tests to get this merged?

afh:
Thanks for your continued work on this @dwt and thanks @Enzime for helping out, all very much appreciated 🙏

@emilazy, you previously requested changes and [opted-in for a friendly ping :)](https://github.com/nix-darwin/nix-darwin/pull/1396#issuecomment-2869951848) if this stalls again. I'm happy to merge this, but wanted to leave an opportunity for you to share your thoughts and expertise on the latest changes to this PR.

modules/system/applications.nix
```diff
@@ -1,18 +1,60 @@
...
-{
-  options = {
-  };
+      if ! ensureAppManagement; then
+        if [[ "$(launchctl managername)" != Aqua ]]; then
+          # It is possible to grant the App Management permission to `sshd-keygen-wrapper`, however
+          # there are many pitfalls like requiring the primary user to grant the permission and to
+          # be logged in when `darwin-rebuild` is run over SSH and it will still fail sometimes...
+          printf >&2 '\e[1;31merror: permission denied when trying to update apps over SSH, aborting activation\e[0m\n'
+          printf >&2 'Apps could not be updated as `darwin-rebuild` requires Full Disk Access to work over SSH.\n'
```
emilazy, line 29: Can we make this conditional on there being any application management to do? (Not a blocker.)

Enzime: That’s a good point, I forgot that in the current form it would mean everyone always needs App Management or FDA for SSH deploys

Sure, we can use the output of `rsync --dry-run` to only test the apps we’re going to actually modify

emilazy: Yeah… in the long run it would be nice to have a system where activation scripts can be modularly composed of dry run + check + execute so that it’s easier to implement these things. Although it might be more ROI to just move things to launchd services/agents in which case it would just work, I think.

Enzime: I was also thinking this but I hadn’t actually tested to figure out if you can grant App Management to a Launch Daemon

The thing that is weird is you need to be `root` to have UNIX permissions to modify installed applications but you need the Mach bootstrap context of a user that granted App Management as this TCC permission is at a user level and I don’t think you can grant it as `root` without disabling SIP (that’s the reason why we recommend FDA over SSH as FDA is granted at a system level not user level)

afh:
Apologies eveeryone, I misspoke :/ I mistook this PR to be in the nixpkgs repo, I do not have commits rights to this repo.

Thanks @emilazy for chiming in here again and sharing your expertise.

johnstegeman:
I've been using this for a few weeks now by referencing dwt/nix-darwin/application-linknig-done-right in my flake, and its working great. Waiting for this to merge

modules/system/applications.nix
dwt: @emilazy I'm. not entirely sure how the rsync output is supposed to disable the application management question. Do you propose to just check if rsync would sync any files (exit code perhaps?) and then not check?

dwt: @emilazy / @Enzime  Would maybe a simpler way to make the permission check optional be to just check if there is anything in the GUI link farm? If you have anything in there, it's going to need the permissions eventually.

dwt: Friendly ping @emilazy @Enzime : what can I do to get this moving forward?

phucisstupid:
merge it

dwt:
> merge it

My understanding is, that the current holdout is that this change right now will require every user (even those that do not install GUI apps) to give application management permissions.

I'm not sure that this is super bad, and would welcome the improvement either here or in a separate merge request to make this permission only be asked if there are actually GUI apps to be installed.

modules/system/applications.nix
dwt: @emilazy @Enzime If I read the current code correctly, the check will only happen _if_ you actually have any GUI applications installed. To me that makes this even less a blocker and something that we could refine in a later pull request?

plxty:
For anyone who wants to test (give it a try :) it without patching or else,

```nix
# flake.nix
inputs = {
  nix-darwin-linking = {
    url = "github:dwt/nix-darwin/application-linking-done-right";
    inputs.nixpkgs.follows = "nixpkgs";
  };
};
```

```nix
darwinSystem {
  modules = [
    { disabledModules = [ "system/applications.nix" ]; }
    "${inputs.nix-darwin-linking}/modules/system/applications.nix"
  ];
}
```

modules/system/applications.nix
dwt: Friendly ping @emilazy  and @Enzime : am I missing something that the permission check will only happen for people who already have gui apps installed? 

If this is not the case or you require the check being conditional on actual changes to apps installed this run, please let me know how to proceed to get this mergeable.

nevir:
One interesting wrinkle with your change is for apps like VS Code that have a companion CLI tool. 

E.g. if I use `code` to launch VS Code, it launches the original copy, rather than the one installed to `/Applications`

phucisstupid (approved):

josh1703658784:
It seems like copying applications was the chosen path forward but for anyone interested I cooked up a trivial alternative which relies on [mkalias](https://github.com/reckenrode/mkalias).

Replacing the final `rsync` command with this loop:
```
  for appPath in '${config.system.build.applications}/Applications/'*; do
    appName="$( basename "$appPath" )"
    realAppPath="$( readlink "$appPath" )"
    ${lib.getExe pkgs.mkalias} "$realAppPath" "$targetFolder/$appName" &
  done
  wait
```

I'm very much a newbie but am curious what it would take to provide an option where users could select between the default symlink, copying (this), or relying on `mkalias`.

dwt:
@josh1703658784 I thought I addressed this in  the description of this pull request (but there is more in the discussion of #214).

> Aliasses: Require either AppleScript (a permission we don't want to have, as it easily bypasses Apples TCC) or extra tools (that would be ok), but also Aliasses are not categorized as 'Application' by SpotLight.

This was pretty much the killer for the alias approach - which would otherwise be vastly superior.

dwt:
> One interesting wrinkle with your change is for apps like VS Code that have a companion CLI tool.
> 
> E.g. if I use `code` to launch VS Code, it launches the original copy, rather than the one installed to `/Applications`

That is interesting. Not sure this is actually fixable in nix though. :/

`code` could fix it, by using something like `open -a VSCode` under the hood, but apparently they do something else?

modules/system/applications.nix
dwt: Friendly ping  @emilazy and @Enzime again. I would like to get this merged, could you please provide another look and feedback to this point:

If I read the current code correctly, the check will only happen if you actually have any GUI applications installed. To me that makes this even less a blocker and something that we could refine in a later pull request?

reckenrode:
> @josh1703658784 I thought I addressed this in the description of this pull request (but there is more in the discussion of #214).
> 
> > Aliasses: Require either AppleScript (a permission we don't want to have, as it easily bypasses Apples TCC) or extra tools (that would be ok), but also Aliasses are not categorized as 'Application' by SpotLight.
> 
> This was pretty much the killer for the alias approach - which would otherwise be vastly superior.

This was asked on Matrix, and I also had the following [answer][1]: 

> The caveats I know off hand: applications show up as aliases not applications; they don’t get picked up for file associations or system services until they are run; and setting an application to open a file will reference a store path instead of a stable path in `/Applications`.

> > One interesting wrinkle with your change is for apps like VS Code that have a companion CLI tool.
> > E.g. if I use `code` to launch VS Code, it launches the original copy, rather than the one installed to `/Applications`
> 
> That is interesting. Not sure this is actually fixable in nix though. :/
> 
> `code` could fix it, by using something like `open -a VSCode` under the hood, but apparently they do something else?

This is a known limitation. It’s possible to work around it (e.g., by having nix-darwin check for references to applications in the store and replace them with `/Applications/Nix Apps`), but that should be a separate PR.

(I never bothered to fix it generally in my module because I had a workaround for VS Code, and then I stopped using VS Code.)

[1]: https://matrix.to/#/!lheuhImcToQZYTQTuI:nixos.org/$XU7g7aT6etr1AmsouQYGR_fgXEMtyx2IowgTi60yttE?via=nixos.org&via=matrix.org&via=nixos.dev

nevir:
One other alternative to consider: is it possible to make the `/nix` volume spotlight-indexable? (…which [might require](https://github.com/nix-darwin/nix-darwin/issues/214#issuecomment-3208289761) switching to an unencrypted APFS volume? unclear why spotlight refuses to enable for it today…)

and then managing spotlight indexing via darwin / home-manager instead? (e.g. by managing indexed paths through `mdimport <path>`, etc)

dwt:
> One other alternative to consider: is it possible to make the `/nix` volume spotlight-indexable? (…which [might require](https://github.com/nix-darwin/nix-darwin/issues/214#issuecomment-3208289761) switching to an unencrypted APFS volume? unclear why spotlight refuses to enable for it today…)
> 
> and then managing spotlight indexing via darwin / home-manager instead? (e.g. by managing indexed paths through `mdimport <path>`, etc)

I do not have enough know how about this to be able to even try. I would welcome an experiment from you and some data so we can ground further discussion. If that works, that could be a path to a better implementation.

josh1703658784:
> @josh1703658784 I thought I addressed this in the description of this pull request (but there is more in the discussion of #214).
> 
> > Aliasses: Require either AppleScript (a permission we don't want to have, as it easily bypasses Apples TCC) or extra tools (that would be ok), but also Aliasses are not categorized as 'Application' by SpotLight.
> 
> This was pretty much the killer for the alias approach - which would otherwise be vastly superior.

I should've been more clear that I just posted it for anyone interested in going done the alias route. Didn't intend to derail the design decisions you made or re-start the conversation that's already been had. Really appreciate your work on this!

nevir:
> > One other alternative to consider: is it possible to make the `/nix` volume spotlight-indexable? (…which [might require](https://github.com/nix-darwin/nix-darwin/issues/214#issuecomment-3208289761) switching to an unencrypted APFS volume? unclear why spotlight refuses to enable for it today…)
> > and then managing spotlight indexing via darwin / home-manager instead? (e.g. by managing indexed paths through `mdimport <path>`, etc)
> 
> I do not have enough know how about this to be able to even try. I would welcome an experiment from you and some data so we can ground further discussion. If that works, that could be a path to a better implementation.

Ok, I did a bunch of investigation w/ that approach, and I think it might be viable. See #1576

phucisstupid:
home-manager app not show in spotlight

dwt:
🎉 @Samasaur1. Thanks!

afh:
Thanks everyone who helped move forward with this PR, very much appreciated!! 🙏

PR #1399, opened 2025-03-24
Convert non-configurable Zsh defaults into configurable options

### LINKED ISSUE #983: Why does `/etc/zshrc` hardcode `HISTFILE`?
Why is generated `/etc/zshrc` opinionated on this? Why can't I opt out?

NixChaos:
This converts the following hard-coded options from /etc/zshrc into configurable options, but sets the previously hard-coded values as default. 
```
SAVEHIST=2000
HISTSIZE=2000
HISTFILE=$HOME/.zsh_history

setopt HIST_IGNORE_DUPS SHARE_HISTORY HIST_FCNTL_LOCK

bindkey -e
```
The exception being that I changed the default `HISTFILE` path to `"${ZDOTDIR:-$HOME}/.zsh_history"
` because that's what Apple sets by default. `${ZDOTDIR:-$HOME}` indicates that it will look for `ZDOTDIR`, then revert to `$HOME`, if it doesn't find it.

The new configurable options are:
`programs.zsh = {`
- `defaultKeymap`
- `history.ignoreDups`
- `history.path`
- `history.save`
- `history.share`
- `history.size`
`};`

Fixes #983.

### DIFF
--- a/modules/programs/zsh/default.nix
+++ b/modules/programs/zsh/default.nix
@@ -38,10 +38,7 @@ let
 
         path = mkOption {
           type = types.str;
-          default = "\${\ZDOTDIR:-$HOME\}/.zsh_history";
-          # defaultText = literalExpression ''
-          #   "''${ZDOTDIR:-$HOME}/.zsh_history"
-          # '';
+          default = "$HOME/.zsh_history";
           example = literalExpression ''"$HOME/.local/share/zsh/zsh_history"'';
           description = "History file location.";
         };


bew:
> Fixes #983.

NOTE: this line needs to be in the PR description (not in a separate comment) to be considered by Github as a command to auto-close the linked issue when this PR is merged.

modules/programs/zsh/default.nix
```diff
@@ -182,14 +249,25 @@ in
...
-      HISTSIZE=2000
-      HISTFILE=$HOME/.zsh_history
-
-      setopt HIST_IGNORE_DUPS SHARE_HISTORY HIST_FCNTL_LOCK
-
-      bindkey -e
+      # History
+      HISTFILE="${cfg.history.path}"
+      HISTSIZE=${toString cfg.history.size}
+      SAVEHIST=${toString cfg.history.save}
```
Samasaur1, line 257: Is there a reason you're reordering these?

modules/programs/zsh/default.nix
```diff
@@ -1,21 +1,71 @@
...
+          };
+
+          share = mkOption {
+            type = types.bool;
+            default = true;
+            description = "Share command history between zsh sessions.";
+          };
+        };
+      }
+    );
```
Samasaur1: To match our existing style, this should go in the `let-in` block at the top of the file, rather than adding a new `let-in` block in `options`

Samasaur1 (changes requested):
The actual content of this PR is great. However, you've also made a bunch of unrelated formatting changes, some of which are even causing merge conflicts. If you can revert those changes, we'd be pretty close to merging this in (see my two inline comments).

modules/programs/zsh/default.nix
NixChaos: Oops. This was my first open source contribution, and I'm still really inexperienced with GitHub so I've accidentally closed it.

modules/programs/zsh/default.nix
NixChaos: Just to match the order in the default file from Apple. The order won't actually change anything, so isn't actually necessary. The defaults:
``` zsh
# Save command history
HISTFILE=${ZDOTDIR:-$HOME}/.zsh_history
HISTSIZE=2000
SAVEHIST=1000
```

NixChaos:
Oops. This was my first open source contribution, and I'm still really inexperienced with GitHub so I've accidentally closed it.

NixChaos:
Figured out how to reopen it and I've redone the changes in the GitHub code editor so my formatter can't go rogue.

Samasaur1:
Okay, in the name of you not having to wait, I think the best next step here is for you to change the `HISTFILE` default back to what we had (`$HOME/.zsh_history`). Then I'll merge this PR so that things are configurable, and if we decide to change the default to match macOS that can be in a follow-up PR

NixChaos:
Thanks. I've done that now.

Something I should maybe draw your attention to is that I based the naming of the options on home-manager; I've recently realised that NixOS has a radically different structure for the options. Would it be a good idea if I updated the PR to be consistent with NixOS rather than home-manager, given that NixOS, like nix-darwin, is a system config?
```nix
# home-manager style
programs.zsh = {
  defaultKeymap = "emacs";
  history = {
    ignoreDups = true;
    path = "$HOME/.zsh_history";
    save = config.programs.zsh.history.size;
    share = true;
    size = 2000;
  };

# NixOS style
programs.zsh = {
  histFile = "$HOME/.zsh_history";
  histSize = 2000; # also used for SAVEHIST
  interactiveShellInit = ''
    bindkey -e # not set in NixOS (but is the default in Zsh)
  '';
  setOptions = [
    "HIST_IGNORE_DUPS"
    "SHARE_HISTORY"
    "HIST_FCNTL_LOCK" # hard-coded in HM style
  ];
};
```

Samasaur1:
Yes, matching the NixOS module as closely as possible would be ideal. Note also that I've just queued a different PR that makes the history-specific options configurable, but I would still appreciate at least the setopt handling

Issue #1397, opened 2025-03-23
Local directory of the linux-builder name and leftover files after disabling


dwt:
I'm currently looking at the linux builder, and have some questions:

- [Why is the working directory set to `/var/lib/darwin-builder`](https://github.com/LnL7/nix-darwin/blob/ebb88c3428dcdd95c06dca4d49b9791a65ab777b/modules/nix/linux-builder.nix#L147), even though the thing is called linux-builder? Wouldn't a matching directory name make more sense?
- After disabling the linux builder it leaves files in that directory behind. As these can be quite sizable, wouldn't it make sense to remove them when `nix.linux-builder.enable = false` is set?

I would be happy to provide patches for both, but I would like to understand the issues at stake first.

Enzime:
It was originally called the `darwin-builder` and the name is a vestige

Deleting that directory would be fine with me on `nix.linux-builder.enable = false`

PR #1400, opened 2025-03-24
Linux builder: working directory rename and cleanup

dwt:
As discussed in #1397

* the name of the working directory of the linux builder is not consistent with the package name
* the linux builder leaves behind state outside of the `/nix/store` when disabled

This tries to solve both by upgrading the old name and removing the directory when the module is disabled.

This doesn't handle updating to the new version and disabling at the same time. Those people might end up with an extra directory, but that is no worse than it is now, which is why I didn't want to take up the extra complexity.

@Enzime  What do you think?

### DIFF
--- a/modules/nix/linux-builder.nix
+++ b/modules/nix/linux-builder.nix
@@ -159,7 +159,13 @@ in
     '';
   };
 
-  config = mkIf cfg.enable {
+  config = mkMerge [
+    (mkIf (!cfg.enable) {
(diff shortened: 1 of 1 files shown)


dwt:
Friendly ping @Enzime - you mentioned you would give this a review, can you perhaps give a timeframe when you might get to this?

Enzime (approved):
LGTM

Thanks for the contribution

PR #1407, opened 2025-03-29
website: try to fix

emilazy:
My excuse for testing in prod is that GitHub Actions makes it a pain to do anything else.


Enzime (approved):
🤣

Issue #1243, opened 2024-12-29
Can't enable firewall via nix-darwin


thislooksfun:
I'm trying to set up my system settings via nix-darwin, but I noticed that the firewall isn't being enabled as expected.

Steps to reproduce:
1. Add `system.defaults.alf.globalstate = 1;` to `flake.nix`
2. Run `darwin-rebuild switch`
3. Check System Settings > Network > Firewall, and see that it is still disabled
4. Can also check by running `sudo /usr/libexec/ApplicationFirewall/socketfilterfw --getglobalstate`

From my testing it seems that running `sudo /usr/libexec/ApplicationFirewall/socketfilterfw --setglobalstate [on/off]` correctly toggles the firewall, but writing to the `com.apple.alf.plist` file doesn't do anything. Same for changing the other settings, like the stealth mode.

Maybe it would be worth replacing `system.defaults.alf.*` with a new `system.firewall.*` block that uses `socketfilterfw` instead of `defaults`? Or am I missing something obvious to get this working?

Tested on macOS 15.1 on an M4 MacBook. I even tried restarting after making the change, but no luck.

Enzime:
Yeah I think it will need a new module as a lot of the `system.defaults` options are quite old and macOS no longer respects some of them, I would probably put it under `networking.firewall`

thislooksfun:
I managed to get my flake to automatically configure the firewall by using `postUserActivation`. I'm hoping to translate this into an actual proper solution in `nix-darwin` at some point, but here's what I have for now:

```nix
{ ... }:
{
  system.activationScripts.postUserActivation.text = ''
    # Configure firewall
    sudo /usr/libexec/ApplicationFirewall/socketfilterfw --setglobalstate on
    sudo /usr/libexec/ApplicationFirewall/socketfilterfw --setallowsigned on
    sudo /usr/libexec/ApplicationFirewall/socketfilterfw --setallowsignedapp on
    sudo /usr/libexec/ApplicationFirewall/socketfilterfw --setstealthmode on
  '';
}
```

emilazy:
Should be fixed by https://github.com/nix-darwin/nix-darwin/pull/1520.

thislooksfun:
That works beautifully, thank you!

PR #1424, opened 2025-04-21
networking.firewall: init module

al3xtjames:
Add a basic module to enable or disable the built-in macOS application firewall. The module also supports configuration of pf with custom rules (based on https://apple.stackexchange.com/a/429972).

Addresses: #1243

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -45,6 +45,7 @@
   ./system/version.nix
   ./time
   ./networking
+  ./networking/firewall.nix
   ./nix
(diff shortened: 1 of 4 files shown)


emilazy:
Closing in favour of https://github.com/nix-darwin/nix-darwin/pull/1520, but thank you!

emilazy:
Actually I missed that this can also configure `pf`. I think it would make sense to have a `networking.firewall` for that alongside `networking.applicationFirewall`, if you‘re still interested in adding that.

PR #1429, opened 2025-04-30
fix: allow path to be used for dock's persistent-apps

### LINKED ISSUE #1428: Type inconsistency for dock's `persistent-apps` option
Prior to https://github.com/nix-darwin/nix-darwin/commit/e21d07988b30adbd5b77d16f9fa40b7d5fc00a09 persistent-apps allowed strings or paths to be used:
https://github.com/nix-darwin/nix-darwin/blob/02ba211ea19fb4e5a5da2e24a6ebbed526f5231d/modules/system/defaults/dock.nix#L130-L145

We can see that the commit attempts to keep consistency with previous behaviour by providing a conversion function `toTagged = path: { app = path; };` and that the type is defined as:
https://github.com/nix-darwin/nix-darwin/blob/4515dacafb0ccd42e5395aacc49fd58a43027e01/modules/system/defaults/dock.nix#L157-L160

However, the type for `app` is defined as string:
https://github.com/nix-darwin/nix-darwin/blob/4515dacafb0ccd42e5395aacc49fd58a43027e01/modules/system/defaults/dock.nix#L135

Which means configs with `persistent-apps = [/User/Desktop];` will fail.

Two solutions:
1. fix the conversion function `toTagged`
2. fix the type declaration to `types.nullOr (types.listOf (types.str simpleType toTagged taggedType)); `

ed9w2in6:
Fixes #1428

This use solution 1 in #1428.

### DIFF
--- a/modules/system/defaults/dock.nix
+++ b/modules/system/defaults/dock.nix
@@ -155,7 +155,7 @@ in {
             };
 
         simpleType = types.either types.str types.path;
-        toTagged = path: { app = path; };
+        toTagged = path: { app = builtins.toString path; };
         in
       types.nullOr (types.listOf (types.coercedTo simpleType toTagged taggedType));
       default = null;


ed9w2in6:
Actually for this I should add test case that uses path.  Do not merge yet.

Enzime:
In future you can set the PR to draft to indicate it's not ready to merge yet

ed9w2in6:
@Enzime I am not sure how to add a test case for paths while not being brittle to any minor changes to the test environment. 

Do you have any idea how to do that?
One idea is to do something with the path of `pkgs.hello`, assuming that won't change much.

I am not too sure about this though.

Issue #527, opened 2022-09-16
error: The option `programs.zsh.enableAutosuggestions' does not exist.


champignoom:
Newbie user trying to set up zsh.

`~/.nixpkgs/darwin-configuration.nix`:
```
{ config, lib, pkgs, ... }:
{
  environment.systemPackages = [ pkgs.zsh-autosuggestions ];
  services.nix-daemon.enable = true;
  programs.zsh.enable = true;
  programs.zsh.enableAutosuggestions = true;
  system.stateVersion = 4;
}

```

```
$ darwin-rebuild switch
error: The option `programs.zsh.enableAutosuggestions' does not exist.
...
```

This happens to `programs.zsh.plugins = [];` as well. It seems that only a fixed set of options (listed in `man configuration.nix`) are available on nix-darwin.

DavSanchez:
Hi,

As far as I can tell,  `programs.zsh.enableAutosuggestions` is an option from [`home-manager`](https://nix-community.github.io/home-manager/options.html#opt-programs.zsh.enableAutosuggestions), so only if you are using that you'll be able to use it. `nix-darwin` does not provide this option _per se_ [according to its manual](https://daiderd.com/nix-darwin/manual/index.html#sec-options).

PR #1433, opened 2025-05-01
feat: add option programs.zsh.enableAutosuggestions

ed9w2in6:
We got `enableSyntaxHighlighting`, so why not `enableAutosuggestions` too?

Ref: #527

I have added tests too.
However, I saw the other "plugin" related options for `zsh` are NOT tested.
Should I remove my tests for any reasons? e.g. resource concern

### DIFF
--- a/modules/programs/zsh/default.nix
+++ b/modules/programs/zsh/default.nix
@@ -102,6 +102,12 @@ in
       description = "Enable fzf keybinding for Ctrl-r history search.";
     };
 
+    programs.zsh.enableAutosuggestions = mkOption {
+      type = types.bool;
+      default = false;
+      description = "Enable zsh-autosuggestions.";
+    };
+
     programs.zsh.enableSyntaxHighlighting = mkOption {
       type = types.bool;
       default = false;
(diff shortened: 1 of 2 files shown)


ed9w2in6:
Personally, I am not using the likes of `home-manager` to manage my zsh configs for the moment.
I use some POSIX shell script to do that and it works well.

Without this option I have to awkwardly add `pkgs.zsh-autosuggestions` to packages via `nix-darwin` or `home-manager` then use some mechanism (e.g. `environment.etc` for `nix-darwin` or `home.file` for `home-manager`) to guarantee a stable path to be sourced.

This is a simple change and given we already added `enableSyntaxHighlighting` I think this similar feature is justified to be added.

ed9w2in6:
However, if we aim for (semi-)compatibility with nixos, then we should remove all these options that does NOT exist on nixos.
ref: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/zsh/zsh.nix

szcharlesji:
second this. Would be nice if nix-darwin's config options are the same as nixos's

PR #1438, opened 2025-05-08
programs/_1password{,-gui}: init modules

Enzime:
Based on @reckenrode's configs: https://github.com/reckenrode/nixos-configs/blob/22b8357fc6ffbd0df5ce50dc417c23a807a268a2/modules/by-name/1p/1password/darwin-module.nix

I didn't add any tests as these packages are unfree and I didn't think there's enough functionality worth testing here

The option names are to match NixOS, ideally the CLI module should probably be `programs._1password-cli` but we'll just wait for upstream on that

Fixes https://github.com/NixOS/nixpkgs/issues/254944

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -100,6 +100,7 @@
   ./services/nextdns
   ./services/jankyborders
   ./programs/_1password.nix
+  ./programs/_1password-gui.nix
   ./programs/bash
   ./programs/direnv.nix
   ./programs/fish.nix
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -99,6 +99,7 @@
   ./services/yabai
   ./services/nextdns
   ./services/jankyborders
+  ./programs/_1password.nix
   ./programs/bash
   ./programs/direnv.nix
   ./programs/fish.nix
--- a/modules/programs/_1password-gui.nix
+++ b/modules/programs/_1password-gui.nix
@@ -1,0 +1,49 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
+
+let
+  cfg = config.programs._1password-gui;
+in
+{
+  options = {
+    programs._1password-gui = {
+      enable = lib.mkEnableOption "the 1Password GUI application";
+
+      package = lib.mkPackageOption pkgs "1Password GUI" {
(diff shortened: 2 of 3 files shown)


reckenrode:
Thanks for upstreaming this. The only issue I’ve had with it is 1Password quits when it’s updated.

Enzime:
Tested this and it works on my machine :)

complexspaces:
The CLI package works great on my system too, thanks :)

remi-gelinas:
@Enzime You were able to get both the GUI and CLI installed with this? GUI seems to work fine, but I get a lot of permissions errors on `/private/tmp` during the CLI install phase that cuts it short, i.e.:
```
couldn't start daemon: Can't continue. We can't safely access "/private/tmp/nix-build-1password-cli-2.31.0.drv-0/com.agilebits.op.SingleUserEnvironment" because it's not owned by the current user. Change the owner or logged in user and try again.
```

Enzime:
Where did you get that error? What command did you run and what lines did you add to your configuration?

remi-gelinas:
> Where did you get that error? What command did you run and what lines did you add to your configuration?

I just updated to the latest master for both `nixpkgs` and `nix-darwin`, and enabled the new modules in my `nix-darwin` config:
```
{
  programs = {
    _1password.enable = true;
    _1password-gui.enable = true;
  };
}
```

I usually just build the system derivation as a smoke test after new changes before I switch - the repro config is in a public branch of my configs, so `nix build github:remi-gelinas/rosetta/70ed007f328ea24d7a9ccbf191f6f5d8c04a3680#darwinConfigurations.fixture.system` will reproduce the problem (or at least, that's the rev that's failing to build on my system - could be system differences causing it).

Enzime:
```
$ nix build --extra-experimental-features pipe-operators github:remi-gelinas/rosetta/70ed007f328ea24d7a9ccbf191f6f5d8c04a3680#darwinConfigurations.fixture.pkgs._1password-cli --override-input fonts github:nix-community/nixvim
```

Your `fonts` flake input is private so I had to override it to build 1Password CLI from your dotfiles, which works fine for me

I'm using Nix 2.28.3, maybe it's because you're using Lix?

PR #1440, opened 2025-05-11
openssh: set fallback `PATH` for running commands over SSH as `root`

Enzime:
By default, `root` uses `/bin/sh` as its default shell on macOS which does not execute any shell initialization code when run non-interactively like `/etc/profile` or `/etc/bashrc`. This means that `ssh root@localhost nix-store` will fail with `sh: nix-store: command not found` as it won't have any Nix managed directories in the `PATH`.

After a lot of investigation, I only managed to find 2 ways to implement a fix for this without needing to change `root`'s default shell, something which I avoided as users may want to SSH into other Apple system users that also have `/bin/sh` as their default shell.

One way I found was using `pam_env` which is included by default with macOS despite it not being in OpenPAM (I'm not sure if they're taking it from Linux-PAM's sources or elsewhere). However, as `pam_env` is not used by default, it would require modifying `/etc/pam.d/pam_sshd` which would require [`kTCCServiceSystemPolicySysAdminFiles` or Full Disk Access](https://github.com/nix-darwin/nix-darwin/pull/1120#discussion_r1817836999) and wouldn't even bring much benefit as macOS includes a super old version of `pam_env` that doesn't support expanding `@{HOME}` so `PATH` entries that refer to `$HOME` can't be expressed.

So in the end I decided against the extra hassle of ensuring `darwin-rebuild` is granted the correct permissions and instead opted to use `SetEnv` in `sshd_config` to set a static list of `PATH` entries that don't depend on any other environment variables like `$HOME` and `$USER`.

This should not lead to any issues as it should usually get overridden like in interactive shells where Bash will source `/etc/profile` and `/etc/bashrc` and Zsh will always source `/etc/zshenv` even in non-interactive shells.

See https://github.com/DeterminateSystems/nix-installer/issues/383 https://github.com/DeterminateSystems/nix-installer/pull/714

### DIFF
--- a/modules/services/openssh.nix
+++ b/modules/services/openssh.nix
@@ -29,5 +29,19 @@ in
         launchctl disable system/com.openssh.sshd
       fi
     '');
+
+    environment.etc."ssh/sshd_config.d/50-nix-path-fallback.conf".text = let
+      pathDirs = lib.splitString ":" config.environment.systemPath;
+      filteredPathDirs = lib.filter (dir: !lib.hasInfix "$" dir) pathDirs;
+      filteredPath = lib.concatStringsSep ":" filteredPathDirs;
+    in ''
+      # Set a fallback PATH that doesn't depend on any environment variables
+      # for when SSH is run with a command i.e. `ssh root@localhost nix-store`
+      # This is necessary for any users who have `/bin/sh` or `/bin/bash` as
+      # their default shells (`root` by default), this isn't necessary for Zsh
+      # as it will still execute `/etc/zshenv` whereas `/bin/sh` will not
+      # execute any files.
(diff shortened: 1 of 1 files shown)


emilazy (commented):
Shouldn’t this be conditional on the `enable` option? Looks okay other than that although I’m not a fan of the filtering.

Enzime:
My personal preference is to keep this unconditional as I think that this fix is without any downsides that I can see currently. I think it'll become more of a prominent issue once we merge #1341 as more people will probably try to deploy as `root` over SSH.

I've also made a comment upstream requesting they also implement the same solution to remove this issue: https://github.com/DeterminateSystems/nix-installer/issues/383#issuecomment-2869543062

PR #1442, opened 2025-05-12
gitlab-runner: write config as toml, don't clobber existing file

lheckemann:
The previous command would fail because of datetimes not being representable as JSON, wiping the config entirely because of the `sponge` invocation that doesn't care whether the program piped in fails.


Enzime:
`set -o pipefail` should work instead of needing to switch out `sponge` but I have no strong opinion about this

lheckemann:
I don't think it would: this would only cause the exit status to be passed on and the script to fail, but sponge would still run and overwrite the file.

Enzime (approved):

PR #1443, opened 2025-05-14
module: add autoraise service

NickHu:
Adds [AutoRaise](https://github.com/sbmpost/AutoRaise) as a service.


modules/services/autoraise.nix
```diff
@@ -0,0 +1,160 @@
...
+            delay = lib.mkOption {
+              type = nullOr int;
+              default = null;
+              description = "Raise delay, specified in units of pollMillis. Disabled if 0. A delay > 1 requires the mouse to stop for a moment before raising.";
+            };
+            focusDelay = lib.mkOption {
+              type = nullOr int;
+              default = null;
+              description = "Focus delay, specified in units of pollMillis. Disabled if 0. A delay > 1 requires the mouse to stop for a moment before focusing.";
+            };
```
Samasaur1, line 50: This option is only supported when AutoRaise is compiled with the `EXPERIMENTAL_FOCUS_FIRST` flag, which the version in Nixpkgs is not. I don't mind keeping this option for people who override the package so long as the description of this option is updated.

modules/services/autoraise.nix
```diff
@@ -0,0 +1,160 @@
...
+    };
+  };
+
+  config = (
+    lib.mkIf (cfg.enable) {
+      environment.systemPackages = [ cfg.package ];
+
+      launchd.user.agents.autoraise = {
+        command =
+          "${cfg.package}/Applications/AutoRaise.app/Contents/MacOS/AutoRaise"
```
Samasaur1:
```suggestion
          "${lib.getExe cfg.package}"
```

modules/services/autoraise.nix
```diff
@@ -0,0 +1,160 @@
...
+          <link xlink:href="https://github.com/sbmpost/AutoRaise"/>
+          for supported values.
+        '';
+      };
+    };
+  };
+
+  config = (
+    lib.mkIf (cfg.enable) {
+      environment.systemPackages = [ cfg.package ];
```
Samasaur1, line 147: Is this necessary? Running a second instance of AutoRaise is known to cause problems, so I'm not sure what the benefit of adding this to `systemPackages` is.

modules/services/autoraise.nix
```diff
@@ -0,0 +1,160 @@
...
+            };
+            invertIgnoreApps = lib.mkOption {
+              type = nullOr bool;
+              default = null;
+              description = "Turns the ignoreApps parameter into an includeApps parameter. The default is false.";
+            };
+            ignoreApps = lib.mkOption {
+              type = nullOr (listOf str);
+              default = null;
+              description = "Comma separated list of apps for which you would like to disable focus/raise.";
```
Samasaur1, line 84: Don't need to specify that the list is comma-separated, since that is abstracted away

modules/services/autoraise.nix
```diff
@@ -0,0 +1,160 @@
...
+            };
+            ignoreTitles = lib.mkOption {
+              type = nullOr (listOf str);
+              default = null;
+              description = "Comma separated list of window titles (a title can be an ICU regular expression) for which you would like to disable focus/raise.";
```
Samasaur1, line 89: Don't need to specify that the list is comma-separated, since that is abstracted away

modules/services/autoraise.nix
```diff
@@ -0,0 +1,160 @@
...
+            };
+            stayFocusedBundleIds = lib.mkOption {
+              type = nullOr (listOf str);
+              default = null;
+              description = "Comma separated list of app bundle identifiers that shouldn't lose focus even when hovering the mouse over another window.";
```
Samasaur1, line 94: Don't need to specify that the list is comma-separated, since that is abstracted away

modules/services/autoraise.nix
```diff
@@ -0,0 +1,160 @@
...
+            mouseDelta = lib.mkOption {
+              type = nullOr float;
+              default = null;
+              description = "Requires the mouse to move a certain distance. 0.0 = most sensitive whereas higher values decrease sensitivity.";
+            };
+            verbose = lib.mkOption {
+              type = nullOr bool;
+              default = null;
+              description = "Set to true to make AutoRaise show a log of events when started in a terminal.";
+            };
```
Samasaur1, line 114: For this option to be useful, the service needs to have `StandardOutPath` set. If you want, you can check whether this option is set and set `StandardOutPath` if so, but it's probably sufficient to just include a note in this description saying that it must be set.

Samasaur1 (changes requested):
This looks mostly good! thank you especially for including a test

modules/services/autoraise.nix
NickHu: I thought that was just standard practice (lets you do `man autoraise` I believe) but I guess I don't mind either way

NickHu:
Thanks for the review @Samasaur1. I just copied in the descriptions directly from upstream; do you really feel like they need rewording?

modules/services/autoraise.nix
Samasaur1: I think we're not super consistent about it, but since AutoRaise doesn't have a man page, I think adding it has more cons than pros (especially since `autoraise --help` just runs it normally)

Samasaur1:
Personally I would change them, but I see your point about just copying descriptions from upstream rather than making any changes ourselves. I'll resolve the comma-separated list comments, but I do think we need to note that `focusDelay` won't work without overriding the package.

modules/services/autoraise.nix
Samasaur1: The README for AutoRaise says

> *Note*: focusDelay is only supported when compiled with the "EXPERIMENTAL_FOCUS_FIRST" flag.

If it were me I would note this here in the description for `focusDelay`, but you could also add it to the description of the `package` option if you prefer

modules/services/autoraise.nix
Samasaur1: In the name of letting the descriptions of these options be identical to upstream, I think you can leave this as is, but should add an option for controlling the log file path like [the ipfs module does](https://github.com/nix-darwin/nix-darwin/blob/master/modules/services/ipfs.nix)

tests/services-autoraise.nix
```diff
@@ -0,0 +1,44 @@
+{ config, pkgs, ... }:
+
+let
+  autoraise = pkgs.runCommand "autoraise-0.0.0" { } "mkdir $out";
+in
+
+{
```
Samasaur1, line 7: We'll also need to set `system.primaryUser` since this is a user-scoped service.
```suggestion
{
  system.primaryUser = "test-autoraise-user";
  
```

Samasaur1 (changes requested):

PR #1446, opened 2025-05-16
defaults: add `com.apple.iCal` for managing Calendar.app

Enzime:


Enzime:
Tested this locally and it works 👍

It's very cool that `Calendar.app` immediately changes when `first day of week` or `TimeZone support enabled` is set through `defaults`

PR #1458, opened 2025-05-19
darwin-rebuild: use `NIX_REMOTE=daemon` even as `root`

emilazy:
Should fix https://github.com/nix-darwin/nix-darwin/issues/1455#issuecomment-2891233216.

### DIFF
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -11,6 +11,10 @@ fi
 export PATH=@path@
 export NIX_PATH=${NIX_PATH:-@nixPath@}
 
+# Use the daemon even as `root` so that resource limits, TLS and proxy
+# configuration, etc. work as expected.
+export NIX_REMOTE=${NIX_REMOTE:-daemon}
+
 showSyntax() {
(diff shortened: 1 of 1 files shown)


Enzime:
Is this being set for `nixos-rebuild`?

emilazy:
No but it probably should be, and it’s compatible with the behaviour from before the Plan. It’s worse on macOS because the default limit for open FDs is 256, which breaks many builds; skipping the daemon breaks those. The Nix default of skipping the daemon if you are `root` is pretty bad, especially for use‐cases like this; I would argue that it should be changed upstream, but this avoids breaking people who have TLS and proxy settings configured for the daemon.

Enzime:
Does `sudo NIX_REMOTE=daemon darwin-rebuild switch` work?

rafaelpirolla:
Seems to work, but https://github.com/zhaofengli/nix-homebrew/issues/83 is broken at the moment. Will test again after the merge.

rafaelpirolla:
Nice, it can use the env var. It works.

Enzime (approved):
LGTM

rafaelpirolla:
But it seems nix-homebrew doesn't... Oh lord.

PR #1459, opened 2025-05-19
linux-builder: Allow custom host & port

phaer:
I think it's useful to allow customization here.

Ran into this while playing around with an alternative implementation of a linux-builder at https://github.com/phaer/nixos-vm-on-macos. The underlying hypervisor doesn't seem to provide an interface for port forwarding, but does provide full virtio-net ips, so need to be able to customize host and port here but could otherwise reuse the module.

### DIFF
--- a/modules/nix/linux-builder.nix
+++ b/modules/nix/linux-builder.nix
@@ -80,6 +80,30 @@ in
       '';
     };
 
+    host = mkOption {
+      type = types.str;
+      default = "localhost";
+      defaultText = literalExpression ''"localhost"'';
+      description = ''
+        The host used for communicating with the build machine via SSH.
+
+        This is localhost for the default linux-builder, but might be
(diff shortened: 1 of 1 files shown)


Enzime (changes requested):
Can you configure the port for the native Linux builder?

phaer:
> Can you configure the port for the native Linux builder?

Not sure, I understand you correctly: Do you mean, getting the default value from there? If so, i just added that. Or do you mean to actually set the native Linux builders port here? That's already customizable upstream.

Samasaur1 (changes requested):
I'm not opposed to this in theory. Mostly I just worry that people may think that these new options set the hostname and port that the builder listens on as well, so I would say explicitly in the descriptions that this is not the case (perhaps with instructions on how to do so if desired?).

Issue #1457, opened 2025-05-19
“system activation must now be run as root”


emilazy:
nix-darwin has recently switched over to running the entire system activation process as `root`.

If you’re getting this error, it means that you need to use `sudo darwin-rebuild switch` instead of `darwin-rebuild switch` from now on.

If you’re using `nix run nix-darwin`, this is not necessary past the initial installation process and we recommend using `darwin-rebuild` directly instead. You can update your system (`nix flake update` or `nix-channel --update && sudo nix-channel --update`), run `darwin-rebuild switch` one more time to activate the new version, and then use `sudo darwin-rebuild switch` after that.

You might also get an error message describing the changes and referencing `system.primaryUser`. Following the instructions in the message should be enough to get your configuration migrated.

If you have any other issues or need clarification about anything in this announcement, please do open a new issue.

PR #1461, opened 2025-05-20
darwin-rebuild: try to re-run as root when needed

bryango:
This would ease the transition to root activation, as many have encountered:
- #1457

Basically, just try to re-run the whole `darwin-rebuild` script as root when necessary.
- Tested and works great on my machine!
- However, is this safe? Maybe there are edge cases that we need to address? I am always afraid of `sudo`...

### DIFF
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -144,7 +144,8 @@ if [ -z "$action" ]; then showSyntax; fi
 
 if [[ $action =~ ^switch|activate|rollback|check$ && $(id -u) -ne 0 ]]; then
   printf >&2 '%s: system activation must now be run as root\n' "$0"
-  exit 1
+  printf >&2 '%s: trying to re-run as root\n' "$0"
+  exec sudo "$0" "${origArgs[@]}"
 fi
 
 flakeFlags=(--extra-experimental-features 'nix-command flakes')


emilazy:
I would rather not do this because people will rely on it rather than fixing their invocations. Maybe we could give a migration hint after the error message, though, at least until 25.11?

emilazy:
> Actually, maybe it's fine to keep the old invocation? Namely, to always allow invoking `switch` without sudo and prompt for password when needed. I actually prefer this behavior as it feels easier for end users. **Then again, I am not sure if this is secure. This is my main concern.**

I don’t think there’s any security issue, but I have two concerns with it:

1. It’s inconsistent with `nixos-rebuild`. I would like our deployment tools to match NixOS as much as possible, to allow reuse of muscle memory and tooling. NixOS doesn’t default to doing this, so I wouldn’t want us to, either, unless NixOS changes that default.

2. I think it is just generally a good thing for system management tasks to expect to be run as `root` and for privilege escalation to be the user’s task. What nix-darwin does is fundamentally *`root`-y*, and command‐line system management tools on macOS tend not to try to elevate permissions implicitly.

Therefore I would be very happy to have better messaging around this in 25.05 for the migration period – it’s my bad for not expecting this to be an error people would have issues with – but I do not want to keep `sudo` in `darwin-rebuild`, especially indefinitely, just because of the strange way it worked previously.

bryango:
These are indeed reasonable concerns! I have updated the patch to only _suggest_ the sudo command. It currently looks like this on my machine:

```
$ darwin-rebuild switch
/run/current-system/sw/bin/darwin-rebuild: system activation must now be run as root. Try:

  sudo /run/current-system/sw/bin/darwin-rebuild switch
```

In this case, `$0` is the full path to the `darwin-rebuild` command so it looks a bit ugly. But maybe this _could_ be a good thing if `$PATH` wasn't properly set up? I am not sure haha.

Issue #1462, opened 2025-05-20
System activation broken


jcf:
With [Phase 1 of The Plan](https://github.com/nix-darwin/nix-darwin/pull/1341) complete, the requirement to activate as root broke my system, and I expect others might run into some of these same issues.

In order to surface the sort of output people might see for search engines (or more likely LLMs these days), and pool our resources towards fixing our systems, I'm opening an issue to document what I've seen so far.

## `nix-darwin switch` must now run as root

To start with, I ran my usual `nix flake update` earlier today with a wrapper script that looks like this:

``` sh
nix \
  --extra-experimental-features "nix-command flakes" \
  run nix-darwin -- --show-trace --flake "$PWD" switch "$@"
```

When invoked, I get the following error:

```
/nix/store/l8xwvdfhy3dkhk1viziiyaqchpimfhaz-darwin-rebuild/bin/darwin-rebuild: system activation must now be run as root
```

Some searching around ultimately led me to the locked #1457, which contains the following advice:

> If you’re using `nix run nix-darwin`, this is not necessary past the initial installation process and we recommend using `darwin-rebuild` directly instead. You can update your system (`nix flake update` or `nix-channel --update && sudo nix-channel --update`), run `darwin-rebuild switch` one more time to activate the new version, and then use `sudo darwin-rebuild switch` after that.

## 'darwin' was not found in the Nix search path

Having already updated my flake inputs, I invoked `darwin-rebuild switch` as instructed:

``` sh
$ darwin-rebuild switch
building the system configuration...
warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels' does not exist, ignoring
error: file 'darwin' was not found in the Nix search path (add it using $NIX_PATH or -I)
```

## Failed assertions

Going back to `nix run nix-darwin` I'm told I have a bunch of configuration that needs to be updated:

```
       error:
       Failed assertions:
       - The `system.activationScripts.postUserActivation` option has
       been removed, as all activation now takes place as `root`. Please
       restructure your custom activation scripts appropriately,
       potentially using `sudo` if you need to run commands as a user.

       - Previously, some nix-darwin options applied to the user running
       `darwin-rebuild`. As part of a long‐term migration to make
       nix-darwin focus on system‐wide activation and support first‐class
       multi‐user setups, all system activation now runs as `root`, and
       these options instead apply to the `system.primaryUser` user.

       You currently have the following primary‐user‐requiring options set:

       * `homebrew.enable`
[...]

PR #1469, opened 2025-05-23
nix.nixPath: Do not use environment.darwinConfig if no primaryUser

roberth:
This combination makes sense for a remotely managed darwin host with channels still enabled on that host, among probably other use cases.


emilazy:
I think we don’t want to do this because it could be a breaking change. Note that `environment.darwinConfig`’s default depends on state version and doesn’t reference the username for newer state versions; I would recommend setting `"/etc/nix-darwin/configuration.nix"` explicitly in many cases, but for `system.stateVersion` < 6, `system.requiresPrimaryUser` will be set in `modules/environment/default.nix` under a complex set of conditions that I think mean that this PR’s attempt to help out will never actually avoid flagging up an error.

But it’s possible I’ve lost track of something here?

roberth:
_Technically_ it can't be a breaking change, because `"${null}"` doesn't evaluate anyway, and that's exactly what my old configuration ran into after I had set `darwinConfig` to `null`.

Unless you want to redesign how these options interact, I don't really see how I could make this fix even less breaking.
I think I'd need some sort of example of what is broken.

roberth:
In hindsight, I made a mistake linking it to #1462. Sorry for making things seem complicated.
I encountered it more or less at the same time, but it's an independent problem that's not related to the triggering of that message.

emilazy:
Ah, is the idea that you are explicitly setting `environment.darwinConfig` to `null` because you don’t want to do any activation on the host (e.g. you set `system.tools.enable = false;`)? Then yes, I can see the use case, but I don’t think it really has anything to do with `system.primaryUser`, so the commit message confuses me.

Is there a reason you don’t set `nix.nixPath` explicitly? I suppose that my worry is that it will accidentally result in a system  that can’t re‐activate itself, but it doesn’t seem too plausible, so perhaps it’s fine. I wouldn’t be opposed to merging this with the commit message, then – but maybe it should float outside of the `cfg.channel.enable` conditional entirely, since after all it could also be useful to have the path set on a system where you aren’t using flakes but also aren’t using mutable channels? (Though you’d have to explicitly set `darwin` in `$NIX_PATH` to use `darwin-rebuild` with such a configuration anyway, so… I’ll leave it up to your preference.)

roberth:
I've left the message a bit more open ended.

> I suppose that my worry is that it will accidentally result in a system that can’t re‐activate itself

I believe that's somewhat up to the user if they choose to mess around with this, but then to prevent this in a comprehensible manner, it should probably be an `assertions` error with some suggestions.

> `system.tools.enable`

I was happy to keep `darwin-rebuild` around, but I suppose `nix shell nix-darwin#darwin-rebuild` is sufficient in case of some kind of emergency where my automation fails. That's a good suggestion!

Samasaur1 (approved):
Seems like the actual content of the PR is unobjectionable (avoiding a potential eval error). @emilazy any objections to merging?

emilazy:
Yeah I think it’s okay. I would prefer to float it outside of the channel conditional but I’m fine with merging this.

Samasaur1:
I'm gonna merge this as-is because it makes no substantive changes. We could move setting `darwin-config` outside the `channel` conditional in a follow-up PR.

PR #1472, opened 2025-05-23
feat(services.openssh): add settings

squat:
This commit introduces the `settings` option to configure OpenSSH.
Most of the configuration options are lifted verbatim from
https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/networking/ssh/sshd.nix
with a few exceptions:
1. `Macs` is renamed to `MACs` to comply with the upstream OpenSSH
   directive name;
1. `PubkeyAuthentication` is introduced; this directive defaults to `no`
   on the OpenSSH version shipped with NixOS but defaults to `yes` on
   the OpenSSH version shipped with macOS;
1. `Include` is introduced in order to mirror the default macOS
   configuration, which differs from the upstream OpenSSH default;
1. `Subsystem` is introduced in order to mirror the default macOS
   configuration, which differs from the upstream OpenSSH default;
1. `AuthorizedKeyFiles` is set to `[ ".ssh/authorized_keys" ]` to mirror
   the default macOS configuration, which differs from the upstream
   OpenSSH default.
1. `UsePAM` defaults to `yes` on NixOS but defaults to `no`
   on the OpenSSH version shipped with NixOS but defaults to `yes` on
   the OpenSSH version shipped with macOS;

Other notable settings:
1. `KeyAlgorithms` uses the NixOS default rather than the default from
   the OpenSSH version that ships with macOS for better security.
1. `MACs` uses the NixOS default rather than the default from
   the OpenSSH version that ships with macOS for better security.

Signed-off-by: squat <lserven@gmail.com>

### DIFF
--- a/modules/services/openssh.nix
+++ b/modules/services/openssh.nix
@@ -1,7 +1,83 @@
-{ config, lib, ... }:
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
 
 let
   cfg = config.services.openssh;
+  # Lifted from nixpkgs: https://github.com/squat/nixpkgs/blob/e52d633a638c607533ececf4eb9653ec1f93e3d4/nixos/modules/services/networking/ssh/sshd.nix#L20-L78.
+  settingsFormat =
+    let
+      # reports boolean as yes / no
+      mkValueString =
+        v:
+        if lib.isInt v then
+          toString v
+        else if lib.isString v then
+          v
+        else if true == v then
+          "yes"
+        else if false == v then
+          "no"
+        else
+          throw "unsupported type ${builtins.typeOf v}: ${(lib.generators.toPretty { }) v}";
+
+      base = pkgs.formats.keyValue {
+        mkKeyValue = lib.generators.mkKeyValueDefault { inherit mkValueString; } " ";
+      };
+      # OpenSSH is very inconsistent with options that can take multiple values.
+      # For some of them, they can simply appear multiple times and are appended, for others the
+      # values must be separated by whitespace or even commas.
+      # Consult either sshd_config(5) or, as last resort, the OpehSSH source for parsing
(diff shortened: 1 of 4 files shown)


squat:
Dangit, tests are failing because the check relies on the `-G` flag, which was added in [OpenSSH 9.3](https://www.openssh.com/txt/release-9.3), but the test environment uses 9.1

Enzime:
The tests still fail because OpenSSH upstream doesn't have the `UsePAM` option available in Apple's OpenSSH. I would suggest not checking if the configuration is correct.

If you really wanted to check, you could package https://github.com/apple-oss-distributions/OpenSSH in Nixpkgs and then use that package with `-G`

squat:
Thanks @Enzime! I'll take a look. I think it should also be possible to do a check pre-9.3 by only using the `-T` flag and adding more test config to satisfy sshd, like host keys. If it's too annoying for the payoff then I'll skip the check

squat:
> The tests still fail because OpenSSH upstream doesn't have the UsePAM option available in Apple's OpenSSH. I would suggest not checking if the configuration is correct.

It looks like this was possible to fix with `pkgs.openssh.override { withPAM = true; }` to configure Nixpkgs OpenSSH with OpenPAM on Darwin.

Checks are passing now. Installation is still failing because a conflict between the sshd_config this module generates and the one on macOS by default. I'll fix that by moving the generated config into one that is included instead.

squat:
Gentle ping @Enzime since tests are passing now

Samasaur1:
Can you undo the formatting-only changes to `release.nix`?

squat:
> ~Can you undo the formatting-only changes to release.nix?~

~Happily, what formatter does the project use? This was all formatted automatically on save with nixfmt-rfc-style~

Done! PTAL @Samasaur1

PR #1490, opened 2025-05-28
darwin-rebuild `--use-remote-sudo` flag

craigfurman:
Similar to the one in nixos-rebuild. Restores behaviour similar to before the switch to running as root, but behind a flag.

I am definitely not sure that this contains no rough edges, or works for most users, hence the draft state. I tested this by running `nix run .#darwin-rebuild -- switch --use-remote-sudo` and validating that it could build and activate my system configuration.

Note that I didn't actually test switching with private flake inputs (yet), I'd be grateful if someone who was already using these and hasn't been able to do this on recent nix-darwin versions, to test this branch 🙏 

I should also mention that I haven't tested the "remote" aspect of this, at all. The flag name here is just cargo-culted from nixos-rebuild as a starting point.

This was developed by reading https://github.com/nix-darwin/nix-darwin/commit/516dbe1fa40548945f18135875ed22228db4ce33 and adding flag-gated sudo (back) to the operations that appear to require it.

My motivation for writing this is to address https://github.com/nix-darwin/nix-darwin/issues/1471 and https://github.com/nix-darwin/nix-darwin/issues/1484, because I also want to make use of private flake inputs, building the config with my unprivileged user that has ssh key access. I didn't dig into it, but if https://github.com/nix-darwin/nix-darwin/issues/1485 is indeed related to the switch to requiring running as root, this could help.

Is this the sort of thing maintainers would be interested in? I did see a comment about eventually switching to nixos-rebuild-ng, but I don't know how far away that is and flag could help paper over the cracks until the project is ready for that work. If so, I'm happy to add docs to get this over the line. Please let me know if anything else is missing too, I didn't spent a long time looking into the repo's CI setup etc so I'm sure there's something!

### DIFF
--- a/pkgs/nix-tools/darwin-rebuild.sh
+++ b/pkgs/nix-tools/darwin-rebuild.sh
@@ -28,8 +28,15 @@ showSyntax() {
   echo "                             [--no-update-lock-file] [--no-write-lock-file]" >&2
   echo "                             [--override-input input flake] [--update-input input]" >&2
   echo "                             [--no-registries] [--offline] [--refresh]]" >&2
+  echo "               [--use-remote-sudo]" >&2
   echo "               [--substituters substituters-list] ..." >&2
   exit 1
+}
+
+sudo() {
+  # We use `env` before our command to ensure the preserved PATH gets checked
+  # when trying to resolve the command to execute
+  command sudo -H --preserve-env=PATH --preserve-env=SSH_CONNECTION env "$@"
 }
 
 # Parse the command line.
@@ -42,6 +49,7 @@ profile=@profile@
 action=
 flake=
 noFlake=
+useRemoteSudo=
 
 while [ $# -gt 0 ]; do
   i=$1; shift 1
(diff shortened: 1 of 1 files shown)


Enzime:
For NixOS, the `--use-remote-sudo` flag is for running Nix commands over SSH on a target host (specified by `--target-host`).

Currently `darwin-rebuild` doesn't support any remote building (`--build-host`) or remote deploying (`--target-host`) like `nixos-rebuild`.

It would be great if you could backport these two flags as well as `--use-remote-sudo`

craigfurman:
Hi @Enzime, thanks for the context! I hadn't actually realised that darwin-rebuild doesn't support remote building, so perhaps `--use-remote-sudo` is a poor name for the flag added here (although it does at at least add some symmetry with nixos-rebuild). I'm open to suggestions there.

re: adding --build-host and --target-host, I am unlikely to find time for that soon. Would maintainers be open to adding this incremental improvement without blocking on these extra features?

I might return to add these flags, but can't make any promises.

Enzime:
Our primary goal is to match `nixos-rebuild` which requires being called with `sudo` if run locally, so I wouldn't accept this PR without it being specific to remote building

PR #1491, opened 2025-05-29
programs/direnv: fix silent option

xuanduc987:
Port of https://github.com/NixOS/nixpkgs/pull/398717


Samasaur1:
I'm also wondering whether we want to do what home-manager does and check the direnv version, since we aren't in sync with the package version like NixOS is

xuanduc987:
@Samasaur1 


>  Specifically, the symlinkJoin has makeBinaryWrapper in nativeBuildInputs but doesn't use it

Thank you, I fixed it.

> and the DIRENV_CONFIG env variable is still set systemwide instead of having a default value only for the direnv program

Because creating direnv wrapper didn't work because the content of shell hook still refer to the original direnv binary.
On nixpkgs, there is a follow up fix which I only notice after I fixed it on my local https://github.com/NixOS/nixpkgs/pull/402399
The `finalPackage` was an attempt to fix direnv shell hook, but it does not works without the global `DIRENV_CONFIG` variable.

> I'm also wondering whether we want to do what home-manager does and check the direnv version, since we aren't in sync with the package version like NixOS is

`nix-darwin` now requires `nixpkgs` with version >= its own version so I think it would be OK to merge this PR into main & 25.05

Samasaur1:
> Because creating direnv wrapper didn't work because the content of shell hook still refer to the original direnv binary. On nixpkgs, there is a follow up fix which I only notice after I fixed it on my local [NixOS/nixpkgs#402399](https://github.com/NixOS/nixpkgs/pull/402399) The `finalPackage` was an attempt to fix direnv shell hook, but it does not works without the global `DIRENV_CONFIG` variable.

Okay, make sense. I'd still prefer that you mirror the use of `finalPackage` as NixOS does, even if it isn't effective.

> > I'm also wondering whether we want to do what home-manager does and check the direnv version, since we aren't in sync with the package version like NixOS is
> 
> `nix-darwin` now requires `nixpkgs` with version >= its own version so I think it would be OK to merge this PR into main & 25.05

Ah yes, I keep forgetting about this.

xuanduc987:
@Samasaur1 I had added `finalPackage` readonly option

Samasaur1 (approved):

PR #1495, opened 2025-05-30
nix: enable `sandbox` by default

Enzime:
This should not cause issues for most builds as of https://github.com/NixOS/nix/pull/12570 which is available in Nix 2.24.13, 2.24.14 and 2.27.0 onwards.

The corresponding change in Lix https://gerrit.lix.systems/c/lix/+/2861 is available in Lix 2.93.0 onwards.


khaneliman:
One thing to consider: https://github.com/NixOS/nixpkgs/issues/394972 we ran into issues with building rust packages with sandbox enabled.

Enzime:
Do you know if adding `/etc/ssl/certs/ca-certificates.crt` to `nix.settings.extra-sandbox-paths` works?

khaneliman:
> Do you know if adding `/etc/ssl/certs/ca-certificates.crt` to `nix.settings.extra-sandbox-paths` works?

Doesn't seem like it, tested enabling sandbox and building `vimPlugins.avante-nvim` and I get the error from the issue. Then I added that to my extra sandbox paths and still get the same error.

Enzime:
What if you add the store path for `/etc/ssl/certs/ca-certificates.crt` instead?

zhaofengli:
Just a note that https://github.com/NixOS/nixpkgs/issues/343576 is another issue that might occasionally bite us.

al3xtjames:
It's also worth noting that Hydra doesn't build with the sandbox enabled (IIRC), so the binary cache will end up hiding build failures that would otherwise occur locally in the sandbox.

mirkolenz:
I've been using `nix.settings.sandbox = "relaxed";` for a few months now and everything worked perfectly. In the last few weeks however, rebuilding my nix-darwin config became incredibly slow. The individual derivations like packages built in regular time, but as soon as the "top-level" derivations could be built (e.g., etc-registry or home-manager-applications), each build process took minutes to complete instead of seconds. I still haven't figured out what exactly led to this behavior (I even uninstalled nix-darwin and nix completely and set up a new nix store, but without success), but today I tried rebuilding without sandbox enabled and the error was gone. For me, doing `darwin-rebuild --option sandbox false/relaxed` is enough to reproduce this behavior. Could of course also be caused by some strange config I have in my setup, just wanted to let you know about my observations regarding the sandbox setting.

PR #1499, opened 2025-06-02
Avoid confusing users with future deprecations

adamcstephens:
I appreciate the spirit of the current wording, but it's confusing to mention options that don't exist and warning about _future_ deprecations. 

When the replacement options are available, we can add back the deprecation as a warning.

### DIFF
--- a/modules/system/primary-user.nix
+++ b/modules/system/primary-user.nix
@@ -1,6 +1,5 @@
 {
   lib,
-  options,
   config,
   ...
 }:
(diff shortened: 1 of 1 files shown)


adamcstephens:
I'll note that I've seen this come up in multiple support channels. Where someone _wants_ to migrate to the new options, but of course they don't exist.

Any thoughts @emilazy ?

emilazy:
I’m okay with this – I’m not sure if it’ll just cause people to ask what the plan for the future is instead, but I’ve seen people do that anyway, so it seems worth trying :)

emilazy:
Do you mind handling the `git cherry-pick -x` for 25.05?

PR #1503, opened 2025-06-06
Preserve `PATH` variable when using `sudo` for Homebrew

bromanko:
Some systems set `secure_path` in sudoers. When this is set
the `PATH` variable is not set in the sudo environment. Using
`--preserve-env=PATH` and wrapping the call with `env` ensures that the PATH env var is set properly in those systems.

This is similar to the issue with [darwin-rebuild](https://github.com/nix-darwin/nix-darwin/issues/798) not working with sudo on these systems.

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -802,6 +802,7 @@ in
       if [ -f "${cfg.brewPrefix}/brew" ]; then
         PATH="${cfg.brewPrefix}:${lib.makeBinPath [ pkgs.mas ]}:$PATH" \
         sudo \
+          --preserve-env=PATH \
           --user=${escapeShellArg cfg.user} \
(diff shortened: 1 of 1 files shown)


Samasaur1 (approved):
This seems reasonable to me. We've just set `PATH` on the previous line and the only command we run is `brew`, which is in that new path entry.

I suppose there's a slight security risk for things that `brew` calls/looks up in `PATH`, but by that point we've dropped down to the user, so that's mostly mitigated.

bromanko:
Just rebased

Enzime:
It looks like the rebase failed, could you try again with `git fetch --all && git rebase -i upstream/master`?

bromanko:
how bout now

PR #1511, opened 2025-06-11
_1password-gui: fix App Management permission missing as `root` over SSH

Enzime:


Enzime:
From my testing, this PR didn't work when used with `clan machines update` 

I'll need to spend more time debugging this later, as a workaround I just granted `Full Disk Access` to `sshd-keygen-wrapper` :)

Enzime:
Related #1396

PR #1518, opened 2025-06-21
users: support changing home directory if currently set to `/var/empty`

Enzime:


modules/users/default.nix
```diff
@@ -302,12 +303,22 @@ in
...
+            # Support updating home directory if the user's home directory is set to
+            # `/var/empty` as there won't be any files to migrate.
+            homeDir=$(dscl . -read ${dsclUser} NFSHomeDirectory)
+            if [[ "$homeDir" == "/var/empty" && "$homeDir" != "${v.home}" ]]; then
+              dscl . -create ${dsclUser} NFSHomeDirectory ${v.home}
+            fi
+
+            # Always create the home directory as this command is idempotent and works
+            # when the user's home directory is set to `/var/empty`. This is necessary
+            # if we just created a user or changed the user's home directory.
```
YorikSar, line 318: Could you please put this comment behind `optionalString` as well to keep the result from becoming confusing when `v.createHome == false`?

modules/users/default.nix
```diff
@@ -302,12 +303,22 @@ in
...
-            # `sysadminctl -addUser` won't create the home directory if we use the `-home`
-            # flag so we need to do it ourselves
-            ${optionalString (v.home != null && v.createHome) "createhomedir -cu ${name} > /dev/null"}
           fi
 
+          ${optionalString (v.home != null) ''
+            # Support updating home directory if the user's home directory is set to
+            # `/var/empty` as there won't be any files to migrate.
+            homeDir=$(dscl . -read ${dsclUser} NFSHomeDirectory)
+            if [[ "$homeDir" == "/var/empty" && "$homeDir" != "${v.home}" ]]; then
```
YorikSar, line 312: I think if we add a check for `$(realpath ${v.home}) == $(realpath $homeDir)` then https://github.com/nix-darwin/nix-darwin/issues/1256 will be solved and https://github.com/nix-darwin/nix-darwin/pull/1599 will be safe to apply.

modules/users/default.nix
YorikSar, line 205: This check also needs to be updated.

PR #1520, opened 2025-06-22
networking.applicationFirewall: init

Prince213:
This PR adds a new module `networking.applicationFirewall` for configuring the Application Layer Firewall.

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -49,6 +49,7 @@
   ./system/version.nix
   ./time
   ./networking
+  ./networking/applicationFirewall.nix
   ./nix
   ./nix/linux-builder.nix
   ./nix/nix-darwin.nix
--- a/modules/networking/applicationFirewall.nix
+++ b/modules/networking/applicationFirewall.nix
@@ -1,0 +1,48 @@
+{ config, lib, ... }:
+let
+  cfg = config.networking.applicationFirewall;
+
+  socketfilterfw =
+    option: value:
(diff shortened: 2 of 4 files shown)


emilazy:
We have `system.defaults.alf.*`; it seems like there’s overlap between the options? (Though IIRC I had troubles using the existing settings, maybe because of permissions stuff.)

Prince213:
> We have `system.defaults.alf.*`; it seems like there’s overlap between the options? (Though IIRC I had troubles using the existing settings, maybe because of permissions stuff.)

Ah right.  I believe that should be removed as it's no longer supported by Apple and should be removed.  What do you think?

https://support.apple.com/en-us/121011

> Application Firewall settings are no longer contained in a property list. If your app or workflow relies on changing Application Firewall settings by modifying /Library/Preferences/com.apple.alf.plist, then you need to make changes to use the `socketfilterfw` command line tool instead.

Enzime:
I wonder if we should name this as `networking.firewall` to more closely match NixOS

Prince213:
> I wonder if we should name this as `networking.firewall` to more closely match NixOS

Yeah it could be, just that I'm not sure about the potential addition of the packet based firewall.  Maybe we could put that under networking.packetFirewall?

Prince213:
Existing `system.defaults.alf` configurations are now migrated to using `networking.applicationFirewall`.  A warning is also added to encourage migration.

Prince213:
Thanks!

PR #1525, opened 2025-06-29
fix: restart Dock with `killall` without linux only `-q` option

ed9w2in6:
Fixes: https://github.com/nix-darwin/nix-darwin/issues/1524

### DIFF
--- a/modules/system/defaults-write.nix
+++ b/modules/system/defaults-write.nix
@@ -153,7 +153,7 @@ in
 
         ${optionalString (length dock > 0) ''
           echo >&2 "restarting Dock..."
-          killall -qu ${escapeShellArg config.system.primaryUser} Dock || true
+          killall -u ${escapeShellArg config.system.primaryUser} Dock || true
         ''}
       '';
 


emilazy:
What problem did you observe with this? What macOS version?

`-q` behaves as expected for me:

```
yuyuko:~
❭ killall -qu emily Dock

yuyuko:~
❭ killall -qu emily non-running-process

yuyuko:~  1
❭ killall -u emily non-running-process
No matching processes belonging to you were found

yuyuko:~  1
❭
```

and is documented in `killall(1)`.

ed9w2in6:
@emilazy I see, I am on Big Sur 11.5 so that should be the reason.

I understand that from 25.11 nix-darwin targets Sonoma 14 or newer, but 25.05 would still support down to Big Sur 11.3.

However the 25.05 branch also contained `-q`, so maybe just a fix to that branch?

https://github.com/nix-darwin/nix-darwin/blob/536f951efb1ccda9b968e3c9dee39fbeb6d3fdeb/modules/system/defaults-write.nix#L159

And is `-q` really necessary?  If I am understanding it correctly the only case where it would fail to find the Dock process is when the `primaryUser` supplied does exist but its NOT the primary user (e.g. `root`), so it should fail, no?

ed9w2in6:
@emilazy 
May I ask if this can be at least merged / cherrypicked to branch `nix-darwin-25.05`?

I believe this should be done so since 25.05 aims to support down to Big Sur 11.3, in which the `-q` flag is not available.

I don't have strong opinion on merging to master but I do believe the `-q` flag is not necessary for master too.

I shall be most grateful if this pull request can be proceeded, closing or merging is both good for me.
I just don't want this hanging in my work log 😂.

PR #1526, opened 2025-06-30
github-runner/service.nix: fix missing argument in workDir assertion

thomasjm:
There's a typo here that was making `workDir` impossible to set to a string.


Samasaur1 (approved):
Thanks for the contribution!

Samasaur1:
not sure what happened there; let's see if trying again fixes it

PR #1529, opened 2025-07-01
launchd+defaults: Escape XML generated by toPlist

linnnus:
**Change**
NixOS/nixpkgs#356502 introduced an argument `escape` to `lib.generators.toPlist` which makes it escape special characters in the XML. This PR makes nix-darwin use it everywhere.

This is a breaking change, as [discussed under the original PR](https://github.com/nixos/nixpkgs/pull/356502#issuecomment-2480637115).

**Rationale**
The fact that contents aren't escaped by default is [rather unintuitive](https://discourse.nixos.org/t/correctly-running-cloned-flake/55960/11), especially when an API consumer isn't even aware that their input is being used to generate XML.

### DIFF
--- a/modules/launchd/default.nix
+++ b/modules/launchd/default.nix
@@ -90,7 +90,7 @@ let
         serviceConfig.ProgramArguments = mkIf (config.command != "") [
           "/bin/sh"
           "-c"
-          "/bin/wait4path /nix/store &amp;&amp; exec ${config.command}"
+          "/bin/wait4path /nix/store && exec ${config.command}"
         ];
         serviceConfig.EnvironmentVariables = mkIf (env != {}) env;
       };
--- a/modules/launchd/default.nix
+++ b/modules/launchd/default.nix
@@ -9,7 +9,7 @@ let
 
   toEnvironmentText = name: value: {
     name = "${value.serviceConfig.Label}.plist";
-    value.text = generators.toPlist { } value.serviceConfig;
+    value.text = generators.toPlist { escape = true; } value.serviceConfig;
   };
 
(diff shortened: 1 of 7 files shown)


linnnus:
@emilazy I've added an evaluation warning [as discussed on Matrix](https://matrix.to/#/!MMDPZwBJsBXZCsLkuj:matrix.org/$iGsiAQpTDgfewpFSB7En60gwnchSRV3pQHW_4cnCKz0?via=nixos.org&via=matrix.org&via=catgirl.cloud). Apologies for the delay.

eblume:
Hello, I am a nix-darwin user trying to track down the source of a few dozen instances of the following during my nix-darwin environment builds:
```
evaluation warning: Using `lib.generators.toPlist` without `escape = true` is deprecated
```
... which showed up some time in the past few days. I think I understand from this PR and the linked nixpkgs issue that this is a known breaking change.

If that is correct, my question is - is there anything I, as a user, need to do other than just wait for this change to make its way out to the master branch? Thanks!

Samasaur1:
> If that is correct, my question is - is there anything I, as a user, need to do other than just wait for this change to make its way out to the master branch? Thanks!

You are correct on both counts.

linnnus:
@eblume, in most cases this should not matter. The only thing to worry about is if you were doing manual escaping of XML special-characters, as those will now be doubly escaped. For example, this was previously a common pattern:

```nix
launchd.agents."some-daemon".serviceConfig.programArguments = [
  "/bin/sh"
  "-c"
  "wait4path /nix/store &amp;&amp; /nix/store/.../bin/some-daemon"
];
```

PR #1530, opened 2025-07-03
Fix sudo instructions in README.md

edejong:
Update instructions to run `darwin-rebuild` with `sudo -i`.

This removes the warnings about files being written to `/var/root`, and general unfriendliness (see #1457)

This is in line with the NixOS documentation: https://nixos.org/manual/nixos/stable/#sec-changing-config

I removed and then avoided nix-darwin for a long time because of these warnings. I think it's important to ensure new users are not presented with warnings like this the first thing they do.

(Related: it would be great if a fresh install of nix-darwin didn't result in errors running `nix config check`: `[FAIL] Multiple versions of nix found in PATH`.

It might be perfectly fine to have multiple Nix versions, but a new user just sees a system that's already highly invasive by requiring root privileges resulting in what seems to be "broken nix installation" no matter what they try.

I don't know how many people who try nix-darwin for the first time are deterred by this, but it can't just be me. I haven't tried using Nix on other platforms yet, do they manage to get around this boot-strapping issue?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,9 +98,9 @@ Unlike NixOS, `nix-darwin` does not have an installer, you can just run `darwin-
 
 ```bash
 # To use Nixpkgs unstable:
-sudo nix run nix-darwin/master#darwin-rebuild -- switch
+sudo -i nix run nix-darwin/master#darwin-rebuild -- switch
 # To use Nixpkgs 25.05:
-sudo nix run nix-darwin/nix-darwin-25.05#darwin-rebuild -- switch
+sudo -i nix run nix-darwin/nix-darwin-25.05#darwin-rebuild -- switch
 ```
 
 ### Step 3. Using `nix-darwin`
@@ -108,7 +108,7 @@ sudo nix run nix-darwin/nix-darwin-25.05#darwin-rebuild -- switch
 After installing, you can run `darwin-rebuild` to apply changes to your system:
 
 ```bash
-sudo darwin-rebuild switch
+sudo -i darwin-rebuild switch
 ```
 
 #### Using flake inputs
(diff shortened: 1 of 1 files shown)


Samasaur1 (approved):

Samasaur1 (commented):
Per discussion on Matrix, we [explicitly handle](https://github.com/nix-darwin/nix-darwin/blob/7220b01d679e93ede8d7b25d6f392855b81dd475/pkgs/nix-tools/darwin-rebuild.sh#L5-L9) the case of running `darwin-rebuild` with `sudo`, so you shouldn't be seeing this warning. Perhaps something else is the root cause of this issue?

edejong:
Happy to dig into it further, but I just installed nix-darwin on a clean install of my MacBook and I get the same warning.

Perhaps it's caused by some subprocess, as `HOME` isn't exported [here](https://github.com/nix-darwin/nix-darwin/blob/7220b01d679e93ede8d7b25d6f392855b81dd475/pkgs/nix-tools/darwin-rebuild.sh#L5-L9)?

edejong:
@Samasaur1 I tested all commands in [README.md](https://github.com/nix-darwin/nix-darwin/blob/master/README.md): `sudo nix run nix-darwin/master#darwin-rebuild -- switch` throws the warning before even fetching nix-darwin, so that’s just what you get when running `sudo nix <anything>`. Same for `sudo nix-channel`.

I updated the PR to no longer suggest using `sudo -i` when running `darwin-build` or `darwin-uninstall` because those indeed do not emit the warning :)

DavidDTA:
I believe the correct command to recommend is
```
nix shell nix-darwin/master#darwin-rebuild --command sudo darwin-rebuild switch
```
This will run nix as the user but darwin-rebuild as root

PR #1531, opened 2025-07-04
treewide: introduce formatter & format all Nix files

Prince213:
This PR introduces a new flake input, treefmt-nix, to do treewide formatting.

Reference: https://github.com/NixOS/rfcs/pull/166


flake.nix
```diff
@@ -3,88 +3,153 @@
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+    treefmt-nix = {
+      url = "github:numtide/treefmt-nix";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
```
khaneliman: In HM I got backlash introducing the dependency to the flake inputs. Opted for a more manual approach to prevent users from having to grab dev dependencies when they are just consuming the flake and not developing for it. 

```nix
        formatter = forAllPkgs (
          pkgs:
          pkgs.treefmt.withConfig {
            runtimeInputs = with pkgs; [
              nixfmt-rfc-style
              deadnix
              keep-sorted
            ];
            settings = pkgs.lib.importTOML ./treefmt.toml;
          }
        );
```

Otherwise, if you introduce partitioned flake attributes you could have a dev flake and house it there to prevent it, as well.

Prince213: Okay, I'll add a dev flake.

Samasaur1:
I think the most important part of doing treewide formatting is to enforce it afterwards, so ideally we'd add a CI check to make sure that new PRs are formatted

Samasaur1:
Also I agree that I would definitely prefer not having an extra flake input. I suppose adding a dev flake is okay, although it feels overkill

Prince213:
> Also I agree that I would definitely prefer not having an extra flake input. I suppose adding a dev flake is okay, although it feels overkill

We could reuse it for CI like Nixpkgs.

mirkolenz:
What about using `nixfmt-tree`? This could remove the need for an extra flake input or a dev flake.

PR #1538, opened 2025-07-12
fix: launchd user agent cleanup

### LINKED ISSUE #1539: User launch agents are not properly cleaned up if all of them are removed
Hi

I believe there is bug in the current implementation of the user launch agent's activation script. Currently, if no `config.environment.userLaunchAgents` is no text regarding controlling the user launch agents is added in the activation scripts (see condition below). Therefore the clean up section is not executed in a system rebuild. Therefore, if the current active generation has some user launch agent active, those agents won't be unloaded or removed.

```nix
    system.activationScripts.userLaunchd.text = let
      user = lib.escapeShellArg config.system.primaryUser;
>> CONDITION
    in mkIf (config.launchd.user.envVariables != { } || userLaunchAgents != [ ]) ''
<<
      # Set up user launchd services in ~/Library/LaunchAgents
      echo "setting up user launchd services..."

      ${concatStringsSep "\n" (launchdVariables "sudo --user=${user} --" config.launchd.user.envVariables)}

      ${optionalString (builtins.length userLaunchAgents > 0) ''
      sudo --user=${user} -- mkdir -p ~${user}/Library/LaunchAgents
      ''}
      ${concatMapStringsSep "\n" (attr: userLaunchdActivation attr.target) userLaunchAgents}
>> CLEANUP
      for f in /run/current-system/user/Library/LaunchAgents/*; do
        [[ -e "$f" ]] || break  # handle when directory is empty
        f=''${f#/run/current-system/user/Library/LaunchAgents/}

        if [[ ! -e "${cfg.build.launchd}/user/Library/LaunchAgents/$f" ]]; then
          echo "removing user service $(basename "$f" .plist)" >&2
          sudo --user=${user} -- launchctl unload ~${user}/Library/LaunchAgents/"$f" || true
          if [[ -e ~${user}/Library/LaunchAgents/"$f" ]]; then
            sudo --user=${user} -- rm -f ~${user}/Library/LaunchAgents/"$f"
          fi
        fi
      done
<<
    '';
```

I don't think this can be the desired behaviour. I created a PR #1538 to fix this and added a test to ensure that the cleanup script is unconditionally present.

This is my first PR and issue in this repo. Please let me know if I missed something or if there is a better way to fix this.

Thank you

A-jay98:
fixes #1539

### DIFF
--- a/modules/system/launchd.nix
+++ b/modules/system/launchd.nix
@@ -136,9 +136,11 @@ in
 
     system.activationScripts.userLaunchd.text = let
       user = lib.escapeShellArg config.system.primaryUser;
-    in mkIf (config.launchd.user.envVariables != { } || userLaunchAgents != [ ]) ''
-      # Set up user launchd services in ~/Library/LaunchAgents
+    in ''
       echo "setting up user launchd services..."
+
+      ${optionalString (config.launchd.user.envVariables != { } || userLaunchAgents != [ ]) ''
+      # Set up user launchd services in ~/Library/LaunchAgents
 
       ${concatStringsSep "\n" (launchdVariables "sudo --user=${user} --" config.launchd.user.envVariables)}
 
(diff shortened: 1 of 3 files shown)


mightyiam:
Does the added test fail when the implementation change is not applied?

A-jay98:
Yes. 
Screenshot 2025-07-20 at 9 29 44 PM

mightyiam (commented):
I think this nesting of one long optional string inside another is not the nicest possible form. How about concatenation instead?

PR #1542, opened 2025-07-16
services/xquartz: init

Feyorsh:
This is a simple module to manage the XQuartz launch daemon/agent with nix-darwin instead of manually running `xquartz-install` (or using Homebrew).

I've been using this module in my personal config for several months without issue, including with X forwarding over SSH.
You can test to make sure it works with a program like `xeyes` or `oneko`.

### DIFF
--- a/modules/services/xquartz.nix
+++ b/modules/services/xquartz.nix
@@ -1,0 +1,62 @@
+{ lib, pkgs, config, ... }:
+with lib;
+let
+  cfg = config.services.xquartz;
+  xauth = pkgs.xorg.xauth;
+in {
+  meta.maintainers = [ lib.maintainers.feyorsh or "feyorsh" ];
+
+  options.services.xquartz = {
+    enable = mkEnableOption "XQuartz";
+    package = mkOption {
+      type = types.package;
+      default = pkgs.xquartz;
+      description = "The XQuartz package to use.";
+    };
+  };
+
+  config = mkIf cfg.enable {
+    environment = let
+      daemon = "org.nixos.xquartz.privileged_startx.plist";
+      agent = "org.nixos.xquartz.startx.plist";
+      plistLocation = name: builtins.head (builtins.match ".*cp ([^[:space:]]+-${name}) .*" (builtins.readFile "${cfg.package}/bin/xquartz-install"));
+    in {
+      systemPackages = [ cfg.package ];
+
+      launchDaemons.${daemon}.source =
+        (pkgs.substitute {
+          src = plistLocation daemon;
+          substitutions = [
+            "--replace-fail"
+            "@PRIVILEGED_STARTX@"
+            "${cfg.package}/libexec/privileged_startx"
+
+            "--replace-fail"
+            "@PRIVILEGED_STARTX_D@"
+            "${cfg.package}/etc/X11/xinit/privileged_startx.d"
+          ];
+        }).outPath;
+
+      launchAgents.${agent}.source =
+        (pkgs.substitute {
+          src = plistLocation agent;
+          substitutions = [
+            "--replace-fail"
+            "@LAUNCHD_STARTX@"
+            "${cfg.package}/libexec/launchd_startx"
+
+            "--replace-fail"
+            "@STARTX@"
+            "${cfg.package}/bin/startx"
+
+            "--replace-fail"
+            "@XQUARTZ@"
+            "${cfg.package}/bin/Xquartz"
+          ];
+        }).outPath;
+        etc."ssh/ssh_config.d/70-xquartz.conf".text = ''
+          XAuthLocation ${xauth}/bin/xauth
+        '';
+    };
+  };
+}
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -101,6 +101,7 @@
   ./services/tailscale.nix
   ./services/trezord.nix
   ./services/wg-quick.nix
+  ./services/xquartz.nix
   ./services/yabai
   ./services/nextdns
   ./services/jankyborders


Feyorsh:
The previous version used `builtins.unsafeGetAttrPos` which might have caused sandboxing issues, fixed to parse the installer for the paths to the launchctl plists

modules/services/xquartz.nix
```diff
@@ -0,0 +1,62 @@
...
+            "@STARTX@"
+            "${cfg.package}/bin/startx"
+
+            "--replace-fail"
+            "@XQUARTZ@"
+            "${cfg.package}/bin/Xquartz"
+          ];
+        }).outPath;
+        etc."ssh/ssh_config.d/70-xquartz.conf".text = ''
+          XAuthLocation ${xauth}/bin/xauth
```
booxter, line 58: nit: use `lib.getBin`

modules/services/xquartz.nix
```diff
@@ -0,0 +1,62 @@
...
+
+  config = mkIf cfg.enable {
+    environment = let
+      daemon = "org.nixos.xquartz.privileged_startx.plist";
+      agent = "org.nixos.xquartz.startx.plist";
+      plistLocation = name: builtins.head (builtins.match ".*cp ([^[:space:]]+-${name}) .*" (builtins.readFile "${cfg.package}/bin/xquartz-install"));
+    in {
+      systemPackages = [ cfg.package ];
+
+      launchDaemons.${daemon}.source =
```
booxter, line 26: This is very brittle.

I think long term we should drop plist files (and xquartz-install script) from nixpkgs package. I think we can do it once nix-darwin module is merged and tested for a bit. So I'd have the nix-darwin module define plists through native means:

```
launchd.agents.xquartz-startx.serviceConfig = { ... };
launchd.daemons.xquartz-privileged-startx.serviceConfig = { ... };
```

modules/services/xquartz.nix
```diff
@@ -0,0 +1,62 @@
...
+            "--replace-fail"
+            "@STARTX@"
+            "${cfg.package}/bin/startx"
+
+            "--replace-fail"
+            "@XQUARTZ@"
+            "${cfg.package}/bin/Xquartz"
+          ];
+        }).outPath;
+        etc."ssh/ssh_config.d/70-xquartz.conf".text = ''
```
booxter, line 57: This is not a runtime requirement to use xquartz if used locally. So I'd rather hide it behind a separate flag, like `services.xquartz.configureSsh`.

modules/services/xquartz.nix
```diff
@@ -0,0 +1,62 @@
+{ lib, pkgs, config, ... }:
+with lib;
+let
+  cfg = config.services.xquartz;
+  xauth = pkgs.xorg.xauth;
```
booxter, line 5: I think `xorg` namespace is now deprecated. Just `pkgs.xauth`.

modules/services/xquartz.nix
booxter, line 57: We should use `programs.ssh.extraConfig` for this.

booxter (changes requested):
It took me a while to get to it, sorry.

booxter:
Also while at it: are there any plans for an HM module that would initialize `DISPLAY` pointing to the X socket from the launchd service? How do you envision users to set it?

booxter:
I have my in-house variation of this module in my flake, leaving for reference:

- https://github.com/booxter/nix/blob/master/darwin/_mixins/xquartz/default.nix
- https://github.com/booxter/nix/blob/master/home-manager/_mixins/xquartz/default.nix

The first piece adds launchd services for DISPLAY socket activation. The second configures shell envs to set DISPLAY referring to the active display socket. With these changes, I am able to start `xterm` or any other X11 program from shell without starting XQuartz.app and/or setting DISPLAY manually.

Let me know what you think.

Feyorsh:
Thanks for the review, you're right that this PR is in need of a significant refactor—will push some changes up soon.

> Also while at it: are there any plans for an HM module that would initialize DISPLAY pointing to the X socket from the launchd service? How do you envision users to set it?

I'm pretty sure `launchd` sets `$DISPLAY` which should get inherited by just about any process launched after the launch daemon has activated.
What does `launchctl getenv DISPLAY` return for you?

I setup XQuartz very recently on a new Mac and I didn't have to putz with `$DISPLAY`, but I'm curious to hear both how you launch your apps (via Spotlight? Dock? I'm talking specifically about processes that have no obvious parent) and if you have any nontrivial configuration/modification to `/etc/bashrc` or `/etc/profile`.

booxter:
> I'm pretty sure launchd sets $DISPLAY which should get inherited by just about any process launched after the launch daemon has activated.

I don't think gui scoped environment (`launchd {g,s}etenv` managed) is automatically exposed to shell. `launchd` settings apply to apps started from Dock / Launchpad, but not from the terminal. (And vice versa.)

> What does launchctl getenv DISPLAY return for you?

Without extracting it from `launchd getenv` and setting it in shell profile? It's empty. And when I try to start e.g. `xterm` from that shell session, it rightfully complains about `DISPLAY` not set. (Which can be confirmed in `env`.)

How do you run X apps? (And which apps did you try?)

Feyorsh:
Looks like [user_startx](https://gitlab.freedesktop.org/xorg/app/xinit/-/blob/master/launchd/user_startx/startx.plist.cpp) is responsible for setting `$DISPLAY`.
From `man launchd.plist`:
> SecureSocketWithKey <string>
> This optional key is a variant of SockPathName. Instead of binding to a known path, a securely generated socket is created and the path is assigned to the environment variable
that is inherited by all jobs spawned in the job's context.

I don't really know what "the job's context" means (launchd _is_ the root process, and obviously other processes I start are not children of startx), but I can confirm that this LaunchAgent is indeed running:
```
$ launchctl list | rg startx
3362    0       org.nixos.xquartz.startx
```

I can `nix run nixpkgs#xterm`, `oneko`, `xeyes` etc without issue (in addition to significantly larger apps).

booxter:
Let me check if I defined this service in my config. I'll get back.

booxter:
@Feyorsh I checked my config. I *do* have the `user_startx` service defined. And it *does* update `launchctl getenv`. The reason I couldn't see `DISPLAY` set in my newly spawned terminal (kitty) was because I use Aerospace to start it, and I had a keybinding for this that omitted LaunchServices, so the terminal inherited `launchctl` environment from Aerospace (that was not restarted).

The gist is that I had to fix my Aerospace config to use `open` instead: https://github.com/booxter/nix/pull/583

So no need for the shell hack. :)

booxter:
@Feyorsh let me know if you want me to take it from here.

Feyorsh:
@booxter go for it, I'm happy to review. Thanks

PR #1543, opened 2025-07-16
services/dictd: init

Feyorsh:
This is a copy of NixOS's [dictd module](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/misc/dictd.nix) with slight tweaks to make it work on macOS (notably, I seem to recall when setting this up that dictd only worked with a user of a specific UID/GID, but I could very easily be mistaken).

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -66,6 +66,7 @@
   ./services/buildkite-agents.nix
   ./services/chunkwm.nix
   ./services/cachix-agent.nix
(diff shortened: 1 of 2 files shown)


Feyorsh:
Using `activationScripts` won't work because `/var/run` doesn't persist across reboots, so I switched to a separate setup daemon—if anyone has a better way of doing this please suggest a change.

PR #1548, opened 2025-07-24
services/timed: init

stepbrobd:
following the NixOS module `networking.timeServers` knob

with this, we can make the macOS default timed to be mutually exclusive with other ntp clients like https://github.com/nix-darwin/nix-darwin/pull/1523


modules/services/timed.nix
```diff
@@ -0,0 +1,45 @@
...
+        The set of NTP servers from which to synchronise.
+      '';
+    };
+  };
+
+  config = {
+    system.activationScripts.networking.text = ''
+      echo "configuring timed..." >&2
+
+      systemsetup -setUsingNetworkTime 'off' &> /dev/null
```
stepbrobd, line 40: turned off deliberately at the beginning to ensure timed restarts

modules/services/timed.nix
```diff
@@ -0,0 +1,45 @@
...
+  onOff = cond: if cond then "on" else "off";
+in
+{
+  meta.maintainers = [
+    (lib.maintainers.stepbrobd or "stepbrobd")
+  ];
+
+  options.services.timed = {
+    enable = lib.mkOption {
+      default = true;
```
stepbrobd, line 19: nixos equivalent of timed is timesyncd, and they have a guard on whether to enable the ntp client by default: https://github.com/NixOS/nixpkgs/blob/527be1432109cfaf305987865bc17cc908e8e77a/nixos/modules/system/boot/timesyncd.nix#L14

do we need such a guard in nix-darwin or we just set it to true like this?

modules/services/timed.nix
```diff
@@ -0,0 +1,45 @@
...
+      default = config.networking.timeServers;
+      defaultText = lib.literalExpression "config.networking.timeServers";
+      type = lib.types.listOf lib.types.str;
+      description = ''
+        The set of NTP servers from which to synchronise.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
```
folliehiyuki: This block will never be executed if `enable = false`, so it does nothing if a user want to disable timed.

modules/services/timed.nix
folliehiyuki: We should keep it enabled by default, since that's the default behavior on MacOS out of the box.

modules/services/timed.nix
stepbrobd: good catch 😅
thx!

PR #1549, opened 2025-07-27
darwin-rebuild: use pkgs.nix for the flake packages output

hgl:
Nix-darwin flake's `packages.${system}.darwin-rebuild` currently uses `null` for `nixPackage`, this leads to the script to use `/var/root/.nix-profile/bin/nix` since it insists on running as root.

I'm not sure why, but on my Mac that file points to an old version of nix, which lacks support for `inputs.self.submodule` for example.

I'm also not sure if this is the right fix, but it's the most direct one.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -51,7 +51,9 @@
     };
 
     overlays.default = final: prev: {
-      inherit (prev.callPackage ./pkgs/nix-tools { }) darwin-rebuild darwin-option darwin-version;
+      inherit (prev.callPackage ./pkgs/nix-tools {
+        nixPackage = prev.nix;
+      }) darwin-rebuild darwin-option darwin-version;
 
       darwin-uninstaller = prev.callPackage ./pkgs/darwin-uninstaller { };
     };


Enzime:
I suspect that this isn't the correct fix

Instead we should add an activation script that uninstalls `pkgs.nix` from the `root` profile and then the uninstaller either restores that exact version or installs `pkgs.nix` back to the `root` profile

hgl:
Sounds reasonable.

Do you have any idea why nix is installed to root's profile but not regular users?

If we can uninstall it, darwin-rebuild then falls back to `/run/current-system/sw/bin`, and it contains the right version, maybe it doesn't need to be reinstalled?

Enzime:
The Nix installer on macOS will install Nix to the root profile by default and configure a launchd service to run the Nix daemon from that Nix, generally it isn't necessary as other users can just use that same Nix so that it matches the version used for the daemon

Yeah it should be fine to not reinstall it unless a user is uninstalling nix-darwin and we want to restore their Nix to a working state

hgl:
@Enzime I'm trying to create a new PR to address this properly. Could use your inputs.

Specifically:

1. I presume uninstalling nix for root should be done in an activation script? Which activation script is the most appropriate one for that? `nixd-daemon` or should I create a new one?
2. Since nix-darwin can manage a user's nix. I guess in that activation script, when `nix.enable` is true, we should also keep uninstalling and reinstalling if the version doesn't match, to manage the root's nix as well?
3. Even if the root's nix is always updated, I still feel a bit weird about the mismatch between `packages.<system>.darwin-rebuild` and the one installed by the module. I think ideally they should behave exactly the same. Thoughts?

Enzime:
Yeah an activation script would be an appropriate place, I think it should be fine to put it in `nix-daemon`.

I don't really understand what you mean by point number 2 and 3

hgl:
For `#2`, I mean, does the following logic of uninstalling and re-installing root's nix in an activation script sound right?
- If `nix.enable` is false, return, since the user doesn't want nix-darwin to manage the nix package.
- If root's installed nix's version is the same as `nix.package`'s version, return
- `nix profile remove nix && nix profile install ${nix.package}`

Every time the user run `darwin-rebuild switch`, the logic is repeated. This basically means we also put root's nix under nix-darwin's management. Does this implementation sound right to you?

For `#3`, I mean, there are at least two ways to access the `darwin-rebuild` cli:

1. When the nix-darwin module is imported, the `darwin-rebuild` package is added to `environment.systemPackages`: https://github.com/nix-darwin/nix-darwin/blob/8df64f819698c1fee0c2969696f54a843b2231e8/modules/nix/nix-darwin.nix#L43
2. nix-darwin's flake also exports  `packages.<system>.darwin-rebuild`.

These two packages are actually different, because the former assigns `nix.package` to `nixPackage`, and the latter always uses the default `nixPackage` value: `null`.

If a downstream tool wants to wrap `darwin-rebuild`, it should to use the flake output, and the resulting cli will be slightly different than the one in the user's `$PATH`. I wonder if you agree these two should actually be identical? If so, do you have any idea how that can be done? I don't think the `nixPackage` value can be overridden by a downstream tool?

PR #1551, opened 2025-07-28
nix: add nix 2.30 to auto-optimise-store assertion

zowoq:
https://github.com/NixOS/nix/commit/b4bea57667eb5413dbc808dd11c10106ef91c770

### DIFF
--- a/modules/nix/default.nix
+++ b/modules/nix/default.nix
@@ -814,9 +814,12 @@ in
         { assertion = config.users.groups ? "nixbld" -> config.users.groups.nixbld.members != []; message = "refusing to remove all members from nixbld group, this would break nix"; }
 
         {
-          # Should be fixed in Lix by https://gerrit.lix.systems/c/lix/+/2100
+          # Should be fixed in Lix by https://gerrit.lix.systems/c/lix/+/2100, cherry picked to nix in https://github.com/NixOS/nix/pull/13241
           # Lix 2.92.0 will set `VERSION_SUFFIX` to `""`; `lib.versionAtLeast "" "pre20241107"` will return `true`.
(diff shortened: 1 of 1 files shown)


modules/nix/default.nix
```diff
@@ -814,9 +814,12 @@ in
         { assertion = config.users.groups ? "nixbld" -> config.users.groups.nixbld.members != []; message = "refusing to remove all members from nixbld group, this would break nix"; }
 
         {
-          # Should be fixed in Lix by https://gerrit.lix.systems/c/lix/+/2100
+          # Should be fixed in Lix by https://gerrit.lix.systems/c/lix/+/2100, Nix by https://github.com/NixOS/nix/commit/d888846b68dd5fad998b84c5cb6246b1b63398cd
           # Lix 2.92.0 will set `VERSION_SUFFIX` to `""`; `lib.versionAtLeast "" "pre20241107"` will return `true`.
-          assertion = cfg.settings.auto-optimise-store -> (cfg.package.pname == "lix" && (isNixAtLeast "2.92.0" && versionAtLeast (strings.removePrefix "-" cfg.package.VERSION_SUFFIX) "pre20241107"));
+          assertion = cfg.settings.auto-optimise-store -> (
+            (cfg.package.pname == "lix" && (isNixAtLeast "2.92.0" && versionAtLeast (strings.removePrefix "-" cfg.package.VERSION_SUFFIX) "pre20241107"))
+            || (cfg.package.pname == "nix" && isNixAtLeast "2.33")
```
zowoq: The fix from 2.33 was also backported to 2.31.3 and 2.32.5, shall I update this to include those as well or just leave it at 2.33?

Enzime: Yep that would be good

zowoq:
I think this should be fine to merge now?

PR #1555, opened 2025-07-31
systems.newsyslog: init module

khaneliman:
Have been meaning to create something like this for a while. Just implements a basic way to generate `newsyslog.d` files to help with log rotation on macos. I plan on continuing with integration with launchd services that have logging setup so you can easily enable logging and rotating logs for launchd agents on your system. 

Tested a little bit real quick with my own config by creating an entry for `nh-clean` and `sudo newsyslog -f /etc/newsyslog.d/nh-clean.conf` to invoke the cleanup. 

Closes https://github.com/nix-darwin/nix-darwin/pull/1843

### DIFF
--- a/modules/system/newsyslog.nix
+++ b/modules/system/newsyslog.nix
@@ -1,0 +1,260 @@
+{ config, lib, ... }:
+
+let
+  inherit (lib) mkOption types;
+
+  cfg = config.system.newsyslog;
+in
+{
+  options = {
+    system.newsyslog = {
+      enable = lib.mkEnableOption "newsyslog configuration management";
+
+      files = mkOption {
+        type = types.attrsOf (
+          types.listOf (
+            types.submodule {
+              options = {
+                logfilename = mkOption {
+                  type = types.str;
+                  description = ''
+                    Path to the log file to rotate, or the literal string `<default>`.
+
+                    The `<default>` entry applies when newsyslog is invoked with an explicit
+                    log file argument that does not match another entry.
+                  '';
+                  example = "/var/log/myapp.log";
+                };
+
+                owner = mkOption {
+                  type = types.nullOr types.str;
+                  default = null;
+                  description = ''
+                    Owner of the log file.
+
+                    If `null`, the owner portion of `owner:group` is left blank.
+                    See `newsyslog.conf(5)` for default ownership behavior (`root:admin`).
+                  '';
+                  example = "myuser";
+                };
+
+                group = mkOption {
+                  type = types.nullOr types.str;
+                  default = null;
+                  description = ''
+                    Group of the log file.
+
+                    If `null`, the group portion of `owner:group` is left blank.
+                    See `newsyslog.conf(5)` for default ownership behavior (`root:admin`).
+                  '';
+                  example = "wheel";
+                };
+
+                mode = mkOption {
+                  type = types.str;
+                  default = "644";
+                  description = ''
+                    File mode (permissions) for the rotated log file.
+                  '';
+                  example = "600";
+                };
+
+                count = mkOption {
+                  type = types.ints.unsigned;
+                  default = 5;
+                  description = ''
+                    Number of rotated log files to keep.
+                  '';
+                };
+
+                size = mkOption {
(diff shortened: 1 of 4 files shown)


modules/system/newsyslog.nix
```diff
@@ -0,0 +1,217 @@
...
+                  example = "/var/log/myapp.log";
+                };
+
+                owner = mkOption {
+                  type = types.nullOr types.str;
+                  default = null;
+                  description = ''
+                    Owner of the log file.
+
+                    If `null`, no owner will be specified.
```
Samasaur1: I wonder whether we want to include the default values for user (`root`) and group (`admin`), or even just mention that you can find them in the `newsy slog.conf(8)`

modules/system/newsyslog.nix
```diff
@@ -0,0 +1,217 @@
...
+                  '';
+                  example = "2048";
+                };
+
+                when = mkOption {
+                  type = types.nullOr types.str;
+                  default = null;
+                  description = ''
+                    Time-based rotation schedule. Can be:
+                    - Hours: @T00, @T06, @T12, @T18 (every 6 hours starting at midnight)
```
Samasaur1: To me this implies that you could set `when = "@T00, @T06, @T12, @T18"` and have the logs rotated every 6 hours, but the man page seems to imply that you can only have one time specifier. Have you tried this? Am I misinterpreting what you wrote?

modules/system/newsyslog.nix
```diff
@@ -0,0 +1,217 @@
...
+                    - Weekly: $W0D00 (weekly on Sunday at midnight)
+                    - Monthly: $M1D00 (monthly on the 1st at midnight)
+
+                    If `null`, time-based rotation is disabled.
+                  '';
+                  example = "@T00";
+                };
+
+                flags = mkOption {
+                  type = types.listOf types.str;
```
Samasaur1: Might be worth making this `types.listOf types.enum`?

Samasaur1: if so I imagine you'd also want to allow the two flags listed in the man page that you haven't documented

modules/system/newsyslog.nix
```diff
@@ -0,0 +1,217 @@
...
+    environment.etc =
+      let
+        formatEntry =
+          entry:
+          let
+            ownerGroup =
+              if entry.owner != null && entry.group != null then
+                "${entry.owner}:${entry.group}"
+              else if entry.owner != null then
+                entry.owner
```
Samasaur1: Per `newsyslog.conf(8)`

> The ‘:’ is essential regardless if the owner or group field is left blank or contains a value.

modules/system/newsyslog.nix
```diff
@@ -0,0 +1,217 @@
...
+            baseParts = [
+              entry.logfilename
+            ]
+            ++ (lib.optional (ownerGroup != "") ownerGroup)
+            ++ [
+              entry.mode
+              (toString entry.count)
+              (if entry.size != null then entry.size else "*")
+              (if entry.when != null then entry.when else "*")
+              (lib.concatStrings entry.flags)
```
Samasaur1: This also needs to support becoming `-` when no flags are specified but a pidfile (and optionally signal) are

modules/system/newsyslog.nix
```diff
@@ -0,0 +1,217 @@
...
+            ];
+
+            optionalParts = lib.filter (p: p != null && p != "") [
+              entry.pidFile
+              entry.signal
```
Samasaur1: By my reading of the man page, you can't have signal set without having a pidfile. Maybe add an assertion for this?

Samasaur1 (changes requested):
In general this looks great. I checked against the man page and some things didn't quite line up, but I haven't actually used this before, so if the man page is inaccurate just let me know

khaneliman:
Thanks for the review, I haven't looked at the man/module in a long time so I'll take some time to make the fixes/updates in a bit.

khaneliman:
Sorry, kept forgetting to come back to this..

Samasaur1 (approved):

PR #1561, opened 2025-08-08
Add kanata module

### LINKED ISSUE #1537: Feature Request: add services kanata
I'd love to see `kanata` in `nix-darwin`.
GitHub: https://github.com/jtroo/kanata

maxRN:
This adds a [kanata](https://github.com/jtroo/kanata) module to nix-darwin. It's heavily inspired by the kanata module from NixOS: https://github.com/NixOS/nixpkgs/blob/35ec96ffd68878831947916da6186d26efc04273/nixos/modules/services/hardware/kanata.nix
(More of a 1:1 copy with linux-specific parts replaced by macOS specifics)

Regarding the test I'm very unsure if this is the correct way to do it. Would definitely love some feedback on this, also this is my first time contributing to nix-darwin so I might have forgotten some steps.

Closes #1537

### DIFF
--- a/modules/services/kanata.nix
+++ b/modules/services/kanata.nix
@@ -49,8 +49,7 @@ let
           example = "danger-enable-cmd yes";
           description = ''
             Configuration of `defcfg` other than `macos-dev-names-include` (generated
-            from the devices option) and
-            `linux-continue-if-no-devs-found` (hardcoded to be yes).
+            from the devices option).
 
             ${upstreamDoc}
           '';
--- a/release.nix
+++ b/release.nix
@@ -105,6 +105,7 @@ in {
   tests.services-dnsmasq = makeTest ./tests/services-dnsmasq.nix;
   tests.services-dnscrypt-proxy = makeTest ./tests/services-dnscrypt-proxy.nix;
   tests.services-eternal-terminal = makeTest ./tests/services-eternal-terminal.nix;
+  tests.services-kanata = makeTest ./tests/services-kanata.nix;
   tests.services-nix-gc = makeTest ./tests/services-nix-gc.nix;
   tests.services-nix-optimise = makeTest ./tests/services-nix-optimise.nix;
   tests.services-nextdns = makeTest ./tests/services-nextdns.nix;
--- a/tests/services-kanata.nix
+++ b/tests/services-kanata.nix
@@ -1,0 +1,93 @@
+{
+  config,
+  pkgs,
+  ...
+}:
+let
+  kanata =
+    pkgs.runCommand "kanata-0.0.0"
+      {
+        # The kanata service runs kanata before building the config to check if it's valid.
+        # Since we don't use the real binary here we point it to a fake-kanata
+        # that always returns true when invoked.
+        meta.mainProgram = "fake-kanata";
+      }
+      ''
+        mkdir $out
+        mkdir $out/bin
+        cat <<EOF > $out/bin/fake-kanata
+        #!/usr/bin/env bash
+        true
+        EOF
+        chmod +x $out/bin/fake-kanata
+      '';
+in
+
+{
+  system.primaryUser = "test-kanata-user";
+
+  services.kanata.enable = true;
+  services.kanata.package = kanata;
+
+  services.kanata.keyboards.internal = {
+    devices = [ "Apple Internal Keyboard / Trackpad" ];
+    extraDefCfg = ''
+      danger-enable-cmd yes
+      process-unmapped-keys yes
+    '';
+    config = ''
+      (defsrc
+          esc  f1   f2   f3   f4   f5   f6   f7   f8   f9   f10  f11  f12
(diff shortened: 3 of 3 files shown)


phucisstupid (approved):

auscyber:
this will not work without #1595 where i add the karabiner-dk module because the driverkit isn't enabled otherwise

modules/services/kanata.nix
```diff
@@ -0,0 +1,171 @@
...
+      '';
+    };
+    keyboards = lib.mkOption {
+      type = lib.types.attrsOf (lib.types.submodule keyboard);
+      default = { };
+      description = "Keyboard configurations.";
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
```
auscyber, line 154: theres no enabling of the driverkit which is required

modules/services/kanata.nix
```diff
@@ -0,0 +1,171 @@
...
+      checkPhase = ''
+        ${lib.getExe cfg.package} --cfg "$target" --check --debug
+      '';
+    };
+
+  mkService =
+    name: keyboard:
+    lib.nameValuePair (mkName name) {
+      command = ''
+        ${lib.getExe cfg.package} \
```
auscyber, line 123: this needs to be run with sudo to actually engage the driverkit-hid-device and will not have access to the path so needs -E ergo it requires a sudoers addition, similar to https://github.com/nix-darwin/nix-darwin/blob/830b3f0b50045cf0bcfd4dab65fad05bf882e196/modules/services/yabai/default.nix#L101C7-L107C10 but also with setEnv so that path is passed through

modules/services/kanata.nix
auscyber: my pr #1595 adds a separate module to enable this which should probably be enabled here and maybe with the src overriden to be 5.0.0 however kanata head is compatible with 6.2.0

modules/services/kanata.nix
```diff
@@ -0,0 +1,171 @@
...
+           --cfg ${keyboard.configFile} \
+           ${lib.optionalString (keyboard.port != null) "--port ${toString keyboard.port} \\"}
+           ${lib.optionalString (keyboard.extraArgs != "") builtins.concatStringsSep " " keyboard.extraArgs}
+      '';
+      serviceConfig = {
```
auscyber, line 128: path should also be passed in from an extraPackages option as commands will become useless unless in /sbin etc

auscyber (changes requested):
see comments

modules/services/kanata.nix
NickHu: Can you expand on this? I'm doing something similar (not exactly using this module) but it seems to work without the `sudo`, however it does require me to manually add the `kanata` binary to System Settings -> Privacy & Security -> Input Monitoring

NickHu: Here is my version of this module btw:

```nix
# adapted from https://github.com/nix-darwin/nix-darwin/blob/7ebf95a73e3b54e0f9c48f50fde29e96257417ac/modules/services/karabiner-elements/default.nix
{ lib, pkgs, ... }:
let
  kbdCfg = builtins.toFile "kanata.kbd" ''
    ...
  '';
  parentAppDir = "/Applications/.Nix-Karabiner";
in
{
  environment.systemPackages = [ pkgs.kanata ];

  system.activationScripts.preActivation.text = ''
    rm -rf ${parentAppDir}
    mkdir -p ${parentAppDir}
    # Kernel extensions must reside inside of /Applications, they cannot be symlinks
    cp -r ${pkgs.karabiner-elements.driver}/Applications/.Karabiner-VirtualHIDDevice-Manager.app ${parentAppDir}
  '';

  # activate extension
  launchd.user.agents.activate_karabiner_system_ext = {
      "${parentAppDir}/.Karabiner-VirtualHIDDevice-Manager.app/Contents/MacOS/Karabiner-VirtualHIDDevice-Manager"
      "activate"
    ];
  };

  launchd.daemons.Karabiner-DriverKit-VirtualHIDDevice-Daemon = {
    command = "\"${pkgs.kanata.passthru.darwinDriver}/Library/Application Support/org.pqrs/Karabiner-DriverKit-VirtualHIDDevice/Applications/Karabiner-VirtualHIDDevice-Daemon.app/Contents/MacOS/Karabiner-VirtualHIDDevice-Daemon\"";
    serviceConfig.ProcessType = "Interactive";
    serviceConfig.Label = "org.pqrs.Karabiner-DriverKit-VirtualHIDDevice-Daemon";
    serviceConfig.KeepAlive = true;
  };

  launchd.daemons.kanata = {
    # also need to add kanata binary to System Settings -> Privacy & Security -> Input Monitoring
    command = "${lib.getExe pkgs.kanata} --cfg ${kbdCfg}";
    serviceConfig.ProcessType = "Interactive";
    serviceConfig.Label = "org.nixos.kanata";
    serviceConfig.KeepAlive = true;
  };
}
```

PR #1562, opened 2025-08-11
networking: add dhcpClientId option to set DHCP Client ID

booxter:


booxter:
I'm not entirely sure this can merge as-is because it would unset a manually set DHCP Client ID. On the other hand, having the new option behave differently from DNS settings may also be convoluted. Is it the case for a new `stateVersion`?

Samasaur1:
putting my question from matrix here:

> any objection to just not setting the client ID at all if `networking.dhcpClientId` is null, and then having users explicitly set the string to "empty" if they want the system default behavior?

booxter:
@Samasaur1 done

modules/networking/default.nix
```diff
@@ -133,6 +136,20 @@ in
...
+      type = types.nullOr types.str;
+      default = null;
+      example = "my-client-id";
+      description = ''
+        The DHCP client identifier to use when requesting an IP address from a DHCP server.
+
+        If this option is set, it will be used by the system when requesting an IP address.
+        If not set, the system will use its default behavior.
+
+        Set to the string "empty" to clear any previously configured client ID.
```
Samasaur1: This isn't accurate anymore, is it?

```suggestion
        If not set, no changes will be made.

        Set to the string "empty" to clear any previously configured client ID and restore the system default behavior.
```

Samasaur1: feel free to reword this if you like

Samasaur1 (approved):

Samasaur1:
Thank you! (and sorry for the delay in getting to this)

PR #1563, opened 2025-08-11
fish: add shellAbbrs config

Eliot00:


modules/programs/fish.nix
```diff
@@ -8,6 +8,10 @@ let
 
   cfg = config.programs.fish;
 
+  fishAbbrs = concatStringsSep "\n" (
+    mapAttrsToList (k: v: "abbr -ag ${k} ${escapeShellArg v}") cfg.shellAbbrs
```
Samasaur1:
```suggestion
    mapAttrsToList (k: v: "abbr -a ${k} -- ${escapeShellArg v}") cfg.shellAbbrs
```
The `-g` flag does not exist in 4.0 and I think it was removed in 3.6 or earlier. The `--` will allow defining expansions that start with `-`

Samasaur1 (changes requested):
I'm kinda tempted to just replicate the interface that HM uses (allowing set-cursor and other abbreviation modifiers), but since this is compatible with a future change let's start with this. I just have two nits here and then I'm happy to merge this

Samasaur1:
looks like that failing test is not your fault but I would like it to be passing before merging as well. Let me take a closer look

Samasaur1:
eh, I think just make the requested changes and hopefully the failing test was intermittent. I'll test locally if it still fails.

Eliot00:
Thanks for your review. The test is now passing.

Samasaur1 (approved):

Samasaur1:
I don't understand why the merge queue hates me

PR #1569, opened 2025-08-15
ci: bump macOS version (13 -> 14)

Samasaur1:
The runners from macOS 14 on default to aarch64, so this will also transition us away from x86_64


Enzime (approved):

Enzime:
cc @emilazy

PR #1574, opened 2025-08-19
Add BatteryShowEnergyMode to controlcenter module

imrunningaround:
Adds option for showing Energy Mode in menu bar, mapping true to 1 and null/false to null, based on plist tests.

### DIFF
--- a/modules/system/defaults/controlcenter.nix
+++ b/modules/system/defaults/controlcenter.nix
@@ -99,7 +99,6 @@
 
         system.defaults.controlcenter.BatteryShowEnergyMode = lib.mkOption {
         type = lib.types.nullOr lib.types.bool;
-        apply = v: if v == null then null else if v then 1 else 0;
         default = null;
         description = ''
             Apple menu > System Settings > Control Center > Battery > Show Energy Mode
--- a/modules/system/defaults/controlcenter.nix
+++ b/modules/system/defaults/controlcenter.nix
@@ -96,5 +96,18 @@
             24 = Hide icon in menu bar
         '';
     };
+
+        system.defaults.controlcenter.BatteryShowEnergyMode = lib.mkOption {
+        type = lib.types.nullOr lib.types.bool;
+        apply = v: if v == true then 1 else null;
+        default = null;
+        description = ''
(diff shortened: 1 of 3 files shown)


modules/system/defaults/controlcenter.nix
```diff
@@ -95,6 +95,19 @@
             18 = Display icon in menu bar
             24 = Hide icon in menu bar
         '';
+    };
+
+        system.defaults.controlcenter.BatteryShowEnergyMode = lib.mkOption {
+        type = lib.types.nullOr lib.types.bool;
+        apply = v: if v == true then 1 else null;
```
Enzime:
```suggestion
```

Setting the value to `null` on the `nix-darwin` side will mean whatever value is there just remains there, we don't currently support deleting values, I would test if `false` works instead

Enzime (changes requested):
Can you also update the tests?

modules/system/defaults/controlcenter.nix
imrunningaround: Thanks for clarifying that nix-darwin doesn't support deleting values.

Just for some context I originally set `false` to `null` because I was following the pattern I saw in other controlcenter options where the "When Active" state maps to `null`. 

The Energy Mode setting in System Settings only has two options (Always vs. When Active) rather than three options like some other controls, with the third "Don't Show in Menu Bar" option.

modules/system/defaults/controlcenter.nix
```diff
@@ -95,6 +95,20 @@
             18 = Display icon in menu bar
             24 = Hide icon in menu bar
         '';
+    };
+
+        system.defaults.controlcenter.BatteryShowEnergyMode = lib.mkOption {
+        type = lib.types.nullOr lib.types.bool;
+        apply = v: if v == null then null else if v then 1 else 0;
```
Enzime:
```suggestion
```

This is unnecessary when you set booleans through nix-darwin they’ll get converted into integers as necessary

phucisstupid (approved):

iamanaws:
Duplicated/Included on https://github.com/nix-darwin/nix-darwin/pull/1347

PR #1581, opened 2025-08-27
`apply` should return `null` if the argument is not a list

n8henrie:
Fixes https://github.com/nix-darwin/nix-darwin/issues/1578

### DIFF
--- a/modules/launchd/launchd.nix
+++ b/modules/launchd/launchd.nix
@@ -121,7 +121,17 @@ in
         key. Please note: many people are confused by this key. Please read `execvp(3)` very carefully!
       '';
       # TODO: Remove this some time after 25.11.
-      apply = map (value: lib.warnIf (lib.hasInfix "&amp;" value) "A value for `ProgramArguments` contains the literal string `&amp;`. This is no longer necessary and will lead to double-escaping, as nix-darwin now automatically escapes special characters." value);
+      apply =
+        val:
+        if (builtins.isNull val) then
+          val
+        else
+          (map (
+            item:
+            lib.warnIf (lib.hasInfix "&amp;" item)
+              "A value for `ProgramArguments` contains the literal string `&amp;`. This is no longer necessary and will lead to double-escaping, as nix-darwin now automatically escapes special characters."
+            item
+          ) val);
     };
 
     EnableGlobbing = mkOption {


n8henrie:
Also applied nixfmt

n8henrie:
@Samasaur1 -- it looks like you may have merged https://github.com/nix-darwin/nix-darwin/pull/1529 which led to this issue. If you have time, I'd appreciate a review and (if appropriate) to get this fix merged.

Samasaur1:
thanks for pinging me, I unfortunately have limited review time and would have missed this otherwise

modules/launchd/launchd.nix
```diff
@@ -121,7 +121,18 @@ in
...
+      apply =
+        val:
+        if (builtins.isNull val) then
+          val
+        else
+          (map (
+            item:
+            lib.warnIf (lib.hasInfix "&amp;" item) ''
+              A value for `ProgramArguments` contains the literal string `&amp;`. This is no longer necessary and
+              will lead to double-escaping, as nix-darwin now automatically escapes special characters.
```
Samasaur1: Can you change the warning string back to a one-line string?
```suggestion
              A value for `ProgramArguments` contains the literal string `&amp;`. This is no longer necessary and will lead to double-escaping, as nix-darwin now automatically escapes special characters.
```

Samasaur1 (changes requested):
Yeah, I see what happened here. I guess I was only testing with services that use `ProgramArguments`. If you make the one requested change I'll merge this

modules/launchd/launchd.nix
n8henrie: Happy to, I am trying to get used to ~this~ EDIT: something similar as a best practice. However, in this case I was trying to match the surrounding formatting -- all analogous strings are hard-wrapped at roughly 120 chars.

Might it be worth standardizing on the "one sentence per line" format, which has been recommended to me by the nix-documentation (forget the official name) team?

n8henrie: Actually, the others are descriptions. I suspect this has to do with hard wraps in the warning resulting in weird linebreaks at runtime.

Samasaur1: We do have some other warnings that are manually split onto multiple lines, so I agree that we should standardize at some point. For now I'm just trying to have as small an impact as possible

Samasaur1:
why does it let me set a message and then ignore that message??? *sigh*

PR #1582, opened 2025-08-29
github-runner: fix the nodeRuntimes option

sandydoo:
The conditional override was intended to be backwards-compatible with older runner packages without the `nodeRuntimes` argument. However, `override` doesn't yield default arguments, so the override was never applied.

This commit removes the conditional entirely since `nodeRuntimes` was added back in 24.05.

GitHub's policy is to deprecate all previous releases 30 days after a new release. In theory, this should be safe to backport to 25.05.


Enzime:
Could you make a backport PR?

PR #1590, opened 2025-09-14
Recommend official experimental installer instead of the detsys one

### LINKED ISSUE #1588: Update installation instructions
The current instructions state:

> The Nix installer from Determinate Systems is only recommended for use with flake-based setups. It can install one of two distributions of Nix: To install the recommended vanilla upstream Nix, you will need to explicitly say no when prompted to install Determinate Nix.

But this will no longer be a supported option: https://determinate.systems/blog/installer-dropping-upstream/

WeetHet:
https://determinate.systems/blog/installer-dropping-upstream

Closes #1588


Samasaur1:
See #1588

PR #1053, opened 2024-09-03
feat(trackpad): add additional trackpad and gesture support

dustypomerleau:
This PR adds additional trackpad and gesture support.

To `system.defaults.trackpad` it adds:

```
ActuateDetents
DragLock
ForceSuppressed
TrackpadCornerSecondaryClick
TrackpadFourFingerHorizSwipeGesture
TrackpadFourFingerPinchGesture
TrackpadFourFingerVertSwipeGesture
TrackpadMomentumScroll
TrackpadPinch
TrackpadRotate
TrackpadThreeFingerHorizSwipeGesture
TrackpadThreeFingerTapGesture
TrackpadThreeFingerVertSwipeGesture
TrackpadTwoFingerDoubleTapGesture
TrackpadTwoFingerFromRightEdgeSwipeGesture
```

and to `system.defaults.dock` it adds:

```
showAppExposeGestureEnabled
showDesktopGestureEnabled
showLaunchpadGestureEnabled
showMissionControlGestureEnabled
```

A few questions:

1. I have tested locally - am I meant to also add these settings to `tests/system-defaults-write.nix`?
1. There is an inconsistency in the way that we represent `defaults write` commands with only 2 options. `defaults read` outputs these as `-int` (0 or 1), even if the user sets them via `-bool`. In this repo, there is a mix of representing them as `types.bool` and `types.enum [ 0 1 ]`. Does the project have a preference? Should we make them all the same?
1. I have a question about formatting, but thought it best to separate that into its own issue (#).

PR #1591, opened 2025-09-18
feat(trackpad): add additional trackpad and gesture support

dustypomerleau:
I have closed #1053 and opened a new PR, because GitHub did not like my creative rebasing. This should be easier to merge.

This PR adds additional trackpad and gesture support.

To `system.defaults.trackpad` it adds:

```
ActuateDetents
DragLock
ForceSuppressed
TrackpadCornerSecondaryClick
TrackpadFourFingerHorizSwipeGesture
TrackpadFourFingerPinchGesture
TrackpadFourFingerVertSwipeGesture
TrackpadMomentumScroll
TrackpadPinch
TrackpadRotate
TrackpadThreeFingerHorizSwipeGesture
TrackpadThreeFingerTapGesture
TrackpadThreeFingerVertSwipeGesture
TrackpadTwoFingerDoubleTapGesture
TrackpadTwoFingerFromRightEdgeSwipeGesture
```

and to `system.defaults.dock` it adds:

```
showAppExposeGestureEnabled
showDesktopGestureEnabled
showLaunchpadGestureEnabled
showMissionControlGestureEnabled
```

Autoformatting is still a pain point when making small changes. If I can help move #1054 forward please let me know.

### DIFF
--- a/modules/system/defaults/trackpad.nix
+++ b/modules/system/defaults/trackpad.nix
@@ -9,7 +9,7 @@ with lib;
       type = types.nullOr types.bool;
       default = null;
       description = ''
-        Whether to enable trackpad tap to click.  The default is false.
+        Whether to enable tap to click. The default is false.
       '';
     };
 
@@ -17,7 +17,7 @@ with lib;
       type = types.nullOr types.bool;
       default = null;
       description = ''
-        Whether to enable tap-to-drag. The default is false.
+        Whether to enable tap to drag. The default is false.
       '';
     };
 
@@ -25,7 +25,8 @@ with lib;
       type = types.nullOr types.bool;
       default = null;
       description = ''
-        Whether to enable trackpad right click.  The default is false.
+        Whether to enable trackpad right click (two-finger tap/click).
+        The default is false.
       '';
     };
 
@@ -33,7 +34,7 @@ with lib;
       type = types.nullOr types.bool;
       default = null;
       description = ''
-        Whether to enable three finger drag.  The default is false.
+        Whether to enable three-finger drag. The default is false.
       '';
     };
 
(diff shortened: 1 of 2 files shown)


modules/system/defaults/trackpad.nix
```diff
@@ -67,10 +68,134 @@ with lib;
...
+      '';
+    };
+
+    system.defaults.trackpad.TrackpadFourFingerVertSwipeGesture = mkOption {
+      type = types.nullOr (types.enum [ 0 2 ]);
+      default = null;
+      description = ''
+        0 to disable four finger vertical swipe gestures, 2 to enable (down for Mission Control, up for App Exposé).
+        The default is 2.
+        When both three- and four-finger vertical swipe gestures are enabled, the three-finger variant takes precedence. This setting interacts with `system.defaults.dock.showAppExposeGestureEnabled` and `system.defaults.dock.showMissionControlGestureEnabled` to determine whether vertical swipe gestures are enabled for App Exposé, Mission Control, or both.
```
Samasaur1, line 134: I kind of hate this but it seems like a macOS problem and not a problem with your work. Alas

Samasaur1 (approved):

modules/system/defaults/trackpad.nix
dustypomerleau: Thank you for spending time reviewing/merging! I agree with you, but I think it's a fundamental philosophical question for nix-darwin - should the project provide a better interface than Apple does, or an interface that matches Apple 1:1? Once you diverge, it increases the maintenance burden for the project, but it could also be very empowering. Apple tends to provide a bunch of boolean-like settings under the hood, and they deal with conflicts or mutually exclusive states by just ignoring some booleans when others are set. It's...not elegant. If nix-darwin wanted to diverge, all of these mutually exclusive states could be represented as sum types, and the total number of settings presented to the user would be reduced. The question is, is it worth it, particularly if the underlying Apple options are a moving target?

PR #1594, opened 2025-09-20
system: add application support linking

auscyber:
Makes activation script also link Library/Application Support

- this is needed because i'm working on getting karabiner/kanata actually working again
- i have this on my machine working fine but it's needed for the daemon


auscyber:
this is where this is incredibly useful to me, and i might add something like this to nix-darwin if ppl want https://github.com/auscyber/dotfiles/blob/21f73a7f101ed5c1b93638265fed1cf38abb2a2e/modules/darwin/keybinds/karabiner_driver/default.nix

Enzime:
Is this useful for anything outside of Karabiner-Elements?

auscyber:
> Is this useful for anything outside of Karabiner-Elements?

well for the hid device of karabiner elements, not sure

auscyber:
it would be a nice mechanism to have for linking application support thats needed for many applications

auscyber:
there are also quite a few packages that link things to library application support, that could use this, but this is useful because i'm also gonna backage the karabiner hid device driver separately and create a pr for it so that kanata can be added to nix-darwin/home-manager and work on macos

auscyber:
just had someone in matrix talk about linking of vst plugins so yes this is needed

auscyber:
https://github.com/NixOS/nixpkgs/pull/451871 this should be merged so that a lot of the jank of copying is fixed and we can actually make it variable with an option such as 
```nix
option.environment.extraPathsToLink = lib.mkOption {
  type = lib.types.listOf lib.types.string;
  default = [];
  description = "Paths to link in the environment such as `/Library/Application Support`"
};
```

Yethal:
@Enzime I'm currently working on nushell module and this PR is necessary for proper operation as generated config file needs to be linked to /Library/Application Support for vendor autoloading

PR #1600, opened 2025-09-25
remove duplicates from brewfile

uPaymeiFixit:
For example, if you add the same cask multiple times from different locations, you'll end up with a brewfile that tries to install that cask multiple times. This removes those duplicates.


dudeofawesome (approved):

Samasaur1:
ack I did not notice that was going to 25.05 and not master

PR #1614, opened 2025-10-23
networking: modify firewall settings only if explicitly set

ryanccn:
The current behavior of `networking.applicationFirewall` modifies system firewall settings according to its defaults on activation regardless of whether the user actually uses the module; this PR changes the defaults of the relevant options to `null` and only applies settings when these options are explicitly set by the user.

Ideally, there should be a module enable option that controls whether any of the module options are applied, but currently `networking.applicationFirewall.enable` controls whether the firewall itself is enabled and I am not entirely sure if it is appropriate to make a breaking change to that behavior.

### DIFF
--- a/tests/networking-firewall.nix
+++ b/tests/networking-firewall.nix
@@ -1,0 +1,17 @@
+{ config, ... }:
+{
+  networking.applicationFirewall = {
+    enable = true;
+    blockAllIncoming = true;
+    allowSignedApp = false;
+    enableStealthMode = null;
+  };
+
+  test = ''
+    echo "checking socketfilterfw calls in /activate" >&2
+    grep "/usr/libexec/ApplicationFirewall/socketfilterfw --setglobalstate on" ${config.out}/activate
+    grep "/usr/libexec/ApplicationFirewall/socketfilterfw --setblockall on" ${config.out}/activate
+    grep "/usr/libexec/ApplicationFirewall/socketfilterfw --setallowsignedapp off" ${config.out}/activate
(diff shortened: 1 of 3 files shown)


Enzime:
Can you update the tests to set one setting to `null` explicitly and confirm that setting doesn't get changed?

ryanccn:
> Can you update the tests to set one setting to `null` explicitly and confirm that setting doesn't get changed?

@Enzime done!

tests/networking-firewall.nix
```diff
@@ -0,0 +1,17 @@
...
+    allowSignedApp = false;
+    enableStealthMode = null;
+  };
+
+  test = ''
+    echo "checking socketfilterfw calls in /activate" >&2
+    grep "/usr/libexec/ApplicationFirewall/socketfilterfw --setglobalstate on" ${config.out}/activate
+    grep "/usr/libexec/ApplicationFirewall/socketfilterfw --setblockall on" ${config.out}/activate
+    grep "/usr/libexec/ApplicationFirewall/socketfilterfw --setallowsignedapp off" ${config.out}/activate
+    ! grep "/usr/libexec/ApplicationFirewall/socketfilterfw --setstealthmode" ${config.out}/activate
```
Enzime: This looks like it's missing parentheses around the line

Can you test if this actually asserts that text isn't present by setting the option to `true`?

ryanccn: You're right, I've added the missing parentheses so it should work as intended now.

PR #1617, opened 2025-10-26
services: add `sleepwatcher` service

virusdave:
This service allows specification of user or system scripts to run at various system event moments, such as "system woke from sleep" or "AC power plugged in", etc.

Requires https://github.com/NixOS/nixpkgs/pull/455927 to be merged (and backported) first.

### DIFF
--- a/modules/services/sleepwatcher/default.nix
+++ b/modules/services/sleepwatcher/default.nix
@@ -19,58 +19,85 @@ in
         description = "Whether to verbosely log actions performed by sleepwatcher.";
       };
 
-      on_sleep = mkOption {
-        type = types.nullOr types.lines;
+      logFile = mkOption {
+        type = types.nullOr types.path;
         default = null;
+        example = "/var/log/sleepwatcher.log";
+        description = ''
+          The log file path for sleepwatcher output.
+          If set, both stdout and stderr will be logged to this file.
+          If not set, logs will be discarded.
+        '';
+      };
+
+      on_sleep = mkOption {
+        type = types.lines;
+        default = "";
         example = "echo 'Going to sleep'";
         description = "Commands to execute when system goes to sleep.";
       };
 
       on_wakeup = mkOption {
-        type = types.nullOr types.lines;
-        default = null;
+        type = types.lines;
+        default = "";
         example = "echo 'Waking from sleep'";
         description = "Commands to execute when system wakes up.";
       };
 
       on_display_sleep = mkOption {
-        type = types.nullOr types.lines;
-        default = null;
+        type = types.lines;
+        default = "";
         example = "echo 'Display sleeping'";
-        description = "Commands to execute when display goes to sleep.";
+        description = ''
+          Commands to execute when display goes to sleep.
+          
+          Note: The --displaysleep option does not work with Apple silicon Macs.
+        '';
       };
 
       on_display_wakeup = mkOption {
-        type = types.nullOr types.lines;
-        default = null;
+        type = types.lines;
+        default = "";
         example = "echo 'Display awake'";
-        description = "Commands to execute when display wakes up.";
+        description = ''
+          Commands to execute when display wakes up.
+          
+          Note: The --displaywakeup option does not work with Apple silicon Macs.
+        '';
       };
 
       on_display_dim = mkOption {
-        type = types.nullOr types.lines;
-        default = null;
+        type = types.lines;
+        default = "";
         example = "echo 'Display dimmed'";
-        description = "Commands to execute when display dims.";
+        description = ''
+          Commands to execute when display dims.
+          
+          Note: The --displaydim option does not work with Apple silicon Macs.
+        '';
       };
 
       on_display_undim = mkOption {
-        type = types.nullOr types.lines;
-        default = null;
+        type = types.lines;
+        default = "";
         example = "echo 'Display undimmed'";
-        description = "Commands to execute when display undims.";
+        description = ''
+          Commands to execute when display undims.
+          
+          Note: The --displayundim option does not work with Apple silicon Macs.
+        '';
       };
 
       on_plug = mkOption {
-        type = types.nullOr types.lines;
-        default = null;
+        type = types.lines;
+        default = "";
         example = "echo 'Power connected'";
         description = "Commands to execute when power adapter is plugged in.";
       };
 
       on_unplug = mkOption {
-        type = types.nullOr types.lines;
-        default = null;
+        type = types.lines;
+        default = "";
         example = "echo 'Power disconnected'";
         description = "Commands to execute when power adapter is unplugged.";
       };
@@ -83,15 +110,15 @@ in
       };
 
       on_idle = mkOption {
-        type = types.nullOr types.lines;
-        default = null;
+        type = types.lines;
+        default = "";
         example = "echo 'User idle'";
         description = "Commands to execute when user becomes idle (requires idle_time to be set).";
       };
 
       on_idle_resume = mkOption {
-        type = types.nullOr types.lines;
-        default = null;
+        type = types.lines;
+        default = "";
         example = "echo 'User active again'";
         description = "Commands to execute when user resumes activity after idle (requires idle_time to be set).";
       };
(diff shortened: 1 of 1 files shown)


modules/services/sleepwatcher/default.nix
```diff
@@ -0,0 +1,155 @@
...
+      plugScript = pkgs.writeShellScript "sleepwatcher-plug" cfg.on_plug;
+      unplugScript = pkgs.writeShellScript "sleepwatcher-unplug" cfg.on_unplug;
+      
+      launchdConfig = {
+        path = [ config.environment.systemPath ];
+        serviceConfig = {
+          ProgramArguments = [
+            "${cfg.package}/bin/sleepwatcher"
+            "-V"
+          ] ++ optionals (cfg.on_sleep != null) [ "-s" "${sleepScript}" ]
```
bew: I'd recommend to use long `--param` names instead of short `-p`, it makes the code much more explicit and readable in the module.

modules/services/sleepwatcher/default.nix
```diff
@@ -0,0 +1,155 @@
...
+with lib;
+
+let
+  cfg = config.services.sleepwatcher;
+in
+
+{
+  options = {
+    services.sleepwatcher = {
+      enable = mkOption {
```
bew: You should use `mkEnableOption` here, it basically does the same thing that you do manually.
ref: https://noogle.dev/f/lib/mkEnableOption

modules/services/sleepwatcher/default.nix
```diff
@@ -0,0 +1,155 @@
...
+        type = types.bool;
+        default = false;
+        description = "Whether to enable the sleepwatcher daemon.";
+      };
+
+      package = mkOption {
```
bew: Similarily there is `mkPackageOption`
ref: https://noogle.dev/f/lib/mkPackageOption

modules/services/sleepwatcher/default.nix
```diff
@@ -0,0 +1,155 @@
...
+        type = types.package;
+        default = pkgs.darwin.sleepwatcher;
+        description = "This option specifies the sleepwatcher package to use.";
+      };
+
+      on_sleep = mkOption {
+        type = types.nullOr types.lines;
+        default = null;
+        example = "pmset sleepnow";
```
bew: I'm surprised by this example, does that mean that the script needs to call that to immediately sleep?
If yes it should be mentioned in the description, if not I think the example should be a simple echo like other examples

modules/services/sleepwatcher/default.nix
```diff
@@ -0,0 +1,155 @@
...
+        default = null;
+        example = "echo 'Power disconnected'";
+        description = "Commands to execute when power adapter is unplugged.";
+      };
+
+      idleTime = mkOption {
+        type = types.nullOr types.int;
+        default = null;
+        example = 600;
+        description = "Idle time in seconds before executing on_idle command.";
```
bew:
```suggestion
        description = "Idle time in seconds before executing on_idle commands.";
```

modules/services/sleepwatcher/default.nix
```diff
@@ -0,0 +1,155 @@
...
+      };
+
+      on_unplug = mkOption {
+        type = types.nullOr types.lines;
+        default = null;
+        example = "echo 'Power disconnected'";
+        description = "Commands to execute when power adapter is unplugged.";
+      };
+
+      idleTime = mkOption {
```
bew: What's your logic for having action options use `snake_case` and few other options `idleTime` & `daemonMode` using `camelCase` format?

modules/services/sleepwatcher/default.nix
```diff
@@ -0,0 +1,155 @@
...
+        type = types.nullOr types.int;
+        default = null;
+        example = 600;
+        description = "Idle time in seconds before executing on_idle command.";
+      };
+
+      daemonMode = mkOption {
```
bew: The naming of this option is misleading, because even when `false` it's still actually a daemon running for the user.
Maybe rename this as `runAsSystem` or `runAsUser` :thinking:

modules/services/sleepwatcher/default.nix
bew: You could write a small function to do the null check & make a script on-demand for the args to the `ProgramArguments`, less code & nicer to read overall :thinking: 

Something like (includes the use of long param names):
```nix
optionalHook = longArg: hookValue: (let
  hookScript = pkgs.writeShellScript "sleepwatcher-${longArg}-hook" hookValue;
in optionals (hookValue != null) [ "--${longArg}" "${hookScript}" ]
```

used like:
```nix
ProgramArguments = [
  # ...
] ++ optionalHook "sleep" cfg.on_sleep
++ optionalHook "wakeup" cfg.on_wakeup
++ optionalHook "displaysleep" cfg.on_display_sleep
# ...
```

virusdave: Excellent suggestions everywhere.  I've applied them all, I believe.

modules/services/sleepwatcher/default.nix
virusdave: Nope; updated the comment.  Poorly chosen example.

modules/services/sleepwatcher/default.nix
virusdave: Good idea.  Had I drafted the original code solely by hand, I probably would have started there :)

modules/services/sleepwatcher/default.nix
virusdave: Unified to snake_case.

modules/services/sleepwatcher/default.nix
```diff
@@ -0,0 +1,148 @@
...
+            ++ optionalHook "idleresume" cfg.on_idle_resume
+            ++ optionalHook "plug" cfg.on_plug
+            ++ optionalHook "unplug" cfg.on_unplug;
+          KeepAlive = true;
+          RunAtLoad = true;
+        };
+      };
+    in
+    {
+      environment.systemPackages = [ cfg.package ];
```
Samasaur1: Does `sleepwatcher` need to be in the system path? It doesn't seem like this is useful given that the daemon is already running.

modules/services/sleepwatcher/default.nix
```diff
@@ -0,0 +1,148 @@
...
+
+      launchd.user.agents.sleepwatcher = mkIf (!cfg.run_as_system) launchdConfig;
```
Samasaur1: User agents in nix-darwin itself need to have a `managedBy` attribute so that we can properly display error messages when `system.primaryUser` is unset.

modules/services/sleepwatcher/default.nix
```diff
@@ -0,0 +1,148 @@
...
+
+      run_as_system = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Whether to run sleepwatcher as a system daemon (true) or user agent (false).
+          System daemon runs as root and can execute system-level commands.
+          User agent runs as the user and is preferred for user-specific actions.
+        '';
+      };
```
Samasaur1, line 134: Is there any case where the user might want to run both a systemwide and a per-user instance?

modules/services/sleepwatcher/default.nix
```diff
@@ -0,0 +1,148 @@
...
+            ++ optionalHook "displaysleep" cfg.on_display_sleep
+            ++ optionalHook "displaywakeup" cfg.on_display_wakeup
+            ++ optionalHook "displaydim" cfg.on_display_dim
+            ++ optionalHook "displayundim" cfg.on_display_undim
+            ++ optionals (cfg.on_idle != null && cfg.idle_time != null)
+              ([ "--timeout" "${toString cfg.idle_time}" ] ++
+              (optionalHook "idle" cfg.on_idle))
+            ++ optionalHook "idleresume" cfg.on_idle_resume
+            ++ optionalHook "plug" cfg.on_plug
+            ++ optionalHook "unplug" cfg.on_unplug;
```
Samasaur1: We prefer that you set `command` instead of `serviceConfig.ProgramArguments` so that a) the service picks up `/bin/wait4path` so that it doesn't fail repeatedly and be put in the penalty box when the nix store is not yet mounted; b) future changes to launchd services apply to them as well

modules/services/sleepwatcher/default.nix
```diff
@@ -0,0 +1,148 @@
...
+            ++ optionalHook "sleep" cfg.on_sleep
+            ++ optionalHook "wakeup" cfg.on_wakeup
+            ++ optionalHook "displaysleep" cfg.on_display_sleep
+            ++ optionalHook "displaywakeup" cfg.on_display_wakeup
+            ++ optionalHook "displaydim" cfg.on_display_dim
+            ++ optionalHook "displayundim" cfg.on_display_undim
+            ++ optionals (cfg.on_idle != null && cfg.idle_time != null)
+              ([ "--timeout" "${toString cfg.idle_time}" ] ++
+              (optionalHook "idle" cfg.on_idle))
+            ++ optionalHook "idleresume" cfg.on_idle_resume
```
Samasaur1: This check checks whether `cfg.on_idle` is null twice, which seems redundant.

Also, I imagine `cfg.on_idle_resume` should also be ignored if `cfg.idle_time` is unset?

modules/services/sleepwatcher/default.nix
```diff
@@ -0,0 +1,148 @@
...
+    services.sleepwatcher = {
+      enable = mkEnableOption "sleepwatcher daemon to react to various system events";
+
+      package = mkPackageOption pkgs [ "darwin" "sleepwatcher" ] { };
+
+      verbose = mkOption {
+        type = types.bool;
+        default = true;
+        description = "Whether to verbosely log actions performed by sleepwatcher.";
+      };
```
Samasaur1, line 20: This option is essentially useless without setting `serviceConfig.StandardOutPath` and/or `serviceConfig.StandardErrorPath` on the service itself, as launchd just swallows the logs.

modules/services/sleepwatcher/default.nix
```diff
@@ -0,0 +1,148 @@
...
+  };
+
+  config = mkIf cfg.enable (
+    let
+      optionalHook = longArg: hookValue:
+        let hookScript = pkgs.writeShellScript "sleepwatcher-${longArg}-hook" hookValue;
+        in optionals (hookValue != null) [ "--${longArg}" "${hookScript}" ];
+
+      launchdConfig = {
+        path = [ config.environment.systemPath ];
```
Samasaur1: What's the reasoning behind this as opposed to having people set the PATH individually on each script that needs it set?

modules/services/sleepwatcher/default.nix
Samasaur1: I'd recommend adding options for these attributes rather than hardcoding them, btw; see some other services that do that.

Samasaur1:
Also note that the homepage for this software says

> The options --displaydim, --displayundim, --displaysleep and --displaywakup do not work with Apple silicon Macs, this issue is still under investigation

which should be reflected somewhere in this service, probably in the description for those options.

modules/services/sleepwatcher/default.nix
virusdave: nope, removed.

modules/services/sleepwatcher/default.nix
virusdave: Fixed 👍

modules/services/sleepwatcher/default.nix
virusdave: Possibly?  I don't have a concrete use case for it, but it's not too hard to manufacture some.  Can we make this a "cross that bridge" concern, or at least I'd prefer to make that change in a followup unless you feel strongly about it (just to minimize the delta between what I know works and the effort to further test what gets submitted quickly).

I promise i'll follow up with that as well, as I probably DO have uses for it that i haven't thought of yet.

modules/services/sleepwatcher/default.nix
virusdave: Done.

modules/services/sleepwatcher/default.nix
virusdave: Fixed

modules/services/sleepwatcher/default.nix
virusdave: Good idea.  Fixed.

modules/services/sleepwatcher/default.nix
virusdave: No particularly strong reason.  Convenience during development that i never returned to to generalize.  Fixed.

virusdave:
> I think this should target `master` instead of `nix-darwin-25.05`; it can be backported if desired.

Done

> 
> Also I didn't leave this as an inline comment because there wasn't one place to put it, but did you consider making the scripts be `types.lines` instead of `types.nullOr types.lines` and checking for `""` instead of `null`? I think it would be functionally the same but seems to match our style better.

Done (let me know if i missed anything)

virusdave:
> Also note that the homepage for this software says
> 
> > The options --displaydim, --displayundim, --displaysleep and --displaywakup do not work with Apple silicon Macs, this issue is still under investigation
> 
> which should be reflected somewhere in this service, probably in the description for those options.

Fixed.

virusdave:
Thanks for the review comments, all.  I had just been wondering whatever happened with this PR.  Let me know if it's good or if it still needs work.

Cheers!

virusdave:
> Thanks for the review comments, all. I had just been wondering whatever happened with this PR. Let me know if it's good or if it still needs work.
> 
> Cheers!

@Samasaur1 github is telling me you are requesting one more change, but for the life of me i can't tell what it is.  Kindly bump whatever i've missed if I've missed something.

Thanks!

virusdave:
Can I get another look at this?  I believe I've made all requested changes.

virusdave:
@Samasaur1 are you happy for this to be merged now?

virusdave:
Does anyone else have review / merge access?  Seems like @Samasaur1 might be away?

PR #1620, opened 2025-11-01
dnsmasq: make the daemon wait for the nix store to be mounted

Boelensman1:
Uses the wait4path mac os util to make the daemon wait for /nix/store to appear.

This resolves the issue I have of the daemon failing to starting because the nix store is not yet mounted.

### DIFF
--- a/modules/services/dnsmasq.nix
+++ b/modules/services/dnsmasq.nix
@@ -69,13 +69,15 @@ in
     environment.systemPackages = [ cfg.package ];
 
     launchd.daemons.dnsmasq = {
-      serviceConfig.ProgramArguments = [
-        "${cfg.package}/bin/dnsmasq"
-        "--listen-address=${cfg.bind}"
-        "--port=${toString cfg.port}"
-        "--keep-in-foreground"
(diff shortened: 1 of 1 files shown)


modules/services/dnsmasq.nix
```diff
@@ -65,17 +65,24 @@ in
...
+      "--listen-address=${cfg.bind}"
+      "--port=${toString cfg.port}"
+      "--keep-in-foreground"
+    ] ++ addressArgs ++ serverArgs;
+    dnsmasqCmd = "${cfg.package}/bin/dnsmasq ${concatStringsSep " " dnsmasqArgs}";
+  in {
     environment.systemPackages = [ cfg.package ];
 
     launchd.daemons.dnsmasq = {
       serviceConfig.ProgramArguments = [
```
Enzime: Instead of manually specifying `wait4path` you can just change this to `launchd.daemons.dnsmasq.command` like other nix-darwin services

Boelensman1: Done. Would you like me to revert the other changes? I updated it for readability, but that is less needed now.

Enzime: Yes please

Boelensman1: Done!

PR #1624, opened 2025-11-05
fix: zsh-fast-syntax-highlighting path

XYenon:
https://github.com/NixOS/nixpkgs/pull/101007


Samasaur1 (approved):
just making a note here that we may want to update the `zsh-autosuggestions` path as well in the future.

Issue #1621, opened 2025-11-01
Tahoe modified /etc/zshrc and /etc/zprofile


peteringram0:
After updating to tahoe i get the following error:

```bash
error: Unexpected files in /etc, aborting activation
The following files have unrecognized content and would be overwritten:

  /etc/zshrc
  /etc/zprofile

Please check there is nothing critical in these files, rename them by adding .before-nix-darwin to the end, and then try again.
```

Is it safe just to prefix these like the suggestion? Will these files get re-generated?

Samasaur1:
Yes, it's safe to follow those instructions, but we try to keep track of these files so you never hit this kind of issue in the first place. Can you answer the following questions for us:

- What version of Tahoe did you upgrade to?
- upload the contents of `/etc/zshrc` and `/etc/zprofile`
- post the SHA256 hashes of both of those files (`shamus -a 256 /etc/zshrc /etc/zprofile`)

peteringram0:
> Yes, it's safe to follow those instructions, but we try to keep track of these files so you never hit this kind of issue in the first place. Can you answer the following questions for us:
> 
> * What version of Tahoe did you upgrade to? - 26.0.1 (25A362)
> * upload the contents of `/etc/zshrc` and `/etc/zprofile`
> * post the SHA256 hashes of both of those files (`shamus -a 256 /etc/zshrc /etc/zprofile`)


```/etc/zshrc

# Nix
if [ -e '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh' ]; then
    . '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
fi
# End Nix

        
# System-wide profile for interactive zsh(1) shells.

# Setup user specific overrides for this in ~/.zshrc. See zshbuiltins(1)
# and zshoptions(1) for more details.

# Correctly display UTF-8 with combining characters.  We'll assume UTF-8 if the
# locale(1) binary is missing entirely.
if [[ ! -x /usr/bin/locale ]] || [[ "$(locale LC_CTYPE)" == "UTF-8" ]]; then
    setopt COMBINING_CHARS
fi

# Disable the log builtin, so we don't conflict with /usr/bin/log
disable log

# Save command history
HISTFILE=${ZDOTDIR:-$HOME}/.zsh_history
HISTSIZE=2000
SAVEHIST=1000

# Beep on error
setopt BEEP

# Use keycodes (generated via zkbd) if present, otherwise fallback on
[...]

PR #1510, opened 2025-06-11
etc: add known hashes for zprofile and zshrc in macOS 26 beta 1

Samasaur1:
Closes #1509

PR #1629, opened 2025-11-11
[25.05] etc: add known hashes for zprofile and zshrc in macOS 26 beta 1

gamoutatsumi:
backport #1510 to 25.05
relation to: #1621


gamoutatsumi:
worked in my machine

PR #1631, opened 2025-11-12
darwin-rebuild: add --target-host support

hgl:
This add --target-host support to nix-darwin.

https://github.com/Thaigersprint/thaigersprint-2025/issues/2

### DIFF
--- a/pkgs/nix-tools/default.nix
+++ b/pkgs/nix-tools/default.nix
@@ -2,6 +2,7 @@
 , coreutils
 , jq
 , git
+, openssh
 , replaceVarsWith
 , stdenv
 , profile ? "/nix/var/nix/profiles/system"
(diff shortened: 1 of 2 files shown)


mightyiam:
@hgl is this ready for review?

hgl:
Yes

mightyiam:
Did anyone happen to tell you that they might review it when it's ready?

hgl:
@Enzime did, but I think he is currently traveling in Australia.

emaiax:
I wonder if a `--build-host` would also make sense for this PR, wdyt @hgl?

pkgs/nix-tools/default.nix
```diff
@@ -61,8 +62,9 @@ in
     src = ./darwin-rebuild.sh;
 
     replacements = {
-      inherit path nixPath profile;
+      inherit nixPath profile;
       inherit (stdenv) shell;
+      path = "${lib.makeBinPath [ openssh ]}:${path}";
```
Enzime: `path` is set above

hgl:
@emaiax with `targetHostCmd` implemented, it should be relatively easy to add `--build-host`. Can be addressed in another PR.

pkgs/nix-tools/default.nix
hgl: I just realized I didn’t add openssh to path at the top because it isn’t needed by other CLIs.

Issue #1632, opened 2025-11-12
Outdated instruction for Determinate Nix


stmotw:
Determinate Nix no longer distributes vanilla nix by default (see [this pr](https://github.com/DeterminateSystems/nix-installer/commit/6c62aa16c257e25ffccd391fe849e8a404fa4f87) from 3 days ago).

So you will not be prompted to install Determinate Nix and cannot chose `no`.

For now you can run
```
curl -fsSL https://install.determinate.systems/nix | sh -s -- install --prefer-upstream-nix
```
but they don't promise support beyond January 1, 2026.

Docs now need new preferred method for getting nix, before installing nix-darwin.

andreacfromtheapp:
I'm new to Nix and nix-darwin but not to Linux and system administration. I've been trying to understand what Determinate Nix is (a fork?), because I wanted to switch to it and learn for servers and work stuff. However, Lix has been working great on my MBP so far. I'll stick with Lix on my Darwin systems. Will leave Determinate Nix for when I'll have Nix machines. I want to switch back to Linux (and move to Nix) ASAP.

Anyway, I guess that this is a crossroad for nix-darwin, where to stick with Lix (and upstream) or go all in with Determinate Nix? 

Either way I'd be happy for my Darwin systems. As long as it would be "stable" going forward.

My 2c

jhrr:
Yeah, I'm also trying to begin using Nix on my MBP and after several weeks of reading about how to best do this and talking myself into it I am now a bit stuck here at installation due to Determinate's latest manoeuvres.

- https://determinate.systems/blog/installer-dropping-upstream/
- https://www.reddit.com/r/NixOS/comments/1ndh3yd/dropping_upstream_nix_from_determinate_nix/

I'd much rather use Nix than Lix ultimately, despite preferring Lix's general OSS position.

Should we switch to the experimental installer?

- https://github.com/NixOS/experimental-nix-installer

What do the nix-darwin maintainers recommend we do now if we want to start using it?

Thanks for all the effort.

andreacfromtheapp:
I see your points and tend to agree with them. I'm not knowledgeable enough to talk about forks and all the shenanigans. Or why Determinate is going to make a better Nix. Given that Eelco is one of the founders, I'd bet on Determinate if it was me.

> I'd much rather use Nix than Lix ultimately, despite preferring Lix's general OSS position.
> 
> Should we switch to the experimental installer?
> 
[...]

PR #1642, opened 2025-11-18
readme: update determinate installer instructions

Samasaur1:
The `--prefer-upstream-nix` flag is only guaranteed to be supported until January 1st, but what's described now doesn't work at all, so this is still an improvement.

See also #1632 and #1588


Samasaur1:
ah this is gonna fail because fish is still broken, isn't it...

PR #1659, opened 2025-12-18
readme: recommend only the Lix installer

### LINKED ISSUE #1632: Outdated instruction for Determinate Nix

Samasaur1:
Putting this up for review now in case things don't come together in time and we have to drop the reference to the Determinate Nix installer before the new upstream installer is ready.

Will close #1632

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,20 +13,9 @@ nix-darwin is built up around [Nixpkgs](https://github.com/NixOS/nixpkgs), quite
 
 The only prerequisite is a Nix implementation; both Nix and Lix are supported.
 
-As the official Nix installer does not include an automated uninstaller, and manual uninstallation on macOS is a complex process, we recommend using one of the following installers instead:
+As the official Nix installer does not include an automated uninstaller, and manual uninstallation on macOS is a complex process, we recommend using the [Lix installer](https://lix.systems/install/#on-any-other-linuxmacos-system), which supports both flake-based and channel-based setups.
 
-* The [Nix installer from Determinate Systems](https://github.com/DeterminateSystems/nix-installer?tab=readme-ov-file#determinate-nix-installer) is only recommended for use with flake-based setups.
-  It can install one of two distributions of Nix:
-
-  * To install the **recommended** vanilla upstream [Nix](https://nixos.org), you will need to explicitly pass the `--prefer-upstream-nix` flag to the installer.
-
-  * When run without the `--prefer-upstream-nix` flag, it will install the [Determinate](https://docs.determinate.systems/) distribution.
-    As Determinate manages the Nix installation itself, you will need to set `nix.enable = false;` in your configuration to disable nix-darwin’s own Nix management.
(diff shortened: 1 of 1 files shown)


Samasaur1:
If we want I can also add a paragraph along the lines of:

> Using the Lix installer does not preclude you from running upstream Nix. nix-darwin manages the Nix installation by default and will default to upstream Nix.

bestlem:
For that paragraph I would  add the actual nix code to do this, especially as this will be the first thing new users have to do so there needs to be the exact commands to set nix-darwin up, perhaps as a link if too much for this part.

Enzime:
I think we should add an extra paragraph that states something like "When you install nix-darwin, it will change your default Nix to upstream Nix, please set `nix.package = pkgs.lix` if you wish to use Lix instead of Nix."

README.md
```diff
@@ -13,20 +13,9 @@ nix-darwin is built up around [Nixpkgs](https://github.com/NixOS/nixpkgs), quite
...
-
-  * When run without the `--prefer-upstream-nix` flag, it will install the [Determinate](https://docs.determinate.systems/) distribution.
-    As Determinate manages the Nix installation itself, you will need to set `nix.enable = false;` in your configuration to disable nix-darwin’s own Nix management.
-    Some nix-darwin functionality that relies on managing the Nix installation, like the `nix.*` options to adjust Nix settings or configure a Linux builder, will be unavailable.
-
-* The [Lix installer](https://lix.systems/install/#on-any-other-linuxmacos-system) supports both flake-based and channel-based setups.
-
+As the official Nix installer does not include an automated uninstaller, and manual uninstallation on macOS is a complex process, we recommend using the [Lix installer](https://lix.systems/install/#on-any-other-linuxmacos-system), which supports both flake-based and channel-based setups.
 
+Using the Lix installer does not preclude you from running upstream Nix. nix-darwin manages the Nix installation by default and will default to upstream Nix. If you wish to use Lix instead of Nix, set `nix.package = pkgs.lix` in your configuration.
```
Enzime:
```suggestion
The installer you use doesn't affect which Nix interpreter your system will use later on. nix-darwin manages the Nix installation by default and will default to upstream Nix. If you wish to use Lix instead of Nix, set `nix.package = pkgs.lix` in your configuration.
```

I think something simpler to understand would be better

This one is just a starting point, I haven't thought that hard about how to word it nicely

Samasaur1: that's much nicer and lines up exactly with what I was thinking

Mic92:
For the record, I am hoping to get the new Nix installer ready for February. We want to get a month of public testing to not rush things.

PR #1662, opened 2025-12-21
security.pam: add U2F authentication support

delafthi:
Adds support for U2F authentication in PAM by introducing a new `security.pam.u2f` module with the same configuration options as on NixOS for hardware security keys like YubiKeys. This enables multi-factor authentication or passwordless authentication for sudo operations on macOS.

The implementation includes:
- New `security.pam.u2f.*` options for configuring pam_u2f module
- `security.pam.services.sudo_local.u2fAuth` option to enable U2F for sudo
- Integrates with existing Touch ID, Apple Watch, and pam-reattach options
- Adds a test for the PAM module

Note: On macOS, the `u2f_keys` file must be placed in a root-accessible directory (e.g., `/etc/u2f_mappings`) since openpam cannot drop privileges to read user files.

I wasn't sure how to format the files and have used `nixfmt`. Though, I saw that certain files follow an other formatting style (maybe alejandra?). It would be great if `nix fmt` would be set up.

### DIFF
--- a/modules/security/pam.nix
+++ b/modules/security/pam.nix
@@ -1,4 +1,9 @@
-{ config, lib, pkgs, ... }:
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
 
 let
   cfg = config.security.pam.services.sudo_local;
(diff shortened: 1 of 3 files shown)


delafthi:
@Samasaur1 just a friendly ping. Would be great if this could be merged. I'm using this now for nearly half a year with no issues at all.

PR #1667, opened 2025-12-25
feat(gnupg): add configurable package option and install it

iamanaws:


modules/programs/gnupg.nix
```diff
@@ -29,9 +41,12 @@ in
   };
 
   config = mkIf cfg.agent.enable {
+    environment.systemPackages = [ cfg.package ];
+
     launchd.user.agents.gnupg-agent.serviceConfig = {
       ProgramArguments = [
-        "${pkgs.gnupg}/bin/gpg-connect-agent" "/bye"
+        "${cfg.package}/bin/gpg-connect-agent"
```
Enzime:
```suggestion
        (lib.getExe' cfg.package "gpg-connect-agent")
```

Can you change everything to use `lib.getExe'`?

iamanaws:
done

PR #1669, opened 2025-12-27
darwin-rebuild: add flake options to zsh completions

iamanaws:
I noticed `darwin-rebuild` was missing completions for flakes (it has been annoying me for a while), so I added the most common ones. 

I left out some of the more niche debugging/CI flags to keep the list from getting too cluttered, but I’m happy to add them if you think it’s better to be exhaustive!

**Left out for now:**
```zsh
  '--no-update-lock-file[Do not update flake.lock]'\
  '--no-write-lock-file[Do not write flake.lock]'\
  '--no-registries[Do not use flake registries]'\
  '--commit-lock-file[Commit flake.lock updates]'\
  '--substituters[Override substituters list]:substituters:'\
```

### DIFF
--- a/pkgs/nix-tools/darwin-rebuild.zsh-completions
+++ b/pkgs/nix-tools/darwin-rebuild.zsh-completions
@@ -21,6 +21,17 @@ _arguments \
   '(--keep-going -k)'{--keep-going,-k}"[keep going until all builds are finished]"\
   '(--keep-failed -K)'{--keep-failed,-K}'[keep failed builds (usually in /tmp)]'\
   '--fallback[If binary download fails, fall back on building from source]'\
+  '--flake[Build from the given flake URI]:flake uri:'\
+  '--no-flake[Disable implicit flake detection]'\
+  '(-L --print-build-logs)'{-L,--print-build-logs}'[Print build logs during the build]'\
+  '--refresh[Refresh flake inputs]'\
+  '--impure[Allow impure evaluation for flakes]'\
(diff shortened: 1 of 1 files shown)


pkgs/nix-tools/darwin-rebuild.zsh-completions
```diff
@@ -21,6 +21,17 @@ _arguments \
   '(--keep-going -k)'{--keep-going,-k}"[keep going until all builds are finished]"\
   '(--keep-failed -K)'{--keep-failed,-K}'[keep failed builds (usually in /tmp)]'\
   '--fallback[If binary download fails, fall back on building from source]'\
+  '--flake[Build from the given flake URI]:flake uri:'\
+  '--no-flake[Disable implicit flake detection]'\
+  '(-L -vL --print-build-logs)'{-L,-vL,--print-build-logs}'[Print build logs during the build]'\
```
Samasaur1: Let's leave out `-vL` here since it's not listed in the output of `--help` and I think we want to discourage it in favor of either `-L` or the long form.

Samasaur1 (changes requested):

pkgs/nix-tools/darwin-rebuild.zsh-completions
iamanaws: done!

Samasaur1 (approved):

Samasaur1:
failing check was due to an intermittent network error; I've kicked it again but feel free to ping me if it fails again (I'll try to keep on top of the github notification but I may miss it)

PR #1670, opened 2025-12-28
Add flake-parts module

mmlb:
Everything was fine using flake-parts and nix-darwin while it was just one machine, but when I tried to add a second machine I'd get the following error:

```
error: The option `flake.darwinConfigurations' is defined multiple times while it's expected to be unique.
...
```

This avoids that issue, and I can't be the only one hitting this so would be good to just have it defined upstream.

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -1,0 +1,11 @@
+{
+  lib,
+  ...
+}:
+{
+  options.flake.darwinConfigurations = lib.mkOption {
+    type = lib.types.lazyAttrsOf lib.types.raw;
+    default = { };
+    description = "Darwin system configurations";
+  };
+}
--- a/flake.nix
+++ b/flake.nix
@@ -56,6 +56,8 @@
       darwin-uninstaller = prev.callPackage ./pkgs/darwin-uninstaller { };
     };
 
+    flakeModules.default = ./flake-module.nix;
+
     darwinModules.hydra = ./modules/examples/hydra.nix;
     darwinModules.lnl = ./modules/examples/lnl.nix;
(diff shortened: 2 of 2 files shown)


flake-module.nix
```diff
@@ -0,0 +1,16 @@
+{
+  lib,
+  flake-parts-lib,
+  ...
+}:
+{
+  options = {
+    flake = flake-parts-lib.mkSubmoduleOptions {
```
Enzime:
```suggestion
    flake = {
```

That function is unnecessary

flake.nix
```diff
@@ -56,6 +56,11 @@
       darwin-uninstaller = prev.callPackage ./pkgs/darwin-uninstaller { };
     };
 
+    flakeModules = {
+      darwin = ./flake-module.nix;
+      default = self.flakeModules.darwin;
+    };
```
Enzime: Can you just expose it as `flakeModules.default`?

Enzime (changes requested):

flake-module.nix
mmlb: dropped

flake.nix
mmlb: done

mmlb:
I'v updated the PR @Enzime with your feedback addressed, PTAL. Thanks!

Enzime (approved):

IanHollow:
@mmlb can you add support for darwinModules like home-manager has homeModules for flake parts. Here is the file to copy as an example. https://github.com/nix-community/home-manager/blob/master/flake-module.nix

PR #1675, opened 2026-01-03
darwin-rebuild: support --log-format

isabelroses:


remi-dupre:
I would love that, it allows integration with alternative frontends such as [nix-output-monitor](https://github.com/maralorn/nix-output-monitor)

Samasaur1 (approved):
Apologies for the delay

Samasaur1:
I think this merge is going to fail because our tests are broken due to an upstream change, but I'll re-queue it if necessary

PR #1677, opened 2026-01-06
feat(screencapture): add `save-selections`

vaaski:
While setting up my Mac it seems like `cmd+shift+5` remembers the last selection by default and upon investigating I found that nix-darwin doesn't seem to have an option for this.

After comparing the output of `defaults read` before and after toggling the option, I found the setting and verified that it works by putting it into `system.defaults.CustomUserPreferences."com.apple.screencapture"."save-selections"`.

This is my first ever Nix-related contribution and I've taken #1242 as inspiration, let me know if I overlooked anything! :)

image

### DIFF
--- a/tests/fixtures/system-defaults-write/user.txt
+++ b/tests/fixtures/system-defaults-write/user.txt
@@ -545,15 +545,15 @@ launchctl asuser "$(id -u -- test-defaults-user)" sudo --user=test-defaults-user
 <plist version="1.0">
 <true/>
 </plist>'
-launchctl asuser "$(id -u -- test-defaults-user)" sudo --user=test-defaults-user -- defaults write com.apple.screencapture save-selections '<?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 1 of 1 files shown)


tests/fixtures/system-defaults-write/user.txt
```diff
@@ -545,6 +545,11 @@ launchctl asuser "$(id -u -- test-defaults-user)" sudo --user=test-defaults-user
 <plist version="1.0">
 <true/>
 </plist>'
+launchctl asuser "$(id -u -- test-defaults-user)" sudo --user=test-defaults-user -- defaults write com.apple.screencapture save-selections '<?xml version="1.0" encoding="UTF-8"?>
```
Enzime: The tests are currently failing

I suspect this is probably in the wrong place in the file, the easiest way to find out is to remove it and build the test and see where it expects this to be in the file

vaaski: I'm not sure what the right way to figure out the correct location is, I just moved it down one "block" and now `nix-build release.nix -A tests.system-defaults-write` passes locally.

> image
>
> I couldn't see where it expected it from these logs.

> image
> 
> I saw that the defaults have `"save-selections"` after location, so that's the first thing I've tried. Is that a way to figure it out?

PR #1681, opened 2026-01-19
tmux: update key binding

ImSingee:
I believe this is a typo. By default, `%` represents a horizontal split, while `"` represents a vertical split. However, this configuration reverses their functionality.

I read the [tmux-sensible](https://github.com/tmux-plugins/tmux-sensible/blob/master/sensible.tmux) config and it doesn't modify this anywhere.


sinrohit (approved):

booxter:
In HM, there's reverse option for this: https://github.com/nix-community/home-manager/blob/175532b6275b34598a0ceb1aef4b9b4006dd4073/modules/programs/tmux.nix#L62-L65 Maybe relevant.

Samasaur1:
Apologies for the delay in getting to this PR. I agree that this was probably a typo, but now that it's out there I'd like to avoid breaking users' existing configs out from under them.

To that end, rather than changing the config directly, you should add a new boolean option that controls whether these key bindings are reversed. It should default to preserving the existing behavior, which you can do either by setting it explicitly to true (assuming the option you create is `reverseSplitBindings` or similar), or you can set it to true if and only if `config.system.stateVersion` is less than or equal to 6. If you go the stateVersion route, you should also bump `maxStateVersion` to 7 and should add an entry to the changelog.

Samasaur1 (changes requested):
see my comment

ImSingee:
@Samasaur1 Thanks for your review! I've updated the code to reserve backward-compatibility

ImSingee:
@Samasaur1 I think the failed test isn't my fault. I create a separate PR for this https://github.com/nix-darwin/nix-darwin/pull/1772

Samasaur1:
The tests are definitely not your fault. I created a new PR to fix them because it's actually a revert of a change we just did recently, and when that (#1773) gets in we can finally merge this

Samasaur1 (approved):

PR #1682, opened 2026-01-26
feat(autoUpgrade): add system.autoUpgrade in nix-darwin

sinrohit:
Resolves https://github.com/nix-darwin/nix-darwin/issues/1665

**Testing:**

I have tested with both flakes and nix channels.

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -16,6 +16,7 @@
   ./system/checks.nix
   ./system/activation-scripts.nix
   ./system/applications.nix
+  ./system/auto-upgrade.nix
   ./system/defaults-write.nix
   ./system/defaults/controlcenter.nix
   ./system/defaults/LaunchServices.nix
--- a/modules/system/auto-upgrade.nix
+++ b/modules/system/auto-upgrade.nix
@@ -1,0 +1,155 @@
+{ config, lib, ...}:
+with lib;
+let
+  cfg = config.system.autoUpgrade;
+  launchdTypes = import ../launchd/types.nix { inherit config lib; };
+
+in
+
+{
+  imports = [
+    (mkRemovedOptionModule [ "autoUpgrade" "dates" ] "Use `autoUpgrade.interval` instead.")
+    (mkRemovedOptionModule [ "autoUpgrade" "fixedRandomDelay" ] "No `nix-darwin` equivalent to this NixOS option.")
+    (mkRemovedOptionModule [ "autoUpgrade" "persistent" ] "No `nix-darwin` equivalent to this NixOS option.")
+    (mkRemovedOptionModule [ "autoUpgrade" "randomizedDelaySec" ] "No `nix-darwin` equivalent to this NixOS option.")
+    (mkRemovedOptionModule [ "autoUpgrade" "reboot" ] "No `nix-darwin` equivalent to this NixOS option.")
+    (mkRemovedOptionModule [ "autoUpgrade" "rebootWindow" ] "No `nix-darwin` equivalent to this NixOS option.")
+    (mkRemovedOptionModule [ "autoUpgrade" "runGarbageCollection" ] "No `nix-darwin` equivalent to this NixOS option.")
+  ];
+
+  options = {
+    system.autoUpgrade = {
+
+      enable = mkEnableOption ''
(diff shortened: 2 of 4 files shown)


booxter:
I have my own implementation of the same idea and I hit this issue:

```
$ tail -f nix-auto-upgrade.log
Do not sleep when on AC power.
Prefer network over sleep.
Activating home-manager configuration for ihrachyshka
Starting Home Manager activation
Activating checkAppManagementPermission
error: permission denied when trying to update apps, aborting activation
home-manager requires permission to update your apps, please accept the notification
and grant the permission for your terminal emulator in System Settings.

If you did not get a notification, you can navigate to System Settings > Privacy & Security > App Management.
```

Looks like this approach is not very compatible with the fact MacOS really wants you to click buttons to grant permissions. :-|

---

Moreover, when it happens, I get a popup on my desktop saying `nix` is trying to manage apps and needs permissions (good) but when I click Allow... it opens the dialog without an option to grant `nix` permissions (the list is empty). In contrast, when this happens when switching from a terminal, I get the same dialog (for `Aerospace` which is the origin of the terminal launch) and when I click that button, I at least see the Aerospace option that I can enable.

sinrohit:
> I have my own implementation of the same idea and I hit this issue:
> 
> ```
> $ tail -f nix-auto-upgrade.log
> Do not sleep when on AC power.
> Prefer network over sleep.
> Activating home-manager configuration for ihrachyshka
> Starting Home Manager activation
> Activating checkAppManagementPermission
> error: permission denied when trying to update apps, aborting activation
> home-manager requires permission to update your apps, please accept the notification
> and grant the permission for your terminal emulator in System Settings.
> 
> If you did not get a notification, you can navigate to System Settings > Privacy & Security > App Management.
> ```
> 
> Looks like this approach is not very compatible with the fact MacOS really wants you to click buttons to grant permissions. :-|
> 
> Moreover, when it happens, I get a popup on my desktop saying `nix` is trying to manage apps and needs permissions (good) but when I click Allow... it opens the dialog without an option to grant `nix` permissions (the list is empty). In contrast, when this happens when switching from a terminal, I get the same dialog (for `Aerospace` which is the origin of the terminal launch) and when I click that button, I at least see the Aerospace option that I can enable.

It seems to run clean for me, maybe because I am not using such programs.

```console
warning: '--update-input' is a deprecated alias for 'flake update' and will be removed in a future version.
unpacking 'github:nixos/nixpkgs/078d69f03934859a181e81ba987c2bb033eebfc5' into the Git cache...
setting up groups...
updating group members _github-runner...
setting up users...
setting up /Applications/Nix Apps...
setting up pam...
applying patches...
setting up /etc...
setting up GitHub Runner 'macos-runner3'...
setting up GitHub Runner 'macos-runner2'...
setting up GitHub Runner 'macos-runner1'...
removing service org.nixos.nix-darwin-upgrade
```

Anyways `launchd` won't run autoUpgrade when Mac is asleep, so the popup things shouldn't be a limitation as such (Mac Enforcing Security). As if it misses to run autoUpgrade during sleep, it runs after waking up.  But the list being empty definitely looks like a problem.

booxter:
It may be that this is incompatible with `copyApps` only (both for HM and nix-darwin modules), so if one does not use it, it should work. If so, maybe there should be a warning or error on attempt to use both.

In your testing, do you use `copyApps`?

eveeifyeve:
A feature that I want to propose, what if nix-darwin could update macos as well, if there is an update?

EDIT: nvm this is already a feature.

Samasaur1:
I would also be concerned about some of the newer system management commands, which I know can't be run over SSH. It's not entirely clear from the Apple docs whether these commands would work in a LaunchDaemon, but we'd want to check those as well

sinrohit:
> It may be that this is incompatible with `copyApps` only (both for HM and nix-darwin modules), so if one does not use it, it should work. If so, maybe there should be a warning or error on attempt to use both.
> 
> In your testing, do you use `copyApps`?

Apologies for the delayed reply, I don't use `copyApps` in home-manager and that could be the reason why I didn't get that error.  

Now that nix-darwin makes packages from `environment.systemPackages` discoverable in spotlight, I should also test that.

PR #1683, opened 2026-01-26
programs/tailscale-gui: init module

ap-1:
Adds a module for the Tailscale GUI application, following the same pattern as `programs._1password-gui`.

The Tailscale GUI for macOS requires running from `/Applications` to function correctly (it uses system extensions for VPN functionality). I think this is the same situation as 1Password GUI (NixOS/nixpkgs#254944). I copied the approach in `programs/_1password-gui`, copying the app to `/Applications` using `rsync`.

I tested by adding this to my config, and things seem to be working as expected:

```nix
programs.tailscale-gui.enable = true;
```

nixpkgs PR adding tailscale-gui package: https://github.com/NixOS/nixpkgs/pull/483910

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -115,6 +115,7 @@
   ./programs/info
   ./programs/nix-index
   ./programs/ssh.nix
+  ./programs/tailscale-gui.nix
   ./programs/tmux.nix
   ./programs/vim.nix
   ./programs/zsh
--- a/modules/programs/tailscale-gui.nix
+++ b/modules/programs/tailscale-gui.nix
@@ -1,0 +1,41 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
+
+let
+  cfg = config.programs.tailscale-gui;
+in
+{
+  options = {
+    programs.tailscale-gui = {
+      enable = lib.mkEnableOption "the Tailscale GUI application";
+      package = lib.mkPackageOption pkgs "Tailscale GUI" {
+        default = [ "tailscale-gui" ];
+      };
+    };
+  };
+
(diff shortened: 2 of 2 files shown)


ap-1:
`environment.systemPackages` now copies Applications to `/Applications`

_Originally posted by @Enzime in https://github.com/nix-darwin/nix-darwin/issues/1684#issuecomment-3805849570_

@Enzime Must this be invoked within a nix-darwin module in this repository? I'm unable to replicate the behavior of this PR from just adding `tailscale-gui` to my personal `environment.systemPackages`, but it did work for `cloudflare-warp` in https://github.com/nix-darwin/nix-darwin/pull/1684 when I added it to that module.

Enzime:
No, it should work, I would build the package and verify that it actually has an `Applications` folder in the store path

ap-1:
> No, it should work, I would build the package and verify that it actually has an `Applications` folder in the store path

@Enzime turns out `environment.systemPackages` actually copies to `/Applications/Nix Apps/`, not `/Applications/`. I didn't catch that in `cloudflare-warp` for another reason: https://github.com/nix-darwin/nix-darwin/pull/1684#issuecomment-3808173910

Anyway, it turns out Tailscale has to run specifically from `/Applications/`:

```
Tailscale/BundleIdentifiers.swift:41: Fatal error: The current bundleIdentifier is unknown to the registry
```

This seems to be working now. Do you think the `rsync` method is the best option here?

Enzime:
Yeah `rsync` is the best way for now, ideally we would build some kind of interface for packages that need to live in `/Applications` directly (or have a way for allowing packages to specify what folder they should be installed in)

PR #1684, opened 2026-01-27
programs/cloudflare-warp: init module

ap-1:
I recently added Darwin support to `cloudflare-warp` in nixpkgs (see https://github.com/NixOS/nixpkgs/pull/481432 and https://github.com/NixOS/nixpkgs/pull/481599). Cloudflare WARP also requires a launchd daemon to be running for the GUI to communicate with, which this module sets up via `launchd.daemons`.

I tested this and it worked for me. However, if you have a previous Cloudflare WARP installation and get "Failed to save to disk" when running `warp-cli registration new`, you may need to clear stale keychain entries:

```bash
sudo security delete-generic-password -s "WARP" /Library/Keychains/System.keychain
sudo launchctl unload /Library/LaunchDaemons/com.cloudflare.1dot1dot1dot1.macos.warp.daemon.plist
sudo launchctl load /Library/LaunchDaemons/com.cloudflare.1dot1dot1dot1.macos.warp.daemon.plist
warp-cli registration new
```

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -108,6 +108,7 @@
   ./programs/_1password-gui.nix
   ./programs/arqbackup.nix
   ./programs/bash
+  ./programs/cloudflare-warp.nix
   ./programs/direnv.nix
   ./programs/fish.nix
   ./programs/gnupg.nix
--- a/modules/programs/cloudflare-warp.nix
+++ b/modules/programs/cloudflare-warp.nix
@@ -1,0 +1,43 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
+
+let
+  cfg = config.programs.cloudflare-warp;
+in
+{
+  options = {
+    programs.cloudflare-warp = {
+      enable = lib.mkEnableOption "the Cloudflare WARP application";
+      package = lib.mkPackageOption pkgs "Cloudflare WARP" {
+        default = [ "cloudflare-warp" ];
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages = [ cfg.package ];
+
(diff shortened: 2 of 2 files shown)


Enzime:
Does `environment.systemPackages = [ pkgs.cloudflare-warp ];` not work?

ap-1:
> Does `environment.systemPackages = [ pkgs.cloudflare-warp ];` not work?

`environment.systemPackages` alone won't work because Cloudflare WARP requires a launchd daemon (`CloudflareWARP`) to be running for the app to function. Without the daemon, the GUI shows "Unable to connect: IPC error".

The `/Applications` copy isn't strictly required for the daemon or GUI to work (they can run from the nix store), but without it the app won't appear in Spotlight/Raycast. This matches the pattern used by `programs._1password-gui`.

Enzime:
`environment.systemPackages` now copies Applications to `/Applications`

ap-1:
> `environment.systemPackages` now copies Applications to `/Applications`

Good to know, this simplified the module a lot

ap-1:
Sorry for the repeated force pushes. I tested using `environment.systemPackages` for the app copy, and adjusted `ProgramArguments` under the assumption that applications were copied to `/Applications`, like you mentioned.

It turns out that it was actually being copied to `/Applications/Nix Apps`. This caused the daemon to fail with exit code 78 after rebuild, but I didn't catch this because I had an manually-started daemon still running that I was using to test the GUI earlier. Then adjusting `ProgramArguments` to `/Applications/Nix Apps` and rebuilding clean still did not fix it because the daemon was trying to start before the activation script finished copying the app.

I fixed it by pointing the daemon at the nix store path while still using `environment.systemPackages` to make the GUI discoverable, and this appears to work now.

PR #1690, opened 2026-02-04
flake-module: add darwinModules option for flake-parts integration

IanHollow:
This PR enhances the flake-parts module to add support for `darwinModules`, following the same pattern as [home-manager's implementation](https://github.com/nix-community/home-manager/blob/master/flake-module.nix). This allows users to expose reusable nix-darwin modules through their flakes when using flake-parts.

This builds off of the previous PR #1670

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -1,11 +1,46 @@
 {
   lib,
+  flake-parts-lib,
+  moduleLocation,
   ...
 }:
+let
+  inherit (lib)
+    mapAttrs
+    mkOption
+    types
+    ;
+in
 {
-  options.flake.darwinConfigurations = lib.mkOption {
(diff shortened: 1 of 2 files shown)


anton-rudeshko:
@Enzime Hi Michael, could you please have a look at this one as well?

drewgingerich:
This would be an amazing addition, I just switched to `flake-parts` and am running into trouble using it with `nix-darwin` because there doesn't appear to be flake-level options for `darwinModules` and `darwinConfigurations`.

CallMeLaNN:
I'm about to upgrade to 26.05 current major release since I'm using stable branch. I thought this was merged but missed. So I have to keep this code that I patch on my flake. Hopefully this will be available on the next major release.

PR #1691, opened 2026-02-07
container: init module

### LINKED ISSUE #1645: Add apple container service
Apple container: https://github.com/apple/container

Can't find the service so request to add apple container services. I might able to do it myself, but put it here just in case anyone interests.

heywoodlh:
This is an initial stab to install and start Apple's [container](https://github.com/apple/container) system.

I think that at some point this could be expanded to replicate NixOS' oci-containers module, i.e. see:
- https://github.com/nix-darwin/nix-darwin/issues/1664
- https://github.com/nix-darwin/nix-darwin/issues/1645#issuecomment-3630202606

Closes https://github.com/nix-darwin/nix-darwin/issues/1645

(But I think there's lots of room for improvement/increased functionality :smile:)

### DIFF
--- a/modules/services/container.nix
+++ b/modules/services/container.nix
@@ -1,0 +1,87 @@
+{ config, pkgs, lib, ... }:
+
+let
+  cfg = config.services.container;
+in
+{
+  meta.maintainers = [
+    lib.maintainers.heywoodlh or "heywoodlh"
+  ];
+
+  options = {
+    services.container = {
+      enable = lib.mkOption {
+        type = lib.types.nullOr lib.types.bool;
+        default = null;
+        description = ''
+          Whether to enable Apple's container service.
+        '';
+      };
+
+      package = lib.mkOption {
+        type = lib.types.package;
+        default = pkgs.container;
+        description = ''
+          Package containing `container` executable.
+        '';
+      };
+
+      enableKernelInstall = lib.mkOption {
+        type = lib.types.bool;
+        default = true;
+        description = ''
+          Whether the default kernel should be installed or not.
+        '';
+      };
+
+      appRoot = lib.mkOption {
+        type = lib.types.nullOr lib.types.str;
+        default = null;
+        description = ''
+          Path to the root directory for application data.
+        '';
+      };
+
+      installRoot = lib.mkOption {
+        type = lib.types.nullOr lib.types.str;
+        default = null;
+        description = ''
+          Path to the root directory for application executables and plugins
+        '';
+      };
+    };
+  };
+
+  config = lib.mkIf (cfg.enable != null) {
+    environment.systemPackages = lib.optionals (cfg.enable) [
+      cfg.package
+    ];
+
+    system.activationScripts.launchd.text =
+    let
+      kernelInstall = if cfg.enableKernelInstall then "--enable-kernel-install" else "--disable-kernel-install";
+      installRootArg = lib.optionalString (cfg.installRoot != null) ''--install-root "${cfg.installRoot}"'';
+      appRootArg = lib.optionalString (cfg.appRoot != null) ''--app-root "${cfg.appRoot}"'';
+    in if cfg.enable then ''
+      # If /run/current-system/sw/bin/container does not match package path, assume upgrade/downgrade
+      _prev=$(readlink -f /run/current-system/sw/bin/container 2>/dev/null || true)
+      _new="${cfg.package}/bin/container"
+      if [ -e /run/current-system/sw/bin/container ] && [ -n "$_prev" ] && [ "$_prev" != "$_new" ]
+      then
+        echo "Apple Container update detected. Stopping container service."
+        # Sometimes container can hang, violently stop it (see https://github.com/apple/container/issues/1329#issuecomment-4095140386)
+        /usr/bin/sudo -u ${config.system.primaryUser} launchctl bootout gui/"$(id -u ${config.system.primaryUser})"/com.apple.container.apiserver &>/dev/null || true
+        /usr/bin/sudo -u ${config.system.primaryUser} launchctl remove com.apple.container.apiserver &>/dev/null || true
+        pkill -9 -f com.apple.container &>/dev/null || true
+        /usr/bin/sudo -u ${config.system.primaryUser} ${cfg.package}/bin/container system stop
+        # -k preserves user data
+        /run/current-system/sw/bin/uninstall-container.sh -k
+      fi
+
+      echo "Starting Apple container service."
+      /usr/bin/sudo -u ${config.system.primaryUser} ${cfg.package}/bin/container system start ${installRootArg} ${appRootArg} ${kernelInstall}
+    '' else ''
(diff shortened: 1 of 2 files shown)


heywoodlh:
Failing at the moment due to a seeming MacOS-level bug: https://github.com/apple/container/issues/1254#issuecomment-3956444216

heywoodlh:
I rebased off of current main today, and can confirm that with MacOS Tahoe 26.5.1 this works as expected:

```bash
reloading service org.nixos.activate-system
reloading service org.nixos.nix-daemon
+ /usr/bin/sudo -u heywoodlh /nix/store/xfzanx6ivax98j4ld69j08ffjs4ymlhy-container-0.12.3/bin/container system start --enable-kernel-install
Registering API server with launchd...
Verifying apiserver is running...


<snip>

❯ container list
ID  IMAGE  OS  ARCH  STATE  ADDR  CPUS  MEMORY  STARTED
```

heywoodlh:
This is ready. I've implemented `appRoot` and `installRoot` options (my original implementation had that as a TODO). Tested with the following configuration:

```bash
services.container = {
  enable = true;
  user = "heywoodlh";
  appRoot = "/tmp/approot";
  installRoot = "/tmp/installroot";
};
```

and

```bash
services.container = {
  enable = true;
  user = "heywoodlh";
};
```

And both worked configurations worked.

modules/services/container.nix
```diff
@@ -0,0 +1,85 @@
...
+      cfg.package
+    ];
+
+    system.activationScripts.launchd.text =
+    let
+      kernelInstall = if cfg.enableKernelInstall then "--enable-kernel-install" else "--disable-kernel-install";
+      installRootArg = lib.optionalString (cfg.installRoot != null) ''--install-root "${cfg.installRoot}"'';
+      appRootArg = lib.optionalString (cfg.appRoot != null) ''--app-root "${cfg.appRoot}"'';
+    in if cfg.enable then ''
+      /usr/bin/sudo -u ${cfg.user} ${cfg.package}/bin/container system start ${installRootArg} ${appRootArg} ${kernelInstall}
```
Samasaur1: I'm pretty sure this isn't actually running the service under the other user, it's just making the other user own the generated plist file (and whatever else is created on startup), which might actually cause problems.

I think what you'd need to use is probably `launchctl asuser`, but I'm also not sure you _want_ to start this under the other user because if you do I'm pretty sure it won't be running (and if it is you won't be able to talk to it since LaunchAgents are user-scoped).

The upstream behavior of container is a weird mix of user-scoped and system-scoped things, so I'm not really sure it's a good fit as-is. It might be nicer (though obviously much more work on our part) to build a custom implementation on top of the container package instead of using the CLI. If we do use the CLI, I think it essentially has a non-obvious dependency on `system.primaryUser` since it'll only be running for the user who runs `darwin-rebuild switch`

Samasaur1 (changes requested):
Left a lengthy comment about the user-scoping issues I think we have to deal with. We'd also probably want to build in a check for when the package updates, as upstream docs indicate that the containers need to be stopped and started when this happens. It's not totally clear whether we'd also need to do this on "new versions" (from Nix's perspective) that are the same upstream version.

modules/services/container.nix
heywoodlh: > We'd also probably want to build in a check for when the package updates, as upstream docs indicate that the containers need to be stopped and started when this happens. It's not totally clear whether we'd also need to do this on "new versions" (from Nix's perspective) that are the same upstream version.

I believe I addressed this here: https://github.com/heywoodlh/nix-darwin/blob/a9668a0110a31494784cacb4a7f3978b39fd68a8/modules/services/container.nix#L66-L72

> I think what you'd need to use is probably launchctl asuser
> I think it essentially has a non-obvious dependency on system.primaryUser

In a9668a0110a31494784cacb4a7f3978b39fd68a8 I removed my original `cfg.user` option and instead replaced with `launchctl asuser ${config.system.primaryUser}` since it makes more sense to just assume that `container system start` should be invoked with the user running `darwin-rebuild.

(I didn't know `launchctl asuser` was a thing before, thanks for the pointer!)

Let me know if you have other feedback or this doesn't address your comments.

EDIT: I'm testing the changes to ensure upgrades work properly -- will follow up with results.

heywoodlh: For whatever reason, running `launchctl asuser ${config.system.primaryUser}` seems to only run commands as `root` so using it to setup the container system just doesn't work for non-root user. I've reverted back to `/usr/bin/sudo -u ${config.system.primaryUser}` and it seems to be coming up healthy.

I simulated an upgrade from `0.6.0` (the version in nixpkgs 25.11) to `0.12.3` and here's how the output looks whilst running `darwin-rebuild`:

```
Apple Container update detected. Stopping container service.
2026-06-11T13:53:41-0600 info com.apple.container.cli: [ContainerCommands] checking if APIServer is alive
2026-06-11T13:53:41-0600 info com.apple.container.cli: [ContainerCommands] APIServer health check failed, skipping bootout
2026-06-11T13:53:41-0600 info com.apple.container.cli: label=user/501/com.apple.container.container-network-vmnet.default [ContainerCommands] stopping service
2026-06-11T13:53:41-0600 info com.apple.container.cli: label=user/501/com.apple.container.container-core-images [ContainerCommands] stopping service
2026-06-11T13:53:41-0600 info com.apple.container.cli: label=user/501/com.apple.container.container-runtime-linux.test [ContainerCommands] stopping service
No receipt for 'com.apple.container-installer' found at '/'.
No receipt for 'com.apple.container-installer' found at '/'.
No receipt for 'com.apple.container-installer' found at '/'.
Removed `container` tool and helpers
Starting Apple container service.
Registering API server with launchd...
Verifying apiserver is running...
```

And here's output from my terminal before and after the upgrade (notice the container I started still exists after the upgrade but is stopped):

```
## Before upgrade

❯ container --version
container CLI version 0.6.0 (build: release, commit: a23bcf0)

❯ container run -d --name=test docker.io/alpine sleep infinity
test

❯ container ls
ID    IMAGE                            OS     ARCH   STATE    ADDR          CPUS  MEMORY
test  docker.io/library/alpine:latest  linux  arm64  running  192.168.64.2  4     1024 MB

## After upgrade

❯ container --version
container CLI version 0.12.3 (build: release, commit: f989901)

❯ container ls --all
ID    IMAGE                            OS     ARCH   STATE    ADDR  CPUS  MEMORY   STARTED
test  docker.io/library/alpine:latest  linux  arm64  stopped        4     1024 MB
```

I'm going to mark this PR as ready but will happily address any more feedback if there is any.

heywoodlh:
With latest change I fixed a pretty dumb bug with the update logic on my part: https://github.com/nix-darwin/nix-darwin/compare/8defc34ad270827c4e485b46b65c027fd5772500..e4f95598d1b30bd59f0c2095df3f235a224d3073

I didn't account for installations that didn't already have `container` installed via Nix (i.e. if `/run/current-system/sw/bin/container` doesn't exist).

Issue #1373, opened 2025-03-06
`Warning: Could not write to Brewfile.lock.json` with recent nix-darwin


booxter:
This is probably related to the recent removal of `--no-lock` from `homebrew`.

---

When I do 

```
$ nix run nix-darwin -- switch --flake .#ihrachys-macpro
```

I get:

```
Homebrew bundle...
Using felixkratz/formulae
Using openssh
Using svim
Using amethyst
Using chatgpt
Using todoist
Warning: Could not write to /nix/store/4yjb9sa24m6ayy0rfvghar4xrvvzckp6-Brewfile.lock.json!
Hide this warning by setting HOMEBREW_BUNDLE_NO_LOCKFILE_WRITE_WARNING.
Hide these hints with HOMEBREW_NO_ENV_HINTS (see `man brew`).
Homebrew Bundle complete! 6 Brewfile dependencies now installed.
```

The build succeeds, so it's probably just a warning. Regardless, it's not nice. Should nix-darwin set the `HOMEBREW_BUNDLE_NO_LOCKFILE_WRITE_WARNING` variable globally to suppress it?

booxter:
You are right; I had `autoUpdate` disabled. Sorry for the noise.

Issue #1369, opened 2025-03-05
Homebrew error with nix-darwin - "Error: invalid option: --no-lock"


victorhooi:
I have a macOS 15.3.1 (Sequioia) machine. I've installed Nix using the Determinate Systems NIx installer (**not** the Determinate distro, but vanilla Nix) and installed nix-darwin per the nix-darwin README.

I then edited `/etc/nix-darwin/flake.nix`, to add a Homebrew section:
```
victorhooi@MacBook-Pro:/etc/nix-darwin/ > cat flake.nix 
{
  description = "Example nix-darwin system flake";

  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
    nix-darwin.url = "github:LnL7/nix-darwin/master";
    nix-darwin.inputs.nixpkgs.follows = "nixpkgs";
  };

  outputs = inputs@{ self, nix-darwin, nixpkgs }:
  let
    configuration = { pkgs, ... }: {
      # List packages installed in system profile. To search by name, run:
      # $ nix-env -qaP | grep wget
      environment.systemPackages =
        [ pkgs.vim
        ];

      homebrew = {
        enable = true;
        # onActivation.cleanup = "uninstall";

        taps = [];
        brews = [ "cowsay" ];
        casks = [];
      };


      # Necessary for using flakes on this system.
      nix.settings.experimental-features = "nix-command flakes";

      # Enable alternative shell support in nix-darwin.
      # programs.fish.enable = true;

      # Set Git commit hash for darwin-version.
      system.configurationRevision = self.rev or self.dirtyRev or null;

      # Used for backwards compatibility, please read the changelog before changing.
      # $ darwin-rebuild changelog
      system.stateVersion = 6;

      # The platform the configuration will be used on.
      nixpkgs.hostPlatform = "aarch64-darwin";
    };
  in
  {
    # Build darwin flake using:
    # $ darwin-rebuild build --flake .#MacBook-Pro
    darwinConfigurations."MacBook-Pro" = nix-darwin.lib.darwinSystem {
      modules = [ configuration ];
    };
  };
}
```
After running `darwin-rebuild switch`, I get an error from Homebrew about `Error: invalid option: --no-lock`:
```
victorhooi@MacBook-Pro:/etc/nix-darwin/ > darwin-rebuild switch
building the system configuration...
Password:
setting up user launchd services...
Homebrew bundle...
Usage: brew bundle [subcommand]

[...]

PR #1692, opened 2026-02-10
modules/homebrew: comprehensive module refresh

### LINKED ISSUE #596: homebrew: default `brewPrefix` should point to the default value of `$(brew --prefix)` not `$(brew --prefix)/bin`
The `brewPrefix` value defaults to `/opt/homebrew/bin` or `/usr/local/bin`, which is a bit misleading considering that a "prefix" should refers to the next level up (`/opt/homebrew` or `/usr/local`), which also happens to align with the output of `brew --prefix`.

This confusing terminology has bitten me a couple times when setting `environment.systemPath = ["${config.homebrew.brewPrefix}/bin"]`, which results in the unexpected (and potentially shell-breaking) value of `/opt/homebrew/bin/bin` added to `$PATH`. Also, side note: shouldn't that value be added to `$PATH` automatically if `config.homebrew.enable == true`?

From [brew(1)](https://docs.brew.sh/Manpage#--prefix---unbrewed---installed-formula-):

> ### [](https://docs.brew.sh/Manpage#--prefix---unbrewed---installed-formula-)`--prefix` \[_`--unbrewed`_\] \[_`--installed`_\] \[_`formula`_ …\]
> 
> Display Homebrew’s install path. _Default:_
> 
> -   macOS Intel: `/usr/local`
> -   macOS ARM: `/opt/homebrew`
> -   Linux: `/home/linuxbrew/.linuxbrew`

And from [GNU Autoconf documentation](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/autoconf.html#Default-Prefix):

> By default, `configure` sets the prefix for files it installs to /usr/local.

Also see:

- [4.9. /usr/local : Local hierarchy – Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#usrlocalLocalHierarchy)

### LINKED ISSUE #1032: `homebrew.onActivation.cleanup`: option to fail activation as part of checks
With https://github.com/Homebrew/homebrew-bundle/issues/1418 there is now an easy command available to check if any brews, casks, vscodes would be removed by `--cleanup`. I was thinking about adding another option for `homebrew.onActivation.cleanup` (perhaps `"check"`?), which simply fails the pre-activation checks if the system is not in sync with the generated Brewfile (could also be used with `darwin-rebuild check` without actually performing uninstalls). For now I have something like this in my own config, but thought it seemed like it might be useful enough to upstream; it also relies on some internal options to work:

```nix
{ pkgs, config, ... }:
let
  brewfileFile = pkgs.writeText "Brewfile" config.homebrew.brewfile;
in
{
  system.checks.text = ''
    PATH="${config.homebrew.brewPrefix}":$PATH brew bundle cleanup --file='${brewfileFile}' --no-lock
  '';
}
```

Would a PR adding this as a first-class option be accepted? If so, I wasn't sure whether it should be another choice for `onActivation.cleanup`, or a separate option like `homebrew.onActivation.checkCleanup`. I could see wanting something like this, but I'm not sure if there's any precedent for options like this that have different outcomes during `check` vs `switch`:
```nix
{
	homebrew.onActivation = {
		checkCleanup = true; # only fails `darwin-rebuild check`
		cleanup = "zap"; # invokes --zap on `darwin-rebuild switch`
	};
}
```

malob:
The `homebrew` module had fallen out of sync with upstream Homebrew in several ways: a few options were missing values, some entry types supported by `brew bundle` weren't exposed, and a couple of options had become dead code or actively broken after upstream removals. I audited the module against the current [`brew bundle` source](https://github.com/Homebrew/brew) and the nix-darwin issue tracker, and this PR addresses everything I found. I've tested these changes extensively on my personal machine (83 Brewfile dependencies) and haven't experienced any issues.

I suggest reviewing the changes by going through the commits one at a time, since each is self-contained and the overall diff is hard to parse. The commits are ordered so that simpler changes come first.

Summary of changes:

* **Missing option values and new brew/cask options.** 
  * Added `restart_service: :always` (the enum was missing this variant), `link: :overwrite` (supports `brew link --overwrite`), and `postinstall` (shell command run after install/upgrade, supported by both formula and cask installers).
  * Commits: 36815b4, a3fd89f, and c65c24c.

* **New entry types: `homebrew.goPackages` and `homebrew.cargoPackages`.**
  * These generate `go "pkg"` and `cargo "pkg"` Brewfile entries, matching the entry types that `brew bundle` supports but the module didn't expose. Homebrew auto-installs the `go`/`rust` formulae as needed.
  * Commits: cbe4a60 and 3479b79.

* **Removed `homebrew.whalebrews`.**
  * Whalebrew support was removed from Homebrew Bundle in 4.7.0 (Nov 2025), so `whalebrew` entries now break `brew bundle`. **This is a breaking change**, but the option was already non-functional on current Homebrew versions. Uses `mkRemovedOptionModule` so existing configs get a clear error explaining the removal.
  * Commit: 65cfceb.

* **Removed `homebrew.global.lockfiles`.**
  * Follows up on #1362 which removed the `--no-lock` flag. The `HOMEBREW_BUNDLE_NO_LOCK` env var that this option set is also ignored by current Homebrew, so the option was silently doing nothing. Both `lockfiles` and its predecessor `noLock` now warn that they have no effect.
  * Related: #1369, and #1373.
  * Commit: 2453101.

* **Replaced `brewPrefix` with `prefix`.**
  * Closes: #596. 
  * The old option (one of my biggest regrets 😛) stored the bin directory (`/opt/homebrew/bin`) rather than the actual prefix (`/opt/homebrew`). The new `homebrew.prefix` has correct semantics matching `brew --prefix`. **This is a breaking change**: the old option is removed via `mkRemovedOptionModule`, which catches both setters and readers. A warning also fires if the new `prefix` ends with `/bin`, catching users who copy the old value verbatim.
  * Commit: 8c29e14.

* **Added shell integration options.**
  * `enableBashIntegration`, `enableFishIntegration`, and `enableZshIntegration` source `brew shellenv` in the appropriate shell init file, setting up `PATH`, `MANPATH`, `INFOPATH`, other Homebrew specific env vars, and completions. Follows the `enable<Shell>Integration` naming convention. Defaults to `false` to avoid surprising existing users who already have manual shell setup.
  * Commit: ca6f860.

* **Added `onActivation.cleanup = "check"`.**
  * Closes: #1032.
  * The issue requested a way to detect Homebrew packages not in the Brewfile without removing them. The proposed approach was a separate boolean, but I went with a new enum value instead, since the cleanup modes form a natural spectrum of aggressiveness (`"none"` → `"check"` → `"uninstall"` → `"zap"`). A separate boolean would create a problematic interaction: `checkCleanup = true` + `cleanup = "zap"` would fail the check before activation ever reaches the cleanup step that would fix it. 
  * I considered only running it during `check`, but that would mean `switch` succeeds while `check` fails on the same system — confusing. The option description warns that a failed check blocks the entire activation, not just Homebrew.
  * How it works:
    - Runs `brew bundle cleanup` (without `--force`) during `system.checks.text` to detect extra packages
    - If extras are found, activation fails with the list and remediation steps (add the packages, remove them, or switch to `"uninstall"`/`"zap"`)
    - Unlike `"uninstall"`/`"zap"`, never removes anything
    - Runs during both `darwin-rebuild check` and `darwin-rebuild switch`, consistent with all other system checks
  * **This is the most opinionated change in the PR** — both the enum-over-boolean design and the check-blocks-switch behavior are judgment calls. Happy to revisit either if there are concerns. 
  * Commit: c68f5d1.

* **Option description polish.** 
  * Fixed typos, grammar, and consistency across the module: all boolean options now document Homebrew's defaults, `force_auto_update` notes that Homebrew Bundle may silently ignore it, and added `meta.maintainers`. 
  * Commit: e0ffd55.

Testing:
* Each commit was individually tested against the homebrew module tests and verified to produce correct Brewfile output (Ruby syntax, symbol vs bool vs string rendering, section ordering, `${HOMEBREW_PREFIX}` preservation in `postinstall` strings)
* All backward-compat error messages and warnings verified: `brewPrefix` removal, `whalebrews` removal, `lockfiles` deprecation, `/bin` suffix warning on `prefix`
* `brewBundleCmd` output verified for all four cleanup modes (none/check/uninstall/zap)
* Options JSON / docs build passes with all new options documented
* All option descriptions audited against the current Homebrew Bundle source for accuracy
* Tested on my personal machine (83 Brewfile entries across taps, formulae, casks, and mas apps): full `darwin-rebuild switch`, shell integration in zsh and fish, and cleanup `"check"` mode correctly blocking activation for an unlisted cask

This PR builds on #1222 (VS Code extensions) and reorders the `vscode` option to match Homebrew Bundle's canonical entry type ordering.

Fixes #596
Fixes #1032

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,32 @@
+2026-02-10
+- Major changes to `homebrew` module
+
+  `homebrew.brewPrefix` was renamed to `homebrew.prefix`, and its semantics
+  changed: the old option pointed to the bin directory (e.g.,
+  `/opt/homebrew/bin`), while the new option points to the Homebrew prefix
+  (e.g., `/opt/homebrew`), matching `brew --prefix`.
+
+  `homebrew.whalebrews` was removed. Whalebrew support was removed from
+  Homebrew Bundle in Homebrew 4.7.0 (Nov 2025), so `whalebrew` entries in a
+  Brewfile now cause `brew bundle` to fail.
+
+  `homebrew.global.lockfiles` and `homebrew.global.noLock` no longer have any
+  effect. Homebrew Bundle removed lockfile support in Homebrew 4.4.0 (Oct 2024).
+
+  `homebrew.onActivation.cleanup` now supports a `"check"` mode, which checks
+  for unlisted packages and aborts activation if any are found, without
+  removing them.
+
+  Shell integration options were added: `homebrew.enableBashIntegration`,
+  `homebrew.enableFishIntegration`, and `homebrew.enableZshIntegration`.
+
+  New Brewfile entry types were added: `homebrew.goPackages`,
+  `homebrew.cargoPackages`, and `homebrew.vscode`.
+
+  New options were added for brews: `postinstall`, `link = "overwrite"`, and
+  `restart_service = "always"`. The `postinstall` option was also added for
+  casks.
+
 2025-01-30
 - Previously, some nix-darwin options applied to the user running
   `darwin-rebuild`. As part of a long‐term migration to make


modules/homebrew.nix
```diff
@@ -557,23 +557,22 @@ in
   imports = [
     (mkRenamedOptionModule [ "homebrew" "autoUpdate" ] [ "homebrew" "onActivation" "autoUpdate" ])
     (mkRenamedOptionModule [ "homebrew" "cleanup" ] [ "homebrew" "onActivation" "cleanup" ])
+    (mkRemovedOptionModule [ "homebrew" "whalebrews" ] "Whalebrew support was removed from Homebrew Bundle in Homebrew 4.7.0 (Nov 2025). `whalebrew` entries in a Brewfile now cause `brew bundle` to fail. Please manage Whalebrew images directly using the `whalebrew` CLI.")
```
Samasaur1, line 568: Given that we used to add the `whalebrew` formula to the user's brews automatically if they had any entries in `homebrew.whalebrews`, it might be nice to explicitly state here that they'll need to (re)install whalebrew themselves, possibly by adding it to `homebrew.brews`.

This is totally a judgement call thing and I am not a homebrew user, so if you feel that it's clear enough you can just mark this as resolved and move on :)

tests/homebrew-shell-integration.nix
```diff
@@ -0,0 +1,27 @@
+{ config, ... }:
+
+{
+  homebrew.enable = true;
+  programs.fish.enable = true;
```
Samasaur1, line 7: I know `programs.bash.enable` and `programs.zsh.enable` default to true, but it would warm my heart if you either set them explicitly here or left a comment explaining why you did not need to do so

modules/homebrew.nix
```diff
@@ -379,19 +383,31 @@ let
...
 
           Homebrew's default is `false`.
         '';
       };
       no_quarantine = mkNullOrBoolOption {
-        description = "Whether to disable quarantining of downloads.";
+        description = ''
+          Whether to disable quarantining of downloads.
+
+          Homebrew's default is `false`.
```
Samasaur1, line 398: Maybe worth noting that this is now deprecated in Homebrew and will likely be removed in the future

Samasaur1 (commented):
I really appreciate both the triage you did on our open Homebrew-related issues and this PR, which was so well structured it was actually enjoyable to read in a way that few PRs are.

This all looks great to me in general. I have a couple nits and I want to raise the question of the now-deprecated `-no-quarantine` option, since it's something we'll likely have to handle in the future, so I'd like to hear your thoughts, but I would also be totally happy merging this PR as-is

modules/homebrew.nix
malob: Good catch! Done.

tests/homebrew-shell-integration.nix
malob: Done.

malob:
@Samasaur1, thanks man :)

I usually try to hold myself to a pretty high standard, especially when making changes this big, and do what I can to make things as easy as possible for maintainers. 

I've been meaning to make a number of these changes, e.g., fixing the Homebrew prefix stuff and adding the shell integrations for a while. Honestly, I probably wouldn't have got around to it, if it wasn't for how much easier/faster doing this kind of work has become with the help of Claude Code. I essentially just tasked it with scouring the Homebrew bundle code base and docs for any feature mismatches etc., as well as the issues on this repo related to the module, then had it come up with an implementation plan and then just work through them systematically with it's help. I mostly sat back and let it do its thing, and just provided it with some good guidance and code review during the process. It was a breeze.

iniw:
I noticed that the shell integration stuff uses `interactiveShellInit`. Could it be changed to use `shellInit` instead? [`nix-homebrew`](https://github.com/zhaofengli/nix-homebrew) made the same mistake and I've tried fixing it there, but the [PR](https://github.com/zhaofengli/nix-homebrew/pull/119) hasn't been accepted yet. The motivation/reasoning for the change is in the PR description :)

bestlem:
@iniw Please not - I use brew only from an interactive command line and only to update it.

I am only using homebrew fopr installing things that are not in nixpkgs which currently are just macOS apps

So do you but your issue is that you are not starting the shell interactively - please do not break usual workflows for this odd case.

iniw:
What? That's _not_ a breaking change. Interactive shells will still execute what we place in `shellInit` - it's non-interactive shells that _don't_ execute `interactiveShellInit`.

bestlem:
That is the breaking change.

If I use a non interactive shell - e.g. to build something or to run something I do not want any homebrew in it.

As I said above I only want homebre in interactive shells.

iniw:
That's fair. I don't think there'll be something that works for both... I'm fine with modifying my own config to do the right thing instead of changing it upstream then :)

PR #1693, opened 2026-02-11
Make github-runners state, log, and work dirs use attr name

michaelvanstraten:
If you are running multiple runners with the same name (e.g. the hostname) in different repositories it will silently crash one of them because the are sharing the same state directories.

This Patch fixes this issue by using the GitHub runners attribute key instead of the `cfg.name` for seeding the log, state and work directories.

This is a breaking change (somewhat).

### DIFF
--- a/modules/services/github-runner/options.nix
+++ b/modules/services/github-runner/options.nix
@@ -128,7 +128,7 @@ in
         name = mkOption {
           type = types.nullOr types.str;
           description = ''
-            Name of the runner to configure. If null, defaults to the hostname.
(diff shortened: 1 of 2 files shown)


michaelvanstraten:
Tested it on a real machine, works. I don't know how we should deal with the breaking change, that's why it's a Draft PR.

My existing runner did not migrate to the new directory without issues, I needed to update the token.

michaelvanstraten:
@Enzime should I bump the `system.stateVersion` for this?

PR #1696, opened 2026-02-13
modules/nix: drop assertion for auto-optimise-store

GaetanLepage:
This assertion is not (anymore?) in the [NixOS module](https://github.com/NixOS/nixpkgs/blob/f240e332610927da7e2b378ee51b2849e3317cc5/nixos/modules/config/nix.nix).

Also, the lix patch was ported to the official Nix implementation (see https://github.com/NixOS/nix/pull/14676).
Hence, I believe that we should drop this assertion.


zowoq:
https://github.com/nix-darwin/nix-darwin/pull/1551

GaetanLepage:
> #1551

Thanks for the heads up. I'll close this PR then.

PR #1701, opened 2026-02-17
services/openssh: manage host keys

Enzime:
TIL `sshd-keygen-wrapper` is responsible for generating host keys on every time the SSH server starts and is open source :p

https://github.com/apple-oss-distributions/OpenSSH/blob/f386b2e948280f6ecac875329c0b56020821d558/sshd-keygen-wrapper/HostKeyManager.swift#L20

### DIFF
--- a/modules/services/openssh.nix
+++ b/modules/services/openssh.nix
@@ -1,7 +1,72 @@
-{ config, lib, ... }:
+{ config, lib, pkgs, ... }:
 
 let
   cfg = config.services.openssh;
+
+  hostKeyOpts = {
+    options = {
+      type = lib.mkOption {
+        type = lib.types.enum [ "dsa" "ecdsa" "ed25519" "rsa" ];
+        description = ''
+          Key type passed to `ssh-keygen -t`.
+        '';
+      };
+
+      path = lib.mkOption {
+        type = lib.types.str;
+        description = ''
+          Path to the private key file.
+        '';
+      };
+
+      bits = lib.mkOption {
+        type = lib.types.nullOr lib.types.int;
+        default = null;
+        description = ''
+          Key size in bits. If `null`, `ssh-keygen` uses the default
+          for the given key type (RSA=3072, ECDSA=256, ED25519=fixed).
+        '';
+      };
+
+      comment = lib.mkOption {
+        type = lib.types.str;
+        default = "";
+        description = ''
+          Comment for the key, passed to `ssh-keygen -C`.
+
+          Defaults to an empty string to match Apple's built-in host key
+          generation and avoid leaking the hostname.
+        '';
+      };
+    };
+  };
+
+  hostKeysConfig = lib.concatMapStringsSep "\n"
+    (k: "HostKey ${k.path}")
+    cfg.hostKeys;
+
+  keygenScript = lib.concatMapStrings (k:
+    let
+      escapedPath = lib.escapeShellArg k.path;
+    in ''
+      if ! [[ -s ${escapedPath} ]]; then
+        if ! [[ -L ${escapedPath} ]]; then
+          rm -f ${escapedPath}
+        fi
+
+        keygenArgs=(
+          -t ${lib.escapeShellArg k.type}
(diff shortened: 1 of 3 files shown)


modules/services/openssh.nix
```diff
@@ -1,7 +1,78 @@
...
+        '';
+      };
+
+      openSSHFormat = lib.mkOption {
+        type = lib.types.bool;
+        default = false;
+        description = ''
+          Whether to use the OpenSSH format (`-o` flag) when generating the key.
+        '';
+      };
```
Samasaur1: This flag doesn't actually do anything anymore, as OpenSSH defaults to its own format and has for a while now (I didn't check all old macOS versions, but it's at least unnecessary by Big Sur).

You _can_ specify the legacy PEM format, but it's with the `-m` option instead (and I don't know why you'd want to do so)

modules/services/openssh.nix
```diff
@@ -1,7 +1,78 @@
...
+    cfg.hostKeys;
+
+  keygenScript = lib.concatMapStrings (k:
+    let
+      escapedPath = lib.escapeShellArg k.path;
+      args = lib.concatStringsSep " " (
+        [ "-t ${lib.escapeShellArg k.type}" ]
+        ++ lib.optionals (k.bits != null) [ "-b ${toString k.bits}" ]
+        ++ [ "-C ${lib.escapeShellArg k.comment}" ]
+        ++ lib.optionals k.openSSHFormat [ "-o" ]
```
Samasaur1: likewise, drop this

modules/services/openssh.nix
```diff
@@ -1,7 +1,78 @@
...
+        ++ [ "-f ${escapedPath}" "-N ''" ]
+      );
+    in ''
+      if ! [ -s ${escapedPath} ]; then
+        if ! [ -h ${escapedPath} ]; then
```
Samasaur1: minor nit but `test(1)` says not to rely on the `-h` flag and to use `-L` instead, which is also what we use elsewhere in the codebase

modules/services/openssh.nix
```diff
@@ -24,6 +95,34 @@ in
...
+        type = lib.types.bool;
+        default = false;
+        description = ''
+          Whether to generate missing SSH host keys during activation.
+
+          Defaults to `false` because macOS automatically generates the
+          standard host keys. Enable this if you use custom key paths or need
+          to regenerate deleted keys.
+        '';
+      };
```
Samasaur1: Is there a way to make this idempotent? Right now if you enable it, activate, and then reactivate the same generation it will blow away your keys.

If not, it might be worth noting this in the description

modules/services/openssh.nix
Enzime: Dropped 👍

Enzime: I just took this option from the NixOS OpenSSH module

modules/services/openssh.nix
Enzime: Fixed 👍

modules/services/openssh.nix
Enzime: I believe this is already idempotent because the keygen script only regenerates the secrets if there's no file present or it's empty

I decided to delete this option as it should probably always be enabled because it shouldn't mangle your existing host keys

modules/services/openssh.nix
```diff
@@ -41,6 +122,12 @@ in
       fi
     '');
 
+    environment.etc."ssh/sshd_config.d/099-host-keys.conf" = lib.mkIf (cfg.hostKeys != []) {
+      text = hostKeysConfig;
+    };
+
     environment.etc."ssh/sshd_config.d/100-nix-darwin.conf".text = cfg.extraConfig;
+
+    system.activationScripts.openssh.text = lib.mkIf (cfg.hostKeys != []) keygenScript;
```
Samasaur1, line 131: this doesn't even need to be `mkIf` because the script is the empty string when `cfg.hostKeys` is empty. purely a stylistic thing though, so feel free to ignore this.

PR #1713, opened 2026-02-25
README: Flakes are recommended (not just for beginners)

msgilligan:
Change the README to recommend flakes for everyone, not just for  beginners.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,10 +19,10 @@ The installer you use doesn't affect which Nix interpreter your system will use 
 
 ## Getting started
 
-Despite being an experimental feature in Nix currently, nix-darwin recommends that beginners use flakes to manage their nix-darwin configurations.
+Despite being an experimental feature in Nix currently, nix-darwin recommends using flakes to manage nix-darwin configurations.
 
 <details>
-<summary>Flakes (Recommended for beginners)</summary>
+<summary>Flakes (Recommended)</summary>
 
 ### Step 1. Creating `flake.nix`
 


Samasaur1:
I'm not sure I agree with this change. Although I personally use flakes, I think the idea of the recommendation being 'for beginners" is that once you are no longer a beginner, you're experienced enough to know the options and make an informed decision. I don't feel incredibly strongly about this so I'm open to being convinced otherwise by you or my fellow maintainers

msgilligan:
(Copying my comment from the "nix-darwin Development" Matrix list...)

Of course, once you're experienced enough to know the options, you're also experienced enough to ignore the recommendation, so it's a not a big change.

I have two arguments for making the change:

1. Even for non-beginners, it's nice to have a recommended/default path when you are installing a new tool. (e.g. you're experienced with Nix/NixOS, but new to Nix Darwin)
2. The new nix-installer configures Nix with Flakes by default.

I don't feel incredibly strongly about this either, but as I'm doing a fresh nix-darwin setup (and this is only my second time) I'm looking at the README and ways to make it easier for people to get started. Given that most people use Flakes these days, a nudge in the direction of Flakes seems helpful.

xokdvium:
> The new nix-installer configures Nix with Flakes by default.

1) The experimental rust-based installer is still in development.

2) The fact that it enables some XP features is unintentional and is certainly just a result of not being careful enough to clean out detsys-specific logic. This is just a bug in the (still experimental) installer and in no way intentional.

msgilligan:
BTW, it looks like flakes will probably remain enabled by default by nixos/nix-installer:

https://github.com/NixOS/nix-installer/issues/156#issuecomment-4076516884

Mic92:
> > The new nix-installer configures Nix with Flakes by default.
> 
> 1. The experimental rust-based installer is still in development.

I would consider the feature set production stable at this point. In fact it has bugfixes that both the lix-installer / detsys installer don't have. And it covers a lot of fixes on macOS that the old installer doesn't have.

> 2. The fact that it enables some XP features is unintentional and is certainly just a result of not being careful enough to clean out detsys-specific logic. This is just a bug in the (still experimental) installer and in no way intentional.

Working on it. Will be merged hour.

PR #1720, opened 2026-02-27
modules/networking/wifi: New Module

### LINKED ISSUE #1716: Feature Request: Options for Wi-Fi settings 'Ask to join networks' and 'Ask to join hotspots'
Image

These notifications drive me up the wall and I would like to be able to configure them with Nix so that I can set them once and forget about them forever. I did some digging and have discovered that there is a preferences plist for these settings, however it is not exposed by the `defaults` database.

It's technically possible to get the existing `CustomSystemPreferences` key to do this like so:
```nix
CustomSystemPreferences = {
  "/Library/Preferences/SystemConfiguration/com.apple.airport.preferences" = {
    AutoHotspotMode = "Never";
    JoinModeFallback = [ "DoNothing" ];
  };
};
```
However, this method is extraordinarily clunky and difficult to discover. An ideal solution would be to have high-level nix-darwin options which abstract over these low-level details.

347Online:
Configures Wi-Fi settings (fixes #1716)

Options for behavior when networks or hotspots are detected

- [x] Document macOS version after which the defined settings are available, as discussed here: https://github.com/nix-darwin/nix-darwin/issues/1716#issuecomment-3969426376

### DIFF
--- a/modules/networking/wifi.nix
+++ b/modules/networking/wifi.nix
@@ -1,0 +1,81 @@
+{ config, lib, ... }:
+let
+  cfg = config.networking.wifi;
+in
+{
+  meta.maintainers = [ lib.maintainers._347Online or "347Online" ];
+
+  options = {
+    networking.wifi = {
+      AskToJoinNetworks = lib.mkOption {
+        type = lib.types.nullOr (
+          lib.types.enum [
+            "Off"
+            "Notify"
+            "Ask"
+          ]
+        );
+        default = null;
+        example = "Notify";
+        description = ''
+          Sets the behavior for when no known Wi-Fi network is available:
+
+              Off
+                  Known networks will be joined automatically.
+                  If no known networks are available, you will have to manually select a network.
+
+              Notify (macOS default)
+                  Known networks will be joined automatically.
+                  If no known networks are available, you will be notified of available networks.
+
+              Ask (macOS Tahoe and later)
+                  Known networks will be joined automatically.
+                  If no known networks are available, you will be asked before joining a new network.
+
+          This is equivalent to the setting in 'System Settings > Wi-Fi'
+        '';
+        apply =
+          value:
+          if value == "Off" then
+            "DoNothing"
+          else if value == "Ask" then
+            "Prompt"
+          else
+            value;
+      };
+      AskToJoinHotspots = lib.mkOption {
+        type = lib.types.nullOr (
+          lib.types.enum [
+            "Never"
+            "AskToJoin"
+            "Automatic"
+          ]
+        );
+        default = null;
+        example = "AskToJoin";
+        description = ''
+          Sets the behavior for when nearby personal hotspots are detected and no Wi-Fi network is available:
+
+              Never
+                  Do nothing
+
+              AskToJoin (macOS default)
+                  Show a notification prompting the user to join personal hotspot if it is available
+
+              Automatic (macOS Tahoe and later)
+                  Connect to personal hotspot automatically
+
+          This is equivalent to the setting in 'System Settings > Wi-Fi'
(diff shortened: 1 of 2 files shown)


modules/networking/wifi.nix
```diff
@@ -0,0 +1,79 @@
...
+              Automatic
+                  Connect to personal hotspot automatically
+
+          This is equivalent to the setting in 'System Settings > Wi-Fi'
+        '';
+      };
+    };
+  };
+
+  config = {
```
347Online, line 74: Not sure if this is the right approach, but this was how I solved this in my config first so I figured I would start with that.

Pros:
Takes advantage of already existing system.defaults functionality, preventing potential collisions as opposed to new activationScript code

Cons:
Technically this prevents the user from setting the option this way themself which wouldn't be so bad, except the presented error is not very friendly, e.g.
```
error: The option `system.defaults.CustomSystemPreferences."/Library/Preferences/SystemConfiguration/com.apple.airport.preferences".AutoHotspotMode` is defined both null and not null, in `/nix/store/q5678p8kyjaqv941znfa56hyqv6hsjk2-source/modules/networking/wifi.nix' and `/nix/store/657hkfiq6gf8s7z0br7l4wmil42xy3md-source/modules/darwin/prefs.nix'.
```

347Online:
I've added a callout to the description of the options which specifies the minimum major macOS version required to use the third options as discussed in #1716

modules/networking/wifi.nix
```diff
@@ -0,0 +1,81 @@
+{ config, lib, ... }:
```
Samasaur1, line 1: nitpick: please add a blank line between the first line and the `let-in` block

modules/networking/wifi.nix
Samasaur1: I think you should add a new `system.defaults.airport` module (look at `system.defaults.smb` or `system.defaults.SoftwareUpdate` as references), and then the `config` block in this module should set the options in the new defaults module. You'll also have to add the new defaults module to `modules/system/defaults-write.nix`.

modules/networking/wifi.nix
```diff
@@ -0,0 +1,81 @@
...
+        '';
+      };
+    };
+  };
+
+  config = {
+    system.defaults.CustomSystemPreferences."/Library/Preferences/SystemConfiguration/com.apple.airport.preferences" =
+      {
+        AutoHotspotMode = cfg.AskToJoinHotspots;
+        JoinModeFallback = [ cfg.AskToJoinNetworks ];
```
Samasaur1, line 78: these could both be guarded by `mkIf`, right? Any reason not to do so?

Samasaur1:
I like the public interface of the new networking module you've added. I'd make some changes about how you implemented this (left comments inline), but in general I feel good about this!

Samasaur1 (changes requested):

modules/networking/wifi.nix
347Online: I don't see why not but what should the condition be then? AFAIK we can't really switch on the macOS version (though I could be mistaken). I couldn't think of a different appropriate predicate 🤔

347Online:
@Samasaur1 I love your suggestions, thanks for the feedback. I will make those changes soon

modules/networking/wifi.nix
Samasaur1: Ah, I mean something like `AutoHotspotMode = lib.mkIf (cfg.AskToJoinHotspots != null) cfg.AskToJoinHotspots`, which should avoid problems where the new options you're defining are not set but someone has set the equivalent defaults using CustomSystemPreferences

PR #1724, opened 2026-03-14
services/openssh: only manage host keys when module is enabled

ryanccn:
I don't think it makes sense for host keys to be generated/managed when the OpenSSH module is not enabled; because `services.openssh.hostKeys` does not default to `[]` the module currently generates a set of host keys by default for anyone who isn't using the module at all.

### DIFF
--- a/modules/services/openssh.nix
+++ b/modules/services/openssh.nix
@@ -122,12 +122,12 @@ in
       fi
     '');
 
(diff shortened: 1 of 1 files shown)


ryanccn:
Alternatively the default could be changed to `[]`, which would solve the problem as well.

NateyB:
Please merge this. Automatically generating host keys when the ssh server has not been enabled looks like malware installing a backdoor.

Issue #737, opened 2023-07-12
error: file 'darwin' was not found in the Nix search path (add it using $NIX_PATH or -I)


dbarrosop:
This started happening all of a sudden.

```shell
$ darwin-rebuild switch
building the system configuration...
error: file 'darwin' was not found in the Nix search path (add it using $NIX_PATH or -I)

       at «none»:0: (source not available)
```

looking at NIX_PATH:

```shell
$ echo $NIX_PATH
darwin-config=/Users/dbarroso/.nixpkgs/darwin-configuration.nix:/nix/var/nix/profiles/per-user/root/channels:nixpkgs-overlays=/Users/dbarroso/.nixpkgs/overlays/:nixpkgs-overlays=/Users/dbarroso/.nixpkgs/overlays/
```

And the configuration file `/Users/dbarroso/.nixpkgs/darwin-configuration.nix` exists:

```shell
$ ls /Users/dbarroso/.nixpkgs/darwin-configuration.nix
/Users/dbarroso/.nixpkgs/darwin-configuration.nix
```

`-I` doesn't help:

```
$ darwin-rebuild switch -I darwin-config=/Users/dbarroso/.nixpkgs/darwin-configuration.nix
building the system configuration...
error: file 'darwin' was not found in the Nix search path (add it using $NIX_PATH or -I)

       at «none»:0: (source not available)
```

Any idea on why is this happening or how to fix it?

emilazy:
Hmm, this is probably because of https://github.com/LnL7/nix-darwin/pull/723 or maybe https://github.com/LnL7/nix-darwin/pull/687, but I'm not sure how. Can you run `nix-channel --list` and `sudo nix-channel --list`?

dbarrosop:
Of course:

``` shell
$ nix-channel --list
darwin https://github.com/LnL7/nix-darwin/archive/master.tar.gz
home-manager https://github.com/nix-community/home-manager/archive/master.tar.gz
nixpkgs https://nixos.org/channels/nixpkgs-unstable

$ sudo nix-channel --list
warning: $HOME ('/Users/dbarroso') is not owned by you, falling back to the one defined in the 'passwd' file ('/var/root')
nixpkgs https://nixos.org/channels/nixpkgs-unstable
```

dbarrosop:
Adding the channels to the root user fixed the issue, but leads to another one:

```
$ darwin-rebuild switch
building the system configuration...
[...]

PR #1726, opened 2026-03-14
README: update with official nix installer

siphc:
For users opting for the channels approach, the official Nix installer is easier to use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ nix-darwin is built up around [Nixpkgs](https://github.com/NixOS/nixpkgs), quite
 
 The only prerequisite is a Nix implementation; both Nix and Lix are supported.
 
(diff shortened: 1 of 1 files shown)


Samasaur1:
As the existing README notes, you can use the channels approach with the Lix installer without any extra work. Additionally, the concerns with the existing installer (primarily uninstalling, as noted in the README, but also some other sharp edges) have not been addressed upstream AFAIK, so removing the warning seems misleading.

Long-term, we'd like to start recommending [the new official installer](https://github.com/NixOS/nix-installer) alongside the Lix installer, but last I heard, there was a desire for some more testing before we started recommending it.

siphc:
@Samasaur1 there is one bit of extra work: the multi-user installation adds the unstable branch to the `nixpkgs` nix-channel, whereas the Lix installer does nothing. There might be some other differences too, but I tried doing a clean install with Lix and it just didn't work (it was giving me something like #737 ). Maybe this should be addressed in the README instead?

Either way, I agree with the uninstaller part.

Issue #1228, opened 2024-12-17
Please remove or rename `nix.conf`, then can't use `nix` command without it


spikespaz:
```
The following files have unrecognized content and would be overwritten:

  /etc/nix/nix.conf

Please check there is nothing critical in these files, rename them by adding .before-nix-darwin to the end, and then try again.

$ sudo mv /etc/nix/nix.conf /etc/nix/nix.conf.before-nix-darwin

$ nix run nix-darwin -- switch --flake ~/Work/flakey-darwin
error: experimental Nix feature 'nix-command' is disabled; add '--extra-experimental-features nix-command' to enable it
```

hoyhoy:
I just hit this on macOS 26.3.1 following the exact install instructions on the Readme.

<img width="550" height="382" alt="Image" src="https://github.com/user-attachments/assets/c72cb358-13f0-468b-afcc-75ac0ed671b2" />

candsastle:
Yeah I notice a bit of a chicken-and-egg situation here. I'm personally trying to use `nh darwin switch . -H hostname` and I'm running across this issue.

CielSoleil:
Happening here too

- Lix latest (installed through the curl command the have on their site)
- macOS Tahoe (Latest)

Followed the README to the letter (new installation)

Edit: The command of the OP is not working.

This worked

```sh
cd /etc/nix
sudo mv nix.conf nix.conf.before-nix-darwin
sudo mv nix.custom.conf nix.custom.conf.before-nix-darwin
sudo nix run --extra-experimental-features nix-command --extra-experimental-features flakes nix-darwin/master#darwin-rebuild -- switch
```

Samasaur1:
Apologies, I have a PR up to fix this but we're running into unrelated test failures that are blocking it. The code excerpts pasted here are the correct way to solve this problem in the short term. (If this is from a fresh install/you don't care about what's in those files, you can even remove `nix.conf` and `nix.custom.conf` instead of backing them up)

PR #1730, opened 2026-03-21
etc: support lix-installer 3.95.0

### LINKED ISSUE #1729: Unable to install with current lix installer
The current lix installer (as of 2026-03-20) appears to generate a nix.conf that isn't in the expected list for nix-darwin. Contents and sha256 attached.

```nix
# Generated by https://install.lix.systems/.
# See `/nix/lix-installer --version` for the version details.

extra-experimental-features = nix-command flakes
always-allow-substitutes = true
extra-trusted-substituters = https://cache.lix.systems
extra-trusted-public-keys = cache.lix.systems:aBnZUw8zA7H35Cz2RyKFVs3H4PlGTLawyY5KRbvJR8o=
max-jobs = auto

!include nix.custom.conf
```

```
daa9d824601c088f52de3da176e14328ba41e7be3eb450fdf8b93c86f236f722  /etc/nix/nix.conf
```

Samasaur1:
Closes #1729.
Should fix the recent problems in #1228.

FWIW, the with-flakes one was directly copied from the output of running the lix-installer in a VM, but when I did so again and chose no flakes it seemed no produce the same file, so I read through the current source code and generated the non-flake one by hand.


Samasaur1:
test failure is unrelated (github-runner is broken on stable)

PR #1731, opened 2026-03-22
etc: support Lix 2.95.1

martin-woo:
Relates to #1729

#1730 solves the first error, while this PR solves the second error. 

After installing lix 2.95.1 and manually renaming /etc/nix/nix.conf to skip the first error, running flakes with nix-darwin errors out:

```
error: custom settings in `/etc/nix/nix.custom.conf`, aborting activation
You will need to migrate these to nix-darwin `nix.*` settings if you
wish to keep them. Check the manual for the appropriate settings and
add them to your system configuration, then run:

  $ sudo mv /etc/nix/nix.custom.conf{,.before-nix-darwin}

and activate your system again.
```

I have added the sha256 for the `nix.custom.conf` file created by lix in the change.

Content is:
```
# Written by https://install.lix.systems/.
# The contents below are based on options specified at installation time.


```

### DIFF
--- a/doc/known-files/a6dee4985bf207d3bec6a3cee28aefb33e60f5d0a91d8c20bbd71b9dadb2e601
+++ b/doc/known-files/a6dee4985bf207d3bec6a3cee28aefb33e60f5d0a91d8c20bbd71b9dadb2e601
@@ -1,0 +1,3 @@
+# Written by https://install.lix.systems/.
+# The contents below are based on options specified at installation time.
+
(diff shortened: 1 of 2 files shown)


Samasaur1 (changes requested):
Can you add the generated file to `doc/known-files`?

martin-woo (commented):
Added to the known-files. Note the generated file has 2 empty lines at the end of the file, which doesn't show up in the Github UI.

Samasaur1 (approved):
earlier test failures have finally been cleared up

PR #1738, opened 2026-03-26
services/wg-quick: use command instead of ProgramArguments

apeschar:
Using `command' automatically adds wait4path, which is needed to ensure that the service starts successfully on boot.


alyraffauf:
Anything preventing this from being merged? It's still an issue on latest macos + nix-darwin.

I see a CI failure but logs have expired.

I backported a similar fix to my flake and it's working very well: https://github.com/alyraffauf/hoenn/blob/a1cc82cc28d78da3a1e76802db6c7ceeda2b0ee0/nix/modules/wireguard-hoenn/default.nix#L56

PR #1739, opened 2026-03-29
prometheus-node-exporter: separate extra flags with a space

booxter:


modules/services/monitoring/prometheus-node-exporter.nix
```diff
@@ -104,7 +104,7 @@ in {
         ]
         ++ (map (collector: "--collector.${collector}") cfg.enabledCollectors)
         ++ (map (collector: "--no-collector.${collector}") cfg.disabledCollectors)
-      ) + escapeShellArgs cfg.extraFlags;
+      ) + " " + escapeShellArgs cfg.extraFlags;
```
Samasaur1:
```suggestion
        ++ (map (collector: "--no-collector.${collector}") cfg.disabledCollectors)
        ++ (map escapeShellArg cfg.extraFlags)
      );
```
(will also need inheriting `escapeShellArg`)

Samasaur1 (changes requested):
just a style change (but given how our merge queue is broken at the moment I guess there's no rush)

Samasaur1 (approved):
LGTM, now just a matter of hoping the merge queue starts working...

PR #1742, opened 2026-04-01
nixpkgs.config: Add allowUnfreePackages option

dwt:
Make sure it merges correctly and verify with a test.

This was added in nixpkgs some time ago and is quite a nice addition as it merges additively (in contrast to allowUnfreePredicate) thus allowing to localize allowUnfreePackage specifications to all the locations in your local config where unfree packages are added.

A test is also provided which can be executed with

```sh
nix-build release.nix -A tests.nixpkgs-config-allow-unfree
```

Please let me know if there is anything else you need to get this merged.

### DIFF
--- a/modules/nix/nixpkgs.nix
+++ b/modules/nix/nixpkgs.nix
@@ -17,10 +17,13 @@ let
   mergeConfig =
     lhs_: rhs_:
     let
-      lhs = optCall lhs_ { inherit pkgs; };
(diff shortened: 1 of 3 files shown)


dwt:
As far as I can see that test is not broken by me? Am I missing something?

Samasaur1 (approved):
The test is almost definitely not broken by you. I'll merge this once we get the tests passing again.

dwt:
Thanks @Samasaur1!

PR #1743, opened 2026-04-02
keyboard: assert source keys are remapped at most once

c31io:
Add a new system.keyboard.swapCapsLockAndEscape option to swap the Caps Lock and Escape keys, complementing the existing remapCapsLockToEscape option which only remaps one direction.
The new option would be useful for multi-linguists.

### DIFF
--- a/modules/system/keyboard.nix
+++ b/modules/system/keyboard.nix
@@ -26,6 +26,12 @@ in
       description = "Whether to remap the Caps Lock key to Escape.";
     };
 
+    system.keyboard.swapCapsLockAndEscape = mkOption {
+      type = types.bool;
+      default = false;
+      description = "Whether to swap the Caps Lock key and Escape key.";
+    };
+
     system.keyboard.nonUS.remapTilde = mkOption {
       type = types.bool;
       default = false;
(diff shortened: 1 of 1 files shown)


skykanin (approved):

modules/system/keyboard.nix
```diff
@@ -26,6 +26,12 @@ in
       description = "Whether to remap the Caps Lock key to Escape.";
     };
 
+    system.keyboard.swapCapsLockAndEscape = mkOption {
+      type = types.bool;
+      default = false;
+      description = "Whether to swap the Caps Lock key and Escape key.";
+    };
```
skykanin: I think we should have some assertions about which options are enabled for example enabling `remapCapsLockToControl` and `swapCapsLockAndEscape` at the same time doesn't make any sense. Same goes for `swapCapsLockAndEscape` and `remapCapsLockToControl` etc.

skykanin (changes requested):

c31io:
This PR originally added `system.keyboard.swapCapsLockAndEscape`, but the feature has since been merged to master (together with `swapRightCommandAndRightOption`), so the branch was rebased onto current master and the original commit dropped.

What's left in this PR is the assertion discussed in the review comments:

- An assertion on `system.keyboard.userKeyMapping` guaranteeing every source key is remapped at most once (no duplicate `HIDKeyboardModifierMappingSrc` values). This prevents contradictory combos such as `remapCapsLockToControl` + `remapCapsLockToEscape`, or `remapCapsLockToControl` + `swapCapsLockAndEscape`, while still allowing multiple keys to produce the same result (e.g. several keys mapped to Escape).
- `tests/system-keyboard-mapping.nix` was adapted so it no longer enables conflicting caps-lock mappings together; `swapCapsLockAndEscape` coverage moved to a new `tests/system-keyboard-mapping-swap.nix`.

Happy to adjust either the assertion or the test split as maintainers see fit.

skykanin (approved):

Issue #746, opened 2023-07-24
nix-gc should imitate NixOS API


srid:
... so I can avoid a conditional here:

```nix
{
  nix = {
    settings.auto-optimise-store = true;
    gc = {
      automatic = true;
      options = "--delete-older-than 30d";
    } // (if pkgs.stdenv.isLinux then {
      dates = "weekly";
    } else {
      interval = { Hour = 0; Minute = 0; };
    });
  };
}
```

emilazy:
This would be nice but it would be difficult to do; the format is [rather complex](https://www.freedesktop.org/software/systemd/man/systemd.time.html), especially to implement in Nix code. I believe there are at least some differences between the systemd and launchd handling of these triggers, too (e.g. I believe that systemd will reschedule missed timers if the system was powered off, but launchd will only do it if the system was sleeping).

I would be happy to review a PR that has a go at implementing this, but a full implementation seems potentially intractable (there's an apparent zoneinfo dependency, for one thing) and I'm not totally sure how I feel about the supposedly-compatible-but-not-actually state of carving out a simpler subset to implement. Maybe we could inject an external command into the build of the `.plist` files so the conversion could be done in a real programming language.

PR #1744, opened 2026-04-02
programs/nh: new module

rajanmaghera:
Adds a new module `programs.nh` based on the NixOS equivalent. `programs.nh.clean` can be used instead of `nix.gc` as a periodic garbage collection sweep with more options.

### DIFF
--- a/modules/programs/nh.nix
+++ b/modules/programs/nh.nix
@@ -1,0 +1,141 @@
+# Based off: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/nh.nix
+# When making changes please try to keep it in sync.
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  cfg = config.programs.nh;
+  launchdTypes = import ../launchd/types.nix { inherit config lib; };
+in
+
+{
+
+  meta.maintainers = [
+    lib.maintainers.rajanmaghera or "rajanmaghera"
+  ];
+
+  imports = [
+    (lib.mkRemovedOptionModule [
+      "programs"
+      "nh"
+      "clean"
+      "dates"
+    ] "Use `programs.nh.clean.interval` instead.")
+  ];
+
+  ###### interface
+
+  options = {
+
+    programs.nh = {
+      enable = lib.mkEnableOption "nh, yet another Nix CLI helper";
+
+      package = lib.mkPackageOption pkgs "nh" { };
+
+      flake = lib.mkOption {
+        type = lib.types.nullOr lib.types.str;
+        default = null;
+        description = ''
+          The string that will be used for the `NH_FLAKE` environment variable.
+
+          `NH_FLAKE` is used by nh as the default flake for performing actions, such as
+          `nh os switch`. This behaviour can be overriden per-command with environment
+          variables that will take priority.
+
+          - `NH_OS_FLAKE`: will take priority for `nh os` commands.
+          - `NH_HOME_FLAKE`: will take priority for `nh home` commands.
+          - `NH_DARWIN_FLAKE`: will take priority for `nh darwin` commands.
+
+          The formerly valid `FLAKE` is now deprecated by nh, and will cause hard errors
+          in future releases if `NH_FLAKE` is not set.
+        '';
+      };
+
+      clean = {
+        enable = lib.mkEnableOption "periodic garbage collection with nh clean all";
+
+        interval = lib.mkOption {
+          type = launchdTypes.StartCalendarInterval;
+          default = [
+            {
+              Weekday = 7;
+              Hour = 3;
+              Minute = 15;
+            }
+          ];
+          description = ''
+            The calendar interval at which the garbage collector will run.
+            See the {option}`serviceConfig.StartCalendarInterval` option of
+            the {option}`launchd` module for more info.
+          '';
+        };
+
+        extraArgs = lib.mkOption {
+          type = lib.types.singleLineStr;
+          default = "";
+          example = "--keep 5 --keep-since 3d";
+          description = ''
+            Options given to nh clean when the service is run automatically.
+
+            See `nh clean all --help` for more information.
+          '';
+        };
+      };
+    };
+  };
+
+  ###### implementation
+
+  config = {
+
+    warnings =
+      if (!(cfg.clean.enable -> !config.nix.gc.automatic)) then
+        [
+          "programs.nh.clean.enable and nix.gc.automatic are both enabled. Please use one or the other to avoid conflict."
+        ]
+      else
+        [ ];
+
+    assertions = [
+      # Not strictly required but probably a good assertion to have
+      {
+        assertion = cfg.clean.enable -> cfg.enable;
+        message = "programs.nh.clean.enable requires programs.nh.enable";
+      }
+
+      {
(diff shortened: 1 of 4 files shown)


dwt:
Would you consider merging all the ideas and feedback from #942? Sadly that other person didn't want to continue the work.

rajanmaghera:
@dwt I've cross-referenced both the home-manager and nixos modules to add all the those fields, as well as your `LowPriorityIO` options (I also don't know if they're effective) from [here](https://github.com/nix-darwin/nix-darwin/pull/942#issuecomment-3674354226). I left out the log fields though, not sure if we'd want an extra option for it.

modules/programs/nh.nix
```diff
@@ -0,0 +1,190 @@
...
+          {env}`NH_DARWIN_FLAKE` is used by nh as the default flake for performing
+          {command}`nh darwin` actions, such as {command}`nh darwin switch`.
+          Setting this will take priority over the `flake` option.
+        '';
+      };
+
+      clean = {
+        enable = lib.mkEnableOption "periodic garbage collection with nh clean all";
+
+        interval = lib.mkOption {
```
dwt, line 61: NixOS uses [dates instead of interval](https://search.nixos.org/options?channel=unstable&query=programs.nh&show=programs.nh.clean.dates) as they use a systemd time definition, I'm pretty sure we cannot reuse it in it's entirety.

Is that the reason why you did go with `interval`? Because the supported type is different anyway?

dwt:
@rajanmaghera Nice! I did add the log because I initially had some trouble with the service where it didn't work - though I never found out why, I guess it was because of some bugs in lix perhaps? Anyway, that's why I added the logs to debug that.

Since logging on darwin is so hard, I would not fault this module for not writing logs - that can always be a followup pull request by someone who knows how to do that.

modules/programs/nh.nix
```diff
@@ -0,0 +1,190 @@
...
+        type = with lib.types; nullOr (either singleLineStr path);
+        default = null;
+        description = ''
+          The string that will be used for the {env}`NH_DARWIN_FLAKE` environment variable.
+
+          {env}`NH_DARWIN_FLAKE` is used by nh as the default flake for performing
+          {command}`nh darwin` actions, such as {command}`nh darwin switch`.
+          Setting this will take priority over the `flake` option.
+        '';
+      };
```
dwt: I don't get what these are for, isn't the only reason for nh.clean to .... clean? I am probably missing something here.

dwt (commented):
I really like this, and would like to help get this merged. We will probably need to attract the attention of either @emilazy, @Enzime or @Samasaur1

modules/programs/nh.nix
rajanmaghera: This was discussed here #746 and this is what `nix.gc` does. Home-manager uses `dates` on darwin, but they map some specific systemd keywords to intervals via `mkCalendarInterval` [here](https://github.com/nix-community/home-manager/blob/9cc761169a1bbf1d9787cdbe9abf07f4bae213a1/modules/lib/darwin.nix#L12) and is used [here](https://github.com/nix-community/home-manager/blob/9cc761169a1bbf1d9787cdbe9abf07f4bae213a1/modules/programs/nh.nix#L161), which is a less than ideal solution IMHO.

modules/programs/nh.nix
rajanmaghera: The `nh` tool has some additional commands outside of cleaning and they use these env vars to define a "system" flake.

These don't affect `nh.clean` though, they're mainly here to be equivalent to the home-manager and nixos modules.

modules/programs/nh.nix
dwt: That makes quite some sense. Would it make sense to perhaps support "dates" with a very limited set of values (like home-manager seems to do) for option compatibility, and assert if the chosen option is not one of the compatible ones?

dwt: `weekly` or `monthly` seem like very sensible values here, so maybe just a very limited subset?

rajanmaghera: Do you suggest that we just import hm's logic for this?

Maybe that could be another PR doing a repo-wide sweep of it?

dwt: A repo wide sweet in a separate pull request is probably a good idea. HomeManager basically just supports very few values too, if I see that correctly they support: `hourly`, `daily`, `weekly`, `monthly`, `semiannually`, `annually` - that is not much, and could be a great way to increase nixos and nix-darwin compatibility.

modules/programs/nh.nix
```diff
@@ -0,0 +1,190 @@
...
+    environment = lib.mkIf cfg.enable {
+      systemPackages = [ cfg.package ];
+      variables = lib.mkMerge [
+        (lib.mkIf (cfg.flake != null) (
+          let
+            packageVersion = lib.getVersion cfg.package;
+            isVersion4OrHigher = lib.versionAtLeast packageVersion "4.0.0";
+          in
+          if isVersion4OrHigher then { NH_FLAKE = cfg.flake; } else { FLAKE = cfg.flake; }
+        ))
```
eveeifyeve: No need for this as both stable and unstable are currently at version 4.3.0

modules/programs/nh.nix
eveeifyeve: We should still follow upstream nixos options to the best of the ability. [Upstream defines it as `programs.nh.flake`](https://search.nixos.org/options?channel=25.11&include_modular_service_options=1&include_nixos_options=1&query=nh&show=programs.nh.flake). I feel we should follow it.

rajanmaghera: Upstream nixpkgs has `flake`. [Upstream home-manager](https://github.com/nix-community/home-manager/blob/3c7524c68348ef79ce48308e0978611a050089b2/modules/programs/nh.nix#L21-L66) has `flake`, `darwinFlake`, `osFlake` and `homeFlake`. This commit also defines `flake` right above where this comment is made.

I don't mind reverting the home-manager changes to make it match nixpkgs upstream as close as possible.

rajanmaghera: I've updated it to match nixpkgs.

modules/programs/nh.nix
rajanmaghera: Removed and updated to match upstream nixpkgs.

Eveeifyeve (approved):

modules/programs/nh.nix
eveeifyeve: I would say do another pr to fix this, it shouldn't really block this pr from being merged.

modules/programs/nh.nix
```diff
@@ -0,0 +1,141 @@
...
+    warnings =
+      if (!(cfg.clean.enable -> !config.nix.gc.automatic)) then
+        [
+          "programs.nh.clean.enable and nix.gc.automatic are both enabled. Please use one or the other to avoid conflict."
+        ]
+      else
+        [ ];
+
+    assertions = [
+      # Not strictly required but probably a good assertion to have
```
catuhana, line 104: Nixpkgs dropped this recently. NixOS/nixpkgs#513757

dwt (approved):

skykanin (approved):
Can any committers get this PR merged? I don't think we should block on the comments that have been made here. Those improvements can be made in a follow-up PR

PR #1745, opened 2026-04-06
homebrew: add onActivation.extraEnv option

mmenanno:
## Summary

Adds a generic `homebrew.onActivation.extraEnv` option — an attrset of environment variables prepended to the `brew bundle` invocation during `darwin-rebuild switch`, alongside the existing `HOMEBREW_NO_AUTO_UPDATE=1`.

```nix
homebrew.onActivation.extraEnv = {
  HOMEBREW_NO_ENV_HINTS = "1";
  HOMEBREW_NO_ANALYTICS = "1";
  HOMEBREW_NO_UPDATE_REPORT_NEW = "1";
};
```

## Motivation

During `darwin-rebuild switch`, `brew bundle` produces noisy output (analytics notices, env hints, new formulae reports). Shell environment variables set in `.zshrc` aren't inherited by the activation script (it runs under sudo), and there was no module-level way to suppress them.

## Design

Follows the approach suggested in [#1745 (comment)](https://github.com/nix-darwin/nix-darwin/pull/1745#issuecomment-4239842172): rather than adding one first-class boolean per `HOMEBREW_NO_*` env var, expose a generic `extraEnv` escape hatch that mirrors the existing `extraFlags` / `extraConfig` patterns and matches the `extraEnv` naming used elsewhere in nixpkgs. This covers the long tail of Homebrew env vars without churning on upstream naming changes.

No `global.extraEnv` equivalent — as noted in the same review thread, `environment.variables` already accepts arbitrary attrsets, so a global wrapper would be pure sugar. The activation variant is load-bearing because it needs module help to get past the sudo boundary.

## Implementation

- `homebrew.onActivation.extraEnv` is an `attrsOf str`, defaulting to `{}`.
- Entries are shell-escaped and prepended to `brewBundleCmd` via `mapAttrsToList`.

## Test plan

- Ran `nix-build release.nix -A tests.homebrew` — passes.
- Evaluated `homebrew.onActivation.brewBundleCmd` with `extraEnv = { HOMEBREW_NO_ENV_HINTS = "1"; HOMEBREW_NO_ANALYTICS = "1"; };` and confirmed the rendered command contains the expected `KEY=VALUE` prefixes before `brew bundle`.
- Tested locally by overriding the nix-darwin flake input to this branch and running `darwin-rebuild switch` — Homebrew output is suppressed during activation.

---

<details>
<summary>Original proposal (superseded after review)</summary>

> ## ~~Summary~~
>
> ~~Adds three new boolean options to both `onActivation` and `global` homebrew submodules, following the existing `autoUpdate` pattern:~~
>
> - ~~**`envHints`** — when `false`, sets `HOMEBREW_NO_ENV_HINTS=1` to suppress "Hide these hints..." messages~~
> - ~~**`analytics`** — when `false`, sets `HOMEBREW_NO_ANALYTICS=1` to suppress analytics notices and disable collection~~
> - ~~**`updateReportNew`** — when `false`, sets `HOMEBREW_NO_UPDATE_REPORT_NEW=1` to suppress "New Formulae/Casks" lists after auto-update~~
>
> ~~All default to `true` to preserve existing behavior (non-breaking).~~
>
> ## ~~Implementation~~
>
> ~~Mirrors the existing `autoUpdate` pattern exactly:~~
>
> - ~~**`onActivation`**: env vars prepended to `brewBundleCmd`, same as `HOMEBREW_NO_AUTO_UPDATE`~~
> - ~~**`global`**: env vars added to `homebrewEnvironmentVariables` → `environment.variables`, same as `HOMEBREW_NO_AUTO_UPDATE`~~
>
> ~~Usage:~~
>
> ```nix
> homebrew = {
>   onActivation = { envHints = false; analytics = false; updateReportNew = false; };
>   global = { envHints = false; analytics = false; updateReportNew = false; };
> };
> ```

Replaced with a single `onActivation.extraEnv` attrset per review discussion above.

</details>

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -142,6 +142,43 @@ let
           activation.
         '';
       };
+      envHints = mkOption {
+        type = types.bool;
+        default = true;
+        description = ''
+          Whether to enable Homebrew to print environment variable hints (e.g.,
+          "Hide these hints with HOMEBREW_NO_ENV_HINTS") during {command}`nix-darwin` system
+          activation.
+
+          Implementation note: when disabled, this option sets the `HOMEBREW_NO_ENV_HINTS`
+          environment variable when {command}`nix-darwin` invokes {command}`brew bundle [install]`
+          during system activation.
+        '';
+      };
+      analytics = mkOption {
+        type = types.bool;
+        default = true;
+        description = ''
+          Whether to enable Homebrew to print analytics notices and collect anonymous analytics
+          during {command}`nix-darwin` system activation.
+
+          Implementation note: when disabled, this option sets the `HOMEBREW_NO_ANALYTICS`
+          environment variable when {command}`nix-darwin` invokes {command}`brew bundle [install]`
+          during system activation.
+        '';
+      };
+      updateReportNew = mkOption {
+        type = types.bool;
+        default = true;
+        description = ''
+          Whether to enable Homebrew to print lists of new formulae and casks after auto-updating
+          during {command}`nix-darwin` system activation.
+
+          Implementation note: when disabled, this option sets the
+          `HOMEBREW_NO_UPDATE_REPORT_NEW` environment variable when {command}`nix-darwin` invokes
+          {command}`brew bundle [install]` during system activation.
+        '';
+      };
       extraFlags = mkOption {
         type = types.listOf types.str;
         default = [ ];
@@ -158,6 +195,9 @@ let
     config = {
       brewBundleCmd = concatStringsSep " " (
         optional (!config.autoUpdate) "HOMEBREW_NO_AUTO_UPDATE=1"
+        ++ optional (!config.envHints) "HOMEBREW_NO_ENV_HINTS=1"
+        ++ optional (!config.analytics) "HOMEBREW_NO_ANALYTICS=1"
+        ++ optional (!config.updateReportNew) "HOMEBREW_NO_UPDATE_REPORT_NEW=1"
         ++ [ "brew bundle --file='${brewfileFile}'" ]
         ++ optional (!config.upgrade) "--no-upgrade"
         ++ optional (config.cleanup == "uninstall") "--cleanup"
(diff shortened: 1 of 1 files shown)


malob:
Appreciate the effort to add these options. They do seem useful.

Reading through though, I started wondering about where the threshold should sit for adding a new first-class option per Homebrew env var vs. going with a generic escape hatch. Homebrew has a lot of `HOMEBREW_NO_*` vars (no-emoji, no-install-cleanup, no-install-from-api, no-colors, etc.), and if each one becomes a PR cycle (new option, new docstring, two scopes, naming bikeshed), the module's going to grow a long list of essentially-identical boolean passthroughs. That also locks the module into Homebrew's naming choices, which churns if upstream renames or deprecates an env var.

The rough mental model I land on: an option earns first-class status if it (a) has behavioral coupling with other options or module state, (b) needs non-trivial documentation beyond "sets env var X," (c) has a meaningful asymmetry between `onActivation` and `global` (different defaults or semantics), (d) has a non-trivial type, or (e) is computed from module state. `autoUpdate`, `upgrade`, `cleanup`, and `brewfile` all hit at least one of those. `envHints`, `analytics`, and `updateReportNew` are pretty clean passthroughs that don't really clear any of those bars.

An alternative that handles the same use case and a lot of others at once would be a generic `onActivation.extraEnv` attrset option. This would mirror the existing `extraFlags` / `extraConfig` escape hatches the module already has, and match the `extraEnv` naming convention used elsewhere in nixpkgs (e.g., `services.hadoop.yarn.*.extraEnv`). Roughly:

```nix
homebrew.onActivation.extraEnv = {
  HOMEBREW_NO_ENV_HINTS = "1";
  HOMEBREW_NO_ANALYTICS = "1";
  HOMEBREW_NO_UPDATE_REPORT_NEW = "1";
};
```

Prepended to `brewBundleCmd` the same way `HOMEBREW_NO_AUTO_UPDATE=1` already is. Same ergonomics as what you've written, covers the long tail, and doesn't churn when Homebrew adds or renames env vars.

One open question: I'm not sure whether a `global.extraEnv` equivalent is worth adding. Since `environment.variables` is itself an attrset the user can already write into directly, a module-scoped version would be pure sugar rather than new capability. The activation case is different, since it genuinely needs module help to get past the sudo wrapper. Curious what you think.

mmenanno:
That makes a lot of sense to me, I appreciate the zoomed out view there, and I agree also with your point about the global version being pure sugar that probably shouldn't be in the scope. I'll adjust the PR to those changes

Samasaur1 (approved):
This revised version looks good to me. @malob does this address your concerns? If so, I can merge it

malob:
Yep, this addresses what I was getting at. The `extraEnv` shape is what I had in mind, and the implementation looks clean. Appreciate you reworking it, @mmenanno.

Samasaur1:
`test-stable` is currently broken which is blocking the merge queue. See #1756

PR #1746, opened 2026-04-07
homebrew: add noEnvHints option to suppress hints

anpryl:
## Summary

- Add `homebrew.onActivation.noEnvHints` to set `HOMEBREW_NO_ENV_HINTS=1` in the `brew bundle` command during system activation
- Add `homebrew.global.noEnvHints` to set `HOMEBREW_NO_ENV_HINTS=1` via `environment.variables` for interactive shell usage

## Motivation

Homebrew prints contextual hints (e.g., "Adjust how often this is run with \$HOMEBREW_AUTO_UPDATE_SECS...") during activation. Setting `environment.variables.HOMEBREW_NO_ENV_HINTS = "1"` doesn't suppress these because the activation script runs `brew bundle` via `sudo --preserve-env=PATH`, forwarding only `PATH` to the subprocess.

The new `onActivation.noEnvHints` option follows the same pattern as `onActivation.autoUpdate`, which prepends `HOMEBREW_NO_AUTO_UPDATE=1` directly to the `brewBundleCmd`.

## Usage

```nix
{
  homebrew.onActivation.noEnvHints = true;
  homebrew.global.noEnvHints = true;
}
```

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -142,6 +142,23 @@ let
           activation.
         '';
       };
+      noEnvHints = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Whether to suppress Homebrew's contextual hints and suggestions during
+          {command}`nix-darwin` system activation. Homebrew normally prints messages like
+          "Adjust how often this is run with `$HOMEBREW_AUTO_UPDATE_SECS`..." after
+          auto-updates. Enabling this option silences those hints.
+
+          To also suppress hints when you manually invoke Homebrew commands, see
+          [](#opt-homebrew.global.noEnvHints).
+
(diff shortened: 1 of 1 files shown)


malob:
This one looks like it's covered by #1745, which proposes the same thing plus a couple of related noise-suppression env vars, so it probably makes sense to consolidate the conversation there.

One small note on naming: `noEnvHints` is negative-prefixed, whereas the existing options in the module are positive-framed even when they set `HOMEBREW_NO_*` vars under the hood (e.g., `autoUpdate = false` sets `HOMEBREW_NO_AUTO_UPDATE=1`). Positive framing is also the more common NixOS convention. #1745 follows that shape (`envHints = false`), which composes better if more of these knobs get added.

PR #1747, opened 2026-04-14
darwin-rebuild: add --target-host and --use-remote-sudo support

NoahBres:
## Summary

This branch builds on top of #1631 (--target-host support by @hgl) and adds `--use-remote-sudo` support, mirroring the pattern from `nixos-rebuild`.

- Ports `--target-host` from #1631 as-is
- Adds `--use-remote-sudo` flag: SSH as a normal user, wrap remote activation commands with `sudo` (avoids requiring root SSH login)
- Copies the system closure to the target before activating using `nix copy -v --no-require-sigs --to "ssh://$targetHost"`
- Shows path count and MiB before copying so there's feedback during what can be a slow transfer
- Allocates a TTY (`ssh -t`) for the sudo case so sudo can prompt for a password

## Usage

```sh
# Connect as root directly:
darwin-rebuild switch --flake .#hostname --target-host root@host.local

# Connect as a normal user with sudo on the remote:
darwin-rebuild switch --flake .#hostname --target-host noah@host.local --use-remote-sudo
```

## Known issues / open questions

- **Determinate Nix signature enforcement**: `nix copy --no-require-sigs --to ssh://` fails on remotes running Determinate Nix even when the connecting user is in `trusted-users`. Determinate Nix appears to enforce `require-sigs = true` unconditionally regardless of user trust level. The workaround is to set up a signing key on the build machine and add its public key to the remote's `trusted-public-keys`. Investigation ongoing.
- The `ssh://` protocol is used (matching `nixos-rebuild`'s default) rather than `ssh-ng://`.
- `--use-remote-sudo` does not yet support `--build-host` (separate machine for building vs activation).

## Test plan

- [ ] `--target-host root@host` activates on remote as root
- [ ] `--target-host user@host --use-remote-sudo` activates on remote via sudo
- [ ] Incremental deploys only copy changed paths (nix copy diff behaviour)
- [ ] Closure copy progress is visible (`-v` flag)
- [ ] Works with standard Nix (non-Determinate)
- [ ] Works with Determinate Nix (pending signing key solution)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


proski:
`--use-remote-sudo` was renamed to simply `--sudo` in nixos-rebuild. I think darwin-rebuild should be using `--sudo` as well.

PR #1764, opened 2026-05-10
defaults: use upstream plist value type

Enzime:
We shouldn't vendor the value type as it can change upstream like in https://github.com/NixOS/nixpkgs/pull/518660

This PR isn't actually necessary for using `mkPlistData` from that PR as it creates an attrset containing strings which is still compatible with the old type


emilazy:
I guess this breaks because of the escaping changes?

PR #1765, opened 2026-05-10
services-modular: add modular service interface for launchd

KiaraGrouwstra:
Introduces a [modular service](https://nixos.org/manual/nixos/stable/#modular-services) interface consisting of `system.services` (LaunchDaemons) and `users.users.<name>.services` (LaunchAgents).
This allows consuming modular services in nixpkgs via `imports = [ pkg.passthru.services.default ]`.

The surface, usable in both `system.services` or `users.users.<name>.services`:

general:
- `process.argv`: program and arguments
- `configData.<name>.{text,source}`: config files under `/etc/system-services/<prefix>/` (system) or `~user/.config/system-services/<prefix>/` (user agents)

darwin-specific:
- `processConfig.{user,group,workingDirectory,environment,standardOutput,standardError}`: translated to launchd plist keys
- `launchd` freeform escape hatch for native plist keys with no portable analogue (e.g. `StartCalendarInterval`)

For what it's worth, the number of modular services available in nixpkgs is currently limited to a handful, whereas the interface exposed for these is also a work in progress (see https://github.com/NixOS/nixpkgs/issues/428084). That said, given portability was the motivating factor in RFC 163 (which modular services were inspired by), being able to get feedback from Darwin users would go a long way in validating this feature as it matures.

While an alternate way to get system-specific service files could be from upstream projects, as noted at https://github.com/NixOS/nixpkgs/issues/428084#issuecomment-3998655081, I would consider these experiments not mutually exclusive.

Note that configuration generation, noted as lower-hanging fruit than service by @emilazy at https://github.com/NixOS/nixpkgs/issues/428084#issuecomment-3994512395, is expected to land in nixpkgs by https://github.com/NixOS/nixpkgs/pull/509595.

Disclaimers:

- I'm not on Darwin personally, so treat this PR accordingly.
- I used a coding agent in the creation of this patch.

## Testing

- [x] `nix-build release.nix -A tests.services-modular-basic`: minimal daemon plist
- [x] `nix-build release.nix -A tests.services-modular-configdata`: configData files at correct etc paths
- [x] `nix-build release.nix -A tests.services-modular-user-agent`: user agent label/path layout
- [x] `nix-build release.nix -A tests.services-modular-upstream`: full ghostunnel plist via upstream `passthru.services`
- [x] `nix-build release.nix -A tests.services-modular-upstream-eval`: eval-only schema check


modules/services-modular/extra-root.nix
```diff
@@ -0,0 +1,93 @@
+{ lib, ... }:
+let
+  inherit (lib) mkOption types;
+in
+{
+  _class = "service";
+
+  options = {
+    processConfig = mkOption {
```
emilazy, line 9: As far as I can tell, these options are only implemented in this module and do not at all map to anything when used with systemd on Linux? That makes the naming quite misleading. If the intention is to add support for systemd, there are issues with that, too; see below.

modules/services-modular/extra-root.nix
```diff
@@ -0,0 +1,93 @@
...
+      type = types.submodule {
+        options = {
+          user = mkOption {
+            type = types.nullOr types.str;
+            default = null;
+            description = ''
+              User to run the service as. Translated to launchd
+              `UserName`. Only honored for system services
+              (LaunchDaemons); for user agents an assertion is raised
+              if a value other than `null` is supplied.
```
emilazy, line 20: This seems like an example of the pitfalls of trying to express system daemons with independent lifetimes and user agents bound to a graphical login session through the same configuration layer.

modules/services-modular/extra-root.nix
```diff
@@ -0,0 +1,93 @@
...
+            '';
+          };
+          standardOutput = mkOption {
+            type = types.nullOr types.str;
+            default = null;
+            description = ''
+              Path of the file to which standard output is
+              redirected. Translated to launchd `StandardOutPath`.
+            '';
+          };
```
emilazy, line 51: I assume the intent is to translate this into systemd `StandardOutput=`. That is unlikely to work properly: launchd `StandardOutPath` and `StandardErrorPath` are created using the owner and group that the service executes with, whereas on systemd I believe they are created using the user of the systemd process.

(Additionally, IIRC systemd will multiplex a single open file when they are set to the same path, but launchd will not.)

modules/services-modular/user.nix
```diff
@@ -0,0 +1,163 @@
+{ config, lib, pkgs, ... }:
+let
+  inherit (lib) concatMapAttrs mapAttrs' mkOption types;
+
+  translate = import ./translate.nix { inherit lib; };
+
+  userRootPath = userName: "/Users/${userName}/.config/system-services";
```
emilazy, line 7: This is a violation of the XDG Base Directory specification: `$XDG_CONFIG_HOME` is not necessarily `~/.config`.

It is also not in line with [macOS platform guidelines](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW1) for native applications, which would call for `~/Library/Application Support/‹bundle identifier›`.

I think it is fine and sometimes preferable for portable code to use the XDG specification on macOS, but not to incorrectly follow the XDG specification or mix it with a directory that is not standard for either XDG (`/etc/system-services`) or the platform.

modules/services-modular/user.nix
```diff
@@ -0,0 +1,163 @@
...
+      {
+        options.services = mkOption {
+          type = types.attrsOf (servicesLibFor name).serviceSubmodule;
+          default = { };
+          description = ''
+            Modular services run as LaunchAgents for this user.
+
+            Only honored for the primary user (`system.primaryUser`);
+            other users' services are evaluated but not loaded since
+            nix-darwin only manages the primary user's launchd domain.
```
emilazy, line 93: This is not an interface that makes sense. New functionality relying on `system.primaryUser` should not be added, and if it’s under `users.users`, it needs to support arbitrary users. The general direction has been towards delegating per‐user management to Home Manager, however.

modules/services-modular/config-data.nix
```diff
@@ -0,0 +1,38 @@
+# Sets `configData.<name>.path` for each service in the tree, mirroring
+# the upstream `nixos/modules/system/service/systemd/config-data-path.nix`
+# pattern but with a configurable root prefix so the same module can be
+# reused for both system services (`/etc/system-services/...`) and user
+# services (`${HOME}/.config/system-services/...`).
```
emilazy, line 5: `/etc/system-services` is an extremely generic name that, to my knowledge, corresponds to no standard. I don’t understand why this bespoke prefix is used or how a user would be expected to know that it maps to nix-darwin services. It sounds like part of the OS.

modules/services-modular/extra-root.nix
```diff
@@ -0,0 +1,93 @@
...
+        };
+      };
+    };
+
+    launchd = mkOption {
+      default = { };
+      type = types.submodule {
+        freeformType = types.attrsOf types.unspecified;
```
emilazy, line 67: This is not the right type.

modules/services-modular/extra-root.nix
```diff
@@ -0,0 +1,93 @@
...
+      description = ''
+        Free-form launchd plist keys merged verbatim into the generated
+        plist. This is the platform escape hatch for anything that has no
+        portable analogue (for example `StartCalendarInterval` or
+        `LimitLoadToSessionType`).
+
+        Mirrors the `systemd.*` namespace on NixOS: upstream service modules
+        guard these keys with
+        `lib.optionalAttrs (options ? systemd) { systemd.* = ...; }`, and the
+        same pattern applies here with `launchd` in place of `systemd`.
```
emilazy, line 78: (Which is a really bad idea, because it means that load‐bearing options will be silently dropped on other systems, leading to broken services or even security vulnerabilities.)

modules/services-modular/extra-root.nix
```diff
@@ -0,0 +1,93 @@
...
+in
+{
+  _class = "service";
+
+  options = {
+    processConfig = mkOption {
+      default = { };
+      type = types.submodule {
+        options = {
+          user = mkOption {
```
emilazy, line 13: `user` and `group` of course have the significant issue that you cannot portably *declare* users between NixOS and nix-darwin, and there are meaningful obstacles to doing so (nix-darwin requires static UID allocation whereas it is strongly discouraged in NixOS; naming conventions for service users differ).

Even if you could, the bigger problem is that on systemd, `DynamicUser` and extensive use of the systemd functionality for sandboxing and state directory handling is idiomatic, whereas these are not present at all on launchd and require services to be structured in different ways. When those systemd options are silently omitted, services will break or pose a security risk.

modules/services-modular/translate.nix
```diff
@@ -0,0 +1,62 @@
...
+
+  toServiceConfig = { kind, service }:
+    let
+      inherit (service) process processConfig;
+      cfg = processConfig;
+      argv = process.argv or [ ];
+      darwinExtras = service.launchd or { };
+    in
+    (optionalAttrs (argv != [ ]) {
+      ProgramArguments = argv;
```
emilazy, line 24: This does not work reliably. The Nix store is not necessarily mounted at boot or even the start of a graphical login session, which is why our `command` option translates into a shell invocation using `wait4path`.

modules/services-modular/translate.nix
```diff
@@ -0,0 +1,62 @@
...
+    optionalAttrs
+    ;
+
+  dash =
+    before: after:
+    if after == "" then before
+    else if before == "" then after
+    else "${before}-${after}";
+
+  toServiceConfig = { kind, service }:
```
emilazy, line 16: Since this does not set `RunAtLoad` or `KeepAlive`, and does not have functionality for the idiomatic macOS approach of socket activation, it looks to me like the daemon definitions will simply do nothing by default. (Another load‐bearing difference in model to systemd.)

Was this tested with any service definition?

emilazy (changes requested):
As I said in https://github.com/NixOS/nixpkgs/issues/428084#issuecomment-3991433341, and elaborated on at length in https://github.com/NixOS/nixpkgs/issues/428084#issuecomment-3994512395 and https://github.com/NixOS/nixpkgs/issues/428084#issuecomment-3998655081:

> I personally believe that it’s unlikely that modular services in their current form will be very useful for reducing maintenance burden for nix-darwin or providing a genuine abstraction for portability between systemd and launchd, unless by bringing NixOS services down to a lowest common denominator, which I don’t think would be a good thing.

launchd in particular is quite idiosyncratic in comparison to other common service managers, as I talked about in my comments on that tracking issue. I’m surprised to see this PR given that there was disappointingly no further engagement on the points I raised there.

As far as I am aware there was no engagement with nix-darwin or, to my knowledge, any active Nix‐on‐macOS maintainers at all when modular services were being developed and promoted as a solution to Linux–macOS portability – e.g. the NixOS manual section mentioning “on `nix-darwin` this is the module wrapping [`launchd`](https://en.wikipedia.org/wiki/Launchd)” despite no such module existing at the time (I found out about this name drop only after it was already merged). I think that’s unfortunate, as it means there was no chance to point out the issues with trying to abstract a lowest common denominator out of systemd and launchd earlier on in the design process.

The service used in the tests, ghostunnel, would currently be very unsafe to use with this: there is no portable interface for setting the user to run a service as (and there cannot idiomatically be, per my review comments), which means that despite the modular service setting `systemd.service.serviceConfig.DynamicUser`, it would run as `root` on nix-darwin. It also silently discards the options to set the certificate on non‐systemd, due to limitations of the modular services interface, which also makes it useless for its central purpose of TLS proxying even if it wasn’t a security vulnerability.

That kind of silent exposure and brokenness is not something we ought to expose to users, and this is before it has even tried to tackle the important issues of dependency handling and readiness notification, for which the systemd and launchd stories are completely incompatible. (Though this is already present in the modular services in Nixpkgs: the `php` modular service sets `Type = "notify";` for systemd and an equivalent for finit, but that’s not a concept that exists in launchd.)

This is a substantial amount of code with some hard blockers in its current state, but in general I don’t believe that this will add value to nix-darwin unless the underlying design questions are addressed. I am not sure they can be addressed in a way that does not impose unreasonable costs on NixOS and split nix-darwin maintenance across the repositories, and I am frankly increasingly concerned that the way the modular services effort has been approached so far indicates a lack of desire to engage about addressing them at all or to take community feedback into account.

KiaraGrouwstra:
@emilazy:

thanks again for your extensive feedback - and apologies, i should have indeed responded to your primary concerns when (if not before) filing this. given they've been relevant to modular services more broadly than the darwin case, i've responded to a few of them there now, hoping this will also provide for some context to my optimism.

that said, as someone on nixos rather than mac, i'll admit i've felt out of my depth here. as such, i hadn't personally felt qualified earlier to respond to your feedback at the tracking thread on the darwin-related bits.
more importantly tho, your points stand, and for the most part i wouldn't be qualified to say otherwise - it's mostly that in the areas of uncertainty i've been felt more hopeful on this.

> This is a substantial amount of code with some hard blockers in its current state

fair - and thank you for your comments. in filing this initial PR i'd hoped to make for a more concrete discussion. taking a step back here tho, let me see if i could first assuage your high-level concerns after all, despite my limited knowledge on darwin.

> As far as I am aware there was no engagement with nix-darwin or, to my knowledge, any active Nix‐on‐macOS maintainers at all when modular services were being developed and promoted as a solution to Linux–macOS portability – e.g. the NixOS manual section mentioning “on `nix-darwin` this is the module wrapping [`launchd`](https://en.wikipedia.org/wiki/Launchd)” despite no such module existing at the time (I found out about this name drop only after it was already merged). I think that’s unfortunate, as it means there was no chance to point out the issues with trying to abstract a lowest common denominator out of systemd and launchd earlier on in the design process.

pinging @roberth here as to the bits mentioned here from before my time on this topic.

> [NixOS/nixpkgs#428084 (comment)](https://github.com/NixOS/nixpkgs/issues/428084#issuecomment-3994512395)

i can't speak for others and the expectations they project toward this feature

> That kind of silent exposure and brokenness is not something we ought to expose to users, and this is before it has even tried to tackle the important issues of dependency handling and readiness notification, for which the systemd and launchd stories are completely incompatible. (Though this is already present in the modular services in Nixpkgs: the `php` modular service sets `Type = "notify";` for systemd and an equivalent for finit, but that’s not a concept that exists in launchd.)

so, maybe this was part of why i'd felt optimistic enough (in spite of my ignorance, and your warnings) to file this, but to users there may be different ways to interface with this:

1. directly configuring the 'lowest common denominator' interface (e.g. `process.argv = ...;`). in practice, i would expect this to happen primarily at service definitions in nixpkgs.
1. importing aforementioned service definitions from nixpkgs (e.g. `imports = [ pkgs.ghostunnel.services.default ];`), which on the side of a consuming system such as nix-darwin would involve processing the info from that lowest common denominator interface in a way sensible on the system in question, for in as far as applicable. as you note, this comes with no particular guarantees that the service in question has been tweaked for use on the consuming system, and as such is only as good as the expressiveness of that generic interface and the consuming system's ability to handle the info it exposes in a useful manner.
1. system-specific options, like some of the ones suggested in this initial draft like `launchd`. for what it's worth, these may just be redundant with the following category.
1. raw config using system configuration outside of modular services interfaces, like the blocks conditional on systemd in the service definitions at nixpkgs.[^native] such config can be either located at system-specific service definitions (as the mentioned nixos-specific blocks - for darwin not touched upon in this PR so far) to fill the gaps by the (current) generic interface, or at the side of the the config of whoever is consuming the system.

[^native]: nixos has so far had the privilege of service definitions having lived in 'their' repo, meaning they haven't had to be in the position of turning out to be among the systems untested for a particular service definition so far.

so to more concretely answer:

> That kind of silent exposure and brokenness is not something we ought to expose to users

i think we could address this gap by communicating to users the distinction between nixpkgs service definitions at best offering the info from the generic interface, vs system-specific modular service definitions that might complement the upstream definitions to fill their system-specific gaps, providing equivalents to the upstream blocks conditional on systemd.

as such, users might be made aware whether the service they wanted to consume has been tested for their system, or whether they might have reached the territory of having to tinker with what was available from outside of their native system.

does that make some sense?

again, sorry for having rushed this out before doing a better job on the communication front.

roberth:
Thanks for the ping. I think @KiaraGrouwstra is making good points so I'll try to build on that to provide just a bit more context.

Modular services is still very much in development. I've intentionally only sprinted the core skeleton of its module system hierarchy so that it can be a community project, not a me project.

I chose to do it that way because a big code dump (like Nix 2.4, flakes) *doesn't* work. Instead, I want it to be a community-driven project, and it's wide open for contributions, as you've rightfully observed are necessary.
So I'm glad we're having this conversation, despite the mismatched expectations.

> point out the issues with trying to abstract a lowest common denominator

The lowest common denominator is just a starting point. Built-in shims, feature checks and assertions can all help to make things not perfect but better.
We'll want to find a balance between those, and in order to get there, a PR like this is very helpful.

> unless the underlying design questions are addressed.

sgtm!

mweinelt:
@KiaraGrouwstra Do you get paid for this work? If so, that should go into the disclaimer as well.

KiaraGrouwstra:
@mweinelt i'll keep that in mind from now, thanks.
to answer the question tho, no, modular services isn't in scope there for now.

PR #1774, opened 2026-05-17
homebrew: fix `onActivation.cleanup = "zap"` after Homebrew bundle refactor

HuaDeity:
## Problem

Homebrew/brew@f38cd4b ("Convert bundle subcommands") moved each `brew bundle` subcommand into its own class with its own option parser. `--zap` is now declared only on `CleanupSubcommand`; the `install` subcommand (which `brew bundle` dispatches to by default) rejects it:

```
Error: Invalid usage: The install subcommand does not accept the --zap switch.
```

This breaks `darwin-rebuild switch` entirely for users with `homebrew.onActivation.cleanup = "zap"`.

## Fix

Run `brew bundle cleanup --force --zap` as a **separate invocation** after the install pass. The install pass must omit `--cleanup` in zap mode — if it ran `--cleanup` first, casks would be uninstalled normally before the zap pass could see them, and `brew uninstall --zap` would never fire.

`"uninstall"` mode keeps its existing single-invocation `--cleanup` flow (the `install` subcommand still accepts `--cleanup`).

## Behaviour matrix

| `cleanup` value | Invocations |
|---|---|
| `"none"` | `brew bundle --file='…' --no-upgrade` |
| `"check"` | checks phase: `brew bundle cleanup --file='…'` (dry-run) |
| `"uninstall"` | `brew bundle --file='…' --no-upgrade --cleanup` |
| `"zap"` | `brew bundle --file='…' --no-upgrade` → `brew bundle cleanup --file='…' --force --zap` |

## Validation

Both `HOMEBREW_NO_AUTO_UPDATE` and `extraEnv` are forwarded to the cleanup invocation. `extraFlags` is install-only (as documented) and is not passed to the cleanup command.

Verified against current Homebrew source (`Library/Homebrew/bundle/subcommand/install.rb` and `cleanup.rb`) that `install` accepts `--cleanup` but not `--zap`, and `cleanup` accepts `--force` and `--zap`.

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -100,10 +100,16 @@ let
           {command}`brew uninstall` is run for those packages.
 
           When set to `"zap"`, {command}`nix-darwin` invokes
-          {command}`brew bundle [install]` with the {command}`--cleanup --zap`
-          flags. This uninstalls all packages not listed in the generated Brewfile, and if the
-          package is a cask, removes all files associated with that cask. In other words,
+          {command}`brew bundle [install]`, followed by a separate
+          {command}`brew bundle cleanup --force --zap` invocation. This uninstalls all
+          packages not listed in the generated Brewfile, and if the package is a cask,
+          removes all files associated with that cask. In other words,
           {command}`brew uninstall --zap` is run for all those packages.
+
+          (Two invocations are required because, since the bundle subcommand refactor
+          in Homebrew (Homebrew/brew@f38cd4b), `--zap` is accepted only by the
+          `cleanup` subcommand and is rejected by the implicit `install` subcommand
+          that `brew bundle` defaults to.)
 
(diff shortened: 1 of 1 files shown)


matsjfunke:
would be really nice to get this

wthueb:
> "uninstall" mode keeps its existing single-invocation --cleanup flow (the install subcommand still accepts --cleanup).

Seems like not quite:
```
Error: Invalid usage: `brew bundle install --cleanup` requires `--force`, `--force-cleanup` or `$HOMEBREW_ASK`.
```

Edit: #1787

Samasaur1:
Looks like https://github.com/Homebrew/brew/commit/11d92383294d890692b79a887bf2b48b7203de0f reintroduced that flag on `brew bundle install`? The current version of `brew/Library/Homebrew/bundle/subcommand/install.rb` seems to handle it correctly. Feel free to reopen this PR if that is not the case.

PR #1712, opened 2026-02-24
README: Update instructions for 25.11

msgilligan:
Replace all references of `25.05` with `25.11`

PR #1782, opened 2026-05-25
version: bump to 26.11

Samasaur1:
à la #1649 and #1712.


Samasaur1:
oh, duh, the new version has made it to nixpkgs master but not yet the latest nixpkgs unstable

PR #1784, opened 2026-05-26
config/terminfo: remove termite

ethancedwards8:
Terminfo was removed from nixpkgs because it was deprecated upstream.

Link: https://github.com/NixOS/nixpkgs/pull/522784

### DIFF
--- a/modules/config/terminfo.nix
+++ b/modules/config/terminfo.nix
@@ -48,7 +48,6 @@
           rxvt-unicode-unwrapped
           rxvt-unicode-unwrapped-emoji
           st
-          termite
           tmux
           wezterm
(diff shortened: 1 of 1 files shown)


Samasaur1:
Ran into the same issue when cutting the 26.05 release so this patch is included in both of my PRs already. I'll leave this PR open for now but I think it'll be superseded

ethancedwards8:
ah thanks

rvolosatovs:
FYI, I'm currently unable to update to `nix-darwin-26.05` branch, due to this patch missing from it: https://github.com/nix-darwin/nix-darwin/blob/56c666e108467d87d13508936aade6d567f2a501/modules/config/terminfo.nix#L51

PR #1789, opened 2026-06-01
homebrew: address new CLI flag requirements

### LINKED ISSUE #1787: homebrew `bundle install --cleanup` changed break `homebrew.onActivation.cleanup`
After this commit https://github.com/Homebrew/brew/commit/e0d818bbbd6c7a28e336f2f89992ad696136d5d9, `homebrew.onActivation.cleanup = "unintall/zap"` need --force, --force-cleanup or $HOMEBREW_ASK set.

stepbrobd:
closes https://github.com/nix-darwin/nix-darwin/issues/1787

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -181,8 +181,8 @@ let
         ++ mapAttrsToList (k: v: "${k}=${escapeShellArg v}") config.extraEnv
         ++ [ "brew bundle --file='${brewfileFile}'" ]
         ++ optional (!config.upgrade) "--no-upgrade"
-        ++ optional (config.cleanup == "uninstall") "--cleanup"
-        ++ optional (config.cleanup == "zap") "--cleanup --zap"
+        ++ optional (config.cleanup == "uninstall") "--force-cleanup"
+        ++ optional (config.cleanup == "zap") "--zap --force-cleanup"
         ++ config.extraFlags
       );
     };
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -274,6 +274,21 @@ let
           for current behavior.
         '';
       };
+      trusted = mkOption {
+        type = types.bool;
+        default = false;
+        example = true;
+        description = ''
+          Whether to trust this tap during {command}`nix-darwin` system activation, by adding the
+          `trusted: true` option to its {command}`brew bundle` Brewfile entry.
+
+          Homebrew 6.0.0 enabled `HOMEBREW_REQUIRE_TAP_TRUST` by default, which refuses to
+          load formulae/casks from non-official taps that haven't been trusted, aborting activation.
+          Set this to `true` for non-official taps you control so their formulae and casks are
+          installed during activation. Official taps are always trusted, so this has no effect only
+          them.
+        '';
+      };
 
       brewfileLine = mkInternalOption { type = types.nullOr types.str; };
     };
@@ -287,7 +302,8 @@ let
           "tap ${sCfg.name}"
           + optionalString (sCfg ? clone_target) ", ${sCfg.clone_target}"
           + optionalString (sCfg ? force_auto_update)
-            ", force_auto_update: ${sCfg.force_auto_update}";
+            ", force_auto_update: ${sCfg.force_auto_update}"
+          + optionalString config.trusted ", trusted: true";
       };
   };
 
@@ -514,6 +530,27 @@ let
       # `version_file` is intentionally not exposed: it writes the installed version to a file
       # path relative to the `brew bundle` working directory, which is not meaningful during
       # nix-darwin system activation.
+      trusted = mkOption {
+        type = types.bool;
+        default = true;
+        example = false;
+        description = ''
+          Whether to trust this formula during {command}`nix-darwin` system activation, by adding the
+          `trusted: true` option to its {command}`brew bundle` Brewfile entry.
+
+          Homebrew 6.0.0 enabled `HOMEBREW_REQUIRE_TAP_TRUST` by default, which refuses to
+          load formulae from non-official taps that haven't been trusted, aborting activation. Set
+          this to `true` for a formula from a non-official tap you control so it is installed during
+          activation.
+
+          This only takes effect when {option}`name` is a fully-qualified name (`user/repo/formula`),
+          since only fully-qualified names map to a tap and can be trusted (and installed) on their
+          own. When {option}`name` is a plain formula name, it is resolved through your tapped
+          repositories, so trust must instead come from the containing tap being marked as trusted
+          (see [](#opt-homebrew.taps)). Official taps are always trusted, so this has no effect on
+          them.
+        '';
+      };
 
       brewfileLine = mkInternalOption { type = types.nullOr types.str; };
     };
@@ -521,14 +558,15 @@ let
     config =
       let
         sCfg = mkProcessedSubmodConfig config;
-        sCfgSubset = removeAttrs sCfg [ "name" "restart_service" "link" ];
+        sCfgSubset = removeAttrs sCfg [ "name" "restart_service" "link" "trusted" ];
       in
       {
         brewfileLine =
           "brew ${sCfg.name}"
           + optionalString (sCfgSubset != { }) ", ${mkBrewfileLineOptionsListString sCfgSubset}"
           + mkBrewfileLineBoolOrSymbolString "link" config sCfg
-          + mkBrewfileLineBoolOrSymbolString "restart_service" config sCfg;
+          + mkBrewfileLineBoolOrSymbolString "restart_service" config sCfg
+          + optionalString config.trusted ", trusted: true";
       };
   };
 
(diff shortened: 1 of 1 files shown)


stepbrobd:
i think this should also supersede https://github.com/nix-darwin/nix-darwin/pull/1774? not exactly sure

stepbrobd:
ci failure should be addressed by https://github.com/nix-darwin/nix-darwin/pull/1784#issuecomment-4546520432

ethancedwards8 (approved):

ethancedwards8:
@stepbrobd maybe we version gate this somehow? this fails on older homebrew versions

stepbrobd:
> @stepbrobd maybe we version gate this somehow? this fails on older homebrew versions

thx for the suggestion

iirc there isn’t any specific state version for homebrew in nix-darwin so we’d do this in a if/else block during activation?

or add a boolean option?

tbh i think the expectation in this repo is everyone have the latest version of homebrew? maybe this should be good as is and just ask users to update...

ethancedwards8:
maybe. i'm not sure what nix-darwin norms are. i agree that its probably best to just update but idk.

stepbrobd:
ping @Samasaur1

stepbrobd:
cleanup flag has been deprecated and trust has been made a requirement in https://github.com/Homebrew/brew/commit/bc44e3d0192104840b4543bf781d7db95814e95d

while waiting for this to be reviewed/merged, users can temporarily repin nix-darwin to `nix-darwin.url = "github:stepbrobd/nix-darwin/bundle";`

modules/homebrew.nix
```diff
@@ -274,6 +274,21 @@ let
           for current behavior.
         '';
       };
+      trusted = mkOption {
+        type = types.bool;
+        default = false;
```
DavSanchez, line 279: Thinking if this should be set to `true` by default.

I consider that, typically, when a `nix-darwin` user simply adds a tap to this list is because the intent is using the formulae it contains, so having it not work on addition feels like the wrong UX to me.

stepbrobd: now you've mentioned this i think it would be better to guard this under state version

heywoodlh:
Can confirm, I cherry-picked the commits from this PR and it resolved https://github.com/nix-darwin/nix-darwin/issues/1794 for me and got Nix-Darwin Homebrew module working correctly.

modules/homebrew.nix
Samasaur1: I would agree that this should default to true, as I can't really imagine a use case for adding a tap that isn't trusted.

modules/homebrew.nix
```diff
@@ -959,10 +974,18 @@ in
       fi
     '';
 
-    system.activationScripts.homebrew.text = mkIf cfg.enable ''
+    system.activationScripts.homebrew.text = let toTrust = map (t: t.name) (filter (t: t.trusted) cfg.taps); in mkIf cfg.enable ''
       # Homebrew Bundle
       echo >&2 "Homebrew bundle..."
       if [ -f "${cfg.prefix}/bin/brew" ]; then
+        ${optionalString (toTrust != [ ]) ''
```
Samasaur1: Instead of running `brew trust` manually, we should use the `trusted: true` Brewfile option (see https://docs.brew.sh/Brew-Bundle-and-Brewfile)

Samasaur1 (changes requested):
The `--cleanup` -> `--force-cleanup` changes look good to me. The trust changes need a little more work

modules/homebrew.nix
stepbrobd: curious if you mean defaulting to true regardless of the state version?

modules/homebrew.nix
khaneliman: I tested the `trusted: true` approach against Homebrew 6.0.0 and found one gap.

With nix-darwin generating:

```ruby
tap "deskflow/homebrew-tap", trusted: true
cask "deskflow", greedy: true
```

activation still fails:

```text
Error: Refusing to load cask deskflow/tap/deskflow from untrusted tap deskflow/tap.
Run `brew trust --cask deskflow/tap/deskflow` or `brew trust deskflow/tap` to trust it.
```

Looking at Homebrew 6.0.0, `trusted:` appears to be acted on for fully-qualified `brew`/`cask` entries, but not for `tap` entries during install. `Library/Homebrew/bundle/tap.rb` includes `trusted: true` when dumping taps, but I do not see an equivalent `Homebrew::Trust.trust!(:tap, ...)` during tap install/preinstall.

I verified these work:

```ruby
cask "deskflow/tap/deskflow", trusted: true
```

and:

```sh
brew trust --tap deskflow/homebrew-tap
```

Given that, I think nix-darwin may still need to run `brew trust --tap` before `brew bundle`, or Homebrew needs to make `tap ..., trusted: true` actually mutate tap trust before loading casks/formulae from that tap.

modules/homebrew.nix
Samasaur1: That is what I meant, though I don't feel strongly about it. I suppose it's probably better to allow for the case where taps were installed and the require trusted flag was already set, it just feels like friction

modules/homebrew.nix
khaneliman: https://github.com/Homebrew/brew/issues/22668 actually looks like the upstream issue

Samasaur1: You bring up a good point though, the `trusted` option should exist directly on formulae and casks so that they can be installed from untrusted taps without needing to have the tap trusted in your config

Samasaur1: Okay the upstream bug has been closed, can someone test whether `trusted: true` now works as expected?

Samasaur1 (changes requested):
Okay here's where I've landed:
1. `cleanup` -> `force-cleanup` still looks fine
2. if `trusted: true` works we should use that; otherwise the `brew trust` invocation is acceptable
3. Taps should default to not trusted for backwards compatibility reasons
4. Casks and formulae should have an identical trusted option, but it should default to true

Sorry to waffle back and forth on the default value for trusted, it's gonna be a little annoying to set them to trusted explicitly but it's plausible that existing configurations rely on the current (lack of) trust behavior.

stepbrobd:
supersedes https://github.com/nix-darwin/nix-darwin/pull/1802

stepbrobd:
tested in my config

```nix
nix-repl> :p darwinConfigurations.macbook.config.homebrew.brewfile
# Created by `nix-darwin`'s `homebrew` module

# Brews
brew "nextfire/tap/apple-music-discord-rpc", start_service: true, restart_service: :changed, trusted: true
brew "deno", trusted: true
brew "readline", trusted: true
brew "sqlite", trusted: true
...

# Casks
cask "stepbrobd/tap/opentabletdriver", trusted: true
cask "resonative/proaudio/ua-connect", trusted: true
cask "lyraphase/av-casks/soundid-reference", trusted: true
cask "alcove", trusted: true
cask "beeper", trusted: true
cask "betterdisplay", trusted: true
cask "container", trusted: true
...
```

homebrew seem to also generate entries in user homedir (tbh not a fan):

```console
$ cat .homebrew/trust.json
{
  "trustedtaps": [
    "nextfire/tap",
    "stepbrobd/tap"
  ],
  "trustedcasks": [
    "lyraphase/av-casks/soundid-reference",
    "resonative/proaudio/ua-connect",
    "stepbrobd/tap/opentabletdriver"
  ]
}
```

i'm keeping the commit history for now in case we need to do more iterations, but after the review i'll squash them

modules/homebrew.nix
stepbrobd: works fine ;)

stepbrobd:
(in future PRs) i think we should add a cron job for integration testing homebrew to catch errors like this earlier

modules/homebrew.nix
```diff
@@ -514,21 +529,43 @@ let
       # `version_file` is intentionally not exposed: it writes the installed version to a file
       # path relative to the `brew bundle` working directory, which is not meaningful during
       # nix-darwin system activation.
+      trusted = mkOption {
+        type = types.bool;
+        default = true;
+        example = false;
+        description = ''
+          Whether to trust this formulae during {command}`nix-darwin` system activation, by adding the
```
Samasaur1: In this description you're using "formulae" as both the singular and plural, but it's only the plural, the singular is formula. Therefore it should be:
- "Whether to trust this formula"
- "Set this to `true` for a formula"
- "`user/repo/formula`"
- "When {option}`name` is a plain formula name"

This isn't a blocker but you mentioned you wanted to do some squashing so I'm pointing it out here

Samasaur1 (approved):
Looks good to me! Thanks for sticking with this to get it over the line.

You mentioned you wanted to squash your commits, so I'm not merging this yet, but if you've changed your mind I am happy to merge this as-is and I can address the typo in a followup PR.

modules/homebrew.nix
stepbrobd: gee, thanks for this!

stepbrobd: addressed

wthueb:
Can this be backported to `nix-darwin-26.05`?

Samasaur1:
> Can this be backported to `nix-darwin-26.05`?

Yeah, it should be. @stepbrobd if you'd like your commits in the backport to show up as "Verified", you can open the backport PR (same commits just targeting `nix-darwin-26.05`); if you don't care, I can do the backport.

stepbrobd:
please go ahead with the backport

thx again for the review!

Samasaur1:
Backport in #1805

PR #1801, opened 2026-06-14
nix: set builders to @/etc/nix/machines when distributedBuilds is enabled

mikemckibben:
When `nix.distributedBuilds` is `true` and `nix.buildMachines` is configured, nix-darwin writes the machine specifications to `/etc/nix/machines` but never adds `builders = @/etc/nix/machines` to the generated `nix.conf`. This means the nix daemon does not read the machines file and distributed builds silently fail.

This fix mirrors the NixOS module's behavior for setting the builders when `distributedBuilds` is `true`.

### DIFF
--- a/modules/nix/default.nix
+++ b/modules/nix/default.nix
@@ -962,6 +962,7 @@ in
         # );
       }
 
+      (mkIf cfg.distributedBuilds { builders = "@/etc/nix/machines"; })
       (mkIf (!cfg.distributedBuilds) { builders = null; })
 
       (mkIf (isNixAtLeast "2.3pre") { sandbox-fallback = false; })


Samasaur1:
`@/etc/nix/machines` is already the default value for `builders` in Nix, so this shouldn't be necessary. The NixOS module also does not set this option explicitly, relying on the default value.

If you're running into issues, you may be better off opening a discussion or reaching out on Matrix.

mikemckibben:
Ok, definitely appears to be an issue from the nix-darwin side, but maybe it's just my flake.lock version. I did try a flake update to see if the latest version fixes it-- it didn't. The builders never gets set to point to the /etc/nix/machines symlink. My lock file has

```
{
        "lastModified": 1781190061,
        "narHash": "sha256-QRMpLbsmlciMGv4yx75FUoIl54K02JbIX1tgdPHPw1s=",
        "owner": "nix-darwin",
        "repo": "nix-darwin",
        "rev": "f2b3fdb347f91dfd344ad196666a0b0fe8aad05c",
        "type": "github"
}
```

Here is the workaround I made on my local configuration to get up and running in case someone from google lands here with the same issue.

```
nix.settings.builders = lib.mkIf config.nix.distributedBuilds "@/etc/nix/machines";
```

Samasaur1:
I'm confused what the actual error behavior you're seeing is.

1. Does `/etc/nix/machines` exist?
2. Does `/etc/nix/nix.conf` have `builders` set to anything?
3. Do remote builds work?

Without your workaround, I would expect the answers to be 1) yes; 2) no; 3) yes. It sounds like you're seeing 1) yes; 2) no; 3) no; in which case I think this is a bug in whatever Nix implementation you're using. If you can let me know what your answers to these questions are I can help point you in the right direction.

PR #1804, opened 2026-06-18
homebrew: pass extraEnv to cleanup checks

bryango:
**Context:** This is tangentially related to #1789 as `brew trust` writes its trust record under `$XDG_CONFIG_HOME` if it is set. So for the commands to be consistent one must set the same, consistent `extraEnv`.

**Disclaimer:** All code was written by hand by me but reviewed and verified by an agent (gpt-5.5 on codex).

When `homebrew.onActivation.cleanup == "check"`, nix-darwin runs a separate `brew bundle cleanup` command during `system.checks`, before the normal Homebrew activation command.

That cleanup check should use the same `homebrew.onActivation.extraEnv` values as the main `brew bundle [install]` command. For example, if `$XDG_CONFIG_HOME` is set via `extraEnv`, it should also be visible to `brew bundle cleanup` during the system check phase.

- **Note:** I intentionally kept `homebrew.onActivation.extraFlags` specific to the `bundle [install]` command, rather than passing those flags to the `bundle cleanup` check command, in case one needs to specifically customize the [install] behavior by adding e.g. `--verbose`. But I’m happy to revisit this behavior if you believe `extraFlags` should also apply to the cleanup check command.

### Commits

It would be easier to review the commits one by one through 👉  https://github.com/nix-darwin/nix-darwin/pull/1804/commits

- The 1st commit contains the intended behavior change: pass `homebrew.onActivation.extraEnv` through to the cleanup check.
- The following 2 commits should be trivial refactors. They move the shared PATH/sudo/env setup into `brewBundleCmd`, then reuse that helper for `system.checks`, without intending to change the actual generated commands apart from whitespace and flag ordering.
- **Note:** If the last 2 commits feel too heavy I am okay to drop them. The important behavioral change is the 1st commit that I would like to upstream.

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -151,7 +151,7 @@ let
         };
         description = ''
           Extra environment variables to set when {command}`nix-darwin` invokes
-          {command}`brew bundle [install]` during system activation.
+          {command}`brew bundle [install]` during system checks and activation.
 
           Useful for setting Homebrew's `HOMEBREW_NO_*` variables (e.g.,
           `HOMEBREW_NO_ENV_HINTS`, `HOMEBREW_NO_ANALYTICS`, `HOMEBREW_NO_UPDATE_REPORT_NEW`)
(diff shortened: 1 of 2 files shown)


bryango:
Friendly ping @Samasaur1 @mmenanno @malob for comments & review 😉

Samasaur1:
> Friendly ping @Samasaur1 @mmenanno @malob for comments & review 😉

I was actually midway through reviewing this when I got your ping :)

bryango:
Wow that's super fast! Thank you!

PR #1809, opened 2026-06-23
services.plugin-playground: init

aspauldingcode:
This PR introduces a new service module for [Plugin Playground](https://github.com/CoreBedtime/playground), a general-purpose runtime tweak system for macOS.

### What changed
- Added `services.plugin-playground` to manage the tweak injection environment.
- Configures default system preferences (`/opt/pluginplayground/current.options`) natively using macOS `defaults`.
- Registers the `grant` utility via a `launchd` daemon to manage process injection.

### Why it changed
To allow declarative Nix configuration of Plugin Playground on macOS (Apple Silicon).

### How it was tested
- Verified module options natively via a local `darwin-rebuild switch`.
- Ran `nix-build release.nix -A tests` locally without errors.

### Dependency
Depends on Nixpkgs PR https://github.com/NixOS/nixpkgs/pull/534452 to provide the underlying `plugin-playground` derivation.

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -104,6 +104,7 @@
   ./services/yabai
   ./services/nextdns
   ./services/jankyborders
+  ./services/plugin-playground
   ./programs/_1password.nix
   ./programs/_1password-gui.nix
   ./programs/arqbackup.nix
--- a/modules/services/plugin-playground/default.nix
+++ b/modules/services/plugin-playground/default.nix
@@ -1,0 +1,73 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.services.plugin-playground;
+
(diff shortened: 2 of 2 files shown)


aspauldingcode:
The new options are:
```
services.plugin-playground = {
  # (bool) Enables the Plugin Playground environment and `grant` daemon.
  enable = true; 
  
  # (package) Override the derivation to use (defaults to pkgs.plugin-playground).
  package = pkgs.plugin-playground; 
  
  # (bool) Disables arm64e PAC signing for spawned processes. 
  # Set to true when compiling without native arm64e ABI.
  disablePAC = true; 
  
  # (bool) Fallback to the legacy path (/private/var/ammonia/core/tweaks/).
  useLegacyAmmonia = false; 
  
  # (bool) Global kill-switch to pause all injection system-wide.
  pauseInjection = false; 
};
```

PR #1820, opened 2026-07-06
Add system.defaults.NSGlobalDomain.NSConvolutionOverride1

yzhou216:
See: <https://github.com/yannbertrand/macos-defaults/issues/450>


thor:
The same setting works for me: allows to override the rounded border radius of windows.

PR #1822, opened 2026-07-07
bump nixpkgs to fix manual/website builds

zowoq:
https://github.com/nix-darwin/nix-darwin/actions/runs/28835626241

Only the website build is run in CI on master but seems all of the checks that include the manual in their closure aren't able to be built unless `--override-input nixpkgs ...` is used.

cc @Samasaur1

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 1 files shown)


Samasaur1:
BLEGH this is exactly the kind of issue i was trying to avoid causing with the fix... thanks for catching this

PR #1830, opened 2026-07-12
homebrew: skip unchanged bundle activation

### LINKED ISSUE #1829: Skip Homebrew activation when Brewfile and managed inventory are unchanged
## Problem

The Homebrew activation script runs `brew bundle` on every `darwin-rebuild switch`, even when the generated Brewfile and Homebrew state have not changed. On a real Apple Silicon workstation with 3 declared formulae, 41 casks, and 21 Mac App Store apps, a warm `brew bundle check` takes 1.16 to 1.46 seconds. The full reconciliation also performs cleanup work.

Homebrew 6.0.7 exposes much cheaper inventory commands on the same machine:

```
brew list --formula -1  0.03s
brew list --cask -1     0.02s
mas list                0.05s
```

Alternative clients do not preserve the module's semantics. zerobrew 0.3.2 supports only casks with `binary` artifacts and does not implement `mas`; malt uses a separate prefix and currently parses but does not install `mas` entries.

## Proposed outcome

Add an opt-in `homebrew.onActivation` fast path that records the generated Brewfile identity and a cheap fingerprint of Homebrew-managed inventories after successful reconciliation. On a later activation, skip `brew bundle` only when both match. Any declared or externally observed inventory change must run the existing command unchanged.

The option should reject or bypass unsupported Brewfile directive types rather than silently weakening their semantics.

Related consumer issue: indexable-inc/index#2881.

(sent by an AI agent via Codex)

andrewgazelka:
Closes #1829.

## Summary

Add opt-in `homebrew.onActivation.fastPath` support. After a successful reconciliation, nix-darwin records:

* an identity covering the generated Brewfile and relevant activation configuration
* sorted inventories for taps, formulae, casks, and Mac App Store applications

A later activation skips `brew bundle` only when both are unchanged. Any configuration or inventory change runs the existing `brew bundle` command unchanged, then refreshes the state atomically.

The option rejects modes and directives whose behavior cannot be represented by these inventories, including upgrades, auto-update, arbitrary flags and Brewfile config, unsupported package categories, and formula link, conflict, or service state options.

## Tests

* `nix build .#checks.aarch64-darwin.homebrew-fast-path --no-link --print-build-logs`
* `nix build .#checks.aarch64-darwin.homebrew .#checks.aarch64-darwin.homebrew-cleanup-check .#checks.aarch64-darwin.homebrew-shell-integration --no-link --print-build-logs`

(sent by an AI agent via Codex)

### DIFF
--- a/modules/homebrew.nix
+++ b/modules/homebrew.nix
@@ -8,6 +8,12 @@ let
 
   brewfileFile = pkgs.writeText "Brewfile" cfg.brewfile;
 
+  fastPathIdentity = pkgs.writeText "homebrew-activation-identity" (builtins.toJSON {
+    version = 1;
+    inherit (cfg) brewfile prefix user;
+    inherit (cfg.onActivation) cleanup extraEnv;
+  });
+
   # Brewfile creation helper functions -------------------------------------------------------------
 
   mkBrewfileSectionString = heading: entries: optionalString (entries != [ ]) ''
@@ -172,7 +178,23 @@ let
         '';
       };
 
+      fastPath = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Whether to skip {command}`brew bundle` when the generated Brewfile, activation options,
+          and installed Homebrew inventory are unchanged since the last successful activation.
+
+          The inventory includes taps, formulae, casks, and Mac App Store applications. The fast
+          path requires [](#opt-homebrew.onActivation.autoUpdate) and
+          [](#opt-homebrew.onActivation.upgrade) to be disabled. It does not support arbitrary
+          Brewfile directives, extra flags, Visual Studio Code extensions, Go packages, Cargo
+          packages, or formula options that reconcile link, conflict, or service state.
+        '';
+      };
+
       brewBundleCmd = mkInternalOption { type = types.functionTo types.str; };
+      inventoryCmd = mkInternalOption { type = types.str; };
     };
 
     config = {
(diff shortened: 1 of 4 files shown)


scriptogre:
Tested on an M1 Pro with Homebrew 6.0.13: 1 formula, 29 casks, 8 MAS apps.

I changed a tracked comment each run to force normal flake evaluation.

| Mode | Average |
|---|---:|
| Current | 10.63s |
| `fastPath` | 8.11s |

So **2.5 seconds, or 24%, faster**, which is pretty nice! 

Cache hits and misses worked as expected.

Suggestions:

- Maybe rename `fastPath` to `reconcileIfChanged`, following Nix's [`restartIfChanged`](https://github.com/NixOS/nixpkgs/blob/b7c2ada94fe99c15b0dbcf4d11fd7850b957a436/nixos/lib/systemd-unit-options.nix#L547-L555) pattern?
 - Maybe we shouldtrack only MAS app IDs, since version changes trigger unnecessary rebuilds?
 - Should probably add behavior tests for cache hits, misses, drift, and failed reconciliation

scriptogre:
I put the follow-up changes in [andrewgazelka/nix-darwin#1](https://github.com/andrewgazelka/nix-darwin/pull/1). All four Homebrew checks pass locally.

The failed upstream check is an unrelated `yabai` linker crash. Merging the follow-up will push a new commit and rerun CI.

scriptogre:
Opened #1849 as a rebased continuation with your commits preserved.

PR #1831, opened 2026-07-16
networking: add declarative /etc/hosts management

### LINKED ISSUE #1035: Option to configure /etc/hosts file
Hi!

In NixOS there is the option `networking.hosts` to configure the `/etc/hosts` file for additional DNS entries.

Example:
```nix
networking.hosts = {
  "127.0.0.2" = ["other-localhost"];
  "192.0.2.1" = ["mail.example.com" "imap.example.com"];
};
```
See [NixOS Wiki - Networking](https://nixos.wiki/wiki/Networking) for more information.

However, a similar options seems to be missing on nix-darwin regarding the listed options here: https://daiderd.com/nix-darwin/manual/index.html

Can we add the `networking.hosts` option to nix-darwin or is there already another way to configure the hosts file?

DavSanchez:
Closes #1035.

Adds `networking.enableHosts`, `networking.hosts`, `networking.extraHosts`, and `networking.hostFiles` for declarative `/etc/hosts` management on darwin.

This implementation avoids the symlink entirely by using an activation script in `system.activationScripts.postActivation` that writes a regular file to `/etc/hosts`, addressing issues with the previous #939 that was reverted. 

On every activation:

1. Read the current `/etc/hosts` (a regular file)
2. Strip any previous `# BEGIN Nix-managed … # END Nix-managed` block
3. Write back the original (non-Nix) content, followed by the generated entries wrapped in the `BEGIN`/`END` markers

Step 3 only adds the Nix block if there's any entries defined in the `nix-darwin` config, which means existing entries added by macOS, Docker Desktop, VPN clients, or manual edits are preserved (well, unless they were added in the middle of a Nix block).

If `networking.enableHosts` is set to `false`, then only steps 1 and 2 are ran.

I wanted to add a `stevenblack` module to `nix-darwin` and noticed that I'd need to add this first. If this or a similar one lands, I'll open another one with the `stevenblack` module almost verbatim from Nixpkgs.


DavSanchez:
Hi @Enzime @Samasaur1 , requesting a review if you have the time. Let me know your thoughts.

DavSanchez:
Noticed there's some tests failing in the CI but they don't seem to be related to my change.

Issue #391, opened 2021-12-06
Respect `/etc/paths.d` and `/etc/manpaths.d`


kidonng:
XCode (`rvictl`), Wireshark (CLI) and some other apps put config in `/etc/paths.d`. By using nix-darwin one has to configure this manually (and may not realize it!)

Ideally nix-darwin should use `/usr/libexec/path_helper` but that's not _functional_ since it depends on unspecified inputs. Is there a better way to get the desired `$PATH` and `$MANPATH` (i.e. respect `/etc/paths.d` and `/etc/manpaths.d`)?

lilyball:
I've noticed this now myself. `/etc/profile` will run `/usr/libexec/path_helper` prior to sourcing `/etc/bashrc`, but then `/etc/bashrc` overwrites it. Unless it's a nested login shell, then `$__NIX_DARWIN_SET_ENVIRONMENT_DONE` is already set and `/etc/bashrc` doesn't modify `$PATH`, thus resulting in the `path_helper` paths moving to the front of `$PATH` and the Nix environment paths showing up afterwards.

(note: the manpage for `path_helper` says it appends the paths, but it doesn't, it prepends them)

I don't use `$MANPATH` myself, but for the `$PATH` issue, I think nix-darwin should do the following:

In the `set-environment` script it currently sets `$PATH`, and the value it sets includes `/usr/local/bin:/usr/bin:/usr/sbin:/bin:/sbin`. This suffix is hardcoded by nix-darwin. Instead of hard-coding this suffix, it should do something like

```bash
export PATH=$HOME/.nix-profile/bin:/etc/profiles/per-user/$USER/bin:/run/current-system/sw/bin:/nix/var/nix/profiles/default/bin
if shopt -q login_shell && [ -x /usr/libexec/path_helper ]; then
    eval `PATH= /usr/libexec/path_helper -s`
    PATH=$HOME/.nix-profile/bin:/etc/profiles/per-user/$USER/bin:/run/current-system/sw/bin:/nix/var/nix/profiles/default/bin:${PATH:-/usr/local/bin:/usr/bin:/usr/sbin:/bin:/sbin}
else
    PATH=$HOME/.nix-profile/bin:/etc/profiles/per-user/$USER/bin:/run/current-system/sw/bin:/nix/var/nix/profiles/default/bin:/usr/local/bin:/usr/bin:/usr/sbin:/bin:/sbin
fi
```

[...]

PR #1832, opened 2026-07-17
feat: Add option `enablePathHelper` for zsh to support macOS `path_helper`

paulchen5:
Without the option certain binaries from various applications aren't getting loaded into the `$PATH` variable. For more information run `man path_helper`.


modules/programs/zsh/default.nix
```diff
@@ -126,6 +126,12 @@ in
       description = "Enable zsh-syntax-highlighting.";
     };
 
+    programs.zsh.enablePathHelper = mkOption {
+      type = types.bool;
+      default = true;
+      description = "Enable the macOS path_helper utility";
```
SuperSandro2000:
```suggestion
      description = "Whether to run the macOS path_helper utility to add extra directories from Applications to PATH";
```

modules/programs/zsh/default.nix
```diff
@@ -191,6 +197,12 @@ in
       if test -f /etc/zprofile.local; then
         source /etc/zprofile.local
       fi
+
+      ${optionalString cfg.enablePathHelper
+        "if [ -x /usr/libexec/path_helper ]; then
+        eval `/usr/libexec/path_helper -s`
+fi" 
+      }
     '';
```
SuperSandro2000:
```suggestion
      ${optionalString cfg.enablePathHelper ''
        if [ -x /usr/libexec/path_helper ]; then
          eval `/usr/libexec/path_helper -s`
        fi 
      ''}
    '';
```

SuperSandro2000:
This will close #391

modules/programs/zsh/default.nix
```diff
@@ -126,6 +126,12 @@ in
       description = "Enable zsh-syntax-highlighting.";
     };
 
+    programs.zsh.enablePathHelper = mkOption {
+      type = types.bool;
+      default = true;
```
Enzime: Not sure if this is a good default

Could you post what the PATH would look like with this option disabled and enabled?

paulchen5: This is `$PATH` without the Option:

`/Users/<user-name>/.local/share/zinit/polaris/bin:/opt/homebrew/bin:/opt/homebrew/sbin:/Users/<user-name>/.nix-profile/bin:/run/current-system/sw/bin:/nix/var/nix/profiles/default/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/<user-name>/go/bin:/Applications/Nix Apps/Ghostty.app/Contents/MacOS`

And this is `$PATH` with the option being enabled:

`/Users/<user-name>/.local/share/zinit/polaris/bin:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/pkg/env/global/bin:/Applications/Privileges.app/Contents/MacOS:/Users/<user-name>/.nix-profile/bin:/run/current-system/sw/bin:/nix/var/nix/profiles/default/bin:/Users/<user-name>/go/bin:/Applications/Nix Apps/Ghostty.app/Contents/MacOS`

Enzime: Looks like it puts a lot of stuff in front of nix-darwin’s PATH entries could you change the default to be false?

paulchen5: Ok I changed it. However this "option" is enabled by default on macOS when you don't configure it via `nix-darwin`.

This ist the content of `/etc/zprofile.before-nix-darwin`

```zsh
# System-wide profile for interactive zsh(1) login shells.

# Setup user specific overrides for this in ~/.zprofile. See zshbuiltins(1)
# and zshoptions(1) for more details.

if [ -z "$LANG" ]; then
        export LANG=C.UTF-8
fi

if [ -x /usr/libexec/path_helper ]; then
        eval `/usr/libexec/path_helper -s`
fi
```

SuperSandro2000: The default should probably be true, even if it is not the nix way, as otherwise existing integration and apps break and it is kinda hard to figure out how and why.

Enzime:
I don't see any option to trigger the checks again, can you rebase this PR and push it again?

PR #1858, opened 2026-08-29
mas: support declaring apps by bundle id and migrate list parsing to JSON

iamanaws:
support declaring apps by bundle id (similar to https://github.com/nix-darwin/nix-darwin/pull/1846) and migrate list parsing to use the new --json output [mas@v7.0.0](https://github.com/mas-cli/mas/releases/tag/v7.0.0)

### DIFF
--- a/modules/programs/mas.nix
+++ b/modules/programs/mas.nix
@@ -27,8 +27,8 @@ let
 
   apps = mapAttrsToList (name: id: { inherit name id; }) cfg.packages;
 
-  desiredIds = map (app: toString app.id) apps;
-  homebrewIds = map toString (attrValues config.homebrew.masApps);
+  desiredIds = map (app: escapeShellArg app.id) apps;
+  homebrewIds = map (id: escapeShellArg id) (attrValues config.homebrew.masApps);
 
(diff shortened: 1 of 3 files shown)


modules/programs/mas.nix
```diff
@@ -27,8 +27,8 @@ let
 
   apps = mapAttrsToList (name: id: { inherit name id; }) cfg.packages;
 
-  desiredIds = map (app: toString app.id) apps;
-  homebrewIds = map toString (attrValues config.homebrew.masApps);
+  desiredIds = map (app: escapeShellArg (toString app.id)) apps;
+  homebrewIds = map (id: escapeShellArg (toString id)) (attrValues config.homebrew.masApps);
```
nrabulinski:
```suggestion
  desiredIds = map (app: escapeShellArg app.id) apps;
  homebrewIds = map escapeShellArg (attrValues config.homebrew.masApps);
```
`escapeShellArg` already calls `toString` on the argument

nrabulinski (approved):
One small nit, besides that I think it's good

modules/programs/mas.nix
iamanaws: fixed, thanks!

PR #1864, opened 2026-09-12
services/xquartz: init

booxter:
Thanks to @Feyorsh for initial inspiration at:
https://github.com/nix-darwin/nix-darwin/pull/1542

This module does *not* rely on plists from xquartz package but defines
services natively. Comparing to upstream packaged plists, I removed a
lot of unnecessary / obsolete / unused attributes and only left those
necessary for the operation of the service.

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -102,6 +102,7 @@
   ./services/tailscale.nix
   ./services/trezord.nix
   ./services/wg-quick.nix
+  ./services/xquartz.nix
   ./services/yabai
   ./services/nextdns
   ./services/jankyborders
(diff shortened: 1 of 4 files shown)


booxter:
I'm not sure I am happy about this module, because it hits https://github.com/nix-darwin/nix-darwin/issues/1255 which results in service failing to start (recovers after relogin) on new plist deployed. Strictly speaking, this is not a bug in the new module but in nix-darwin, but what I know is that if I were to deploy the GUI side of the service via Home Manager (something I am doing in my private config), the service has no such issue.

I wonder if this module should be two separate modules: one for system side (in nix-darwin) and another for user side (in HM). Or maybe we can tie it so that for the user side we use HM launchd implementation that doesn't have this issue...

